PR #81, opened 2024-03-05
Refactor

Rockindash:
Refactor:
- Models
- Extensions
- View Models
- States

Localization


johnnyluo (approved):
Thanks for the refactor @Rockindash , LGTM

PR #82, opened 2024-03-05
Construct unique device name for keygen and keysign

johnnyluo:
Currently we are using `UIDevice.current.name` as local party's identifier when keygen and keysign , but since IOS15 , `UIDevice.current.name` will only return a generic name , like `iphone` or `ipad` , instead the name user put in their settings

If we get the device name in user's settings , then we have to request a special entitlement. 

I don't want to request the entitlement , just I updated the view and also the way how we get the local party identifier

Also , currently for the back button on the top navigation bar , we are using 
```
ToolbarItem(placement: .navigationBarLeading) {
                NavigationButtons.backButton(presentationStack: self.$presentationStack)
            }
```
I removed this , just let it to use the standard one , it says `back` , but it is ok

### DIFF
--- a/voltixApp/VoltixApp.xcodeproj/xcshareddata/xcschemes/VoltixApp.xcscheme
+++ b/voltixApp/VoltixApp.xcodeproj/xcshareddata/xcschemes/VoltixApp.xcscheme
@@ -17,7 +17,7 @@
                BlueprintIdentifier = "DE49A1622B65F6D9000F3AFB"
                BuildableName = "VoltixApp.app"
                BlueprintName = "VoltixApp"
-               ReferencedContainer = "container:VoltixApp.xcodeproj">
+               ReferencedContainer = "container:voltixApp.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
       </BuildActionEntries>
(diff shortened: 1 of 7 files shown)


voltixApp/VoltixApp.xcodeproj/xcshareddata/xcschemes/VoltixApp.xcscheme
```diff
@@ -17,7 +17,7 @@
                BlueprintIdentifier = "DE49A1622B65F6D9000F3AFB"
                BuildableName = "VoltixApp.app"
                BlueprintName = "VoltixApp"
-               ReferencedContainer = "container:VoltixApp.xcodeproj">
```
johnnyluo, line 20: @Rockindash , I am not sure why xcode fighting with this , do you know?

enriquesouza: I noticed that as well. Sometimes it is low case and sometimes it is capital.

enriquesouza (approved):

voltixApp/VoltixApp.xcodeproj/xcshareddata/xcschemes/VoltixApp.xcscheme
Rockindash: So, the closet reference I could find to this issue was this forum
https://forums.developer.apple.com/forums/thread/663975
Seems like, it's how Xcode modifies the scheme to run it over different devices (simulators or physical devices).
We don't need to worry about it since its another of those annoying Xcode's issues. But if needed we can add this file to git ignore.

PR #84, opened 2024-03-06
USDC fixed removal - Support any ERC20

enriquesouza:
Removal of USDC hardcode. 

Transaction tested and sent.

image


voltixApp/VoltixApp/Views/Send/SendInputDetailsView.swift
```diff
@@ -209,7 +210,8 @@ struct SendInputDetailsView: View {
                                                         if tx.coin.ticker.uppercased() == "ETH" {
                                                             tx.amount = eth.addressInfo?.ETH.getAmountInEth(newValueDouble) ?? ""
                                                         } else {
-                                                            if let tokenInfo = eth.addressInfo?.tokens?.first(where: {$0.tokenInfo.symbol == "USDC"}) {
+                                                            if let tokenInfo = tx.token {
```
Rockindash, line 213: This can be rewritten as:
else if let tokenInfo = tx.token {
// Code here
}

enriquesouza: Good idea.

Rockindash (approved):

PR #85, opened 2024-03-06
Thorchain: Support to balance, prices, and accounts.

enriquesouza:
Thorchain support to balance, prices, and accounts.

image

### DIFF
--- a/voltixApp/VoltixApp/View Models/Services/Thorchain/ThorchainService.swift
+++ b/voltixApp/VoltixApp/View Models/Services/Thorchain/ThorchainService.swift
@@ -44,7 +44,7 @@ class ThorchainService: ObservableObject {
     var formattedRuneBalance: String? {
         guard let balances = balances else { return nil }
         for balance in balances {
-            if balance.denom == "rune" {
+            if balance.denom.lowercased() == Chain.THORChain.ticker.lowercased() {
                 guard let runeAmount = Double(balance.amount) else { return "Invalid balance" }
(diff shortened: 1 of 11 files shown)


voltixApp/VoltixApp/View Models/Services/Thorchain/ThorchainService.swift
```diff
@@ -0,0 +1,186 @@
...
+                return formatter.string(from: NSNumber(value: balanceRune))
+            }
+        }
+        return "Balance not available"
+    }
+    
+    var runeBalance: String? {
+        guard let balances = balances else { return nil }
+        for balance in balances {
+            if balance.denom == "rune" {
```
Rockindash, line 66: Is this checking for selected currency??
If so, can we use an enum instead? It gets prone to make a mistake here when using plain String.
We can remove any margin of error by using enums.

enriquesouza: It is not the selected coin, this is specific for the RUNE, but we can always create enums, not sure where should we place it. The coin at the moment is called THORCHAIN, but in the blockchain, it has another name. Maybe we can rename the coin to RUNE and keep the chain as Thorchain @johnnyluo?

enriquesouza: I will use the ticker instead @Rockindash. It will solve this issue

PR #95, opened 2024-03-12
Solana

enriquesouza:
This is the Solana's implementation.

Pasted Graphic
+0 00068 SOL

https://solscan.io/tx/5jXU9CzaZut44jXuVRFD9HVhXuYxd2wupMoBfR2eji46CCAVJ82RUNX85qN46tmGzZgGq7iDCJnYcB9hZhckXSZ9


Rockindash (approved):

voltixApp/VoltixApp/Model/Solana/SolanaRecentBlockhash.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+extension SolanaService {
+	
+	class SolanaRecentBlockhashResponse: Codable {
```
johnnyluo, line 5: I don't understand why these classes are under SolanaService extension. can you explain?

voltixApp/VoltixApp/Views/Send/SendVerifyView.swift
```diff
@@ -1,290 +1,317 @@
...
+					await unspentOutputsService.fetchUnspentOutputs(for: tx.fromAddress)
+				}
+			}
+		} else if tx.coin.chain.name.lowercased() == Chain.THORChain.name.lowercased() {
+			Task {
+				await thor.fetchAccountNumber(tx.fromAddress)
+			}
+		} else if tx.coin.chain.name.lowercased() == Chain.Solana.name.lowercased() {
+			Task {
+				await sol.getSolanaBalance(account: tx.fromAddress)
```
johnnyluo, line 38: I think you can fire off multiple async at the same time , and then wait for all the result at the end.
FYI: https://www.swiftbysundell.com/articles/swift-concurrency-multiple-tasks-in-parallel/

voltixApp/VoltixApp/Services/Solana/SolanaService.swift
```diff
@@ -0,0 +1,128 @@
...
+				"id": 1,
+				"method": "sendTransaction",
+				"params": [encodedTransaction]
+			]
+			let data = try await postRequest(with: requestBody)
+			
+			print(String(data: data, encoding: String.Encoding.utf8))
+			
+			let response = try jsonDecoder.decode(SolanaRPCResponse<String>.self, from: data)
+			DispatchQueue.main.async { [weak self] in
```
johnnyluo, line 72: if I understand it correctly , update a published field doesn't need to be on the main thread,  which means you don't need to use DispatchQueue.  

Also you decorate the class with `@MainActor`  , which means it should be running on the main thread anyway

@Rockindash  is my understanding correct?

Rockindash: Yes, if the class has the macro @MainActor, you don't need DispatchQueue at all. It's already running the tasks in the main thread.

PR #112, opened 2024-03-18
Add viewmodel to keygen/keysign peer discovery view and join keygen / keysign view

johnnyluo:


voltixApp/VoltixApp/Views/Keysign/KeysignDiscoveryView.swift
```diff
@@ -179,45 +133,24 @@ struct KeysignDiscoveryView: View {
     
     var bottomButtons: some View {
         Button(action: {
-            self.startKeysign(allParticipants: self.selections.map { $0 })
-            self.currentState = .Keysign
-            self.participantDiscovery.stop()
+            self.viewModel.startKeysign()
```
Rockindash, line 136: You can remove self when referencing the view models. 
Same goes for other classes as well.

voltixApp/VoltixApp/View Models/KeysignDiscoveryViewModel.swift
```diff
@@ -0,0 +1,104 @@
...
+
+enum KeysignDiscoveryStatus {
+    case WaitingForDevices
+    case FailToStart
+    case Keysign
+}
+
+class KeysignDiscoveryViewModel: ObservableObject {
+    private let logger = Logger(subsystem: "keysign-discovery", category: "viewmodel")
+    var vault: Vault
```
Rockindash, line 18: Can use let instead of var here for the declaration.

Rockindash (approved):

voltixApp/VoltixApp/Views/Keysign/KeysignDiscoveryView.swift
johnnyluo: yeah, I know , I am using swiftformat , it will automatically add it back for me , so I will keep it.

voltixApp/VoltixApp/View Models/KeysignDiscoveryViewModel.swift
johnnyluo: need to change it at `setData` , so have to use var

PR #119, opened 2024-03-21
EVM transaction broadcast moved to infura + refactories

enriquesouza:


johnnyluo:
Not relevant anymore , as we move to use etherscan

PR #121, opened 2024-03-22
Refactor ios

Rockindash:
Refactored the following view for the send crypto flow
- Key sign View
- Join Keysign


johnnyluo (approved):
LGTM

PR #129, opened 2024-03-24
Open transaction link on keysign done view

johnnyluo:
The PR will add the feature , on keysign done view , when user click the link button , it will open the transaction on a block explorer on safari

### DIFF
--- a/voltixApp/VoltixApp/View Models/KeysignViewModel.swift
+++ b/voltixApp/VoltixApp/View Models/KeysignViewModel.swift
@@ -65,7 +65,12 @@ class KeysignViewModel: ObservableObject {
         self.vault = vault
         self.keysignPayload = keysignPayload
     }
-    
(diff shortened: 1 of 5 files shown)


voltixApp/VoltixApp/View Models/KeysignViewModel.swift
```diff
@@ -65,7 +65,12 @@ class KeysignViewModel: ObservableObject {
         self.vault = vault
         self.keysignPayload = keysignPayload
     }
-    
+    func getTransactionExplorerURL(txid: String) -> String{
+        if let keysignPayload {
```
Rockindash, line 69: guard would be a better check in this case.

Rockindash (approved):

voltixApp/VoltixApp/View Models/KeysignViewModel.swift
johnnyluo: fixed

PR #130, opened 2024-03-25
Ios entry point

Rockindash:
Setup a new entry point with the following scenarios:
- Show Onboarding view to first-time users
- Show Create Vault view to new users
- Show Vaults view to existing users


enriquesouza (approved):
LGTM

PR #176, opened 2024-04-02
Avax rpc

enriquesouza:
Avax support and ARC20 support


voltixApp/VoltixApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,41 +0,0 @@
-{
-  "pins" : [
-    {
```
johnnyluo, line 3: I don't think you can remove this one , otherwise build will fail

voltixApp/VoltixApp/Model/Coin.swift
```diff
@@ -95,7 +96,12 @@ class Coin: Codable, Hashable {
     }
     
     func getMaxValue(_ fee: BigInt) -> Decimal {
-        let maxValue = (BigInt(rawBalance, radix: 10) ?? 0) - fee
+        
+        // USDC has 6 decimals
```
johnnyluo, line 100: I believe this is not generic , and won't work for other coins , for example 
BTC is using 1E8 
ATOM is using 1E6

enriquesouza: UTXO does not use this, only EVM

voltixApp/VoltixApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
enriquesouza: It is generated in runtime

voltixApp/VoltixApp/Model/Coin.swift
johnnyluo: then , maybe rename it to something more EVM specific , for example: `getEVMMaxValue`

johnnyluo (approved):
LGTM

voltixApp/VoltixApp/Model/Coin.swift
enriquesouza: Thanks for your catch. I am fixing it.

enriquesouza: Fixed and tested on:
Avax
USDC-avax
Thor RUNE
Sol

PR #191, opened 2024-04-03
Given the congestions on solana it was not broadcasting any transacti…

enriquesouza:
…on. So I had to usin 7k lamports + 500k microLamports so it could go through.


voltixApp/VoltixApp/Services/Solana/SolanaService.swift
```diff
@@ -106,15 +106,14 @@ class SolanaService: ObservableObject {
...
                 "params": [["commitment": "finalized"]]
             ]
 			
             let data = try await postRequest(with: requestBody)
-            let response = try jsonDecoder.decode(SolanaRPCResponse<SolanaRecentBlockhashResponse>.self, from: data)
-			
-            self.recentBlockHash = response.result.value.blockhash
-            self.feeInLamports = String(response.result.value.feeCalculator.lamportsPerSignature)            
+            
+            self.recentBlockHash = Utils.extractResultFromJson(fromData: data, path: "result.value.blockhash") as? String
```
flypaper0, line 115: Why we changing service class state instead of returning values from method?

Rockindash (approved):

voltixApp/VoltixApp/View Models/SendCryptoVerifyViewModel.swift
```diff
@@ -216,7 +216,7 @@ class SendCryptoVerifyViewModel: ObservableObject {
                 coin: tx.coin,
                 toAddress: tx.toAddress,
                 toAmount: tx.amountInLamports,
-                chainSpecific: BlockChainSpecific.Solana(recentBlockHash: recentBlockHash, priorityFee: 0),
+                chainSpecific: BlockChainSpecific.Solana(recentBlockHash: recentBlockHash, priorityFee: 500_000),
```
johnnyluo, line 219: is there a way to get the priority fee from block explorer or node?

PR #194, opened 2024-04-04
[Swap] Fetching quotes 

flypaper0:


voltixApp/VoltixApp/Model/Chain.swift
```diff
@@ -70,6 +70,31 @@ enum Chain: String, Codable, Hashable, CaseIterable {
         }
     }
 
+    var asset: String {
```
johnnyluo, line 73: why `.solana` cause a fatal error?  is it specific for swap? maybe we should name it like `swapAsset` ?

johnnyluo (approved):
LGTM

voltixApp/VoltixApp/Model/Chain.swift
johnnyluo: you could raise follow up PR to address it

PR #196, opened 2024-04-04
[Swap] Assets fixed

flypaper0:


voltixApp/VoltixApp/Model/Chain.swift
```diff
@@ -70,7 +70,7 @@ enum Chain: String, Codable, Hashable, CaseIterable {
         }
     }
 
-    var asset: String {
+    var swapAsset: String {
```
enriquesouza, line 73: Why is it renamed to swapAsset?

PR #200, opened 2024-04-05
Evm rpc public node

enriquesouza:
Migration to use public RPCs, refractory on EVMs, and clean up the code


voltixApp/VoltixApp/Services/RpcEvmService.swift
```diff
@@ -105,8 +104,17 @@ class RpcEvmService {
         return try await intRpcCall(method: "eth_getBalance", params: [address, "latest"])
     }
     
-    private func fetchMaxPriorityFeePerGas() async throws -> BigInt {
-        return try await intRpcCall(method: "eth_maxPriorityFeePerGas", params: [])
+    func fetchMaxPriorityFeePerGas() async throws -> BigInt {
+        let feeInGwei = try await intRpcCall(method: "eth_maxPriorityFeePerGas", params: [])
+        
+        let feeInGweiString = feeInGwei.description
```
johnnyluo, line 110: I think this is wrong , the result return from the RPC is in Wei , so we always need to convert it to GWei here , and also we should maybe keep the minimum as 1 Gwei

voltixApp/VoltixApp/Services/BalanceService.swift
```diff
@@ -8,76 +8,60 @@
...
-
+    
     func balance(for coin: Coin) async throws -> (coinBalance: String, balanceUSD: String) {
         await CryptoPriceService.shared.fetchCryptoPrices()
-
-        switch coin.chain {
-        case .bitcoin, .bitcoinCash, .litecoin, .dogecoin:
+        
+        switch coin.chainType {
+        case .UTXO:
```
johnnyluo, line 23: I prefer to use switch on `coin.chain` , it is much better than using `chainType`  , chainType is a really weird property , for example , bitcoin is UTXO chain , that's fine , but .THORChain is not a chain type , Solana is not a chain type either  
Could you please revert  it to use `coin.chain`?   and also since `coin.chain` is enum , so if the case is not exhaustive , it will give us an error

johnnyluo (changes requested):

voltixApp/VoltixApp/Services/RpcEvmService.swift
enriquesouza: **BSC:**
priorityFeeGwei: 1000000000
TICKER: USDT - BSC
gas: 1
nonce: 2

priorityFeeGwei: 1000000000
TICKER: BNB
gas: 1
nonce: 2

**AVAX**

TICKER: AVAX
gas: 25
nonce: 6
priorityFeeGwei: 0

TICKER: USDC - AVAX
gas: 25
nonce: 6
priorityFeeGwei: 0

**Ethereum**

TICKER: ETH
gas: 12
nonce: 2
priorityFeeGwei: 1000000

TICKER: USDC
gas: 13
nonce: 2
priorityFeeGwei: 970000

Rockindash (approved):

voltixApp/VoltixApp/Services/RpcEvmService.swift
johnnyluo: Ethereum

TICKER: ETH
gas: 12
nonce: 2
priorityFeeGwei: 1000000

TICKER: USDC
gas: 13
nonce: 2
priorityFeeGwei: 970000

This means ETH's priority fee is less then 1 gwei , then we can just use 1gwei

enriquesouza: Yep. I've done that.

voltixApp/VoltixApp/Services/BalanceService.swift
enriquesouza: Done

johnnyluo (approved):

PR #201, opened 2024-04-05
Ios UI fixes

Rockindash:
Fixed the following issues:
- Identify SELF on tx list
- All addresss txId is hyper-linked (open browser)
- Unnecessary Camera Screen
- Device Select List is Cards (not ROW)


johnnyluo (approved):
LGTM

PR #206, opened 2024-04-05
Explorers links

enriquesouza:


voltixApp/VoltixApp/Utils/Endpoint.swift
```diff
@@ -132,6 +132,34 @@ class Endpoint {
             return nil
         }
     }
+    
+    static func getExplorerByAddressURLByGroup(groupName: String, address: String) -> String? {
+        switch groupName {
```
flypaper0, line 137: is to possible to use Chain enum here?

flypaper0 (approved):

voltixApp/VoltixApp/Utils/Endpoint.swift
enriquesouza: Yes, I will use it.

PR #225, opened 2024-04-09
Hide main device when keygen/keysign

johnnyluo:
1. Hide main device when keygen/keysign , and add it to selection list by default
2. When keysign finished / failed retry , go back to vault detail view
3. Show 2/3 vault , 3/4 vault on keygen

### DIFF
--- a/voltixApp/VoltixApp/Views/Keysign/ParticipantDiscovery.swift
+++ b/voltixApp/VoltixApp/Views/Keysign/ParticipantDiscovery.swift
@@ -17,7 +17,7 @@ class ParticipantDiscovery: ObservableObject {
         self.discoverying = false
     }
     
-    func getParticipants(serverAddr: String, sessionID: String) {
+    func getParticipants(serverAddr: String, sessionID: String, localParty: String) {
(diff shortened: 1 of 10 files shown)


voltixApp/VoltixApp/Views/Keysign/ParticipantDiscovery.swift
```diff
@@ -33,6 +33,9 @@ class ParticipantDiscovery: ObservableObject {
                             let peers = try decoder.decode([String].self, from: data)
                             DispatchQueue.main.async {
                                 for peer in peers {
```
flypaper0, line 35: ```for peer in peers where peer != localParty {```

enriquesouza (approved):
LGTM

voltixApp/VoltixApp/Views/Keysign/ParticipantDiscovery.swift
enriquesouza: I believe it is faster without filtering since you run the for only once, if you filter like this you run a for to remove the unwanted then another for to run on wanted. This way with continue you run only once and skip unwanted.

PR #253, opened 2024-04-13
prices failover, if coin gecko fails, we fetch it from coin paprika

enriquesouza:
The price was sometimes failing and returning ZERO, so if we reach the rate limit, or any error occurs we try to fetch it from coin paprika.


voltixApp/VoltixApp/Stores/Price/coinpaprika_coins.json
johnnyluo, line 1: I didn't see this file get used anywhere , why we need this one?

johnnyluo (approved):
LGTM , one question

voltixApp/VoltixApp/Stores/Price/coinpaprika_coins.json
enriquesouza: This one contains all coins from coin paprika. In case we don't find a coin in the other list this one is a sure. I will remove and add it once we need it.

PR #257, opened 2024-04-15
Update Keygen & Keysign to support Premium

johnnyluo:


enriquesouza (approved):

voltixApp/VoltixApp/View Models/KeysignViewModel.swift
```diff
@@ -86,10 +86,19 @@ class KeysignViewModel: ObservableObject {
         for msg in self.messsageToSign {
             logger.info("signing message:\(msg)")
             let msgHash = Utils.getMessageBodyHash(msg: msg)
+            var pubkey = ""
```
flypaper0, line 89:
```suggestion
            let pubkey: String
```

johnnyluo: PubKey will be changed depends on which keysign type is , but I can move this logic out of the loop though

Issue #295, opened 2024-04-18
[BUG] Access Dictionary from multiple thread cause the app to crash sometimes


johnnyluo:
Currently when we open a vault , it load crypto balance for all the coins at the same time ,  however a lot places we use dictionary to cache data , and dictionary is not threadsafe , and not safe to be accessed from multiple thread

johnnyluo:
This has been fixed

PR #297, opened 2024-04-18
Add Threadsafe dictionary , and update blockchair service to use it

johnnyluo:
1. Add a threadsafe dictionary , and update blockchair service to use it , to avoid random crash caused by accessing dict from multiple thread
2. Update Homeview a little bit , remove 'selectedVault' from View , which is not need


flypaper0 (approved):

enriquesouza (approved):

johnnyluo:
#295

PR #312, opened 2024-04-22
Pulse Chain

enriquesouza:
I've added Pulse chain, but I saw WC does not support it, therefore we can refractory the code to add it via Chain ID not relying on WalletCore.CoinType.


enriquesouza:
It is not supported by Wallet Core

PR #326, opened 2024-04-25
issues/274

enriquesouza:
Now it validates the max and does not allow you to send if you don't have enough fees to cover the costs. Also, it will show the error messages, so you can see you don't have enough funds to send.

https://github.com/voltix-vault/voltix-ios/issues/274


johnnyluo (approved):
LGTM

PR #327, opened 2024-04-25
issues/284: amounts validations

enriquesouza:
https://github.com/voltix-vault/voltix-ios/issues/284

To avoid errors while sending we are now validating the full transaction cost before moving to review. It should prevent issues with non broadcasted transactions.


johnnyluo:
Is this one still relevant? it seems you already include all the changes in  #328

enriquesouza:
Already merged.

PR #328, opened 2024-04-25
Issues/273 - QrCode Scan issues

enriquesouza:
https://github.com/voltix-vault/voltix-ios/issues/273


voltixApp/VoltixApp/View Models/SendCryptoViewModel.swift
```diff
@@ -224,48 +224,13 @@ class SendCryptoViewModel: ObservableObject, TransferViewModel {
...
         }
         
-        let coinName = tx.coin.chain.name.lowercased()
-        let key: String = "\(tx.fromAddress)-\(coinName)"
-        
-        if  tx.coin.chain.chainType == ChainType.UTXO {
-            let walletBalanceInSats = utxo.blockchairData.get(key)?.address?.balance ?? 0
-            let totalTransactionCostInSats = tx.amountInSats + BigInt(tx.feeInSats)
-            print("Total transaction cost: \(totalTransactionCostInSats)")
+        if tx.isAmountExceeded && tx.coin.isNativeToken {
```
johnnyluo, line 227: a few questions about this:
1. What happens if the tx.coin.isNativeToken is false, for example ,ERC20/BEP20 etc , do we also check their send amount is larger than balance
2.  What if user trying to send ERC20 , but they don't have ETH available?

enriquesouza: 1 - ERC20 uses Native Tokens (Coins) to pay the fees. So they always can send the full amount of ERC20 Token.
2 - Then we will have an error of not enough fees. We will need to raise another PR to show the error in the tx complete screen.

enriquesouza: 1 - ERC20 uses Native Tokens (Coins) to pay the fees. So they always can send the full amount of ERC20 Token.
2 - We are now validating if they have native tokens to pay the fees

voltixApp/VoltixApp/View Models/SendTransaction.swift
```diff
@@ -37,21 +37,75 @@ class SendTransaction: ObservableObject, Hashable {
...
+        let totalTransactionCost = amountInRaw + gasInt
+        
+        return totalTransactionCost > totalBalance
+        
+    }
+    
+    var hasEnoughNativeTokensToPayTheFees: Bool {
+        var gasInt = BigInt(gas) ?? BigInt.zero
+        
+        if coin.chainType == .EVM {
```
johnnyluo, line 64: This is not correct , many blockchains have multiple tokens , not only EVM.

For example solana has many tokens.  mayachain has both CACAO and Maya token

you should take the chain from the coin they are sending , and then find out whether they have any native token on that blockchain

enriquesouza: Yes, I can check a nonnative and get the native . It will work for all

voltixApp/VoltixApp/View Models/SendTransaction.swift
```diff
@@ -37,21 +37,90 @@ class SendTransaction: ObservableObject, Hashable {
...
+        
+    func hasEnoughNativeTokensToPayTheFees() async -> Bool {
+        guard !coin.isNativeToken else { return true } 
+
+        var gasPriceBigInt = BigInt(gas) ?? BigInt.zero
+        if let gasLimitBigInt = BigInt(coin.feeDefault) {
+            if coin.chainType == .EVM {
+                gasPriceBigInt *= gasLimitBigInt
+            } 
+            if let vault = ApplicationState.shared.currentVault {
```
johnnyluo, line 69: We don't need to do a RPC call to check the balance , balance has been checked in the vault detail view,  so just remove the RPC call , and check the balance field in the coin should be enough.

enriquesouza: The balance is empty because we get it from the vault store and the vault store has an empty balance. We call the get balance for the native coin only if it has a cache then it will return from the cache.

johnnyluo (approved):
LGTM

Issue #311, opened 2024-04-21
[Improvement] Add Vault Pubkey to Keysign QRCode


johnnyluo:
At the moment , when doing keysign , the QRCode doesn't have the vault's pubkey in it ,  thus if user choose the wrong vault on the join device , keysign will continue but it will fail of course , but that's not very user friendly.

Let's add vault's pubkey to the QRCode , and when join device scan the barcode , make sure it is the correct vault , if it is not , show the following screen

<img width="283" alt="image" src="https://github.com/voltix-vault/voltix/assets/749608/b73124d2-5621-424e-955b-a64862af4546">

PR #334, opened 2024-04-26
Keysign qr code

Rockindash:
Add Vault Pubkey to Keysign QRCode, Issue #311


voltixApp/VoltixApp/View Models/SendCryptoViewModel.swift
```diff
@@ -69,7 +69,8 @@ class SendCryptoViewModel: ObservableObject, TransferViewModel {
             chainSpecific: BlockChainSpecific.UTXO(byteFee: BigInt(tx.feeInSats)),
             utxos: utxoInfo,
             memo: tx.memo,
-            swapPayload: nil
+            swapPayload: nil, 
+            vaultPubKeyECDSA: vault.pubKeyECDSA
         )
         
         if let vault = ApplicationState.shared.currentVault {
```
johnnyluo, line 76: since we pass `vault` into this method now , so we don't need to take vault from ApplicationState.shared.currentVault anymore , could you please update this as well?

johnnyluo (approved):
LGTM

PR #361, opened 2024-05-02
Fix the issue with KeygenVerify , also add a test to ensure encryptio…

johnnyluo:
…n/description work across both IOS and android


enriquesouza (approved):
LGTM

PR #362, opened 2024-05-02
[Registry] CoinActions

flypaper0:


voltixApp/VoltixApp/Services/Actions/ChainAction.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+//  Created by Artur Guseinov on 01.05.2024.
+//
+
+import SwiftUI
+
+enum CoinAction: String, Codable {
+    case send
+    case swap
+    case bond
```
johnnyluo, line 13: I think `bond`, `unbond`,`leave` and `custom` all belongs to `deposit`
when user click `deposit` , then they will have four options , 
- bond , you can bond a node / add a bond provider 
- unbond , you need to put in node address, and amount to unbond 
- leave, you need to put in a node address

enriquesouza (approved):

PR #367, opened 2024-05-02
issue 366

enriquesouza:
Auto-select the device if only one is selected.

https://github.com/voltix-vault/voltix-ios/issues/366


voltixApp/VoltixApp/Views/Keygen/PeerDiscoveryView.swift
```diff
@@ -188,6 +188,11 @@ struct PeerDiscoveryView: View {
             } label: {
                 PeerCell(id: peer, isSelected: viewModel.selections.contains(peer))
             }
+            .onAppear {
```
johnnyluo, line 191: The issue you linked , it was talking during keysign  , the change you are making is keygen

enriquesouza: I will apply it to both. It makes sense.

johnnyluo (approved):
LGTM

PR #370, opened 2024-05-02
337 various fixes

enriquesouza:
A few more small fixes from the issue 337

### DIFF
--- a/voltixApp/VoltixApp/Views/Vault/EditVaultView.swift
+++ b/voltixApp/VoltixApp/Views/Vault/EditVaultView.swift
@@ -67,6 +67,7 @@ struct EditVaultView: View {
     var view: some View {
         ScrollView {
             VStack(spacing: 16) {
+                deviceName
                 vaultDetails
                 backupVault
                 editVault
@@ -74,6 +75,10 @@ struct EditVaultView: View {
                 deleteVault
             }
         }
+    }
+    
+    var deviceName: some View {
+        Text(Utils.getLocalDeviceIdentity()).padding(.top, 30)
     }
     
     var vaultDetails: some View {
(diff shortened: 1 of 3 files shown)


voltixApp/VoltixApp/Views/Vault/VaultPairDetailView.swift
```diff
@@ -55,7 +55,11 @@ struct VaultPairDetailView: View {
                     .frame(maxWidth: .infinity, alignment: .center)
                 
                 ForEach(devicesInfo, id: \.Index) { device in
-                    VaultPairDetailCell(title: device.Signer, description: .empty)
+                    if device.Signer == Utils.getLocalDeviceIdentity() {
```
johnnyluo, line 58: This is not correct, instead you should compare `device.Signer` , `vault.localPartyID`

voltixApp/VoltixApp/Views/Vault/EditVaultView.swift
```diff
@@ -76,6 +77,10 @@ struct EditVaultView: View {
         }
     }
     
+    var deviceName: some View {
+        Text(Utils.getLocalDeviceIdentity()).padding(.top, 30)
```
johnnyluo, line 81: Here as well , you should not call `Utils.getLocalDeviceIdentity()` after the vault is already created , use `vault.localPartyID` instead

voltixApp/VoltixApp/Views/Vault/VaultPairDetailView.swift
enriquesouza: Changing now

voltixApp/VoltixApp/Views/Vault/EditVaultView.swift
enriquesouza: Changing now

johnnyluo (approved):
LGTM

PR #390, opened 2024-05-04
Set max amount when sending for UTXO.

enriquesouza:
This PR has the purpose of fixing the issue with MAX amounts on UTXO, there was an disabled flag on helper that was not allowing it to work properly, now it is passed via parameter. 

It will only work if you hit the button send MAX, otherwise it will validate normally.

### DIFF
--- a/voltixApp/VoltixApp/Services/RpcService.swift
+++ b/voltixApp/VoltixApp/Services/RpcService.swift
@@ -50,7 +50,7 @@ class RpcService {
             
             if let error = response["error"] as? [String: Any], let message = error["message"] as? String {
                 
-                if message.contains("known") || message.contains("already known") || message.contains("Transaction is temporarily banned") {
+                if message.contains("known") || message.contains("already known") || message.contains("Transaction is temporarily banned") || message.contains("nonce too low: next nonce"){
                     return try decode("Transaction already broadcasted.")
                 }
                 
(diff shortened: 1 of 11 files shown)


voltixApp/VoltixApp/Services/RpcService.swift
```diff
@@ -50,7 +50,7 @@ class RpcService {
             
             if let error = response["error"] as? [String: Any], let message = error["message"] as? String {
                 
-                if message.contains("known") || message.contains("already known") || message.contains("Transaction is temporarily banned") {
+                if message.contains("known") || message.contains("already known") || message.contains("Transaction is temporarily banned") || message.contains("nonce too low: next nonce"){
```
johnnyluo, line 53: `nonce too low`, this one it depends , it could be the transaction already broadcast successfully and committed to block , or it could be we didn't get the latest nonce when we create the transaction

Are you sure we should treat as success?

johnnyluo (approved):
LGTM

voltixApp/VoltixApp/Services/RpcService.swift
enriquesouza: We always get the latest nonce before sending, so I believe this error will only happen if it tries to broadcast the same transaction 2x. The full message is ```nonce too low: next nonce 3 of 4```.

PR #420, opened 2024-05-11
Total balance double count issue fixed

Rockindash:
Total balance double count issue for Vault balance fixed, Issue # 401


enriquesouza (approved):
LGTM

PR #455, opened 2024-05-15
Deposits & Transaction Memos

enriquesouza:
Creating the Transaction Memos for Thor Chain and Deposits. Using as reference the TC documentation:

https://dev.thorchain.org/concepts/memos.html

We should be able to create memos easily to perform actions on ThorChain using Vultisg Wallet.


VultisigApp/VultisigApp/Views/TransactionMemo/TransactionMemoVerifyView.swift
```diff
@@ -0,0 +1,162 @@
...
+            
+            Separator()
+            getDetailsCell(for: "gas", with: tx.gasInReadable)
+        }
+        .padding(16)
+        .background(Color.blue600)
+        .cornerRadius(10)
+    }
+    
+    //    var checkboxes: some View {
```
johnnyluo, line 83: remove this

johnnyluo (approved):
LGTM

PR #473, opened 2024-05-18
Refresh balance

enriquesouza:


VultisigApp/VultisigApp/Views/Vault/ChainDetailView.swift
```diff
@@ -1,22 +1,16 @@
...
     let vault: Vault
-    let balanceInFiat: String?
+    @State var balanceInFiat: String?
     
     @State var showSheet = false
     @State var tokens: [Coin] = []
     @State var actions: [CoinAction] = []
     @State var isLoading = false
     @StateObject var sendTx = SendTransaction()
+    @State var coinViewModels: [String: CoinViewModel] = [:]
```
johnnyluo, line 13: I don't think we should store a map of viewmodels here , given we already have a list of coins , can we extend the service to calculate it directly?

Rockindash (approved):

PR #481, opened 2024-05-20
[Swap] Swaps bug fixes

flypaper0:


VultisigApp/VultisigApp/Views/Keysign/KeysignSwapConfirmView.swift
```diff
@@ -35,6 +28,8 @@ struct KeysignSwapConfirmView: View {
 
     var summary: some View {
         VStack(spacing: 16) {
+            getValueCell(for: "Action", with: getAction())
```
johnnyluo, line 31: Can you please specify the swap provider here? like THORChain or 1Inch or else

VultisigApp/VultisigApp/View Models/KeysignViewModel.swift
```diff
@@ -88,21 +88,23 @@ class KeysignViewModel: ObservableObject {
 
     func startKeysign() async {
         defer {
-            self.messagePuller?.stop()
+            messagePuller?.stop()
         }
-        for msg in self.messsageToSign {
+        for msg in messsageToSign {
             do {
+                try await verifyAllowance()
```
johnnyluo, line 95: Could you please do this check before keysign? Like in confirm screen? 

Because if we check here , and the approval is not committed ,then user might need to refill the form again

VultisigApp/VultisigApp/Chains/THORChainSwaps.swift
```diff
@@ -8,6 +8,17 @@ import Tss
 import WalletCore
 
 class THORChainSwaps {
+
+    static var affiliateFeeRateBp: String {
+#if DEBUG
+        return "0"
+#else
+        return "70"
```
johnnyluo, line 16: Can you please update this to `50`?

VultisigApp/VultisigApp/View Models/KeysignViewModel.swift
flypaper0: I we'll check this before keysing that check will always failed because transaction just sent

johnnyluo (approved):
LGTM

Issue #496, opened 2024-05-22
Updating UI on main page


realpaaao:
1. Please update the UI of main page to be concruent with figma design so we only have one settings button top left and one re-arrange vault on the top right. 
2. Add Send and Receive Buttons to be cuncruent with figma
3. Ammend settings page so it is also congruent.

![image](https://github.com/vultisig/vultisig-ios/assets/167348323/d8e775ef-7ca4-4fa9-a73b-0472434b4e1b)

PR #508, opened 2024-05-23
Main view UI update

Rockindash:
Main View UI updates as per new design & edit vault option moved to settings, Issue #496


johnnyluo (approved):
LGTM

PR #529, opened 2024-05-26
[Swap] Do not include affiliate if amount below 100$

flypaper0:


VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
```diff
@@ -33,7 +33,8 @@ struct OneInchService {
             from: from,
             slippage: "0.5",
             referrer: referrerAddress,
-            fee: 0.7
+            fee: 0.7,
```
johnnyluo, line 36: Can you please change the fee to `0.5`

johnnyluo (changes requested):
You will also need to make a change in `THORChainSwaps.swift`  https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Chains/THORChainSwaps.swift#L45

johnnyluo (approved):
LGTM

PR #539, opened 2024-05-27
Case added for Mac devices using the iPad app

Rockindash:
"Mac" added as Prefix added for Macs running iPad/iPhone app


johnnyluo (approved):
LGTM

PR #543, opened 2024-05-28
Home actions edit

Rockindash:
Added Swap action to Home View in case even a single selected Coin supports Swap.

### DIFF
--- a/VultisigApp/VultisigApp/View Models/VaultDetailViewModel.swift
+++ b/VultisigApp/VultisigApp/View Models/VaultDetailViewModel.swift
@@ -11,6 +11,7 @@ import SwiftUI
 @MainActor
 class VaultDetailViewModel: ObservableObject {
     @Published var coinsGroupedByChains = [GroupedChain]()
+    @Published var selectedGroup: GroupedChain? = nil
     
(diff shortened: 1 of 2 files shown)


VultisigApp/VultisigApp/View Models/VaultDetailViewModel.swift
```diff
@@ -11,6 +11,7 @@ import SwiftUI
 @MainActor
 class VaultDetailViewModel: ObservableObject {
     @Published var coinsGroupedByChains = [GroupedChain]()
+    @Published var selectedGroup: GroupedChain? = nil
```
flypaper0, line 14: why we need published property? 

I think we can just have one computed property, that return array of actions

Rockindash: Its there to fetch the Chain with both Send and Swap option. Just to keep the preset Coin consistent for both Send and Swap view.
Also, had to set this as a Published property since sometimes it'll try to search for a chain before the Categorization logic is finished and wont show the actions at all. Its there to rerender the component after that sorting it complete.

johnnyluo (approved):
LGTM

PR #548, opened 2024-05-28
thorchain dynamic fee

enriquesouza:
https://github.com/vultisig/vultisig-ios/issues/483

### DIFF
--- a/VultisigApp/VultisigApp/Chains/thorchain.swift
+++ b/VultisigApp/VultisigApp/Chains/thorchain.swift
@@ -8,7 +8,41 @@ import Tss
 import WalletCore
 
 enum THORChainHelper {
-    static let THORChainGas: UInt64 = 2000000
+    static let THORChainGas: UInt64 = getTHORChainGasPrice()
+    
+    private static var cachedTHORChainGas: UInt64?
+    
+    private static func fetchTHORChainGas(completion: @escaping (UInt64) -> Void) {
+        if let gas = cachedTHORChainGas {
+            completion(gas)
+        } else {
+            Task {
+                do {
+                    let feePrice = try await ThorchainService.shared.fetchFeePrice()
+                    cachedTHORChainGas = feePrice
+                    completion(feePrice)
+                } catch {
+                    print("Failed to fetch THORChain gas price: \(error)")
+                    completion(0) // or any default value you see fit
+                }
+            }
+        }
+    }
+    
+    static func getTHORChainGasPrice() -> UInt64 {
+        let semaphore = DispatchSemaphore(value: 0)
+        var gasPrice: UInt64 = 0
+        
+        fetchTHORChainGas { price in
+            gasPrice = price
+            semaphore.signal()
+        }
+        
+        _ = semaphore.wait(timeout: .distantFuture)
+        return gasPrice
+    }
+    
+    
     static func getRUNECoin(hexPubKey: String, hexChainCode: String) -> Result<Coin, Error> {
         let derivePubKey = PublicKeyHelper.getDerivedPubKey(hexPubKey: hexPubKey,
                                                             hexChainCode: hexChainCode,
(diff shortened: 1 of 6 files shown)


VultisigApp/VultisigApp/Chains/thorchain.swift
```diff
@@ -8,7 +8,41 @@ import Tss
 import WalletCore
 
 enum THORChainHelper {
-    static let THORChainGas: UInt64 = 2000000
+    static let THORChainGas: UInt64 = getTHORChainGasPrice()
+    
+    private static var cachedTHORChainGas: UInt64?
+    
+    private static func fetchTHORChainGas(completion: @escaping (UInt64) -> Void) {
```
johnnyluo, line 15: This logic doesn't belongs to THORChainHelper , it should be in THORChainService, like all the other chains

VultisigApp/VultisigApp/Chains/thorchain.swift
```diff
@@ -47,12 +81,11 @@ enum THORChainHelper {
         input.accountNumber = accountNumber
         input.sequence = sequence
         input.mode = .sync
-        // THORChain fee is 0.02 RUNE
         input.fee = CosmosFee.with {
-            $0.gas = 20000000
+            $0.gas = THORChainHelper.THORChainGas
```
johnnyluo, line 85: gas will not be changed , actually thorchain doesn't have gas , we can continue to hard code it

VultisigApp/VultisigApp/Chains/thorchain.swift
```diff
@@ -47,12 +81,11 @@ enum THORChainHelper {
...
             $0.amounts = [CosmosAmount.with {
                 $0.denom = "rune"
-                $0.amount = "2000000"
+                $0.amount = THORChainHelper.THORChainGas.description
```
johnnyluo, line 88: This amount should be put into `KeysignPayload.BlockSpecific.THORChain` , otherwise , in some case , main device and pair device will not agree with each other , this will cause tx can't be signed

johnnyluo (changes requested):

VultisigApp/VultisigApp/Chains/thorchain.swift
enriquesouza: Yes, I am moving it to Thorchain Service.

VultisigApp/VultisigApp/Chains/thorchain.swift
enriquesouza: Hardcoded.

VultisigApp/VultisigApp/Chains/thorchain.swift
enriquesouza: Moved to ```KeysignPayload.BlockSpecific.THORChain```

VultisigApp/VultisigApp/Chains/thorchain.swift
```diff
@@ -8,7 +8,8 @@ import Tss
 import WalletCore
 
 enum THORChainHelper {
-    static let THORChainGas: UInt64 = 2000000
+    static let THORChainGas: UInt64 = ThorchainService.shared.fetchFeePrice()
```
johnnyluo, line 11: You probably shouldn't have this one

VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
```diff
@@ -75,4 +77,49 @@ class ThorchainService {
...
+            self.cacheFeePrice[cacheKey] = (data: resultNumber, timestamp: Date())
+            return resultNumber
+        } else {
+            print("JSON decoding error")
+        }
+        
+        return .zero
+    }
+    
+    private func fetchFeePrice(completion: @escaping (UInt64) -> Void) {
```
johnnyluo, line 101: Can you please remove this function , and also `func fetchFeePrice` below as well 
This is not good design , task in a callback function , and also semaphore..

johnnyluo (changes requested):

enriquesouza (commented):
fixed

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
```diff
@@ -75,4 +77,24 @@ class ThorchainService {
         
         return cacheEntry.balances
     }
+    
+    func fetchFeePrice() async throws -> UInt64 {
+        let cacheKey = "thorchain-fee-price"
+        if let cachedData: UInt64 = await Utils.getCachedData(cacheKey: cacheKey, cache: cacheFeePrice, timeInSeconds: 60*5) {
```
flypaper0, line 83: we can use URLRequest cache for that kind of things. 
https://developer.apple.com/documentation/foundation/nsurlrequest/cachepolicy

flypaper0 (approved):

PR #561, opened 2024-05-29
bch prefix removal

enriquesouza:


VultisigApp/VultisigApp/Chains/UTXOChainsHelper.swift
```diff
@@ -69,8 +69,13 @@ class UTXOChainsHelper {
         else {
             return .failure(HelperError.runtimeError("public key \(derivedPubKey) is invalid"))
         }
+        let address = coin.deriveAddressFromPublicKey(publicKey: publicKey)
```
flypaper0, line 72: do we need to add prefix again when we setting address into WalletCore tx input?

enriquesouza: No, it works with prefixes and without the prefix for sending and receiving.

johnnyluo (approved):
LGTM

PR #562, opened 2024-05-29
Sum base fee + priority fee

enriquesouza:


VultisigApp/VultisigApp/Services/RpcEvmService.swift
```diff
@@ -37,9 +37,12 @@ class RpcEvmService: RpcService {
...
         
-        let adjustedGasPrice = BigInt(Double(try await gasPrice) * multiplier)
-            
-        return (adjustedGasPrice, try await priorityFee, Int64(try await nonce))
+        let gasPriceValue = try await gasPrice
+        let priorityFeeValue = try await priorityFee
+        
+        let adjustedGasPrice = BigInt(Double(gasPriceValue + priorityFeeValue) * multiplier)
+        
+        return (adjustedGasPrice, priorityFeeValue, Int64(try await nonce))
```
johnnyluo, line 45: This way , isn't we are going to pay (basefee + priority fee) * 1.5 + priority fee?

enriquesouza: Yes, we will. You are right, We need to add it only to display because we already set it in the input. So it is the tip for the miners.

johnnyluo (approved):
LGTM

Rockindash (approved):

Issue #549, opened 2024-05-28
[BUG] Hide `Gas(auto)` from send crypto screen


johnnyluo:
**Describe the bug**
Currently send polkadot screen , it show `Gas(Auto)` as 1 DOT , which is not correct
When send Dot transaction , we don't need to estimate gas fee , the blockchain itself will figure it out and deduct it automatically , so we can hide `Gas(Auto)` field , otherwise it confuse user.

Also DOT blockchain will purge accounts that have less than 1 DOT , so we need to remind user if they left less than 1 DOT in their account , they could lose it.

enriquesouza:
I will leave a fee of 0.025 which is the average charged by DOT in 2023
https://polkadot.network/blog/polkadot_q4_update_data

For Existential Deposit I followed their docs
https://support.polkadot.network/support/solutions/articles/65000168651-what-is-the-existential-deposit-

PR #568, opened 2024-05-30
Existential deposit messages

enriquesouza:
This should resolve the POLKA dot issue for ED and fees.

https://github.com/vultisig/vultisig-ios/issues/549


Rockindash (approved):
LGTM

Issue #566, opened 2024-05-29
[ADD] Check for input on deposit view before continue


realpaaao:
**Is your feature request related to a problem? Please describe.**
At the moment it is possible to continue on `message deposit view` without or incorrect input.

**Describe the solution you'd like**
Check for following input in text fields:
- [x] Amount: only positive, on Mac negative numbers are possible.
- [x] Node address: only viable address format for respective chain
- [x] Provider: having description `optional` as a placeholder and checking for correct address format
- [x] Fee: having description `optional` as a placeholder and only positive numbers

if not, prompt user error message to complete missing or bad input.

enriquesouza:
Regarding the amount field, it is highly unlikely that users will enter a negative number for the bond, especially since we already use a numeric keypad to input the amount. Therefore, this scenario should not occur. Moreover, implementing this kind of validation would significantly increase the complexity of the deposit process due to its dynamic nature.

The node address is already being validated for Thor and Maya.

PR #571, opened 2024-05-30
Check for input on deposit view before continue #566

enriquesouza:
https://github.com/vultisig/vultisig-ios/issues/566

This should add optional fields for deposits fields


johnnyluo (approved):
LGTM

PR #576, opened 2024-05-30
[Coin] Dynamic tokens support

flypaper0:


johnnyluo (dismissed):
LGTM

johnnyluo:
I wasn't able to choose token for Ethereum , but I can choose for BSC and AVAX , could you please test Ethereum and layer 2 as well

VultisigApp/VultisigApp/Views/Vault/TokenSelectionView.swift
```diff
@@ -28,31 +32,33 @@ struct TokenSelectionView: View {
                 NavigationBackSheetButton(showSheet: $showTokenSelectionSheet)
             }
         }
+        .task {
+            try? await tokenViewModel.loadData(chain: group.chain)
```
johnnyluo, line 36: don't use `try?` , please capture the error appropriately , it swallow the error

VultisigApp/VultisigApp/Services/1inch/OneInchToken.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import Foundation
+
+struct OneInchToken: Codable, Hashable {
+    let address: String
+    let symbol: String
+    let name: String
+    let decimals: Int
+    let logoURI: URL
```
johnnyluo, line 15: `logoURI` can be nil , so you need to make it optional

enriquesouza (commented):
I got this error a few times if we try to add multiple tokens. It is possibly a rate limit. 

```
nw_flow_add_write_request [C331 tokens-data.1inch.io:443 failed parent-flow (satisfied (Path is satisfied), interface: en0[802.11], uses wifi)] cannot accept write requests
```

enriquesouza:
When selecting Cacao, I can't find the MAYA token anymore. Maybe we should only check this search only for EVMs.

image

image

enriquesouza:
I already have Base added, like in this image, but when I try to add a new chain it does not recognize anymore it is already added. 
image
image

enriquesouza (changes requested):
I found a few bugs in this branch and left some comments. Please have a look.

enriquesouza (changes requested):

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/View Models/CoinSelectionViewModel.swift
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        return nil
+    }
+    private func addToChain(asset: Coin, to vault: Vault) async {
+        do{
+            if let newCoin = getNewCoin(asset: asset, vault: vault) {
+                // Fetch priceProviderId for EVM tokens
+                if !newCoin.isNativeToken, asset.chainType == .EVM {
+                    newCoin.priceProviderId = try await priceService.fetchCoingeckoId(
```
enriquesouza, line 197: failed to save coin to model context The data couldn’t be read because it isn’t in the correct format.
failed to save coin to model context The data couldn’t be read because it isn’t in the correct format.

We can add a thread sleep of 1s to add the tokens if N. It is probably hitting the limit.

enriquesouza (changes requested):
We still can't add coins. So maybe we can add a retry with 1s wait for each token.

enriquesouza (approved):
LGTM, just a small fix when selecting the token should go selected to the send page, but should not stop the merge.

Issue #436, opened 2024-05-13
ADD: Keysign/Keygen QR code auto detect


jpthor:
# Current

Use main vault camera icon on a key-gen QR code - fail, error message not helpful
Use keygen `JOIN` icon on a key-sign QR code - fail, error message not helpful

# Desired

QR code should have a flag encoded to flip between key-sign and key-gen appropriately

Rockindash:
Will instead implemented a universal scanner for the app since we already have a tag that states if the type is Keysign or Keygen. This way, no additional step is required by the users.

PR #577, opened 2024-05-31
Central scan view

Rockindash:
Universal scanner added to auto-detect both Keygen and Keysign, Issue #436


enriquesouza (changes requested):
I've made 3 tests to generate the keys and 2 crashed the app. I used the cam of the keysign in the first 2. In the end the app crashed. 

This is my third one. It didn't crash but it shows like no chain selected. 


image

enriquesouza (approved):
LGTM, and tested.

Issue #521, opened 2024-05-25
[BUG] Back button from vault detail view , after create a new vault , always go to the newly created vault


johnnyluo:
**Describe the bug**
Back button from chain detail view , doesn't go back to correct vault, after create a new vault

**To Reproduce**
1. Create a 2/2 vault, after the vault is created , the app automatically open the newly created vault , assume the vault name is `NewVault-1`
2. Change to another vault
3. Click a chain to go to the chain detail view
4. click "back" button , it go to `NewVault-1` , not the vault I choose at step 2

**Expected behavior**
"back" button go back to the correct vault

PR #578, opened 2024-05-31
Back button bug

Rockindash:
Back button bug where it always goes back to the previously redirected vault fixed, Issue #521

### DIFF
--- a/VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
+++ b/VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
@@ -26,15 +26,11 @@ struct VaultDetailView: View {
             view
             scanButton
         }
-        .onAppear {
+        .onFirstAppear {
             setData()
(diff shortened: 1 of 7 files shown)


VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
```diff
@@ -150,6 +146,8 @@ struct VaultDetailView: View {
         viewModel.setOrder()
         viewModel.updateBalance()
         viewModel.getGroupAsync(tokenSelectionViewModel)
+        viewModel.setDefaultCoins(for: vault)
+        appState.currentVault = vault
```
johnnyluo, line 150: `viewModel.setDefaultCoins` if put in here , it will be called multiple time , which will cause it to crash , you need to put it to onAppear, and only call it once

johnnyluo (approved):
LGTM

PR #579, opened 2024-05-31
Thor and Maya charging fees twice

enriquesouza:
we were charging twice the fees for both Maya and Thor. Also, the camera had some issues to key sign so i reverted the changes.

### DIFF
--- a/VultisigApp/VultisigApp/Chains/thorchain.swift
+++ b/VultisigApp/VultisigApp/Chains/thorchain.swift
@@ -48,10 +48,6 @@ enum THORChainHelper {
         input.mode = .sync
         input.fee = CosmosFee.with {
             $0.gas = 20000000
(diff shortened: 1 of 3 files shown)


VultisigApp/VultisigApp/Chains/thorchain.swift
```diff
@@ -137,13 +133,11 @@ enum THORChainHelper {
...
             $0.fee = CosmosFee.with {
                 $0.gas = 20000000
-                $0.amounts = [CosmosAmount.with {
-                    $0.denom = "rune"
-                    $0.amount = fee.description
-                }]
             }
         }
         
+        print(input.debugDescription)
```
Rockindash, line 139: Print statement can be removed.

enriquesouza: Tks, Removed

Issue #565, opened 2024-05-29
[BUG] crash when using exisitng vault name in vault creation


realpaaao:
**Describe the bug**
When creating a new vault with a name that already exists, the initiating device crashes while the pairing device creates the vault.

**To Reproduce**
1. Create vault with existing name
2. do keygen
3. app crashes

**Expected behavior**
Check on Vault creation in `RenameVaultView` if vault name already exists, if prompt user an error message to change vault name.

johnnyluo:
Let's make sure the vault name is unique

realpaaao:
@Rockindash solved this issue with 1890dedda4fbc3b3b32415785ccfecd4991a889e . Thanks!

PR #582, opened 2024-06-01
Unique vault name

Rockindash:
Check for unique vault's name implemented, will prevent more than one Vault to have the same name, Issue #565


johnnyluo (approved):
LGTM

PR #593, opened 2024-06-02
Move set default coins for vault to keygen and import vault

johnnyluo:
Currently after a vault get created or when delete a vault , it often cause the app to crash , because the logic in vaultdetailview depends on the change on vault , and apply set default coin for vault twice. 

This PR move the logic to set vault's default coin to keygen success , or import vault success stage , which ensure it only apply once

This also means user can remove all the chains if they want


enriquesouza (approved):
LGTM and tested.

Issue #400, opened 2024-05-05
Reordering Chains is not permanent 


SamYap0:
**Describe the bug**
After reordering chains (e.g move THORChain to the top, on top of BTC), once we move to another screen and back, it reverts to default ordering.

**Expected behavior**
Ordered chains to remain.

**Smartphone (please complete the following information):**
 - Device: Both iPhone 11 Pro & 14
 - OS: 17.4.1
 - Version 24

realpaaao:
@Rockindash would it be possible to also implement this request [https://github.com/vultisig/vultisig-ios/issues/449](url) ?
As than we could remove the re-arrange button on the right on main to be concruent with figma design.

johnnyluo:
Instead of allow user to manually order the tokens, let's do this
1. Order by token balance , descending

PR #609, opened 2024-06-05
Chain sort

Rockindash:
Chains on Vault Detail View are being sorted based on total balance, Issue #400

### DIFF
--- a/VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
+++ b/VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
@@ -85,7 +85,7 @@ struct VaultDetailView: View {
     var list: some View {
         List {
             ForEach(viewModel.coinsGroupedByChains.sorted(by: {
-                $0.totalBalanceInFiatString > $1.totalBalanceInFiatString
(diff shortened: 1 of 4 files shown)


VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
```diff
@@ -86,15 +85,13 @@ struct VaultDetailView: View {
     var list: some View {
         List {
             ForEach(viewModel.coinsGroupedByChains.sorted(by: {
-                $0.order < $1.order
+                $0.totalBalanceInFiatString > $1.totalBalanceInFiatString
```
johnnyluo, line 88: I don't think sort by string is correct , because 2 will be larger than 10, is that right?

Rockindash: Oh no, this is the other way around. The sorting order is in descending order. Chain with higher balance will be on the top.

Rockindash: Let me fix that.

johnnyluo (approved):
LGTM

Issue #515, opened 2024-05-24
ADD: Swipe down for balance refresh on main page


realpaaao:
To improve UX have posibility to swipe down on main page and refresh balances.

PR #621, opened 2024-06-06
Swipe refresh

Rockindash:
Pull to refresh balance added to Home View, Issue #515


johnnyluo:
When I checkout this branch , and trying to pull and refresh , it changed to show first vault's chains

johnnyluo (approved):
LGTM

PR #647, opened 2024-06-09
[Swaps] Mayachain support 

flypaper0:


VultisigApp/VultisigApp/Services/SwapService/SwapService.swift
```diff
@@ -11,15 +11,42 @@ struct SwapService {
...
+        {
+            return try await fetchOneInchQuote(
+                chain: fromChainID,
+                amount: amount, fromCoin: fromCoin,
+                toCoin: toCoin, isAffiliate: isAffiliate
+            )
+        }
+
+        // Mayachain resolver
+        if fromCoin.chain == .mayaChain || toCoin.chain == .mayaChain {
```
johnnyluo, line 32: I don't think this check is correct , on MayaChain , you can swap from RUNE<->CACAO, but you can also swap Between KUJI<-> Dash , if you trying to swap between KUJI <-> DASH , this will not able to get a quote

VultisigApp/VultisigApp/Services/SwapService/SwapService.swift
```diff
@@ -39,9 +66,15 @@ private extension SwapService {
         }
     }
 
-    func fetchThorchainQuote(amount: Decimal, fromCoin: Coin, toCoin: Coin, isAffiliate: Bool) async throws -> SwapQuote {
+    func fetchThorchainQuote(
```
johnnyluo, line 69: maybe make it more generic name , `fetchCrossChainQuote` ?

VultisigApp/VultisigApp/Services/SwapService/SwapService.swift
flypaper0: We'll toggle KUJI<-> Dash swaps in next PR. I think PR is already big. I'm still not sure that it's working need to do separate testing and may be fix something. In current PR is only RUNE->CACAO  CACAO->RUNE

enriquesouza (approved):

PR #677, opened 2024-06-13
Search ux

enriquesouza:
Some improvements on search UX


Rockindash:
Couple of comments regarding the UI:
Simulator Screenshot - iPhone 15 Pro - 2024-06-13 at 11 29 31

- Search bar can be moved to 2nd row. (Will be less cramped)
- Done button can be removed since back button shares the same functionality.
- Instead of search bar title can be used i.e. "Choose Tokens".
- S in search can be capitalized.

enriquesouza:
image
Changed it to Save and fixed the Search capitalization.

enriquesouza:
This is what we have on main branch.

image

I've added a 'Save' button because the current 'Cancel' button is confusing.

Here's the issue:
1. To save your selections, you currently have to cancel the search, which shows the back button.
2. This action erases everything you've selected, making it unclear whether your selections were saved or not.
3. You end up clicking 'Back', but it's unclear what actions this should trigger.

The new 'Save' button directly saves your selections, eliminating this confusion and ensuring users know their choices are saved.

johnnyluo (approved):
LGTM

PR #684, opened 2024-06-14
Replace walletcore to local SPM package

johnnyluo:


enriquesouza (approved):
We need to make sure we merge the latest from WC, because they are always fixing and adding new tokens.

PR #689, opened 2024-06-14
Arb requires a higher gas limit

enriquesouza:
It seems to be a common issue on Arbitrum. It requires a much higher gas limit. 

https://ethereum.stackexchange.com/questions/142867/gas-too-low-for-transaction-arbitrum


johnnyluo (approved):
LGTM

Issue #405, opened 2024-05-07
[Feature] Add support for macOS natively


Rockindash:
Add macOS native support to the app. Remove the current "Designed for iPad" build and replace it with macOS target.

**Issue**
Currently TrustWallet Core library does not support macOS platform. So this option has been used as an alterantive.

johnnyluo:
Closing , the first version of MacOS is out

PR #702, opened 2024-06-17
Mac support/UI support

Rockindash:
- If checks added for iOS to compile the project for native macOS platform, Issue #405
- Onboarding view adapted for macOS.

*Please note that as of now the build does not work for macOS.*
Main objective here was to adapt the WalletCore framework for macOS and compile the project.


johnnyluo (approved):
LGTM

Issue #703, opened 2024-06-17
[BUG] When adding some chains app crashes


realpaaao:
**Describe the bug**
On the newest update with my iPad the app crashes when I want to select MAYA Chain on one specific vault. Other vaults work fine.
Another user reported that crash when selecting BTC or CRO.
So far could not reproduce with any other vault.

See the crash log:
[TestFlight feedback (1).zip](https://github.com/user-attachments/files/15873338/TestFlight.feedback.1.zip)

**Expected behavior**
no crash on selecting chain

**Smartphone (please complete the following information):**
 - Device: iPad
 - OS: 17.5.1
 - Version 31

realpaaao:
Also happens on Mac when importing a vault and adding BTC. doens't seem to happen for one specific chain or vault...

realpaaao:
Could also recreate on iPhone with creating new vault and enabling CRO.

https://github.com/vultisig/vultisig-ios/assets/167348323/8b1e52e7-8a24-4a44-873e-909e190156c9

realpaaao:
Also on iPad with BTC chain

PR #721, opened 2024-06-19
[BUG] When adding some chains app crashes #703

flypaper0:


enriquesouza (commented):
We should not add if it exists, correct?

johnnyluo (approved):
LGTM

Issue #642, opened 2024-06-08
IMPROVE: Padding on Chain selection iPad/Mac 


realpaaao:
**Is your feature request related to a problem? Please describe.**
The padding is missing on Chain selection after last chain on iPad and Mac

**Describe the solution you'd like**
Add padding

![image](https://github.com/vultisig/vultisig-ios/assets/167348323/4d6e86ca-707e-416b-b81b-7d6ae397642a)

PR #731, opened 2024-06-20
Padding added for iPad on Chain selection view

Rockindash:
Padding added for iPad on Chain selection view and increase vertically, Issue #642


johnnyluo (approved):
LGTM

PR #732, opened 2024-06-20
cache for balance

enriquesouza:
In this update, I've added a caching mechanism to the BalanceService to improve performance by storing frequently accessed data in memory. 

This change reduces the need for slow disk I/O operations caused by try await Storage.shared.save(coin). 

We aim to achieve faster data retrieval and enhance our app's responsiveness by minimizing these disk accesses. 

Let's test this implementation to evaluate its effectiveness.

### DIFF
--- a/VultisigApp/VultisigApp/Services/BalanceService.swift
+++ b/VultisigApp/VultisigApp/Services/BalanceService.swift
@@ -41,7 +41,6 @@ class BalanceService {
         do {
             let cacheKey = "\(coin.ticker)-\(coin.contractAddress)-\(coin.chain.rawValue)-\(coin.address)"
             if let cachedData = await Utils.getCachedData(cacheKey: cacheKey, cache: cache, timeInSeconds: CACHE_TIMEOUT_IN_SECONDS) {
-                try await updateCoin(coin, rawBalance: cachedData.balance, priceRate: cachedData.priceRate)
                 return
             }
             
@@ -93,10 +92,8 @@ class BalanceService {
                 (rawBalance, priceRate) = try await dot.getBalance(coin: coin)
             }
             
-            if rawBalance != .zero, priceRate > 0 {
-                try await updateCoin(coin, rawBalance: rawBalance, priceRate: priceRate)
-                cache.set(cacheKey, (data: (balance: rawBalance, priceRate), timestamp: Date()))
-            }
+            try await updateCoin(coin, rawBalance: rawBalance, priceRate: priceRate)
+            cache.set(cacheKey, (data: (balance: rawBalance, priceRate), timestamp: Date()))
         } catch {
             print("BalanceService error: \(error.localizedDescription)")
         }
--- a/VultisigApp/VultisigApp/Services/BalanceService.swift
+++ b/VultisigApp/VultisigApp/Services/BalanceService.swift
@@ -20,11 +20,15 @@ class BalanceService {
     private let kuji = KujiraService.shared
     private let maya = MayachainService.shared
     private let dot = PolkadotService.shared
-
+    
+    private var cache: ThreadSafeDictionary<String, (data: (balance: String, priceRate: Double), timestamp: Date)> = ThreadSafeDictionary()
+    
+    private let CACHE_TIMEOUT_IN_SECONDS: Double = 120 // 2 minutes
+    
     func updateBalances(coins: [Coin]) async {
         await withTaskGroup(of: Void.self) { group in
             for coin in coins {
-                group.addTask { [unowned self]  in
+                group.addTask { [unowned self] in
                     if !Task.isCancelled {
                         await updateBalance(for: coin)
                     }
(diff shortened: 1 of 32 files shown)


VultisigApp/VultisigApp/Views/Keygen/MessagePuller.swift
```diff
@@ -73,7 +73,8 @@ class MessagePuller: ObservableObject {
                     print("Response: \(String(data:data,encoding: .utf8) ?? "")")
                     let decoder = JSONDecoder()
                     let msgs = try decoder.decode([Message].self, from: data)
-                    for msg in msgs.sorted(by: { $0.sequence_no < $1.sequence_no }) {
+                    let sortedMsgs = msgs.sorted(by: { $0.sequence_no < $1.sequence_no })
```
johnnyluo, line 76: why change this? I don't see any difference

enriquesouza: Just sorting outside the loop for performance improvement.

VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
```diff
@@ -41,6 +41,8 @@ struct VaultDetailView: View {
         }
         .onChange(of: vault.coins) {
             viewModel.fetchCoins(for: vault)
+            // Without the set data it does not update the balance
+            setData()
```
johnnyluo, line 45: I removed that ,  because if I add 10 coins for ETH , this method will be called 10 times , that means balance will updated 10 times

VultisigApp/VultisigApp/Services/BalanceService.swift
```diff
@@ -20,78 +20,95 @@ class BalanceService {
...
                     }
                 }
             }
         }
     }
-
+    
     @MainActor func updateBalance(for coin: Coin) async {
         do {
+            let cacheKey = "\(coin.ticker)-\(coin.contractAddress)-\(coin.chain.rawValue)-\(coin.address)"
```
johnnyluo, line 42: Please don't do this right now , there is no indication that getting balance or any api calls cause the hang

johnnyluo: Maybe just record what is the last time we do the balance check , if it is less than 1 minute , then we skip

VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
enriquesouza: Removed

VultisigApp/VultisigApp/Services/BalanceService.swift
enriquesouza: I will make it simple just setting the amount of coin rawBalance and PriceRate to match with the cached amount, just in case it missed somewhere.

enriquesouza:
```
private var cache: ThreadSafeDictionary<String, (data: Bool, timestamp: Date)> = ThreadSafeDictionary()
let cacheKey = "\(coin.ticker)-\(coin.contractAddress)-\(coin.chain.rawValue)-\(coin.address)"
            if let cachedData = await Utils.getCachedData(cacheKey: cacheKey, cache: cache, timeInSeconds: CACHE_TIMEOUT_IN_SECONDS) {
                return
            }
```
I'm just using the cache to avoid going to APIs for few minutes and return whatever is already persisted in the coin.

VultisigApp/VultisigApp/Services/BalanceService.swift
```diff
@@ -20,78 +20,92 @@ class BalanceService {
     private let kuji = KujiraService.shared
     private let maya = MayachainService.shared
     private let dot = PolkadotService.shared
-
+    
+    private var cache: ThreadSafeDictionary<String, (data: Bool, timestamp: Date)> = ThreadSafeDictionary()
```
johnnyluo, line 24: can you remove the cache? as discussed , we could just keep when is the last time user check balance , if it is less than 1 minutes , we could skip it

VultisigApp/VultisigApp/Views/Vault/ChainDetailView.swift
```diff
@@ -132,14 +132,22 @@ struct ChainDetailView: View {
         ChainHeaderCell(group: group, isLoading: $isLoading)
     }
     
+    var sortedCoins: [Coin] {
```
johnnyluo, line 135: As discussed yesterday , this doesn't help with the performance , can you move it into viewmodel as a published property instead , only update it when the balance get changed

VultisigApp/VultisigApp/Services/BalanceService.swift
enriquesouza: yes, I've done that i just forgot to push

VultisigApp/VultisigApp/Views/Vault/ChainDetailView.swift
enriquesouza: I will remove that.

VultisigApp/VultisigApp/Services/BalanceService.swift
johnnyluo, line 24: We are not using the cache, can you please just remove it , use a date instead?

enriquesouza: I will add a var then.

johnnyluo: nevermind , I misunderstand it

johnnyluo (approved):
LGTM

PR #739, opened 2024-06-22
Ensure vault name is unique for both keygen and reshare

### LINKED ISSUE #737: [BUG] Possible to delete existing vault with pairing same vault name
**Describe the bug**
When performing a keygen, if the initiating device attempts to create a vault with a name that already exists on the pairing device, it can will overwrite the existing vault on the pairing device and the funds will be lost.

**To Reproduce**
1. Create a vault where the name is the same as a vault that already exisits on pairing device
2. pair with other device
3. let keygen finish
4. app crashes on pairing device
5. see error

**Expected behavior**
The initiating device or the pairing devices should check on joining the session if the name already ay exists in one of the vaults and abort if that is the case.

**Screenshots**
before:
IMG_0113
after:
IMG_0114

all devices on version 31

johnnyluo:
fixes #737

### DIFF
--- a/VultisigApp/VultisigApp/View Models/JoinKeygenViewModel.swift
+++ b/VultisigApp/VultisigApp/View Models/JoinKeygenViewModel.swift
@@ -180,6 +180,14 @@ class JoinKeygenViewModel: ObservableObject {
         }
     }
 #endif
+    func isVaultNameAlreadyExist(name: String) -> Bool  {
+        for v in self.vaults {
+            if v.name == name && !v.pubKeyECDSA.isEmpty {
+                return true
+            }
+        }
+        return false
+    }
     
     func handleQrCodeSuccessResult(scanData: Data) {
         var useVultisigRelay = false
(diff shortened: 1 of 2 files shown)


Rockindash (approved):

VultisigApp/VultisigApp/View Models/JoinKeygenViewModel.swift
```diff
@@ -180,6 +180,14 @@ class JoinKeygenViewModel: ObservableObject {
         }
     }
 #endif
+    func isVaultNameAlreadyExist(name: String) -> Bool  {
```
enriquesouza, line 183: I would use contains intead

```
func isVaultNameAlreadyExist(name: String) -> Bool {
    return self.vaults.contains { v in
        v.name == name && !v.pubKeyECDSA.isEmpty
    }
}
```

enriquesouza (approved):
LGTM, just a a few suggestions

PR #741, opened 2024-06-22
Mac support/setup views

Rockindash:
UI Update for Mac, Issue #405
- Fixed font issue on macOS target.
- Set min size for windows
- Updated the UI for setup views: Onboarding, CreateVaultView and ImportWalletView


johnnyluo (approved):
LGTM

PR #751, opened 2024-06-24
add vault name to reshare message

johnnyluo:


VultisigApp/VultisigApp/View Models/KeygenViewModel.swift
```diff
@@ -121,6 +121,8 @@ class KeygenViewModel: ObservableObject {
                 // if local party is not in the old committee , then he is the new guy , need to add the vault
                 // otherwise , they previously have the vault
                 if !self.vaultOldCommittee.contains(self.vault.localPartyID) {
+                    VaultDefaultCoinService(context: context)
+                        .setDefaultCoinsOnce(vault: self.vault)
                     context.insert(self.vault)
```
flypaper0, line 126: No need to save context after context.insert. 

I think we also should take a look on autosaveEnabled context property, and disable it to have more control 
https://developer.apple.com/documentation/swiftdata/modelcontext/autosaveenabled

flypaper0 (approved):

PR #774, opened 2024-06-27
[Swap] Inbound address check removed

flypaper0:
Inbound address check removed because it's can be MsgDeposit and we're building tx with WalletCore


johnnyluo (approved):
LGTM

PR #786, opened 2024-06-28
Mac support/import views

Rockindash:
Added the following views and error state for Mac:
- New Import QR Code Keygen View implemented
- New Import QR Code Keysign View implemented
- Error state added for nandleQrCodeFromImage under Utils
(To display errors as an alert message)


johnnyluo (approved):
LGTM

PR #791, opened 2024-06-28
[Bugfix] Selected tokens is not appears on MacOS

flypaper0:


johnnyluo:
This cause another bug , like if I add avalanche , then "token selection" button doesn't exist
image

VultisigApp/VultisigApp/Views/Vault/ChainDetailView.swift
```diff
@@ -2,9 +2,16 @@ import SwiftUI
...
     @ObservedObject var group: GroupedChain
-    let vault: Vault
     
-    @State var tokens: [CoinMeta] = []
+    let vault: Vault
+
+    var tokens: [Coin] {
+        return vault.coins
+            .filter { $0.chain == group.chain }
+            .filter { !$0.isNativeToken }
```
johnnyluo, line 11: I think you should remove `.filter{ !$0.isNativeToken}` , we need to show the balance of native token
the balance you see on the top is actually the total balance of all tokens

johnnyluo (approved):
LGTM

Issue #617, opened 2024-06-05
ADD: Validity check of input on text fields


realpaaao:
Please implement the validity check for valid input on text field directly when leaving the text field after entering it. Furthermore restrict maximum characters to something reasonable, like 50. 
Therefore the user won't have to enter all details to see if all inputs are correct just to do it again if wrong.

enriquesouza:
https://github.com/vultisig/vultisig-ios/pull/799

PR #799, opened 2024-07-01
max length extension

enriquesouza:
https://github.com/vultisig/vultisig-ios/issues/617


VultisigApp/VultisigApp/Extensions/TextField.swift
```diff
@@ -0,0 +1,56 @@
...
+import SwiftUI
+import Combine
+
+#if canImport(UIKit)
+import UIKit
+
+private var maxLengthKey: UInt8 = 0
+
+extension UITextField {
+    @IBInspectable var maxLength: Int {
```
flypaper0, line 19: IBInspectable can be removed, we're do not use interface builder

VultisigApp/VultisigApp/Extensions/TextField.swift
```diff
@@ -0,0 +1,56 @@
...
+        get {
+            return objc_getAssociatedObject(self, &maxLengthKey) as? Int ?? 50
```
flypaper0, line 21: why 50 is default value?

VultisigApp/VultisigApp/Extensions/TextField.swift
```diff
@@ -0,0 +1,56 @@
...
+import Foundation
+import SwiftUI
+import Combine
+
+#if canImport(UIKit)
+import UIKit
+
+private var maxLengthKey: UInt8 = 0
+
+extension UITextField {
```
flypaper0, line 18: This extension is unused. Let's remove as we're not using UITextFields from UIKit

flypaper0 (approved):

VultisigApp/VultisigApp/Extensions/TextField.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+#endif
+
+struct MaxLengthModifier: ViewModifier {
+    @Binding var text: String
+    var maxLength: Int = 50
+    
+    func body(content: Content) -> some View {
+        content
+            .onChange(of: text) { newValue in
```
flypaper0, line 44: Fix warning please 
`'onChange(of:perform:)' was deprecated in iOS 17.0: Use `onChange` with a two or zero parameter action closure instead.`

VultisigApp/VultisigApp/Extensions/TextField.swift
enriquesouza: It is used to set max length to both macOs and IOs.

VultisigApp/VultisigApp/Extensions/TextField.swift
enriquesouza: Ah, sure

VultisigApp/VultisigApp/Extensions/TextField.swift
enriquesouza: It is used to set max length to both macOs and IOs.

johnnyluo: I think what @flypaper0 means is `@IBInspectable` is used for interface builder , we are not using that , so the property wrapper should be removed

enriquesouza: Hum... let me check

VultisigApp/VultisigApp/Extensions/TextField.swift
enriquesouza: You are right.

VultisigApp/VultisigApp/Extensions/TextField.swift
enriquesouza: He is right, we don't need that the extension resolves that. Thanks @flypaper0

PR #823, opened 2024-07-03
get tokens using one inch

enriquesouza:


VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
```diff
@@ -36,26 +36,57 @@ struct OneInchService {
...
         let response: OneInchTokensResponse = try await Utils.fetchObject(from: Endpoint.fetchTokens(chain: chain))
         let tokens = Array(arrayLiteral: response.tokens.values).reduce([], +)
         return tokens
     }
+    
+    func fetchTokensBalance(chainId: Int, walletAddress: String) async throws -> [OneInchToken] {
+        let oneInchChainId = String(describing: chainId)
+        let balanceData = try await Utils.asyncGetRequest(urlString: Endpoint.fetch1InchsTokensBalance(chain: oneInchChainId, address: walletAddress), headers: [:])
+        
+        if let balanceDictionary = try JSONSerialization.jsonObject(with: balanceData, options: []) as? [String: String] {
```
flypaper0, line 60: is it possible to use JSONDecoder?

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
```diff
@@ -36,26 +36,57 @@ struct OneInchService {
...
+            
+            var tokens: [OneInchToken] = []
+            for tokenAddress in nonZeroBalanceTokenAddresses {
+                if let tokenInfo: OneInchToken = Utils.extractResultFromJson(fromData: tokenInfoData, path: tokenAddress, type: OneInchToken.self) {
+                    tokens.append(tokenInfo)
+                }
+            }
+            return tokens;
+        }
+        return []
```
flypaper0, line 83: This is throwing function. Can we do not return empty array if something went wrong in parsing?

VultisigApp/VultisigApp/Services/RpcEvmService.swift
```diff
@@ -253,46 +254,18 @@ class RpcEvmService: RpcService {
...
-                    if let image = mutableToken.tokenInfo.image{
-                        mutableToken.tokenInfo.setImage(image: "\(extractTokenDomainURL(from: urlString))\(image)" )
-                    }
-                    return mutableToken
-                }
-                cacheTokens.set(cacheKey, (data: tokens, timestamp: Date()))
-                return tokens
-            } else {
-                return []
+            let oneInchTokens = try await oneInchService.fetchTokensBalance(chainId: nativeToken.chain.chainID ?? 0, walletAddress: nativeToken.address)
```
flypaper0, line 260: Didn't got the idea. We're calling `fetchTokensBalance` that returns array of 1inch tokens, but after mapping it into array of CoinMeta that does not have balance property. Do we really fetch balances here?

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
```diff
@@ -36,26 +36,57 @@ struct OneInchService {
...
+            }
+            
+            let nonZeroBalanceTokenAddresses = tokensWithNonZeroBalance.map { $0.key }
+            
+            let tokenInfoData = try await Utils.asyncGetRequest(
+                urlString: Endpoint.fetch1InchsTokensInfo(chain: oneInchChainId, addresses: nonZeroBalanceTokenAddresses), headers: [:]
+            )
+            
+            var tokens: [OneInchToken] = []
+            for tokenAddress in nonZeroBalanceTokenAddresses {
```
flypaper0, line 76: `reduce` can be used here

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
```diff
@@ -36,26 +36,57 @@ struct OneInchService {
...
+                if let tokenInfo: OneInchToken = Utils.extractResultFromJson(fromData: tokenInfoData, path: tokenAddress, type: OneInchToken.self) {
```
flypaper0, line 77: Can you parse in array? [OneInchToken].self
But better to use JSONDecoder imo

VultisigApp/VultisigApp/Services/RpcEvmService.swift
flypaper0: May be just wrong `fetchTokensBalance` function name

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
enriquesouza: in this case, it is not possible, because it is not an array returned from 1inch, each address is the key, then the value is the ```OneInchToken```

```
{
  "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee": {
    "address": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
    "chainId": 1,
    "decimals": 18,
    "name": "Ether",
    "symbol": "ETH",
    "providers": [
      "1inch",
      "Curve Token List"
    ],
    "logoURI": "https://tokens.1inch.io/0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee.png",
    "eip2612": false,
    "tags": [
      "native",
      "PEG:ETH"
    ],
    "rating": 2
  },
  "0x85f17cf997934a597031b2e18a9ab6ebd4b9f6a4": {
    "address": "0x85f17cf997934a597031b2e18a9ab6ebd4b9f6a4",
    "chainId": 1,
    "decimals": 24,
    "name": "NEAR Protocol",
    "symbol": "NEAR",
    "providers": [
      "1inch",
      "CoinGecko",
      "Kleros Tokens"
    ],
    "logoURI": "https://tokens.1inch.io/0x85f17cf997934a597031b2e18a9ab6ebd4b9f6a4.png",
    "eip2612": false,
    "tags": [
      "tokens"
    ],
    "rating": 3
  }
}
```

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
enriquesouza: Changed to use reduce

VultisigApp/VultisigApp/Services/RpcEvmService.swift
enriquesouza: This method discovers the tokens the user has on its address

```
static func fetch1InchsTokensBalance(chain: String, address: String) -> String {
        return "\(vultisigApiProxy)/1inch/balance/v1.2/\(chain)/balances/\(address)"
    }
 ```
 
 so it returns something like:
 
 ```
 {
 "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee":0,
 "0x85f17cf997934a597031b2e18a9ab6ebd4b9f6a4":18371273
 }
```

Then I filter by the contracts they have balances to consider it a token

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
enriquesouza: Yep. We are allowing it to throw, we only return [] if the user has no tokens or if any of them has no balance, therefore if any error happens on try ... it is throwing.

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
enriquesouza: Changed to use ```let balanceDictionary = try JSONDecoder().decode([String: String].self, from: balanceData)```

VultisigApp/VultisigApp/Services/RpcEvmService.swift
enriquesouza: I also renamed the function

VultisigApp/VultisigApp/Services/1inch/OneInchService.swift
enriquesouza: I've changed to use the compactMap.

VultisigApp/VultisigApp/View Models/VaultDefaultCoinService.swift
```diff
@@ -39,6 +39,10 @@ class VaultDefaultCoinService {
             for coin in coins {
                 self.context.insert(coin)
                 vault.coins.append(coin)
+
+                Task {
+                    try await CoinSelectionViewModel().addDiscoveredTokens(nativeToken: coin, to: vault)
```
johnnyluo, line 44: This is weird , create a `CoinSelectionViewModel()` here ,

johnnyluo (approved):

VultisigApp/VultisigApp/View Models/VaultDefaultCoinService.swift
enriquesouza: I was also confused, but then I could see this service is in the same folder as all view models, so maybe we can create a service responsible for adding all coins and removing this view model.

PR #825, opened 2024-07-03
Sui has an address higher than 50, it is actually ~67.

enriquesouza:
Sui has an address higher than 50, it is actually ~67.


johnnyluo (approved):
LGTM

enriquesouza:
@realpaaao

PR #837, opened 2024-07-05
Coin Service

enriquesouza:
CoinService Implementation

We are centralizing all coin manipulation logic within a dedicated CoinService. This service will be responsible for handling operations such as adding, deleting, or updating coins. It will be designed to interact seamlessly with other services but will remain completely decoupled from any UI-related responsibilities.

The CoinService will receive the necessary data and commands to perform its tasks, ensuring a clean separation of concerns. The ViewModel will act as the intermediary between the UI and the CoinService, solely focusing on UI interactions and delegating all coin manipulation tasks to the CoinService.


VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -0,0 +1,106 @@
...
+//
+//  Created by Enrique Souza Soares on 05/07/24.
+//
+
+import Foundation
+
+@MainActor
+class CoinService {
+    
+    static let shared = CoinService()
```
johnnyluo, line 13: This service doesn't have any state , it just a bunch of functions , don't make it singleton , let's create it when we need it.

VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -0,0 +1,106 @@
...
+    
+    public func removeCoins(coins: [Coin], vault: Vault) async throws {
```
johnnyluo, line 15: why we need to make it `public`?  the same for all other methods

VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -0,0 +1,106 @@
...
+            
+            try await addToChain(assets: newCoins, to: vault)
+            
+        } catch {
+            print("fail to save asset,\(error)")
+        }
+    }
+    
+    public func addToChain(assets: [CoinMeta], to vault: Vault) async throws {
+        if let coin = assets.first, coin.chain.chainType == .EVM, !coin.isNativeToken {
```
johnnyluo, line 61: this is very confusing , need to add some comments. 
inside the if , is add ERC20/BEP20 tokens
inside else is add chain

VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -0,0 +1,106 @@
...
+        if let priceProviderId {
+            newCoin.priceProviderId = priceProviderId
+        }
+        // Save the new coin first
+        try await Storage.shared.save(newCoin)
+        vault.coins.append(newCoin)
+        return newCoin
+    }
+    
+    public func addDiscoveredTokens(nativeToken: Coin, to vault: Vault) async throws  {
```
johnnyluo, line 86: this function doesn't throw error at all ,  either remove the outer layer of do/catch, or remove `throws`

johnnyluo (changes requested):
some comments

VultisigApp/VultisigApp/Services/CoinService.swift
enriquesouza: It can be static as well, but we should avoid creating a new instance every time. It interacts directly with the storage, using Swift data. Additionally, it follows the same pattern as all other services.

VultisigApp/VultisigApp/Services/CoinService.swift
enriquesouza: I didn't change that. I only moved. I will separate the logic soon using a factory and some methods to instantiate the coins. So we can kill those if EVM, etc.

VultisigApp/VultisigApp/Services/CoinService/VaultDefaultCoinService.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 import SwiftData
 
-
+// Should I move this to Vault service?
```
johnnyluo, line 11: is it still a question?

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
```diff
@@ -0,0 +1,45 @@
...
+                } catch {
+                    print("Error adding the token \(token.ticker) service: \(error.localizedDescription)")
+                }
+            }
+        } catch {
+            print("Error fetching service: \(error.localizedDescription)")
+        }
+    }
+    
+    override func addToChain(assets: [CoinMeta], to vault: Vault) async throws {
```
johnnyluo, line 30: Honestly I am not a fan of this , see `addToChain` , this method exist on many places....  two in it's parent.  And override here as well, it is super confusing ,  you need to think about it multiple times to figure which method it is actually calling...

enriquesouza: What is your suggestion?

VultisigApp/VultisigApp/Services/CoinService/VaultDefaultCoinService.swift
enriquesouza: I'm still thinking about this one. What is your suggestion?

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
johnnyluo: maybe don't put any implementation on the parent class `CoinService`

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
```diff
@@ -0,0 +1,45 @@
...
+//  EvmCoinService.swift
+//  VultisigApp
+//
+//  Created by Enrique Souza Soares on 05/07/24.
+//
+
+import Foundation
+
+@MainActor
+class EvmCoinService: CoinService {
```
flypaper0, line 11: Can we do not use inheritance? We can do it with protocols

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    override func addDiscoveredTokens(nativeToken: Coin, to vault: Vault) async {
```
flypaper0, line 13: Its better to make CoinService methods `async throws` and handle errors outside

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
enriquesouza: What about the code duplication?

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
enriquesouza: What about the code duplication?

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
enriquesouza: For discovering tokens there is not need to throw it. We can simply don't add the problematic item IMO.

VultisigApp/VultisigApp/Services/CoinService/EvmCoinService.swift
flypaper0: or you can always use CoinService instance. add there addDiscoveredTokens and addTokenToChain

enriquesouza: Evms have different implementations. That is why I am using polymorphism.

johnnyluo:
@enriquesouza , let's close this PR for now , @flypaper0 is planning to refactor our storage layer using a new library , and given this doesn't really make the logic simpler , just move to different places. 
Let's keep the branch , so we can get it back if we need in the future

enriquesouza:
That is fine.

PR #851, opened 2024-07-07
select the correct coin

enriquesouza:


VultisigApp/VultisigApp/Views/Vault/CoinDetailView.swift
```diff
@@ -46,8 +46,14 @@ struct CoinDetailView: View {
         .navigationBarBackButtonHidden(true)
         .navigationTitle(NSLocalizedString(coin.ticker, comment: ""))
         .onAppear {
+            sendTx.coin = coin
```
flypaper0, line 49: Do we need to set coin here? `.reset` is setting coin

flypaper0 (approved):

VultisigApp/VultisigApp/Views/Vault/CoinDetailView.swift
enriquesouza: Changed to use reset only on change and it also works. Thanks.

PR #860, opened 2024-07-09
solana tokens rpc

enriquesouza:
### Solana Tokens PR

This PR is quite sizeable due to the several complex features we've had to tackle:

* **Token discovery via RPC**: We've utilized the RPC for token discovery purposes.
* **Token info fetch from Solana.FM**: We're now retrieving token information from Solana.FM, a different source.
* **protobuf updates for token-associated addresses**: To support sending tokens, we've had to modify the protobuf to accommodate token-associated addresses.
* **Handling 3 types of token sending on Solana**:
	+ Native tokens
	+ Sending to addresses with existing token associations
	+ Sending tokens to addresses without prior token associations


VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -69,6 +70,59 @@ class SolanaService {
...
+            "params": [
+                walletAddress,
+                ["programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"], // SPL Token Program ID
+                ["encoding": "jsonParsed"]
+            ]
+        ]
+        do {
+            let data = try await postRequest(with: requestBody)
+            
+            if let jsonResponse = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any],
```
flypaper0, line 89: can we use JSONDecoder and Codable please?

enriquesouza: Yes, sure.

VultisigApp/VultisigApp/Model/Solana/TokenAccountsByOwner.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    class SolanaTokenAmount: Decodable {
+        let amount: String
+        let decimals: Int
+        let uiAmount: Double
+        let uiAmountString: String
+    }
+    
+    func parseSolanaTokenResponse(jsonData: Data) throws -> SolanaService.SolanaDetailedRPCResult<[SolanaService.SolanaTokenAccount]> {
+        return try JSONDecoder().decode(SolanaService.SolanaDetailedRPCResult<[SolanaService.SolanaTokenAccount]>.self, from: jsonData)
```
flypaper0, line 52: let's move all Solana related network requests into SolanaService?

flypaper0: not only parsing

VultisigApp/VultisigApp/Services/Keysign/BlockChainSpecific.swift
```diff
@@ -14,10 +14,10 @@ enum BlockChainSpecific: Codable, Hashable {
     case THORChain(accountNumber: UInt64, sequence: UInt64, fee: UInt64)
     case MayaChain(accountNumber: UInt64, sequence: UInt64)
     case Cosmos(accountNumber: UInt64, sequence: UInt64, gas: UInt64)
-    case Solana(recentBlockHash: String, priorityFee: BigInt) // priority fee is in microlamports
+    case Solana(recentBlockHash: String, priorityFee: BigInt, fromAddressPubKey: String?, toAddressPubKey: String?) // priority fee is in microlamports
```
flypaper0, line 17: what you think about separating Solana blockchainSpecific into: .solana, .solanaToken, .solanaCreateToken

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -8,6 +9,12 @@ class SolanaService {
     private let rpcURL = URL(string: Endpoint.solanaServiceRpc)!
     private let jsonDecoder = JSONDecoder()
     
+    private var cacheTokenAccountsByOwner: ThreadSafeDictionary<String, (data: [SolanaTokenAccount], timestamp: Date)> = ThreadSafeDictionary()
```
flypaper0, line 12: lets stop using dictionary cache? It's slow because of sync with .barrier and not convenient to use imo. We can use URL cache. Just setup requestPolicy for request you need to cache

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -26,29 +33,35 @@ class SolanaService {
...
+                
+                rawBalance = try await fetchTokenBalance(for: coin.address, contractAddress: coin.contractAddress)
+                
+            }
         } catch {
             print("Error fetching balance: \(error.localizedDescription)")
             throw error
         }
-        return (rawBalance,priceRateFiat)
+        return (rawBalance, priceRateFiat)
```
flypaper0, line 64: we should not return zero balances on fetching error

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -26,29 +33,35 @@ class SolanaService {
...
+                    rawBalance = totalBalance.description
+                }
+                
+            } else {
+                
+                rawBalance = try await fetchTokenBalance(for: coin.address, contractAddress: coin.contractAddress)
+                
+            }
         } catch {
             print("Error fetching balance: \(error.localizedDescription)")
```
flypaper0, line 61: if you handling error just for printing, need to do it on the top level, not in SolanaService

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -26,29 +33,35 @@ class SolanaService {
...
         
         do {
-            let data = try await Utils.PostRequestRpc(rpcURL: rpcURL, method: "getBalance", params:  [coin.address])
             
-            if let totalBalance = Utils.extractResultFromJson(fromData: data, path: "result.value") as? Int64 {
-                rawBalance = totalBalance.description
-            }            
+            if coin.isNativeToken {
+                
+                let data = try await Utils.PostRequestRpc(rpcURL: rpcURL, method: "getBalance", params: [coin.address])
```
flypaper0, line 49: use Codable and JSONDecoder

flypaper0: I've personally remove this Utils.PostRequestRpc, to not work with raw data anymore

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -69,6 +82,135 @@ class SolanaService {
...
+            
+            // This is the address we are sending from and to
+            let pubkey = associatedAccount.pubkey
+            
+            cacheTokenAccountsByOwnerMint.set(cache_key, (data: pubkey, timestamp: Date()))
+            
+            return pubkey
+        } catch {
+            print("Error fetching tokens: \(error.localizedDescription)")
+            throw error
```
flypaper0, line 131: same

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -69,6 +82,135 @@ class SolanaService {
...
+    func fetchTokenBalance(for walletAddress: String, contractAddress: String) async throws -> String {
+        
+        do {
+            
+            let accounts: [SolanaTokenAccount] = try await fetchTokenAccountsByOwner(for: walletAddress)
+            
+            if let token = accounts.first(where: { $0.account.data.parsed.info.mint == contractAddress }) {
+                return token.account.data.parsed.info.tokenAmount.amount
+            } else {
+                return .zero
```
flypaper0, line 177: Let's make the "Empty string value = nil" rule please? To not do isEmpty checks.

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
enriquesouza: I don't think it's worth the level of effort to keep strong typing everything. In this case, we need one property only. It is a base type, so I would need to create a structure to map one item. Well, we can do it, but ...

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
enriquesouza: Ok, I've created a Cache Service to handle that. We will use the ```CachedURLResponse```

VultisigApp/VultisigApp/Model/Solana/TokenAccountsByOwner.swift
enriquesouza: Moved.

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
enriquesouza: I will allow it to throw the error always for all inside the Solana Service.

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
enriquesouza: It is now throwing instead.

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
enriquesouza: Done.

VultisigApp/VultisigApp/Services/CacheService.swift
```diff
@@ -0,0 +1,41 @@
...
+    
+    private let cache: URLCache
+    private let cacheTimeout: TimeInterval
+    
+    private init(cache: URLCache = URLCache.shared, cacheTimeout: TimeInterval = 120) {
+        self.cache = cache
+        self.cacheTimeout = cacheTimeout
+    }
+    
+    func getCachedData<T: Codable>(for key: String, type: T.Type) -> T? {
```
flypaper0, line 22: check .returnCacheDataElseLoad cache policy. 
No need to cache anything manually, just setup this policy for requests you need

https://developer.apple.com/documentation/foundation/nsurlrequest/cachepolicy

enriquesouza: I've changed to use the cache policy with some expiration time as suggested.

VultisigApp/VultisigApp/Services/Keysign/BlockChainSpecific.swift
enriquesouza: I believe it will bring more complexity to our code. So I would recommend us to keep it in the same service. It already handles it for Native, New Token Accounts, and Already associated Token acounts.

VultisigApp/VultisigApp/Services/Solana/SolanaService.swift
```diff
@@ -211,29 +162,22 @@ class SolanaService {
...
-        
         return UInt64(highPriorityFee)
-        
     }
     
     private func postRequest(with body: [String: Any]) async throws -> Data {
-        
         var request = URLRequest(url: rpcURL)
-        
+        request.cachePolicy = .returnCacheDataElseLoad
```
flypaper0, line 171: It should be configurable for specific request. Otherwise it can return outdated data for state specific data, like swap quote for ex

flypaper0: Or we want to cache all solana rpc requests?

enriquesouza: We want to cache them all for 2 min only so we don't reach the request limits.

flypaper0 (approved):

Issue #869, opened 2024-07-11
[BUG] Black font on error


realpaaao:
**Describe the bug**
wrong font on error "This device... Join keygen"
Probably due to color scheme

**Expected behavior**
Have it in white

**Screenshots**
![image](https://github.com/vultisig/vultisig-ios/assets/167348323/0d628c95-b682-45d6-8a5b-0a5db434c792)

PR #870, opened 2024-07-12
Static dark mode color scheme added to the app

Rockindash:
Static dark mode color scheme added to the app, Issue #869


johnnyluo (approved):
LGTM

Issue #863, opened 2024-07-10
[BUG] Adding custom token crashes app


realpaaao:
**Describe the bug**
Adding a custom token results in a crash on macOS app.
After restarting the token got successfully added.

**To Reproduce**
1. add custom token on any chain
2. See error

**Expected behavior**

**Screenshots**
<img width="898" alt="image" src="https://github.com/vultisig/vultisig-ios/assets/167348323/8922ac16-8096-4d5d-a8ba-284123efb036">

**Desktop (please complete the following information):**
 - OS: 14.5
 - Version 37

enriquesouza:
https://github.com/vultisig/vultisig-ios/pull/887/files

PR #887, opened 2024-07-16
fix [BUG] Adding custom token crashes app #863

enriquesouza:
It should fix the error:

https://github.com/vultisig/vultisig-ios/issues/863

### DIFF
--- a/VultisigApp/VultisigApp/Views/Vault/CustomTokenView.swift
+++ b/VultisigApp/VultisigApp/Views/Vault/CustomTokenView.swift
@@ -210,7 +210,8 @@ struct CustomTokenView: View {
                 
                 try await Task.sleep(nanoseconds: 1_000_000_000)
                 isLoading = false
-                chainDetailView.sheetType = nil
+                self.chainDetailView.sheetType = nil
+                dismiss()
             }
         }
     }


VultisigApp/VultisigApp/Views/Vault/CustomTokenView.swift
```diff
@@ -210,7 +210,8 @@ struct CustomTokenView: View {
                 
                 try await Task.sleep(nanoseconds: 1_000_000_000)
                 isLoading = false
-                chainDetailView.sheetType = nil
+                self.chainDetailView.sheetType = nil
+                dismiss()
```
johnnyluo, line 214: I am not sure why this will cause a crash , and also why this will solve the issue, can you please explain?

enriquesouza: It is not a crash. It is not dismissing the window, so it doesn’t come back.

johnnyluo, line 214: I think it should be something like 
#if os(macOS)
dismiss()
#else
self.chainDetailView.sheetType = nil
#endif

johnnyluo (approved):
LGTM

Issue #465, opened 2024-05-17
Default to Wide-Angle lense


TForge1:
**Describe the bug**
The camera defaults to wide-angle, making it very hard to scan the QR code. 
wide-angle lense has too much iso and makes a blurry/grainy image/is not good with light management.
So its blurry and takes a long time to scan.

**Expected behavior**
 Would prefer general camera lense/narrow camera lense.

**Smartphone (please complete the following information):**
 - iphone 15

PR #894, opened 2024-07-17
Primary camera set as default for QRCode scanner

Rockindash:
Primary camera set as default for QRCode scanner, Issue #465


johnnyluo (approved):
LGTM

PR #897, opened 2024-07-17
change the paired devices messages and validation

enriquesouza:


VultisigApp/VultisigApp/Views/Keygen/PeerDiscoveryView.swift
```diff
@@ -382,6 +401,29 @@ struct PeerDiscoveryView: View {
         " " +
         NSLocalizedString("vault", comment: "")
     }
+    
+    var invalidNumberOfSelectedDevices: Alert {
```
johnnyluo, line 405: I don't like this prompt up message ,  just disable "continue" button

johnnyluo (approved):
LGTM

PR #898, opened 2024-07-17
Add more information to Coin protobuf def

johnnyluo:
This PR avoid to match coin in vault , which means even the coin doesn't added to the given vault, it should still be able to co-sign


enriquesouza (approved):
LGTM and tested.

PR #899, opened 2024-07-17
fix on m of n since it was allowing to proceed with only one device

enriquesouza:
fix on m of n since it was allowing to proceed with only one device


johnnyluo (approved):
LGTM

Issue #781, opened 2024-06-27
ADD: Default chains in settings


realpaaao:
Please add the possibility to override the hardcoded default chains in the settings and make them global for all vaults that get created. Include all supported chains.
If user deselects all default chains, enforce the hardcoded chains.

Use [Figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=6892-6496&t=xfnbvFHmouveBm9k-4) as reference.

johnnyluo:
fixed #947

PR #906, opened 2024-07-18
Default chains

Rockindash:
Default chains option added to settings, Issue #781


VultisigApp/VultisigApp/Model/CoinMeta.swift
```diff
@@ -7,8 +7,10 @@
 
 import Foundation
 import WalletCore
+import SwiftData
 
-struct CoinMeta : Hashable,Codable{
+@Model
+class CoinMeta : Hashable, Codable {
```
johnnyluo, line 13: Can we please don't save `CoinMeta` into swiftdata , the reason we add CoinMeta is because we have too much crash issues with SwiftData and Coin ,  

If we do this , we will get the issue again

Issue #849, opened 2024-07-06
ADD: Sharing Vault UID via QR for web


realpaaao:
As @johnnyluo already raised for iOS #836 ,  we need a QR Code that exports Vault details, so users can import in web and join airdrop.

Please use the #836 for functionality reference.

And [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=7592-14093&t=lr7S9OUCYrUaHdfT-4) for UI reference, as we need a way to export the Vault information in app.

Therefore change main vault view so it includes the QR share option in top right, as we have pulldown feature to refresh and we don't need refresh button, can be replaced.

The saved QR-code needs name format: Vultisig-{VaultName}-{xyz}.png . `XYZ is last 3 digits of UID`
With the `save or share` button give option to store on device or share directly on other apps/users.

![image](https://github.com/vultisig/vultisig-ios/assets/167348323/2481c2cd-28fa-4992-bae9-d34ab46c5ce5)

johnnyluo:
implemented in #910

Issue #836, opened 2024-07-05
feature: Add QR Code show vault public keys


johnnyluo:
Add an option to export vault's public keys in QR code , the content of the QR code will be     
       
- uid , `sha256({vault name} - {public_key_ecdsa} - {public_key_eddsa} - {hex_chain_code})`
- public_key_ecdsa
- public_key_eddsa
- hex_chain_code

For example:
```json
{
   "uid": "fd3f9f0c0a6fb3ded9f9ac3473a5d1c42a0d5b35d011cc9729cf4c44429be58f",
   "name" : "vault name",
  "public_key_ecdsa": "0353e0f3cf78007d40200b2f71d8c7f6e8be1d4c30ba01881ee10816c09b5107f8",
  "public_key_eddsa": "ae257d1bf86387ce432410ad430ad41d7e9b371a80d5d0f0d11f9d466be79ba9",
  "hex_chain_code": "d5b95a8576bf7e4a9061fcb70a4b38ebbefbb0355cb76ba6ca8fa7b77e100f75"
}
```

The QR Code should be saved in `.png` format

Based on these information , integration partner can use bip39 unhardened derivation to get the address on all the blockchains.

johnnyluo:
implemented in #910

PR #910, opened 2024-07-18
Public key qr

Rockindash:
Added QR Code to show vault public keys, Issue #836 and #849


flypaper0 (approved):

VultisigApp/VultisigApp/Views/Components/VaultDetailQRCode.swift
```diff
@@ -0,0 +1,119 @@
...
+                .opacity(0.7)
+        }
+    }
+    
+    func getQRCode(vault: Vault) -> Image {
+        let name = vault.name
+        let ecdsaKey = vault.pubKeyECDSA
+        let eddsaKey = vault.pubKeyEdDSA
+        let hexCode = vault.hexChainCode
+        let id = "sha256(\(name) - \(ecdsaKey) - \(eddsaKey) - \(hexCode))"
```
johnnyluo, line 101: @Rockindash , id , I mean we should compute the SHA256 hash of "name-ecdsakey-eddsakey-hexode" , not plain text

johnnyluo (changes requested):
see comments

enriquesouza (changes requested):

VultisigApp/VultisigApp/Views/Components/VaultDetailQRCode.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+    
+    func getQRCode(vault: Vault) -> Image {
+        let name = vault.name
+        let ecdsaKey = vault.pubKeyECDSA
+        let eddsaKey = vault.pubKeyEdDSA
+        let hexCode = vault.hexChainCode
+        let id = "\(name)-\(ecdsaKey)-\(eddsaKey)-\(hexCode)".sha256()
+        
+        let data = """
```
johnnyluo, line 103: This generate a string , which is not a valid json , can we please define a struct , and use json encode?

johnnyluo (approved):
LGTM

enriquesouza (approved):
LGTM

PR #922, opened 2024-07-22
blowfish

enriquesouza:


Rockindash (approved):

VultisigApp/VultisigApp/Views/Components/BlowfishWarningInformationNote.swift
```diff
@@ -0,0 +1,83 @@
...
+            Image(systemName: "exclamationmark.triangle")
+                .foregroundColor(Color.warningYellow)
+        }
+    }
+    
+    var text: some View {
+        VStack(alignment: .leading, spacing: 8) {
+            
+            if blowfishMessages.isEmpty {
+                Text("Transaction Scanned Successfully! You are cleared to proceed with safety.") // Assuming `message` has a `message` property
```
johnnyluo, line 60: If the response from blowfish is empty , then please don't show anything , this message give user a sense guarantee , which is not good

enriquesouza: Shouldn’t we show we scanned the transaction? I can hide it as well.

VultisigApp/VultisigApp/View Models/SendCryptoVerifyViewModel.swift
```diff
@@ -36,6 +36,35 @@ class SendCryptoVerifyViewModel: ObservableObject {
     func amount(for coin: Coin, tx: SendTransaction) -> BigInt {
         return tx.amountInRaw
     }
+    
+    func blowfishEVMTransactionScan(tx: SendTransaction) async throws -> BlowfishResponse? {
```
johnnyluo, line 40: We also need this information on `KeysignMessageConfirmView` , on pair device , it is important to scan the tx

johnnyluo (changes requested):
Make sure we do the scan on pair device as well, and when blowfish returns empty , then don't show any messages , don't give false guarantee

VultisigApp/VultisigApp/Views/Components/BlowfishWarningInformationNote.swift
johnnyluo: Well , or you can just say "Transaction scanned by Blowfish" , and make sure it is translated

enriquesouza: I will do that "Transaction scanned by Blowfish"

enriquesouza: image

@johnnyluo

VultisigApp/VultisigApp/View Models/SendCryptoVerifyViewModel.swift
enriquesouza: Working on it

enriquesouza: image

enriquesouza:
> Make sure we do the scan on pair device as well, and when blowfish returns empty , then don't show any messages , don't give false guarantee

Showing Logic

Nothing: When blowfishResponse is nil.
Green Success: When blowfishResponse is not nil and has no warnings.
List of Warnings: When blowfishResponse is not nil and has warnings.

johnnyluo (approved):
I think this is fine for sending , but I don't think this implementation cover swap / bridging  , given we only support swap through THORChain/1inch/LI.FI at the moment , not sure we need it either

enriquesouza:
For swapping it is a different implementation, we will have to check for approvals as well

PR #928, opened 2024-07-23
Update swap progress url to use runescan

### LINKED ISSUE #896: [IMPROVE] On crosschain SWAPS, use RUNESCAN instead of FROM chain explorer
**Is your feature request related to a problem? Please describe.**
Currently a cross-chain SWAP tx hash is the FROM tx from the FROM chain explorer. This is not ideal as the full picture of the cross-chain swap is missing

**Describe the solution you'd like**
Best would be to use directly the TC tx hash view on RUNESCAN. As additional point, please consider using RUNESCAN instead of TRACK.NINEREALMS, as I think RUNESCAN is overall better, even though it depends on what you prefer as well ..

**Note: please do it for all OS, not only iOS**

johnnyluo:
fixes #896


enriquesouza (approved):
LGTM

PR #940, opened 2024-07-25
Retrieve network id from THORChain

### LINKED ISSUE #930: [BUG] THORChain retrieve ChainID from RPC endpoint
**Describe the bug**
Currently,  for THORChain , chain-id is set using WalletCore's coin.ChainID , which is stored in walletcore , usually it doesn't change , but THORChain is going to do a hardfork soon , thus the chainID will be changed

Retrieve THORChain's chain-ID from RPC endpoint  https://rpc.ninerealms.com/status

johnnyluo:
fixes #930 

When THORChain hardfork , it will change the chain id , this allow Vultisig continue to work without an update

### DIFF
--- a/VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
+++ b/VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
@@ -8,6 +8,7 @@
 import Foundation
 
 class ThorchainService: ThorchainSwapProvider {
+    var network: String = ""
     static let shared = ThorchainService()
     
     private var cacheFeePrice: [String: (data: UInt64, timestamp: Date)] = [:]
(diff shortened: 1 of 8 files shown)


VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
```diff
@@ -97,4 +98,14 @@ class ThorchainService: ThorchainSwapProvider {
         
         return .zero
     }
+    
+    func getTHORChainChainID() async throws -> String  {
+        if !network.isEmpty {
+            return network
+        }
+        let (data, _) = try await URLSession.shared.data(from: Endpoint.thorchainNetworkInfo)
```
enriquesouza, line 106: Can you cache it for a long time?

enriquesouza (approved):
LGTM, but we should cache it

VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
johnnyluo: it only does this once

PR #947, opened 2024-07-25
Default chains new

Rockindash:
Default chains added to settings, Issue #781


enriquesouza (approved):
LGTM

Select OK
UnSelect OK
Search OK
Create a new Vault, expect to have the default chains
Result -> OK

Issue #878, opened 2024-07-15
ADD: Save button for custom tokens on mac


realpaaao:
On mac the save button in the Token selection view is still missing.
Please add.
<img width="875" alt="image" src="https://github.com/user-attachments/assets/23380e49-343b-4b67-8a4a-53dec849c762">

PR #949, opened 2024-07-26
Save button added on Mac for Custom Token Sheet

Rockindash:
Save button added on Mac for Custom Token Sheet, Issue #878


johnnyluo (approved):
LGTM

PR #963, opened 2024-07-29
Label to Cosmos

enriquesouza:


johnnyluo:
Let's hold this one for next release , need to coordinate with android side , otherwise cosmos won't be able signed with Ios and Android

flypaper0 (approved):

VultisigApp/VultisigApp/Model/Chain.swift
```diff
@@ -119,7 +119,7 @@ enum Chain: String, Codable, Hashable, CaseIterable {
         case .bitcoinCash: return "BCH"
         case .litecoin: return "LTC"
         case .dogecoin: return "DOGE"
-        case .gaiaChain: return "GAIA"
```
johnnyluo, line 122: This one can't be changed , THORChain call it `GAIA`

PR #964, opened 2024-07-30
Blowfish solana

enriquesouza:
This is the implementation of Blowfish for Solana.


Rockindash (approved):

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
```diff
@@ -44,6 +44,54 @@ struct BlowfishService {
...
+        let body = try JSONEncoder().encode(blowfishRequest)
+        let dataResponse = try await Utils.asyncPostRequest(urlString: endpoint, headers: headers, body: body)
+        
+        do {
+            let response = try JSONDecoder().decode(BlowfishResponse.self, from: dataResponse)
+            return response
+        } catch let DecodingError.dataCorrupted(context) {
+            print("Data corrupted: \(context)")
+            throw DecodingCustomError.dataCorrupted(context)
+        } catch let DecodingError.keyNotFound(key, context) {
```
flypaper0, line 78: why you handling all decoding errors?

VultisigApp/VultisigApp/View Models/JoinKeysignViewModel.swift
```diff
@@ -273,4 +273,23 @@ class JoinKeysignViewModel: ObservableObject {
             chain: payload.coin.chain
         )
     }
+    
+    func blowfishSolanaTransactionScan() async throws -> BlowfishResponse? {
```
flypaper0, line 277: throwing function should not return optional value

VultisigApp/VultisigApp/View Models/SendCryptoVerifyViewModel.swift
```diff
@@ -43,6 +43,52 @@ class SendCryptoVerifyViewModel: ObservableObject {
...
+            }
+            
+            
+            return try await BlowfishService.shared.blowfishSolanaTransactionScan(
+                fromAddress: tx.fromAddress, zeroSignedTransaction: zeroSignedTransaction
+            )
+            
+        } catch {
+            switch error {
+            case KeysignPayloadFactory.Errors.notEnoughBalanceError:
```
flypaper0, line 75: KeysignPayloadFactory.Errors implement LocalizedError. You can use: self.errorMessage = error.localizedDescription

VultisigApp/VultisigApp/Views/Components/BlowfishSolanaWarningInformationNote.swift
```diff
@@ -0,0 +1,73 @@
...
+            HStack(spacing: 12) {
+                icon(response: response)
+                text(response: response)
+            }
+                .padding(12)
+                .background((response.aggregated?.warnings?.isEmpty ?? true) ? Color.green.opacity(0.35) : Color.warningYellow.opacity(0.35))
+                .cornerRadius(12)
+                .overlay(
+                    RoundedRectangle(cornerRadius: 12)
+                        .stroke((response.aggregated?.warnings?.isEmpty ?? true) ? Color.green : Color.warningYellow, lineWidth: lineWidth)
```
flypaper0, line 29: ViewModel can make BlowfishSolanaWarningInformationNote much simpler

VultisigApp/VultisigApp/Views/Keysign/KeysignMessageConfirmView.swift
```diff
@@ -26,7 +26,10 @@ struct KeysignMessageConfirmView: View {
                 
                 HStack {
                     Spacer()
-                    if (viewModel.keysignPayload?.coin.chainType == .EVM && blowfishResponse != nil) {
+                    if (
+                        (viewModel.keysignPayload?.coin.chainType == .EVM || viewModel.keysignPayload?.coin.chainType == .Solana) &&
```
flypaper0, line 30: Views should not contain logic. Only read properties from viewModels

VultisigApp/VultisigApp/Views/Keysign/KeysignMessageConfirmView.swift
```diff
@@ -47,6 +50,17 @@ struct KeysignMessageConfirmView: View {
                             isLoading = false
                         }
                     }
+                } else if (viewModel.keysignPayload?.coin.chainType == .Solana) {
+                    isLoading = true
+                    Task{
```
flypaper0, line 55: same here, that should be in viewModel

VultisigApp/VultisigApp/Views/Send/SendCryptoVerifyView.swift
```diff
@@ -44,6 +44,19 @@ struct SendCryptoVerifyView: View {
                         isLoading = false
                     }
                 }
+            } else if (tx.coin.chainType == .Solana) {
```
flypaper0, line 47: viewModel

VultisigApp/VultisigApp/Views/Send/SendCryptoVerifyView.swift
```diff
@@ -52,15 +65,24 @@ struct SendCryptoVerifyView: View {
...
+                
+            }
         }
     }
     
     var view: some View {
         VStack {
             fields
-            if (tx.coin.chainType == .EVM && blowfishResponse != nil) {
+            if ((tx.coin.chainType == .EVM || tx.coin.chainType == .Solana) && blowfishResponse != nil) {
```
flypaper0, line 85: pass blowfishResponse as method argument, to avoid optional checks

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
enriquesouza: It is because they don't have a typed payload, and I never know what is coming and or what is missing. So that helps whenever it breaks.

VultisigApp/VultisigApp/View Models/JoinKeysignViewModel.swift
enriquesouza: Removed the throws then

VultisigApp/VultisigApp/Views/Keysign/KeysignMessageConfirmView.swift
enriquesouza: Moved to the view model.

VultisigApp/VultisigApp/View Models/SendCryptoVerifyViewModel.swift
enriquesouza: Done

VultisigApp/VultisigApp/Chains/Solana.swift
```diff
@@ -112,6 +112,39 @@ enum SolanaHelper {
...
+        guard let pubkeyData = Data(hexString: vaultHexPubKey) else {
+            throw HelperError.runtimeError("public key \(vaultHexPubKey) is invalid")
+        }
+        
+        let inputData = try getPreSignedInputData(keysignPayload: keysignPayload)
+        let allSignatures = DataVector()
+        let publicKeys = DataVector()
+        
+        // Compile with zero signature
+        let signature = Data(hexString: "00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")
```
flypaper0, line 128: String(repeating: "0"

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
```diff
@@ -18,7 +18,7 @@ struct BlowfishService {
         origin: String,
         txObjects: [BlowfishRequest.BlowfishTxObject],
         simulatorConfig: BlowfishRequest.BlowfishSimulatorConfig? = nil
-    ) async throws -> BlowfishResponse? {
+    ) async -> BlowfishResponse? {
```
flypaper0, line 21: throws was fine here, let's not return optional instead?

flypaper0: `do` `catch`  should be on the top level only. eg in viewModel

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
```diff
@@ -28,30 +28,96 @@ struct BlowfishService {
...
+            
+            let endpoint = Endpoint.fetchBlowfishSolanaTransactions()
+            let headers = ["X-Api-Version" : "2023-06-05"]
+            let body = try JSONEncoder().encode(blowfishRequest)
+            let dataResponse = try await Utils.asyncPostRequest(urlString: endpoint, headers: headers, body: body)
+            
+            do {
+                let response = try JSONDecoder().decode(BlowfishResponse.self, from: dataResponse)
+                return response
+            } catch let DecodingError.dataCorrupted(context) {
```
flypaper0, line 86: still do not understand this json decoding handlers here

flypaper0: errorDescription has same text inside

enriquesouza: That is because Blowfish does not have a pattern or return in their JSON; each transaction returns a different JSON, like a tx for EVM, which is different from an EVM token, which is different from SOLANA, which is different for SOLANA tokens.
Sometimes, I need to find out where the error is when mapping the JSON to a struct, which helps me out.

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
enriquesouza: Blowfish is a 3rd party API; we have no control over it. I prefer to return it as nil since it was frequently breaking, so in that case, for Blowfish, it is better to only show something to break our transaction flow.

VultisigApp/VultisigApp/Views/Keysign/KeysignMessageConfirmView.swift
```diff
@@ -36,18 +36,10 @@ struct KeysignMessageConfirmView: View {
...
-                            isLoading = false
-                        } catch {
-                            print(error.localizedDescription)
-                            isLoading = false
-                        }
-                    }
-                } else {
+                isLoading = true
+                Task {
+                    try await viewModel.blowfishTransactionScan()
```
johnnyluo, line 41: if there is an exception somehow , this might cause the app the crash , because the exception has not been captured ,correct?

enriquesouza: Tks, I've fixed that one.

enriquesouza:
Tests:

Forcing warnings:

Sending:
image

Keysigning:
image

enriquesouza:
Success:
image
image

VultisigApp/VultisigApp/Views/Send/SendCryptoVerifyView.swift
```diff
@@ -33,35 +34,30 @@ struct SendCryptoVerifyView: View {
...
-                isLoading = true
-                Task{
-                    do {
-                        blowfishResponse = try await sendCryptoVerifyViewModel.blowfishEVMTransactionScan(tx: tx)
-                        isLoading = false
-                    } catch {
-                        print(error.localizedDescription)
-                        isLoading = false
-                    }
+            isLoading = true
```
Rockindash, line 37: All the methods can be moved to a separate method (we usually set ut as setData()) to make the code cleaner.

VultisigApp/VultisigApp/Services/Blowfish/BlowfishService.swift
```diff
@@ -11,22 +11,21 @@ import BigInt
 struct BlowfishService {
     static let shared = BlowfishService()
     
-    func scanTransactions
-    (
+    enum BlowfishServiceError: Error {
```
Rockindash, line 14: Enums within this struct can be moved to an extension of BlowfishService. Makes the more cleaner and structures since enums are not accessed that frequently.

VultisigApp/VultisigApp/Views/Send/SendCryptoVerifyView.swift
enriquesouza: done

Issue #955, opened 2024-07-28
[BUG] Wrong autodiscover on vault creation


realpaaao:
**Describe the bug**
When a Vault is created the autodiscover feature falsely discovers tokens on iOS:
10 assets for ETH
3 assets for BSC
without having any balance on these assets

**To Reproduce**
1. Create Vault
2. See error

**Expected behavior**
Doesn't dicover tokens if there is no balance

**Screenshots**
![image](https://github.com/user-attachments/assets/8e8b8e2f-7616-416e-88a8-3d754d314af0)

-version 42

enriquesouza:
Are you sure you don't have any dust?

realpaaao:
@enriquesouza just verified, that only happens when no default chains are selected at all.

enriquesouza:
https://github.com/vultisig/vultisig-ios/pull/975

PR #975, opened 2024-08-01
Auto discover with only native coins

enriquesouza:


enriquesouza:
https://github.com/vultisig/vultisig-ios/issues/955

Rockindash (approved):

PR #976, opened 2024-08-01
Added check for physical device

Rockindash:
Added additional check for physical devices to bypass password requirements for simulators.


enriquesouza (approved):
LGTM

Issue #667, opened 2024-06-11
ADD UX Flow: scanning address opens send view


realpaaao:
For easier UX flow please add a logic that when a QR code of an address gets scanned with the QR scanner it automatically opens the send view with the correct chain selected.

PR #978, opened 2024-08-02
Send QR code

Rockindash:
Send flow added to QR code scan, Issue #667

### DIFF
--- a/VultisigApp/VultisigApp/Views/GeneralCodeScannerView.swift
+++ b/VultisigApp/VultisigApp/Views/GeneralCodeScannerView.swift
@@ -19,6 +19,8 @@ struct GeneralCodeScannerView: View {
     @Binding var shouldJoinKeygen: Bool
     @Binding var shouldKeysignTransaction: Bool
     @Binding var shouldSendCrypto: Bool
+    @Binding var selectedChain: Chain?
+    let sendTX: SendTransaction
     
     @State var isGalleryPresented = false
     @State var isFilePresented = false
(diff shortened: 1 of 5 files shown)


VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
```diff
@@ -179,9 +181,34 @@ struct SendCryptoView: View {
...
         await sendCryptoViewModel.loadGasInfoForSending(tx: tx)
     }
     
+    private func presetData() {
+        guard let chain = selectedChain else {
+            selectedChain = nil
+            return
+        }
+        
+        let selectedCoin = vault.coins
```
enriquesouza, line 194:
```
guard let selectedCoin = vault.coins.first(where: { $0.chain == chain && $0.isNativeToken }) else {
    selectedChain = nil
    return
}
    ```

File: VultisigApp/VultisigApp/Views/GeneralCodeScannerView.swift [hunk 2]
```diff
@@ -114,28 +118,59 @@ struct GeneralCodeScannerView: View {
...
+        shouldJoinKeygen = false
+        showSheet = false
+        checkForAddress()
+    }
+    
+    private func checkForAddress() {
+        let address = deeplinkViewModel.address ?? ""
+        sendTX.toAddress = address
+        
+        for asset in settingsDefaultChainViewModel.baseChains.sorted(by: {
```
enriquesouza, line 149: it is better for performance if you set the sorted value in a variable before interact with it. 

like 

```
let sortedAssets = settingsDefaultChainViewModel.baseChains.sorted ...
```

enriquesouza (changes requested):
Just a few improvements IMO

enriquesouza (approved):
LGTM

Issue #916, opened 2024-07-20
IMPPROVE: Vault backup export


realpaaao:
When backing up the vault please open up the `share sheet` instead saving it directly to the device.
So the user can decide to directly share the vault share or save it to device. Similar as on the share for QR codes.
![image](https://github.com/user-attachments/assets/af3d8f54-8eac-4351-864c-173fa5b31fa4)

johnnyluo:
fixed in #982

PR #982, opened 2024-08-03
Vault backup options

Rockindash:
Share sheet added for Vault Backup Share, Issue #916


VultisigApp/VultisigApp/Views/Vault/BackupPasswordSetupView.swift
```diff
@@ -57,23 +63,6 @@ struct BackupPasswordSetupView: View {
 #if os(macOS)
         .padding(.horizontal, 25)
 #endif
-        .fileExporter(
```
johnnyluo, line 60: once this get removed , on IOS , the file is named always "file.bak" , you need to update the one in `exportFile`

johnnyluo (changes requested):
see comments

johnnyluo (approved):
LGTM

PR #986, opened 2024-08-04
Solana tokens issues (fix)

enriquesouza:
Adding another end point since we make a lot of calls to RPC in solana, so we can split the calls and then we don't hit the limit. Also added try catches because it is being impossible to find the source of the error.


johnnyluo (approved):
LGTM

PR #987, opened 2024-08-04
ens resolver

enriquesouza:
This resolves the ENS name to send

### DIFF
--- a/VultisigApp/VultisigApp/Extensions/RpcEvmServiceEnsServiceExtension.swift
+++ b/VultisigApp/VultisigApp/Extensions/RpcEvmServiceEnsServiceExtension.swift
@@ -60,25 +60,3 @@ extension RpcEvmService {
         }
     }
 }
-
-extension String {
-    func namehash() -> String {
-        // Split the ENS name into labels
-        let labels = self.split(separator: ".").reversed()
-        
-        // Initialize the node as 32 bytes of zero data
-        var node: Data = Data(repeating: 0, count: 32)
-        
-        for label in labels {
-            // Convert the label to Data, hash it, and get the hex representation
-            let labelData = Data(label.utf8)
-            let labelHash = labelData.sha3(.keccak256)
-            
-            // Combine the current node hash with the label hash and hash again
-            node = (node + labelHash).sha3(.keccak256)
-        }
-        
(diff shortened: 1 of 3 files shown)


enriquesouza:
Tests:
https://etherscan.io/name-lookup-search?id=vitalik.eth

Resolved Address:
 Address0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 
 
 Test code:
 ```
 Task {
            let ensName = "vitalik.eth"
            let namehash = ensName.namehash()
            print("Namehash for \(ensName): \(namehash)")
            
            let factory = try EvmServiceFactory.getService(forChain: Chain.ethereum)
            let address = try await factory.resolveENS(ensName: ensName)
            
            print("Resolved address \(address)")
        }
 ```

VultisigApp/VultisigApp/Extensions/RpcEvmServiceEnsServiceExtension.swift
```diff
@@ -0,0 +1,84 @@
...
+            
+            // Convert the address data to a hex string and return
+            return "0x" + addressData.toHexString()
+        } else {
+            throw RpcEvmServiceError.rpcError(code: -1, message: "Invalid address data")
+        }
+    }
+}
+
+extension String {
```
johnnyluo, line 64: we have a file call "StringExtensions.swift" why not put this method over there?

johnnyluo (approved):
This mostly looks OK to me , but I don't think this PR is complete , you still need to have logic on the UI part to actually resolve the input name to address.

I suppose you can raise another PR for that , I will approve this one for now

VultisigApp/VultisigApp/Extensions/RpcEvmServiceEnsServiceExtension.swift
enriquesouza: Moved

enriquesouza:
> This mostly looks OK to me , but I don't think this PR is complete , you still need to have logic on the UI part to actually resolve the input name to address.
> 
> I suppose you can raise another PR for that , I will approve this one for now

I will break it down in back end and UI, easier to review. I will have to change in many places to validate the address.

PR #991, opened 2024-08-05
ens ui

enriquesouza:


VultisigApp/VultisigApp/View Models/SendTransaction.swift
```diff
@@ -20,10 +20,37 @@ class SendTransaction: ObservableObject, Hashable {
...
+            let address = try await factory.resolveENS(ensName: ensName)
+            
+            print("Resolved address \(address)")
+            
+            return address
+            
+        } catch {
+            
+            print("Error to extract the DOMAIN ADDRESS: \(error.localizedDescription)")
+            return ""
```
johnnyluo, line 49: maybe just return "toAddresss" here

johnnyluo:
shouldn't we resolve the ENS name before user click "Next" button on send view?

VultisigApp/VultisigApp/View Models/SendTransaction.swift
enriquesouza: I will return toAddress from now.

enriquesouza:
> shouldn't we resolve the ENS name before user click "Next" button on send view?

In that scenario I will create an Address Service, so we can call it and remove duplication

Rockindash (approved):

PR #1000, opened 2024-08-07
Address Service

enriquesouza:


VultisigApp/VultisigApp/Services/AddressService.swift
```diff
@@ -0,0 +1,89 @@
...
+//  Created by Enrique Souza Soares on 06/08/24.
+//
+
+import Foundation
+import SwiftUI
+import WalletCore
+
+public class AddressService: ObservableObject {
+    
+    static let shared = AddressService()
```
johnnyluo, line 14: Please remove this singleton implementation , this one doesn't need to be singleton. We should avoid to use singleton whenever we can, not a good pattern

VultisigApp/VultisigApp/Services/AddressService.swift
```diff
@@ -0,0 +1,89 @@
...
+//  VultisigApp
+//
+//  Created by Enrique Souza Soares on 06/08/24.
+//
+
+import Foundation
+import SwiftUI
+import WalletCore
+
+public class AddressService: ObservableObject {
```
johnnyluo, line 12: Why it inherit from `ObservableObject` ? there is no properties on this service

VultisigApp/VultisigApp/Services/AddressService.swift
```diff
@@ -0,0 +1,89 @@
...
+            
+        } catch {
+            
+            print("Error to extract the DOMAIN ADDRESS: \(error.localizedDescription)")
+            return address
+            
+        }
+    }
+    
+    func validateAddress(address: String, chain: Chain) -> Bool {
```
johnnyluo, line 45: This method is misleading to say at least , `isNameService()` will cause this method to by pass the validation as a whole ,  even something fit the `nameService` pattern , it doesn't mean it is a valid address. 

Instead , I think we should resolve the name service , and then perform address validation

johnnyluo (changes requested):
check comments

VultisigApp/VultisigApp/Services/AddressService.swift
enriquesouza: I only resolve it on the validated form; otherwise, I will need to make them all async, and we will make more calls to RPC.

johnnyluo (approved):
LGTM

PR #1008, opened 2024-08-08
[Core] RatesProvider

flypaper0:


VultisigApp/VultisigApp/Services/CryptoPriceService.swift
```diff
@@ -1,120 +1,124 @@
...
+            currencies: currencies
+        )
+
+        let (data, _) = try await URLSession.shared.data(from: url)
+        let response = try JSONDecoder().decode([String: [String: Double]].self, from: data)
+
+        try await RateProvider.shared.save(rates: mapRates(response: response))
+    }
+
+    func fetchPrices(contracts: [String], chain: Chain) async throws {
```
johnnyluo, line 91: Previously , we cache the price for a while , with this new implementation , rates are stored in database , and how often do we refresh the rates?

VultisigApp/VultisigApp/Views/Components/VaultDetailBalanceContent.swift
```diff
@@ -8,8 +8,8 @@
 import SwiftUI
 
 struct VaultDetailBalanceContent: View {
-    let vault: Vault
-    
+    @ObservedObject var vault: Vault
```
johnnyluo, line 11: why we need to have vault as an ObservableObject? will it be changed ?

johnnyluo:
Good work @flypaper0  , it is a pretty big PR , let's do more tests.

VultisigApp/VultisigApp/Services/CryptoPriceService.swift
flypaper0: We're refreshing rates in same places where we refreshing balances. Check BalanceService
If we fetch balances for whole value - fetching all rates
For single coin - fetching rate for this coin only

VultisigApp/VultisigApp/Views/Components/VaultDetailBalanceContent.swift
flypaper0: As we're not storing rate inside coin anymore, we should observe vault to track changes. 
I'm trigger vault update with `vault.objectWillChange.send()` in CryptoPriceService

VultisigApp/VultisigApp/Model/Coin.swift
```diff
@@ -172,11 +173,11 @@ class Coin: ObservableObject, Codable, Hashable {
     
     func fiat(value: BigInt) -> Decimal {
         let decimal = decimal(for: value)
-        return decimal * Decimal(priceRate)
+        return RateProvider.shared.fiatBalance(value: decimal, coin: self)
```
enriquesouza, line 176: I would declare the RateProvider as a var, so we don't have to change everywhere if we need to change it.

enriquesouza (approved):

PR #1009, opened 2024-08-08
CoinService

enriquesouza:
This should separate the logic to add, remove coins. So we can reuse it in different places.

### DIFF
--- a/VultisigApp/VultisigApp/Services/CoinService.swift
+++ b/VultisigApp/VultisigApp/Services/CoinService.swift
@@ -1,0 +1,108 @@
+//
+//  CoinService.swift
+//  VultisigApp
+//
+//  Created by Enrique Souza Soares on 07/08/24.
+//
+
+import Foundation
+
+struct CoinService {
+    
+    static func removeCoins(coins: [Coin], vault: Vault) async throws {
+        for coin in coins {
+            if let idx = vault.coins.firstIndex(where: { $0.ticker == coin.ticker && $0.chain == coin.chain }) {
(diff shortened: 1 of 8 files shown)


VultisigApp/VultisigApp/Views/Vault/ChainDetailView.swift
```diff
@@ -273,6 +273,10 @@ struct ChainDetailView: View {
             for coin in group.coins {
                 await viewModel.loadData(coin: coin)
             }
+            
+            for coin in group.coins where coin.isNativeToken {
+                try await CoinService.addDiscoveredTokens(nativeToken: coin, to: vault)
```
johnnyluo, line 278: This is a try , but didn't catch error , which might cause crash

Rockindash (approved):

VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -0,0 +1,109 @@
...
+        if let priceProviderId {
+            newCoin.priceProviderId = priceProviderId
+        }
+        // Save the new coin first
+        try await Storage.shared.save(newCoin)
+        vault.coins.append(newCoin)
+        return newCoin
+    }
+    
+    static func addDiscoveredTokens(nativeToken: Coin, to vault: Vault) async throws  {
```
johnnyluo, line 84: this function doesn't throw exception , why it has been marked as `throws`?

johnnyluo (changes requested):

johnnyluo (approved):
LGTM

Issue #980, opened 2024-08-02
[BUG] Move titles out of Tab bar on Mac


Rockindash:
**Describe the bug**
All the buttons, the full heading bar should be moved down to the primary view on Mac. Don't put anything on the chrome (see top left and top right). It should just be the app name.

**Screenshots**
![image](https://github.com/user-attachments/assets/a4db89dc-7094-4698-bc7b-83e715805c08)

PR #1017, opened 2024-08-09
Mac new header

Rockindash:
Navigation header moved to main screen from Toolbar on Mac, Issue  #980


enriquesouza (approved):
LGTM, but we should use a base class for headers so we DRY.

Issue #1028, opened 2024-08-12
[BUG] iPad address QR code cut off


realpaaao:
**Describe the bug**
When iPad is in landscape mode, the QR code doesn't scale and is cut off

**Expected behavior**
Have QR code visable

**Screenshots**
![image](https://github.com/user-attachments/assets/2a154c8a-ece7-49ce-b984-c16a76209452)

-v45

PR #1031, opened 2024-08-13
QR code cut off issue fixed for iPad Landscape

Rockindash:
QR code cut off issue fixed for iPad Landscape, Issue #1028


johnnyluo (approved):
LGTM

Issue #1010, opened 2024-08-08
[BUG] Change pop up message for Copy


realpaaao:
**Describe the bug**
If copying a chain address ther should not be an pop up enforced. 
Please change that just a small notification pops up that it is copied, instead of blocking the window with a pop-up.

**Screenshots**
![image](https://github.com/user-attachments/assets/5600ac4c-1511-4e75-9f7a-ff3ba00c2138)

-v44

PR #1039, opened 2024-08-14
Popup cell added to other copy options

Rockindash:
Popup capsule popup added for copy options, Issue #1010


johnnyluo (approved):
LGTM

johnnyluo:
Also keep in mind , we can long press the chain cell to copy address as well , in the vault view
That one seems still show an alert

PR #1045, opened 2024-08-15
Fix bug the refresh on EVM detail page cause it to crash

johnnyluo:


VultisigApp/VultisigApp/Services/CoinService.swift
```diff
@@ -96,6 +96,10 @@ struct CoinService {
             
             for token in tokens {
                 do {
+                    let existingCoin =  vault.coins.first{$0.chain == token.chain && $0.ticker == token.ticker}
```
flypaper0, line 99: we're doing this check a lot. What you think about helper like: vault.isExist(coin: CoinMeta) -> Bool ?

flypaper0 (approved):

VultisigApp/VultisigApp/Services/CoinService.swift
johnnyluo: Good idea , we can add it

johnnyluo: `vault.coin(for: token)` , there is already one

PR #1048, opened 2024-08-16
[Send] Single domain resolver with TNS support 

flypaper0:


Rockindash (approved):

johnnyluo:
@flypaper0 , based on this PR , only when people are sending transactions on THORChain it will use TNS ,  But in reality is , TNS can setup a name which point to address on all the chains THORChain can support .
For example, the following name 
https://midgard.ninerealms.com/v2/thorname/lookup/vi

it has an address for ETH , and also an address for THORChain, which means name `vi` can be used for both THORChain and ETH chain
Which means for any address that is not valid address, we can potentially send a request to TNS to try to resolve it

Ideally, I think we should have an UI in the app , allow user to type a THORName to search for the address , instead of implicitly resolve it from TNS

Anyway , what you got here is fine, but it can be improved

PR #1066, opened 2024-08-21
[Send] GasLimit estimations + fiat fee

flypaper0:

### DIFF
--- a/VultisigApp/VultisigApp/Services/BlockChainService.swift
+++ b/VultisigApp/VultisigApp/Services/BlockChainService.swift
@@ -127,7 +127,7 @@ final class BlockChainService {
             
         case .zksync:
             let service = try EvmServiceFactory.getService(forChain: coin.chain)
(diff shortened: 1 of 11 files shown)


enriquesouza (approved):
LGTM

VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -121,13 +121,13 @@ final class BlockChainService {
         case .ethereum, .avalanche, .bscChain, .arbitrum, .base, .optimism, .polygon, .blast, .cronosChain:
             let service = try EvmServiceFactory.getService(forChain: coin.chain)
             let (gasPrice, priorityFee, nonce) = try await service.getGasInfo(fromAddress: coin.address)
-            let gasLimit = normalizeGasLimit(coin: coin, action: action)
+            let gasLimit = gasLimit ?? normalizeGasLimit(coin: coin, action: action)
```
johnnyluo, line 124: Since `gasLimit` it is just a limit , if the transaction spend less then the `gasLimit` the remainder will be returned, thus let's do this
the `gasLimit` we got from `estimate_gas`, and our default value , whichever is larger , let's use the larger one

PR #1071, opened 2024-08-22
[Core] L2 chain icons logic

flypaper0:


johnnyluo:
Hey @flypaper0 , so when user are choosing a chain ,  chain selection list , can we show L2 with L2 chain logo? 

image
In this case , for arbitrum , can we just show arbitrum logo?

enriquesouza (approved):
LGTM

PR #1075, opened 2024-08-23
Fix a bug that cause sending issue

johnnyluo:


enriquesouza (approved):
LGTM

VultisigApp/VultisigApp/View Models/SendCryptoVerifyViewModel.swift
```diff
@@ -126,7 +126,7 @@ class SendCryptoVerifyViewModel: ObservableObject {
                 transactionType: tx.transactionType
             )
 
-            let keysignPayload = try await KeysignPayloadFactory().buildTransfer(
+            keysignPayload = try await KeysignPayloadFactory().buildTransfer(
```
flypaper0, line 129: we can just return keysignPayload here without setting it into var

flypaper0 (approved):

Issue #1046, opened 2024-08-16
[Feature] Implement drag and drop for Mac Import options


Rockindash:
**Is your feature request related to a problem? Please describe.**
Currently, for Mac import options file importer is the only option to import files.

**Describe the solution you'd like**
Implement drag and drop for a much convenient flow.

PR #1096, opened 2024-08-28
Mac drag and drop

Rockindash:
Drag and Drop to upload file option added for Mac, Issue #1046


Rockindash:
SCREENSHOTS:

- File importer
Screenshot 2024-08-28 at 1 30 56 AM

- Text Fields
Screenshot 2024-08-28 at 1 30 38 AM

johnnyluo:
Good work @Rockindash, Can you please investigate , is there a way we don't need to remove the sandbox?

Rockindash:
We only need to remove it while running it via Xcode. The reason is Apple does not allow file access due to some security reasons while on a sandbox environment. For users it shouldn’t be a problem.
It seems like Apple doesn’t has this restriction while using file importer but for drag and drop they do.

johnnyluo (approved):
LGTM

PR #1101, opened 2024-08-28
[Send] Advanced gas settings

flypaper0:


flypaper0:
> it doesn't compile yet , can you please fix it first?

Just sussesfully sent transaction with increased gasLimit and low txMode. 
https://etherscan.io/tx/0x8262d4474510c05c1ce9dc4fef9244c74054da23917b378efa9b50a351deecef

Or you talking about MacOS version? I'll fix it asap

johnnyluo:
This is what I got when trying to send USDC
image

see the USD amount

VultisigApp/VultisigApp/Chains/evm.swift
```diff
@@ -77,18 +77,19 @@ class EVMHelper {
...
         }
-        guard case .Ethereum(let maxFeePerGasWei,
+        guard case .Ethereum(let baseFee,
                              let priorityFeeWei,
                              let nonce,
                              let gasLimit) = keysignPayload.chainSpecific
         else {
             throw HelperError.runtimeError("fail to get Ethereum chain specific")
         }
+        let maxFeePerGas = baseFee + priorityFeeWei
```
johnnyluo, line 87: I think this is the place that will break IOS~Android cross signing, correct? 
Why not calculate maxFeePerGas from outside , and then pass it in , like we did before?

VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
```diff
@@ -54,8 +56,12 @@ struct SendCryptoView: View {
                     backButton
                 }
             }
-            
-            if sendCryptoViewModel.currentIndex==3 {
+            if showFeeSettings {
```
johnnyluo, line 59: Is there a reason why we don't add this to MacOS?

VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -37,136 +41,175 @@ final class BlockChainService {
...
-            let (gasPrice, priorityFee, nonce) = try await service.getGasInfo(fromAddress: coin.address)
+            let baseFee = try await service.getBaseFee()
+            let (_, defaultPriorityFee, nonce) = try await service.getGasInfo(fromAddress: coin.address, mode: feeMode)
             let gasLimit = gasLimit ?? normalizeGasLimit(coin: coin, action: action)
-            let normalizedGasPrice = normalize(gasPrice, action: action)
-            return .Ethereum(maxFeePerGasWei: normalizedGasPrice, priorityFeeWei: normalizePriorityFee(priorityFee,coin.chain), nonce: nonce, gasLimit: gasLimit)
-            
+            let priorityFeesMap = try await service.fetchMaxPriorityFeesPerGas()
+            let priorityFee = priorityFeesMap[feeMode] ?? defaultPriorityFee
+            return .Ethereum(baseFee: baseFee, priorityFeeWei: priorityFee, nonce: nonce, gasLimit: gasLimit)
```
johnnyluo, line 169: if you do `return .Ethereum(baseFee:  (baseFee+priorityFee) , priorityFeeWei: priorityFee, nonce: nonce, gasLimit: gasLimit)` then we probably don't need to change it in `evm.swift`

VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
flypaper0: just want to make it in separate PR

VultisigApp/VultisigApp/Services/BlockChainService.swift
flypaper0: should work

johnnyluo (approved):
LGTM

Issue #1113, opened 2024-08-31
[ADD] QR export name standard


realpaaao:
Include a standard for naming QR code on exports:

Vault QR:
`VultisigQR-{VaultName}-{UID}.png`

keygen QR:
`VaultKeygen-{VaultName}-{UID}-{date}.png`

keysign QR:
`VaultSend-{VaultName}-{UID}-{date}.png`

PR #1124, opened 2024-09-05
Qr name standard

Rockindash:
QR export name updated as per the standard, Issue #1113

### DIFF
--- a/VultisigApp/VultisigApp/Views/Keygen/KeygenView.swift
+++ b/VultisigApp/VultisigApp/Views/Keygen/KeygenView.swift
@@ -26,6 +26,7 @@ struct KeygenView: View {
     
     let progressTotalCount: Double = 4
     let version = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String
+    let build = Bundle.main.infoDictionary?["CFBundleVersion"] as? String
 #if os(iOS)
     private var idiom : UIUserInterfaceIdiom { UIDevice.current.userInterfaceIdiom }
 #endif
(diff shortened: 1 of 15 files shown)


VultisigApp/VultisigApp/Views/Keygen/PeerDiscoveryView.swift
```diff
@@ -61,7 +61,11 @@ struct PeerDiscoveryView: View {
             // only show the QR share button when it is in peer discovery
             if viewModel.status == .WaitingForDevices {
                 ToolbarItem(placement: Placement.topBarTrailing.getPlacement()) {
-                    NavigationQRShareButton(title: "joinKeygen", renderedImage: shareSheetViewModel.renderedImage)
+                    NavigationQRShareButton(
+                        vault: vault,
```
johnnyluo, line 65: Keep in mind , during keygen , `vault` still don't have public key ecdsa and public key eddsa, please make sure it still works in that scenario

VultisigApp/VultisigApp/Views/Keygen/KeygenView.swift
```diff
@@ -115,14 +116,17 @@ struct KeygenView: View {
     }
     
     var appVersion: some View {
-        Text("Vultisig APP V\(version ?? "1")")
-            .textCase(.uppercase)
-            .font(.body14Menlo)
-            .foregroundColor(.turquoise600)
+        return VStack {
```
johnnyluo, line 119: can we please add this to keysign as well

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Views/Keygen/KeygenView.swift
Rockindash: Created a separate issue for that since it'll require shuffling a bit of stuff around from UI perspective.
https://github.com/vultisig/vultisig-ios/issues/1126

Issue #1106, opened 2024-08-29
[ADD] Number abbrevations for tokens


realpaaao:
When having tokens where the amount is very big the screen gets too small and the numbers will have a conflit.
To have that congruent over all platforms and mobile, desktop, please introduce number abbrevations.
1,000,000 => 1M
1,000,000,000 => 1B

Reference from Web on mobile :

![image](https://github.com/user-attachments/assets/d75b7a2d-8111-4ef0-b33f-fb6091d59dbf)

PR #1130, opened 2024-09-05
Abbreviations added for coin balance and fiat value (M and B)

Rockindash:
Abbreviations added for coin balance and fiat value (M and B), Issue #1106


johnnyluo (approved):
LGTM

PR #1163, opened 2024-09-16
[FastVault] EdDSA fix

flypaper0:


VultisigApp/VultisigApp/Model/Curve.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Curve.swift
+//  VultisigApp
+//
+//  Created by Artur Guseinov on 16.09.2024.
+//
+
+import Foundation
+
+enum Curve {
```
johnnyluo, line 10: We already have a type call `KeyType` , why create a new one here?  you can get it from using `chain.signingKeyType`

johnnyluo (approved):
LGTM

PR #1164, opened 2024-09-16
[FastVault] Simple onboarding flow 

flypaper0:


VultisigApp/VultisigApp/Views/New Wallet/CreateVaultView.swift
```diff
@@ -41,7 +44,8 @@ struct CreateVaultView: View {
     
     var newVaultButton: some View {
         NavigationLink {
-            SetupCardsView(tssType: .Keygen)
+            let vaultName = Vault.getUniqueVaultName(modelContext: modelContext)
```
johnnyluo, line 47: Please don't put any logic related stuff in `NavigationLink` , this might get executed many many times. 
Sometimes it might even cause the app to freeze.

Please define a variable , and make sure it get set when screen appears

johnnyluo (changes requested):
see

VultisigApp/VultisigApp/Views/New Wallet/CreateVaultView.swift
flypaper0: interesting, fixed

PR #1175, opened 2024-09-18
[FastVault] Skip QR code step for Fast mode

flypaper0:


johnnyluo (approved):
LGTM

johnnyluo (changes requested):
actually , realize on confirm screen , we should disable "continue" button , if server didn't join, can you please update that?

johnnyluo (approved):
LGTM

Issue #1132, opened 2024-09-08
[ADD]: Part-assigned Vaults


jpthor:
# Context

Vult shares are hard to reason about, we should think about the UX around them.
People may confuse vault shares and get stuck, thinking they are backed up. 

- [ ] Part-assigned generation
- [ ] Vult Share Exports have Part
- [ ] Part Labels

# Desired

Use the concept of parts to help users visually identify them. Each vault share is assigned a part, which is in the name as well as label.

# Rules

## New vaults
1) Starting device is always `part1`
2) Assign the next parts in order of joining session
3) Vault Part can never be re-assigned. 

## Reshare: 
1) Starting device is new `part1`
2) Assign the next parts in order of joining session

<img width="424" alt="image" src="https://github.com/user-attachments/assets/786d1143-2265-4454-a50f-8c9403706586">

johnnyluo:
discussion happens https://discord.com/channels/1203844257220395078/1282301697967788134

SxMShaDoW:
I wonder if adding colors as a secondary indicator could be useful. With that being said - colors run into a "color blind" scenario - so we would probably run out of colors for N of N but if it wasn't a main indicator (where parts is) could still be useful. Just my random 2 cents.

PR #1184, opened 2024-09-19
Part assigned label

Rockindash:
Part-assigned label added to VaultsView, Issue #1132


Rockindash:
I updated the label arrangement to utilize the limited sapce we have for the Vault Cell.
For the fast sign label, we might or might not have it visible. But for part assigned labels they are always visible. With placing part assigned label in the middle we only have less than half of cell's width. Which is not ideal since vault's name can be lenghtly. SO I updated them accordingly, as seen in the screenshot.

IMG_2351

johnnyluo:
can you please rebase it @Rockindash ? also can you please change "Fast Sign" -> "Fast"

johnnyluo (approved):
LGTM

PR #1188, opened 2024-09-20
[MacOS] Advanced gas settings

flypaper0:


VultisigApp/VultisigApp/Views/Send/SendGasSettingsView.swift
```diff
@@ -15,12 +15,15 @@ protocol SendGasSettingsOutput {
 struct SendGasSettingsView: View {
     @Environment(\.presentationMode) var presentationMode
 
-    @ObservedObject var viewModel: SendGasSettingsViewModel
+    @StateObject var viewModel: SendGasSettingsViewModel
```
Rockindash, line 18: When passing a ViewModel reference it is preferred to use ObservedObject instead.

flypaper0: that's how I fixed bug from this thread https://discord.com/channels/1203844257220395078/1203844459687706644/1287481437997629584

It looks like swiftUI issue on macOS, I'll show on sync

Rockindash (approved):

Issue #1133, opened 2024-09-08
[ADD]: Custom File Extension


jpthor:
# Vult share has custom extension `.vult`

This will allow "open in" iOS, MAC:

https://chatgpt.com/share/1c3342cd-81bb-4e5c-b0a0-c47b167847d5

johnnyluo:
implemented at #1210

PR #1210, opened 2024-09-25
Vult extension

Rockindash:
VULT extension support implemented for Mac app, Issue #1133


VultisigApp/VultisigApp/macOS/View/VultisigApp+macOS.swift
```diff
@@ -9,36 +9,24 @@
...
+            .environmentObject(homeViewModel)
+            .environmentObject(settingsDefaultChainViewModel)
+            .environmentObject(vultExtensionViewModel)
+            .environmentObject(macCheckUpdateViewModel)
+            .environmentObject(macCameraServiceViewModel)
+            .buttonStyle(BorderlessButtonStyle())
+            .frame(minWidth: 900, minHeight: 600)
+            .onAppear{
+                NSWindow.allowsAutomaticWindowTabbing = false
+            }
```
johnnyluo, line 29: so we don't need to monitor `scenePhase`?

Rockindash: We do but only for iOS. We use that observer to lock the app when user dismisses the app. On Mac, user can have multiple apps opened at once.

johnnyluo:
Just tested,  on MacOS , my backup file still have three .bak
image

Also the file extention is not in `.vult`

PR #1264, opened 2024-10-04
[Swap] Cacao Eth route

flypaper0:


johnnyluo:
If this PR is to add ETH -> Cacao , I think it is ok

But I think we need to fix the Maya swap route appropriately, can you also check the following? 
Cacao -> ETH-USDT
Cacao -> ETH-USDC
Cacao -> ARB
Cacao -> ERC20 on ARB 
Can you please go through them to get it fixed?

flypaper0:
> If this PR is to add ETH -> Cacao , I think it is ok
> 
> But I think we need to fix the Maya swap route appropriately, can you also check the following? Cacao -> ETH-USDT Cacao -> ETH-USDC Cacao -> ARB Cacao -> ERC20 on ARB Can you please go through them to get it fixed?

yes, here is only ETH -> CACAO

johnnyluo (approved):
LGTM

PR #1297, opened 2024-10-09
Vaults filter logic updated

Rockindash:
Vaults filter logic updated to hide vaults that are already in a folder


johnnyluo (approved):
LGTM

Issue #1156, opened 2024-09-15
[BUG] No automatic sorting after balance detection


realpaaao:
**Describe the bug**
When new balance is detected, the chains do not get sorted until entering another view and retuning to main view.

https://github.com/user-attachments/assets/24ee4959-f7e4-4a8c-b5c9-7445fee41019

-v51

realpaaao:
QA done.

PR #1322, opened 2024-10-12
Reorder list logic updated

Rockindash:
Reorder list after balance fetch on Home View logic updated, Issue #1156


VultisigApp/VultisigApp/iOS/View/Vault/VaultDetailView+iOS.swift
```diff
@@ -40,7 +40,9 @@ extension VaultDetailView {
     }
     
     var cells: some View {
-        return ForEach(viewModel.groups, id: \.id) { group in
+        return ForEach(viewModel.groups.sorted(by: {
```
flypaper0, line 43: Groups is sorting inside `categorizeCoins`. Do we need to do it one more time here?

johnnyluo (approved):
LGTM

PR #1326, opened 2024-10-13
Update RegisterVaultView.swift

realpaaao:
@Rockindash my bad of not updating the figma label. This should be the label.

### DIFF
--- a/VultisigApp/VultisigApp/Views/Vault/RegisterVaultView.swift
+++ b/VultisigApp/VultisigApp/Views/Vault/RegisterVaultView.swift
@@ -67,7 +67,7 @@ struct RegisterVaultView: View {
     }
     
     var label: some View {
-        FilledButton(title: "complete")
+        FilledButton(title: "Save Vault QR")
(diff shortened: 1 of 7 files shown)


VultisigApp/VultisigApp/Views/Vault/RegisterVaultView.swift
```diff
@@ -67,7 +67,7 @@ struct RegisterVaultView: View {
     }
     
     var label: some View {
-        FilledButton(title: "complete")
+        FilledButton(title: "Save Vault QR")
```
johnnyluo, line 70: You need to create an item in `Localizable.strings` , and use the name here , so the text is translated into other languages

Rockindash: Updated the localizable files for all the languages.

johnnyluo (approved):
LGTM

johnnyluo:
@Rockindash can you resolve the conflict in this PR , after that , you can merge it

PR #1333, opened 2024-10-14
[FastVault] Biometry

flypaper0:


VultisigApp/VultisigApp/Views/Keygen/FastVaultEnterPasswordView.swift
```diff
@@ -19,13 +19,19 @@ struct FastVaultEnterPasswordView: View {
...
+
     var view: some View {
         VStack {
             passwordField
             Spacer(minLength: 20)
             disclaimer
             buttons
         }
+        .onAppear {
+            tryAuthenticate()
```
johnnyluo, line 33: 1. Not everyone want to save their password , so we have to provide an option , a checkbox on the screen , when user tick it , after they put in their password for the first time  then we save the password to keychain , otherwise keep as it is
2. we need to save the information somewhere , if user didn't save their password to keychain , then don't request user to authenticate

johnnyluo: image
image

johnnyluo (changes requested):
see the comments

VultisigApp/VultisigApp/Views/Keygen/FastVaultEnterPasswordView.swift
johnnyluo: I think here , the issue still has not been addressed , like if I didn't enable Biometrics , then I should not been prompt about Permission to access keychain 

If I already enable BioMetrics , then we should prompt permission to access keychain , but no password dialog is needed

VultisigApp/VultisigApp/iOS/View/EditVaultView+iOS.swift
```diff
@@ -26,6 +26,7 @@ extension EditVaultView {
                 backupVault
                 editVault
                 reshareVault
+                biometrySelectionCell
```
johnnyluo, line 29: my macOS app doesn't have biometric setting

johnnyluo (approved):
LGTM

Issue #1278, opened 2024-10-07
[IMPROVE] Error message on tokens


realpaaao:
If you want to send tokens of a repsective chain and don't have the native asset/not enough to cover the gas fees, an unappropriate error gets prompted: 
![Image](https://github.com/user-attachments/assets/608d8cc9-0dfe-480e-9604-7b9876031d49)

Please change the error prompt to `Not enough gas to cover transaction`

realpaaao:
QA done.

PR #1342, opened 2024-10-16
Localization value for "nonNegativeFeeError" updated.

Rockindash:
Error message value updated for insufficient tokens, Issue #1278


johnnyluo (approved):
LGTM

Issue #1272, opened 2024-10-06
[UPDATE] logic of check for asset amount on send


realpaaao:
Please update the logic for the amount check of assets.
Currently: Checks after each input, leading to annoying UX when inserting `.` or erasing everything.
Wanted: check once when user leaves input label

https://github.com/user-attachments/assets/f46766ab-003b-43b9-a795-8e73895eda99

realpaaao:
QA done.

PR #1350, opened 2024-10-17
Decimal check moved to button tap

Rockindash:
Decimal check moved to button tap instead of onChange, Issue #1272


johnnyluo (approved):
LGTM

Issue #1325, opened 2024-10-13
[BUG] Able to add Vault multiple to multiple folders


realpaaao:
**Describe the bug**
When creating or editing vault folders, there is no check if a vault is already in a different vault. 
Therefore it is possible to add vaults to multiple folders

**To Reproduce**
1. Add `vault#1 `to `folder#1`
2. see `vault#1 `when creating `folder#2 `
3. add `vault#1` to `folder#2`

**Expected behavior**
Implement a check if a vault is already added to another vault and remove this vault from new folder creation and edit of folders.

**Screenshots**
![Image](https://github.com/user-attachments/assets/ec2bf91b-f1e5-4a64-ba53-8896eab73f17)
![Image](https://github.com/user-attachments/assets/bae8c181-eb3c-4ef5-a337-a7a9e7e04215)

 - Version 1.2.2.

realpaaao:
@Rockindash that still happens when creating new folders or rearranging on `vault view`

https://github.com/user-attachments/assets/545e77a9-5030-4c74-9696-8dcf4f1fcce8

realpaaao:
@Rockindash also happens on macOS

realpaaao:
QA done.

PR #1353, opened 2024-10-18
Vaults filter logic updated

Rockindash:
Vault filter logic updated to prevent it being in multiple folders, Issue #1325


johnnyluo (approved):
LGTM

Issue #1296, opened 2024-10-08
[ADD] Always on screen on iOS when waiting for keygen


realpaaao:
Please add always on screen on iOS when waiting for keygen

johnnyluo:
Fixed in #1366

realpaaao:
QA done.

PR #1366, opened 2024-10-19
Screen stays on for Keygen, Send and Swap

Rockindash:
Stays stays active for Kegen, Send and Swap flow, Issue #1296


johnnyluo (approved):
LGTM

PR #1367, opened 2024-10-19
Mac freeze bug

Rockindash:
Mac on launch freeze issue fixed


VultisigApp/VultisigApp/macOS/View/VultisigApp+macOS.swift
```diff
@@ -24,9 +24,18 @@ extension VultisigApp {
             .environmentObject(macCameraServiceViewModel)
             .buttonStyle(BorderlessButtonStyle())
             .frame(minWidth: 900, minHeight: 600)
+            .onChange(of: scenePhase) {
+                switch scenePhase {
```
johnnyluo, line 28: Current MacOS doesn't freeze at launch , only  "Mac Designed for Ipad" will freeze at launch , so I don't think this is the fix as this only apply to MacOS app

PR #1371, opened 2024-10-19
Update check logic updated to fix repeated alerts

Rockindash:
Update check logic updated to fix repeated alerts


johnnyluo (approved):
LGTM

Issue #1395, opened 2024-10-21
[IMPROVE] remove RUNEPool deposit and withdrawal form MAYA


realpaaao:
When opening the Functions for MAYA Protocol it is possible to select RUNEPool interactions on the MAYA Chain.

Please remove both options.

![Image](https://github.com/user-attachments/assets/0ed9f961-8dbe-407a-9926-212453015918)

realpaaao:
QA done.

PR #1408, opened 2024-10-22
maya cases are now separated

enriquesouza:
image

This solves the issue https://github.com/vultisig/vultisig-ios/issues/1395


realpaaao (approved):
LGTM

Issue #1250, opened 2024-10-01
[BUG] Occasionally receiving "already mined" error


realpaaao:
**Describe the bug**
When doing multiple transactions after another the error "already mined" comes up on initating device while the tx gets properly displayed on pairing device.

**Expected behavior**
Consistently display tx on all devices

**Screenshots**
![Image](https://github.com/user-attachments/assets/d3c5b7e5-316d-40f8-ab0a-6c53cb902e9e)

-v1.1.1

realpaaao:
QA done. will observe further

PR #1409, opened 2024-10-22
transactions already mined

enriquesouza:
This fixes the issue https://github.com/vultisig/vultisig-ios/issues/1250

I could not reproduce it, so I assumed that this should solve


VultisigApp/VultisigApp/Services/RpcService.swift
```diff
@@ -57,6 +57,8 @@ class RpcService {
                     || message.contains("nonce too high")
                     || message.contains("transaction already exists")
                     || message.contains("many requests for a specific RPC call")
+                    || message.contains("already")
```
flypaper0, line 60: why wee need "message.contains("already")" check?

enriquesouza: In case the message returns "already .... anything."

PR #1414, opened 2024-10-23
Filtered vaults added to CreateFolderView

### LINKED ISSUE #1325: [BUG] Able to add Vault multiple to multiple folders

Rockindash:
Vault filter logic updated for CreateFolderView, Fixes #1325

While rearranging all the vaults need to be visible to maintain their order. If we only show the remaining vaults it introduces the possibility of having multiple vaults on a single order value.


johnnyluo (approved):
LGTM

realpaaao:
@Rockindash Can we not make different orders for vaults and no vaults? In android it is super clean but on iOS it is very confusing and not good UX.

PR #1417, opened 2024-10-23
set bounce as true

enriquesouza:
For security reasons I am making the set bounce as true by default so in case of any errors of the transactions, the funds will come back to the wallet. 

https://tonscan.org/address/Ef8t6cZkqFuHjJ_a_ydEK_tu3LHWRA4JZXRyewLY4j8FZ6B5


johnnyluo (approved):
LGTM

PR #1423, opened 2024-10-24
ton stake func

enriquesouza:
Based on TON's specification, I've created a function for TON where you send coins to a POOL with the memo "d" for deposit and "w" for withdrawal;

https://blog.ton.cat/how-to-stake/

All of our transactions are **bounceable**, so if the pool refuses, the amount should be returned. You can see my tests here:
https://tonscan.org/pool/Ef8t6cZkqFuHjJ_a_ydEK_tu3LHWRA4JZXRyewLY4j8FZ6B5

You can also specify the POOL address. Therefore, I left the following POOL address as a default:

Ef8t6cZkqFuHjJ_a_ydEK_tu3LHWRA4JZXRyewLY4j8FZ6B5

The minimum amount for this POOL is **1 003 TON**, so I also left it as a default. 

I also left the same POOL for withdrawal, but to withdraw, we must send 1 TON always, and it will return all funds. There are no other options based on the spec above.


johnnyluo (approved):
LGTM

PR #1431, opened 2024-10-26
Update Ton.swift

realpaaao:
to be able to send to new addresses


johnnyluo:
Closing this one , let @enriquesouza to investigate it , and then fix it appropriately , maybe we can set based on an amount threshold , like if it larger than 5 TOn , then it is bounceable , if less than it is not bounceable

PR #1443, opened 2024-10-28
[FastSign] Biometry settings for MacOS

flypaper0:


VultisigApp/VultisigApp/Views/Settings/SettingsBiometryView.swift
```diff
@@ -28,6 +28,19 @@ struct SettingsBiometryView: View {
     }
 
     var main: some View {
+        VStack {
+            headerMac
```
johnnyluo, line 32: You should only apply `headerMac` when it is running on a Mac , Not IOS

image

johnnyluo (changes requested):
only apply headerMac when it is running on a Mac , Not IOS

johnnyluo (approved):
LGTM

PR #1451, opened 2024-10-29
custom-tokens

enriquesouza:


VultisigApp/VultisigApp/iOS/View/ChainDetailView+iOS.swift
```diff
@@ -43,6 +43,10 @@ extension ChainDetailView {
                 if viewModel.hasTokens(chain: group.chain) {
                     addButton
                 }
+
```
johnnyluo, line 46: you need to add this to MacOS as well

VultisigApp/VultisigApp/iOS/View/ChainDetailView+iOS.swift
```diff
@@ -60,6 +64,14 @@ extension ChainDetailView {
         }
     }
     
+    var addCustomTokenButton: some View {
+        Button {
+            sheetType = .customToken
```
johnnyluo, line 69: Also , can we also hardcode some of the most common tokens on Solana into our token store?

enriquesouza: Yes, def we can.

VultisigApp/VultisigApp/iOS/View/ChainDetailView+iOS.swift
enriquesouza: Tks

johnnyluo (approved):
LGTM

PR #1458, opened 2024-10-29
Retry keysign when timeout

johnnyluo:
1. When keysign timeout, Tss doesn't return an error, instead it return empty TssKeysignResponse , throw an exception when the signature is empty , so it can retry to sign it again
2. When signing with multiple UTXOs , we need to sort the keysign images , so two device can agree on to sign the messages in the same order


enriquesouza (approved):
LGTM

PR #1470, opened 2024-10-31
[FastSign] Password hint

flypaper0:


realpaaao:
@flypaper0 why clear password hint on biometrics disable? As it is only needed when biometrics are not available?

please always have password hint enabled, as it is only needed when biometrics are disabled

PR #1493, opened 2024-11-05
VaultsList reorder logic updated to only include filtered vaults

### LINKED ISSUE #1488: [IMPROVE] Reorder vaults, shows all vaults on vault selection
It is confusing for the user that when re-ordering vaults on vault selection view all vaults (even when in folders) are shown.

Please improve the logic that when a vault is in a folder, it will not be shown in the vault selection view when re-order is toggeled.

Rockindash:
VaultsList reorder logic updated to only include filtered vaults, Fixes #1488


johnnyluo (approved):
LGTM

Issue #1267, opened 2024-10-04
[BUG] Not able to drag an drop backups into macOS


realpaaao:
**Describe the bug**
It is not possible to drag-and-drop backup files into the app to import on the `import` view

**Expected behavior**
Able to drag and drop

**Screenshots**
![Image](https://github.com/user-attachments/assets/c0d432ff-89b3-42b1-b0ba-b2bcbf809fda)

System:

 - Device: macOS
 - Version 1.1.1

johnnyluo:
on IOS app , when open .vult file , it doesn't open import vault view

Rockindash:
@johnnyluo created a follow-up issue for the same.
https://github.com/vultisig/vultisig-ios/issues/1514

realpaaao:
@Rockindash somehow .dat and .bak files are not possible to import via drag-and drop.

PR #1513, opened 2024-11-08
Drag and Drop disabled for macCatalyst assp

Rockindash:
Drag an drop permission fix for macCatalyst, #1267
iPad app running on Mac would have permission issue for drag and drop. So it should only be limited to native Mac app.


johnnyluo (approved):
LGTM

PR #1519, opened 2024-11-09
Vult iOS

### LINKED ISSUE #1397: [IMPROVE] Opening .vult backups
When opening .vult backups on iOS it opens the app but not the import page.
Improve that when opening a .cult backup it directly imports it into the app and open the import view.

Rockindash:
Deeplink updated to support VULT documents on iOS, Fixes #1397


johnnyluo (approved):
LGTM

PR #1524, opened 2024-11-11
it should increase the fee from 10 to 100 LUNA CLASSIC ~1 cent of dollars

enriquesouza:
it should increase the fee from 10 to 100 LUNA CLASSIC ~1 cent of dollars;

I also replaced the logo in SVG, but it was not good per the PNG one, which is ok now.


johnnyluo (approved):
LGTM

PR #1527, opened 2024-11-11
[Keysign] Function decoding

flypaper0:


VultisigApp/VultisigApp/Utils/Endpoint.swift
```diff
@@ -137,6 +137,10 @@ class Endpoint {
         return "https://api.vultisig.com/ton/v3/addressInformation?address=\(address)&use_v2=false";
     }
     
+    static func fetchMemoInfo(hash: String) -> URL {
+        return "https://api.etherface.io/v1/signatures/hash/all/\(hash)/1".asUrl
```
johnnyluo, line 141: For this one , we pass the whole memo in a url , which might not be enough , some memo might be very long,

Can you please check whether it can use http `POST`

johnnyluo (approved):
LGTM

PR #1533, opened 2024-11-12
[Core] GasLimit estimation fixed

flypaper0:


VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -54,25 +54,17 @@ final class BlockChainService {
     
     
     func fetchSpecific(tx: SendTransaction) async throws -> BlockChainSpecific {
-        guard !tx.coin.isNativeToken, tx.coin.chainType == .EVM else {
```
johnnyluo, line 57: this break none EVM chains

johnnyluo (changes requested):
This cause problem for none EVM chains, please do some tests with 
1. EVM 
2. THORChain
3. UTXO chain 
etc etc.

VultisigApp/VultisigApp/Services/BlockChainService.swift
flypaper0: fixed non evm chains

johnnyluo (approved):
LGTM

Issue #1522, opened 2024-11-11
[UPDATE] Deletion screen


realpaaao:
Please update the vault deletion screen to reflect figma. Change `Vault Type` to `Vault Part` , so user knows what part he is deleting:

![Image](https://github.com/user-attachments/assets/31e788c0-0274-4f94-b09b-f846c5cac0a4)

realpaaao:
QA done.

PR #1535, opened 2024-11-12
Updated Vault Deletion screen

Rockindash:
Updated Vault Deletion screen, Issue #1522

### DIFF
--- a/VultisigApp/VultisigApp/Views/Components/VaultDeletionDetails.swift
+++ b/VultisigApp/VultisigApp/Views/Components/VaultDeletionDetails.swift
@@ -91,7 +91,7 @@ struct VaultDeletionDetails: View {
     }
     
     private func getVaultType() -> String {
-        return "\(vault.getThreshold() + 1) of \(vault.signers.count) Vaults"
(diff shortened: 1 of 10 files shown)


VultisigApp/VultisigApp/Views/Components/VaultDeletionDetails.swift
```diff
@@ -91,7 +91,7 @@ struct VaultDeletionDetails: View {
     }
     
     private func getVaultType() -> String {
-        return "\(vault.getThreshold() + 1) of \(vault.signers.count) Vaults"
+        return "Part \(vault.getThreshold() + 1) of \(vault.signers.count)"
```
johnnyluo, line 94: This one is wrong , so user want to know which part they are deleting , not the vault type. 
In this case , we should get the index of local party in the vault.signers , instead of get threshold. 

User want to know which part he/she is deleting , like Part1of3 or Part2of3, or Part3of3

johnnyluo (changes requested):
see comments

johnnyluo (approved):
LGTM

PR #1537, opened 2024-11-13
Cosmos tokens

enriquesouza:


VultisigApp/VultisigApp/Chains/cosmos/cosmos.swift
```diff
@@ -0,0 +1,246 @@
...
+                }
+            }
+            
+            return try input.serializedData()
+            
+        }
+        
+        // IT seems this is only used when literally I want to move the coin from one blockchain to another
+        // IBCs are working as coins so we can simply use the send.
+//        else {
```
johnnyluo, line 100: clean up these

VultisigApp/VultisigApp/Chains/cosmos/cosmos.swift
```diff
@@ -0,0 +1,246 @@
...
+    func getPreSignedInputData(keysignPayload: KeysignPayload) throws -> Data {
+        guard case .Cosmos(let accountNumber, let sequence , let gas, _, let ibc) = keysignPayload.chainSpecific else {
+            throw HelperError.runtimeError("getPreSignedInputData: fail to get account number and sequence")
+        }
+        guard let pubKeyData = Data(hexString: keysignPayload.coin.hexPublicKey) else {
+            throw HelperError.runtimeError("getPreSignedInputData: invalid hex public key")
+        }
+        let coin = self.coinType
+        
+        if keysignPayload.coin.isNativeToken
```
johnnyluo, line 59: for this one , like on Terra , UST , we can't set it as native token , and it is not IBC token either ,this will fail , correct?

johnnyluo: nevermind , terra is in a seperate one

VultisigApp/VultisigApp/Chains/cosmos/cosmos.swift
enriquesouza: Do we want to implement sending from one chain to another? I was thinking about add this feature to cosmos tomorrow. 

Cosmos1…. To osmo1… address

enriquesouza: https://github.com/vultisig/vultisig-ios/issues/1570 

Opened an issue and cleaned it up.

johnnyluo (approved):
LGTM

PR #1572, opened 2024-11-21
memo on terra

enriquesouza:


VultisigApp/VultisigApp/iOS/View/Send/SendCryptoDetailsView+iOS.swift
```diff
@@ -57,11 +57,7 @@ extension SendCryptoDetailsView {
                     coinSelector
                     fromField
                     toField
-                    
-                    if tx.coin.isNativeToken {
-                        memoField
-                    }
-                    
+                    memoField
```
johnnyluo, line 60: This is not correct , because non-native token , the `data` field of transaction has been used to store the contract call , so it can't have custom memo

I guess in your case , USTC is a different case , maybe just need to make an exception for that

johnnyluo:
do you still want to update the tax for USTC?

PR #1580, opened 2024-11-22
Server backup verification

### LINKED ISSUE #1567: [IMPROVE] Server backup verification
Please improve server backup verification that: 

- [x] Going back, brings the user to the setup view again and deletes the created vault. this is needed in case the Vault was not created on the Server side.
- [x] Improve error message when continuing without verification code: grey out `continue` button until input, if wrong input prompt "Wrong verification code, please enter the correct one"
- [x] Make font of numbers on input blank white
- [x] Change the logic on backup verification that it not checks the code when creating a active vault and only on fast vaults

Rockindash:
Server backup verification view updated, Fixes #1567


johnnyluo (approved):
LGTM

PR #1586, opened 2024-11-24
Send refresh

### LINKED ISSUE #1557: [ADD] Refresh on send view
Please add `pulldown` for refresh on send view, so user can update Gas prices/fetch again when in send screen without the need to leave the view

Rockindash:
Refresh option added to SendCryptoView, Fixes #1557


johnnyluo (approved):
LGTM

PR #1632, opened 2024-12-03
FromAddress logic updated for SendCryptoView

### LINKED ISSUE #1346: [BUG] scanning QR code from camera doesn't update FROM address
**Describe the bug**
When scanning a address QR code from the camera on `main view` opens the send screen but has a EVM address selected instead of correctly changing the FROM address to the respective chain. It is possible to proceed and scan the pairing QR

**To Reproduce**
1. Go to main view and open camera
2. scan a QR for addresses

**Expected behavior**
Adjust the `FROM `address to the correct chain selected.

**Screenshots**
[Video](https://github.com/user-attachments/assets/3c561a44-34ea-411a-ae69-66937ca7800b)

 - Version 1.3

**Additional context**
Pairing with another device, the information on the pairing device for the `FROM `label is correct.

Rockindash:
From address logic updated for SendCryptoView, Fixes #1346

### DIFF
--- a/VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
+++ b/VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
@@ -203,6 +203,7 @@ struct SendCryptoView: View {
         }
         
         tx.coin = selectedCoin
+        tx.fromAddress = selectedCoin.address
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #1640, opened 2024-12-05
Add lib type

johnnyluo:
this PR is to add a lib type to vault , keygen message and reshare message 

prepare to add DKLS vault


enriquesouza (approved):
LGTM

PR #1647, opened 2024-12-07
Address check logic updated for GeneralCodeScannerView

### LINKED ISSUE #1047: [BUG] QR invoice scan not correct
**Describe the bug**
When scanning an Invoice QR code with camera the app doesn't behave well:
1. It doesn't select the correct chain or asset when scanning from `main view ` or `send view`
2. When an amount is included in the QR the camera on `main view` doesn't register it but on `send view` it does

**To Reproduce**
1. Scan QR Code with camera on main screen or send screen
2. See error

**Expected behavior**
The App correctly selects the chain, asset and if included price

**Screenshots**
Invoice QR
image

Invoice QR with amount
image

-v46

Rockindash:
Address check logic updated for GeneralCodeScannerView, Fixed #1047 and #1346


johnnyluo (approved):
LGTM

PR #1652, opened 2024-12-09
ripple

enriquesouza:
**What is done and tested:**

[ x ] - Adding / Removing the coin XRP on Token Selection
[ x ] - It generates the ECDSA address
[ x ] - Showing the balance
[ x ] - Pre-Image hashing as well 
[ x ]  - Signed with DER

**Transactions:**

[x] - New account, min of 10 XRP
[x] - Old account, any amount
[x] - Accounts with MEMO (Destination Tag)
[x] - Fast Vault
[x] - Pair devices
[x] - Tx hashes are matching in both devices if already broadcasted.
[x] -  Create a Ripple Chain Specific Prototype, for testing I am reusing this from Cosmos.

Tests:
https://xrpscan.com/account/rKQ3hEqW1MVokhRwUnugjLsVypdP3mrmPA

Send to a new account: OK
It must be a minimum of 10 XRP to open a new account OK
https://xrpscan.com/account/rMY6uEev8PiZzVisDKyh9XKwWVEBrYnb9L

Send to an exchange with MEMO (Destination TAG) OK
Send to an existent account OK

**Prints:**

image

image

image

image


VultisigApp/VultisigApp/Chains/common.swift
```diff
@@ -68,6 +68,9 @@ struct SignatureProvider {
             switch sigResult {
             case .success(let sigData):
                 logger.info("successfully get signature")
+                
+                print("successfully get signature")
```
johnnyluo, line 72: clean up here?

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Chains/common.swift
enriquesouza: Oh, I will )

enriquesouza: I've cleaned in this other one. https://github.com/vultisig/vultisig-ios/pull/1659

PR #1653, opened 2024-12-09
[Core] Matic to pol

flypaper0:


VultisigApp/VultisigApp/View Models/VaultDetailViewModel.swift
```diff
@@ -15,7 +15,15 @@ class VaultDetailViewModel: ObservableObject {
 
     private let balanceService = BalanceService.shared
     private var updateBalanceTask: Task<Void, Never>?
-    
+
+    func migrate(vault: Vault) {
```
johnnyluo, line 19: can you do a data migration with swiftdata?

VultisigApp/VultisigApp/Views/Vault/VaultDetailView.swift
```diff
@@ -44,6 +44,9 @@ struct VaultDetailView: View {
             popup
             shadowView
         }
+        .task {
+            viewModel.migrate(vault: vault)
```
johnnyluo, line 48: this is checking everytime when user open the vault , which is not good

enriquesouza:
@flypaper0 @johnnyluo I've checked that and I would recommend us to not migrate like this. Only create a new Token called POL, pretty much duplicate the Matic. We have a bunch of old wallets and those wallets are using MATIC. If we rename it they will need to remove it and re add it. Or as you are doing a swift data migration it is way more risky than simply create both of them. Coingecko also does that because they have different market cap and tickers. 

So I recommend us to duplicate Matic. Another issue is that all devices must match and once we deploy on IOS we won't be able to sign Android against IOS or Win.

enriquesouza (changes requested):
Please check my comments.

johnnyluo:
Let's close this one for now then

PR #1666, opened 2024-12-13
Fix thorchain send/swap after hardfork

johnnyluo:
https://thornode.ninerealms.com/auth/accounts/thor1vzltn37rqccwk95tny657au9j2z072dhgstcmn 
The response from this endpoint is different than before after the hard fork , thus we need to adjust it to fetch account number and sequence no

swap: https://thorchain.net/tx/6747192E801B9C6058E54E911300897C45FAB77BEE994C56B2DF0604867FAB74
send: https://thorchain.net/tx/4D089C9D1B267BAD3E00A2448708301C5490AAE6B0101E14C8A2F160CBD18ED4


enriquesouza (approved):
LGTM

PR #1691, opened 2024-12-19
Fix/solana spl issues

enriquesouza:
This is related to SPL issues. It is a must to pair alongside Windows. Also it is necessary to change on Android.

**This is the source address**
E778kcLYJ58HSLAPBWiXFp94mmYXkoq4pbp6wcbRnULr

https://solscan.io/account/E778kcLYJ58HSLAPBWiXFp94mmYXkoq4pbp6wcbRnULr

**This is the destination address** 
https://explorer.solana.com/address/CVAPxRchnZnxnyTmzY7K6oVRvDnYSEo26CXnSB5wgL7

CVAPxRchnZnxnyTmzY7K6oVRvDnYSEo26CXnSB5wgL7

### DIFF
--- a/VultisigApp/VultisigApp/Chains/Solana.swift
+++ b/VultisigApp/VultisigApp/Chains/Solana.swift
@@ -35,7 +35,7 @@ enum SolanaHelper {
                         $0.memo = memo
                     }
                 }
-                $0.recentBlockhash = recentBlockHash
+                $0.recentBlockhash = recentBlockHash // DKLS should fix it. Using the same, since fetching the latest block hash won't match with Win and Android
                 $0.sender = keysignPayload.coin.address
                 $0.priorityFeePrice = SolanaPriorityFeePrice.with {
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Chains/Solana.swift
```diff
@@ -23,6 +23,9 @@ enum SolanaHelper {
             throw HelperError.runtimeError("fail to get to address")
         }
         
+        let priorityFeePrice = 1_000_000; // Turbo fee in lamports, around 5 cents
```
johnnyluo, line 26: this hardcode the priority fee , previous we get it from the block chain , are you sure this is the right fix?

enriquesouza: I had to. The blockchain returns a very low one.

PR #1728, opened 2025-01-06
Tron2025

enriquesouza:
We are now able to send TRX and TRC20 Tokens. I only added USDT which is the mainly feature used.

To send TRX is cheap, but to send TRC20 I had to set the limit to 50TRX, which is the only way of it to go through.

Address I am interacting with:
https://tronscan.org/#/address/TVNtPmF7JWw4xoA8GAxEZhCaw2khYn8viH

Transactions:

TRX:
https://tronscan.org/#/transaction/a6d69e607feea108d93f70052f5cd5d50aff1d03c1e2725f10ad2c01d04a8c21

TRC20:
https://tronscan.org/#/transaction/bbdc82bf48f7ab515b83b20d9dd54dcf7939c0127fccb569726ed35bb66cedfb


VultisigApp/VultisigApp/Model/Coin.swift
```diff
@@ -173,6 +175,8 @@ class Coin: ObservableObject, Codable, Hashable {
             return "180000"
         case .akash:
             return "200000"
+        case .tron:
+            return "100000" // TODO: double check this
```
johnnyluo, line 179: do you still need to double check? It seems to me we can't hardcode the fee, as fee depends on multiple factors on TRON chain

johnnyluo (commented):
mostly LGTM , check two comments there

johnnyluo (approved):
LGTM

PR #1766, opened 2025-01-14
macOS token change bug fixed

### LINKED ISSUE #1752: [BUG:macOS] token change to native EVM token on macOS
**Describe the bug**
The bug where when sending and preselecting a token on any EVM chain changes to the selection of the native token still happens on macOS once after the first start of the app. see issue #1624 as reference

**To Reproduce**
1. Use mac and select any token on any EVM chain
2. Click on send when in detail view of this token
3. See error when filling out `address` and `amount`

**Expected behavior**
Keep preselection of token

System:

 - Version 1.10.1

Rockindash:
Token auto change to native bug fixed on macOS, Fixes #1752


johnnyluo (approved):
LGTM

PR #1818, opened 2025-02-02
Fast vault overview

Rockindash:
- Implemented Fast Vault Setup Flow
- Implemented Setup Vault Setup Flow
- Loading heavy animation after view load to avoid screen freeze


johnnyluo (approved):
LGTM

PR #1912, opened 2025-03-06
Display the balance for the unbound fn

enriquesouza:
image

## Summary by CodeRabbit

- **New Features**
	- The unbonding transaction interface now displays real-time balance feedback by showing a formatted balance (including coin ticker) as part of the input field.
	- The transaction processing workflow for unbonding operations has been enhanced with improved data integration, ensuring users have immediate and clear context during transactions.


johnnyluo:
Actually , please don't display balance with unbond , because it cause confusion , user might misunderstand it as the amount they can unbond

Issue #1300, opened 2024-10-09
[BUG] Camera button opens verify view on macOS


realpaaao:
**Describe the bug**
When doing 2 consecutive transactions and signing with macOS, for the second tx if opening the camera view on macOS to sign the transaction opens the last transaction verification view.

**To Reproduce**
1. Make a transaction co signed with macOS
2. initiate a scond tx on iPhone
3. open camera on macOS to join keysign

**Expected behavior**
Open camera and not the last verification view

 - Version 1.2.1

johnnyluo:
@realpaaao can this one still be reproducable?

realpaaao:
@Rockindash this still happens on version 1.10.1

enriquesouza:
I am working on it.

johnnyluo:
fixed in #1916

johnnyluo:
@realpaaao can you QA on this?

realpaaao:
QA done.

PR #1916, opened 2025-03-07
reset the tx before going home

enriquesouza:
This should reset the tx object before going back to home. It should fix the issue #1300

### DIFF
--- a/VultisigApp/VultisigApp/iOS/View/Send/SendCryptoDoneView+iOS.swift
+++ b/VultisigApp/VultisigApp/iOS/View/Send/SendCryptoDoneView+iOS.swift
@@ -9,10 +9,16 @@
 import SwiftUI
 
 extension SendCryptoDoneView {
+    
     var view: some View {
         VStack {
             cards
             continueButton
+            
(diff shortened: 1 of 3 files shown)


VultisigApp/VultisigApp/iOS/View/Send/SendCryptoDoneView+iOS.swift
```diff
@@ -9,10 +9,16 @@
...
+    
     var view: some View {
         VStack {
             cards
             continueButton
+            
+            NavigationLink(destination: HomeView(selectedVault: vault), isActive: $navigateToHome) {
+                EmptyView()
+            }
+            .hidden()
```
Rockindash, line 21: Since, we are using isActive trigger here. It can be replaced by .navigationDesination modifier here.

enriquesouza: Done

enriquesouza (commented):
@Rockindash I've done your suggestions

Rockindash (approved):

PR #1926, opened 2025-03-10
Remove the loader

enriquesouza:
I removed the loader as suggested. It still gets blur, but no loader. We should get used with a top limit between 1 and 5 secs to load all info. 

## Summary by CodeRabbit

- **Refactor**
  - Removed the transient loading indicator from the send crypto screen, resulting in a cleaner and more direct user experience without the previous visual feedback during loading states.

### DIFF
--- a/VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
+++ b/VultisigApp/VultisigApp/Views/Send/SendCryptoView.swift
@@ -32,10 +32,6 @@ struct SendCryptoView: View {
         ZStack {
             Background()
             main
-            
-            if sendCryptoViewModel.isLoading || sendCryptoVerifyViewModel.isLoading {
-                loader
-            }
         }
         .ignoresSafeArea(.keyboard)
         .onAppear {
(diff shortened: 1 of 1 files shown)


johnnyluo (changes requested):
Could you also remove the blur 
https://github.com/vultisig/vultisig-ios/pull/1926/files#diff-a544b6ba475db291218265c74773fb13494be563d00f12451598fb6f1497eefbR72

enriquesouza:
> Could you also remove the blur https://github.com/vultisig/vultisig-ios/pull/1926/files#diff-a544b6ba475db291218265c74773fb13494be563d00f12451598fb6f1497eefbR72

Sure. I got the resume of the meeting and there is a suggestion to keep the loader inside the button, should I move forward with that?

- Enrique is asked to remove the busy indicator and prevent users from clicking the next button if guest information is not fetched.



> Could you also remove the blur https://github.com/vultisig/vultisig-ios/pull/1926/files#diff-a544b6ba475db291218265c74773fb13494be563d00f12451598fb6f1497eefbR72

Sure. Removed

johnnyluo:
I am working on a PR that avoid to fetch blockspecific multiple times, which means we don't need to remove the busy indicator

PR #1928, opened 2025-03-11
Use chain not coin info for address book

enriquesouza:
I've changed to use the chain while adding addresses to address book since it is per chain not coin. 

image

## Summary by CodeRabbit

- **Refactor**
  - Improved the address book selection experience so that ticker symbols and logos display accurately and consistently.
  - Enhanced naming conventions to streamline how coin information is referenced during selection.


johnnyluo (approved):
LGTM

PR #2051, opened 2025-04-14
switch func page

enriquesouza:


VultisigApp/VultisigApp/Model/TransactionMemo/TransactionMemoCosmosSwitch.swift
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+class TransactionMemoCosmosSwitch: TransactionMemoAddressable, ObservableObject {
+    @Published var amount: Double = 0.0
+    @Published var destinationAddress: String = "cosmos144za3huzmfl3k6hge487a6gnu06vwk6t2hfk53"
+    @Published var thorAddress: String = ""
+    
+    @Published var amountValid: Bool = false
+    @Published var destinationAddressValid: Bool = true
+    @Published var thorchainAddressValid: Bool = false
```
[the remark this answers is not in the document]
enriquesouza: It is as it should be

VultisigApp/VultisigApp/Model/TransactionMemo/TransactionMemoCosmosSwitch.swift
```diff
@@ -0,0 +1,142 @@
...
+
+ Action:
+ → Send MsgSend from COSMOS to THORChain vault
+ → Include memo: SWITCH:<thorAddress>
+ 
+ */
+
+class TransactionMemoCosmosSwitch: TransactionMemoAddressable, ObservableObject {
+    @Published var amount: Double = 0.0
+    @Published var destinationAddress: String = "cosmos144za3huzmfl3k6hge487a6gnu06vwk6t2hfk53"
```
johnnyluo, line 30: I think this destination address should be read from `https://thornode.ninerealms.com//thorchain/inbound_addresses` 
we should never hardcode the inbound address , as THORChain will churn every three days, if we hard coded it , user will lose fund

johnnyluo (changes requested):
see comments

VultisigApp/VultisigApp/Model/TransactionMemo/TransactionMemoCosmosSwitch.swift
```diff
@@ -0,0 +1,150 @@
...
+ 
+ */
+
+class TransactionMemoCosmosSwitch: TransactionMemoAddressable, ObservableObject {
+    @Published var amount: Double = 0.0
+    @Published var destinationAddress: String = ""
+    @Published var thorAddress: String = ""
+    
+    @Published var amountValid: Bool = false
+    @Published var destinationAddressValid: Bool = true
```
johnnyluo, line 34: since `destinationAddress` is empty at the beginning , this should be set to `false` 
and only set it true once we get a valid inbound address from THORChain endpoint

VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
```diff
@@ -177,6 +178,34 @@ class ThorchainService: ThorchainSwapProvider {
...
+                cache: cacheInboundAddresses,
+                timeInSeconds: 60 * 5
+            ) {
+                return cachedData
+            }
+            
+            let urlString = Endpoint.fetchThorchainInboundAddressesNineRealms
+            let data = try await Utils.asyncGetRequest(urlString: urlString, headers: [:])
+            
+            if let json = try JSONSerialization.jsonObject(with: data, options: []) as? [[String: Any]] {
```
johnnyluo, line 196: can you please define a proper structure , and then decoded it?  don't store everything as jsonObject

johnnyluo: Also , keep in mind, the chain can be halted , check field 
```
"halted": false,
    "global_trading_paused": false,
    "chain_trading_paused": false,
```
any of these fields are true , we should not allow user to send switch

enriquesouza:
```
Task {
            let addresses = await ThorchainService.shared.fetchThorchainInboundAddress()
            
            if let match = addresses.first(where: { ($0["chain"] as? String)?.uppercased() == "GAIA" }) {
                let halted = match["halted"] as? Bool ?? false
                let globalPaused = match["global_trading_paused"] as? Bool ?? false
                let chainPaused = match["chain_trading_paused"] as? Bool ?? false
                
                if halted || globalPaused || chainPaused {
                    print("Chain is halted or paused. Cannot proceed with switch.")
                    return
                }
                
                self.destinationAddress = match["address"] as? String ?? ""
            }
        }
        ```

johnnyluo (approved):
LGTM

PR #2077, opened 2025-04-21
Migration part share

### LINKED ISSUE #2072: [IMPROVE] Migration flow
- [x] Please make sure the pairing view reflects how many devices need to be present in the `Devices(1/x)`, see figma 

Image

- [x] The pairing device ALWAYS has wrong naming of backup file:
- always still has `part` instead of `share`
- the share number and amount of shares is always wrong ( in a 2-2 vault after upgrading the pairing devices has `part2of4` or differently, never correct)

Rockindash:
Fixes the following migration flow bug, Fixes #2072
- Device count on PeerDiscoveryView required for migration
- Removed Part from vault naming
- Double header issue fixed for Mac

**Screenshot**
Screenshot 2025-04-21 at 11 21 54 AM


VultisigApp/VultisigApp/Extensions/VaultExtension.swift
```diff
@@ -41,13 +41,8 @@ extension Vault {
...
         }
         let cleanVaultName = vaultName.replacingOccurrences(of: "/", with: "-")
-        var partName = ""
-        switch self.libType {
-        case .DKLS:
-            partName = "share"
-        default:
-            partName = "part"
-        }
+        var partName = "share"
```
johnnyluo, line 44: @Rockindash why we remove this logic? I think we would like to keep it , so we can figure out whether it is GG20 or DKLS by looking at the backup name

Rockindash: Just reverted back this change and fixed the device count of PeerDiscoveryView.

VultisigApp/VultisigApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -41,17 +41,17 @@
...
       }
     },
     {
       "identity" : "rive-ios",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/rive-app/rive-ios",
       "state" : {
-        "revision" : "f6ddbd9b5c59ddd3322c7aedcbce5d30ed19929f",
-        "version" : "6.7.5"
+        "revision" : "635cdbdc47c31154d9409ab4b9d4ed14bedb48f4",
```
johnnyluo, line 53: @Rockindash can you please revert this one , don't want to downgrade the dependencies

PR #2108, opened 2025-04-24
allow zero to unbond all

enriquesouza:
It allows to set the amount as zero, so you can unbond everything at once.

## Summary by CodeRabbit

- **Refactor**
  - Updated the unbond amount input field to always appear valid, regardless of user input. The amount is now always accepted, including zero, which represents unbonding all.


johnnyluo (approved):
LGTM

PR #2121, opened 2025-04-30
Add vault type to keysign message , to avoid people mismatch DKLS with GG20

### LINKED ISSUE #2015: [BUG] Should Provide error message when user trying to sign transaction with two different type of vault share
**Describe the bug**
When trying to swap with a GG20 and DKLS vault share due to wrong import or a reshare with different amount of signers involved, it crashes the app

**Expected behavior**
Please display an error that the Vault shares are not matching and therefore no signing is possible. Apply it for different Vault types and signers involved

System:
 - Version 15

johnnyluo:
fixes #2015 

Currently we have two types of vault (GG20/DKLS) , sometimes user might get their devices mixed with GG20 and DKLS vault

For example , user have a GG20 vault , then upgraded it to DKLS  , initiate a send/swap from a DKLS vault share

Then pair with a device , however the share on the device it might be still GG20 , in that case , we should fail user early

## Summary by CodeRabbit

- **New Features**
  - Added a new error message and view to inform users when there is a vault type mismatch during key signing or vault operations.
  - Enhanced validation to detect and prevent mismatched vault types when joining or migrating vaults.
- **Improvements**
  - Updated app and marketing version numbers.
  - Improved localization with new translations for the vault type mismatch error in multiple languages.
- **Bug Fixes**
  - Ensured consistent handling of vault type information across key signing and transaction flows.
- **Tests**
  - Updated test cases to support the new vault type parameter.


enriquesouza (approved):
LGTM

PR #2188, opened 2025-05-06
Fix local mode switch

yvebe:
## Description

Fixed local mode switching on iOS because it didn't switch QR code.

## Which feature is affected?
- [ ] Keygen peer discovery
- [ ] Keysign peer discovery

## Summary by CodeRabbit

- **New Features**
  - Network selection in peer and participant discovery views now dynamically synchronizes with the relay connection state, ensuring consistent behavior when switching between Internet and Local modes.

- **Bug Fixes**
  - Improved reliability of participant discovery and data refresh when changing network selection.

- **Style**
  - Removed debug print statements for a cleaner user experience.


johnnyluo (approved):
LGTM

PR #2190, opened 2025-05-07
tcy price from the pool

enriquesouza:
We should be able to fetch the token price from CoinGecko or from the pool also for Thorchain

image

## Summary by CodeRabbit

- **New Features**
  - Added support for fetching and displaying USD prices for THORChain assets.
  - THORChain asset prices are now cached for improved performance and reduced network usage.
- **Improvements**
  - Enhanced asset price retrieval logic to include THORChain alongside other supported chains.


VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
```diff
@@ -216,10 +221,119 @@ class ThorchainService: ThorchainSwapProvider {
...
+        
+        // Fetch fresh data if cache expired or doesn't exist
+        do {
+            let price = try await fetchAssetPrice(assetName: assetName)
+            if price > 0 {
+                self.cacheAssetPrices.set(cacheKey, (data: price, timestamp: Date()))
+            }
+            return price
+        } catch {
+            return 0.0
```
johnnyluo, line 246: can you please print out the error, so we know if it failed?

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Services/Thorchain/ThorchainService.swift
enriquesouza: Sure.

PR #2239, opened 2025-05-14
TCY native token price

enriquesouza:
this fixes the tcy token price, it does not come from cg so we can't allow any provider id
image

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the price provider information for the "TCY" token on ThorChain to improve accuracy in token data display.


realpaaao (approved):
LGTM

PR #2262, opened 2025-05-18
fix ubond

enriquesouza:
Fix the format for unbound. It was formatting to locale, but we need that only in big integer 

## Summary by CodeRabbit

- **Refactor**
  - Updated the display format for amounts to use plain numeric strings instead of locale-specific formatting.


realpaaao (approved):
LGTM

PR #2276, opened 2025-05-22
list unbond

enriquesouza:
This PR is used to load all the nodes you have bonded assets, it should only load the dropdown if you have any. Once it loads then you can select the node, it will autofill the fields, and you can proceed to unbond it. 

https://runescan.io/address/thor103xklz882ffz6vwjwcrawwt8tn57ngrhpfq3cl

image

image

image

image


VultisigApp/VultisigApp/Services/Thorchain/ThorchainService+TCYStake.swift
```diff
@@ -118,4 +118,53 @@ extension ThorchainService {
...
     
+    func fetchNodeBonds(address: String, completion: @escaping ([ThorchainActiveNodeBondResponse]) -> Void) {
+        let urlString = Endpoint.fetchNodeBonds(address: address)
+        guard let url = URL(string: urlString) else {
+            completion([])
+            return
+        }
+
+        let task = URLSession.shared.dataTask(with: URLRequest(url: url)) { data, response, error in
+            if let error = error {
```
johnnyluo, line 129: can you please do something like 
```
var result: [ThorchainActiveNodeBondResponse] = []
defer {
                completion(result)
}
```

johnnyluo (changes requested):
CleanShot 2025-05-27 at 09 34 30@2x

1. The UI is not friendly , for example , in this case , the node is in active state , which can't unbond , but as a result I can't see it on the list
2. the amount input field has weird style

Please work with Felix and Mia to get a better design for this feature

Issue #1690, opened 2024-12-18
[IMPROVE] SOL Blockhash error message


realpaaao:
Please improve the error message on SOL "Blockhash not found" to "Transaction time out due to slow signing. Please try again."

This makes it more clear for the user what to do.

![Image](https://github.com/user-attachments/assets/953bf3e6-f576-458f-beb2-80381f3220ef)

enriquesouza:
https://github.com/vultisig/vultisig-ios/pull/2330

PR #2330, opened 2025-06-06
fix: handle specific error messages for transaction timeouts 

enriquesouza:
This should fix https://github.com/vultisig/vultisig-ios/issues/1690


johnnyluo (approved):
LGTM

PR #2331, opened 2025-06-06
SUI known tokens prices and prices providers 

enriquesouza:
We are basically looking for the price provider for known tokens that were minted on SUI as well. We have many tokens in our Token Store, so we can search it in our own code. Once we find it we reuse it. 

@realpaaao to reflect it you will have to remove the tokens, the chain and add them back.

image

## Summary by CodeRabbit

- **New Features**
  - Token metadata now includes a price provider ID when available, enabling more detailed information for each token.


johnnyluo (approved):
LGTM

PR #2341, opened 2025-06-10
filter out spam tokens by validating price provider ID for EVM and Solana

enriquesouza:
Filter only for auto discovery the spams only for EVM and SOLANA, Thor and sui don't have this issue


johnnyluo (approved):
LGTM

PR #2445, opened 2025-06-30
Enhance TON bounceable logic

enriquesouza:


VultisigApp/VultisigApp/Chains/Ton.swift
```diff
@@ -43,14 +43,19 @@ enum TonHelper {
             $0.amount = UInt64(keysignPayload.toAmount.description) ?? 0
             $0.mode = sendMode
             
+            // Set bounceable based on chainSpecific parameter
```
johnnyluo, line 46: @enriquesouza , when `sendMaxAmount` is true , do we need to override the amount to 0? 
since the sendMode will be `attachAllContractBalance`

PR #2522, opened 2025-07-17
Add TRON freeze and unfreeze functionality

enriquesouza:


enriquesouza:
It is ready and done.

Issue #2853, opened 2025-08-29
[BUG Release v1.24] - Swaps from Mantle to Major Chains (Solana/Blast/Base/Avalanche) Fail with “Failed to Forward Tx to Sequencer” Error For All Possible Amount


braja555:
**Summary:**
All cross-chain swaps originating from Mantle (MNT) to major chains such as Avalanche, Base, Blast, or Solana fail for all possible amount. The error returned on success screen generated txn hash is:

> “failed to forward tx to sequencer, err: 'failed to forward tx to sequencer, err: 'failed to forward tx to sequencer, err: 'intrinsic gas too low: gas 536900, minimum needed 101277320’”

**Steps to Reproduce:**

1. Open Vultisig app on iOS (Fast Vault v1.24.0).
2. Navigate to Swap tab.
3. Select:

   * From Chain: Mantle
   * From Token: MNT
   * To Chain: Any major destination chain (e.g., Avalanche, Base, Solana)
4. Enter any amount.
5. Proceed through the confirmation and signing flow.
6. Observe the error on success screen txn hash.

![image](https://github.com/user-attachments/assets/bec4faa1-de88-4d0b-86a9-eaa008d554b6)
![image](https://github.com/user-attachments/assets/b8fe3e6c-d69c-46b1-adec-51f53c17056d)
![image](https://github.com/user-attachments/assets/ecc6f7e5-025e-4c02-9fa5-37169a5c10a1)
![image](https://github.com/user-attachments/assets/57f10f5f-19bf-40ad-a1b6-cc1f57114730)

PR #2860, opened 2025-08-30
[FIX] - Swaps from Mantle to Major Chains (Solana/Blast/Base/Avalanche) Fail with “Failed to Forward Tx to Sequencer” Error For All Possible Amount #2853

enriquesouza:
## Summary by CodeRabbit

- Bug Fixes
  - Ethereum swaps now respect provided gas limits during processing.
  - Corrected Mantle gas calculations by applying dedicated default limits for swaps and transfers.
- Refactor
  - Separated Mantle from base-like chains to apply its own gas defaults across token types.
  - Simplified gas handling paths while keeping behavior unchanged for other chains.


johnnyluo (approved):
LGTM

PR #2969, opened 2025-09-11
Add transaction status handling and localization updates

enriquesouza:
Tests

// Covered chains in checkPendingTransactions():
let cosmosChains: [Chain] = [
    .thorChain,      // ✅ Tested OK
    .mayaChain,      // ✅ Tested OK
    .gaiaChain,      // ✅ Tested OK (Cosmos Hub)
    .kujira,         // ✅ Tested OK  
    .osmosis,        // ✅ Tested OK
    .dydx,           // 🧪 Not tested
    .terra,          // ✅ Implemented
    .terraClassic,   // ✅ Implemented
    .noble,          // 🧪 Not tested
    .akash           // ✅ Tested OK
]

## Summary by CodeRabbit

- New Features
  - Pending-transaction tracking across supported Cosmos-family chains (including THORChain and MayaChain) with automatic per-chain polling and UI refresh.
  - Send screen now shows a live countdown for the oldest pending transaction and updates the primary action button and loading state.
  - Broadcasts enqueue pending transactions for nonce tracking.

- Bug Fixes
  - Prevents nonce conflicts by pausing sends while pending transactions exist.
  - Improves nonce freshness by bypassing/clearing cache for chains with pending-transaction support.


johnnyluo:
The PR works mostly , but I am not sure what kind of issue this PR is trying to fix , or what goal it is. 
- If we would like to poll the transaction status and show it to user , then we shall just do that in the transaction done screen 
- If we would like to allow user to sign and broadcast multiple transactions back to back , then you could simply increase the sequence no for every transaction your broadcast

This PR however update sending / swap screen to stop and wait for the previous transaction is commit to block and then continue , which to me it seems not what normal user will care about

enriquesouza:
> The PR works mostly , but I am not sure what kind of issue this PR is trying to fix , or what goal it is.
> 
> * If we would like to poll the transaction status and show it to user , then we shall just do that in the transaction done screen
> * If we would like to allow user to sign and broadcast multiple transactions back to back , then you could simply increase the sequence no for every transaction your broadcast
> 
> This PR however update sending / swap screen to stop and wait for the previous transaction is commit to block and then continue , which to me it seems not what normal user will care about

Bug Fixes

Prevents nonce conflicts by pausing new sends while a pending transaction exists on supported chains.
Improves nonce freshness by selectively bypassing cache on chains with pending-transaction support and clearing cache after confirmation.

VultisigApp/VultisigApp/Services/PendingTransactionManager.swift
```diff
@@ -0,0 +1,335 @@
...
+        case .gaiaChain, .kujira, .osmosis, .dydx, .terra, .terraClassic, .noble, .akash:
+            // Use nonce-based checking for other Cosmos chains
+            return try await checkCosmosNonceChanged(transaction: pendingTransactions[txHash], chain: chain)
+        default:
+            return false
+        }
+    }
+    
+    /// Check if THORChain nonce has incremented (indicating transaction confirmation)
+    private func checkThorchainNonceChanged(transaction: PendingTransaction?) async throws -> Bool {
```
gastonm5, line 236: Let's move these methods to each Blockchain specific service?

VultisigApp/VultisigApp/Views/Send/Screens/SendDetailsScreen.swift
```diff
@@ -40,27 +40,60 @@ struct SendDetailsScreen: View {
...
         
         container
         .disabled(sendCryptoViewModel.showLoader)
         .overlay(sendCryptoViewModel.showLoader ? Loader() : nil)
+        .onAppear {
+            // Initialize button state immediately based on chain
+            if tx.coin.chain.supportsPendingTransactions {
+                sendCryptoViewModel.isCheckingPendingTransactions = true
+            } else {
+                sendCryptoViewModel.isCheckingPendingTransactions = false
```
gastonm5, line 55: Let's move this to viewModel and remove duplication below?

VultisigApp/VultisigApp/Views/Send/Screens/SendDetailsScreen.swift
```diff
@@ -272,11 +354,13 @@ struct SendDetailsScreen: View {
     private func onRefresh() async {
         async let gas: Void = sendCryptoViewModel.loadGasInfoForSending(tx: tx)
         async let bal: Void = BalanceService.shared.updateBalance(for: tx.coin)
-        _ = await (gas, bal)
+        async let pendingCheck: Void = PendingTransactionManager.shared.forceCheckPendingTransactions()
+        _ = await (gas, bal, pendingCheck)
```
gastonm5, line 358: Do we need to check pending transactions for all chains here?

VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -78,11 +86,14 @@ final class BlockChainService {
...
-            if localCacheItem.date.addingTimeInterval(cacheSeconds) > Date() {
-                return localCacheItem.blockSpecific
+        // Skip cache for chains that support pending transactions to ensure fresh nonce
+        if !tx.fromCoin.chain.supportsPendingTransactions {
+            if let localCacheItem =  self.localCache.get(cacheKey) {
+                let cacheSeconds = getCacheSeconds(chain: tx.fromCoin.chain)
+                // use the cache item
+                if localCacheItem.date.addingTimeInterval(cacheSeconds) > Date() {
+                    return localCacheItem.blockSpecific
+                }
```
gastonm5, line 96: Can we try to move this to a separate method so we prevent code duplication in case we need to add it for another tx type?

VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -101,7 +112,10 @@ final class BlockChainService {
             memo: nil,  // Swaps don't have memos
             feeMode: .fast
         )
-        self.localCache.set(cacheKey, BlockSpecificCacheItem(blockSpecific: specific, date: Date()))
+        // Only cache for chains that don't support pending transactions
+        if !tx.fromCoin.chain.supportsPendingTransactions {
+            self.localCache.set(cacheKey, BlockSpecificCacheItem(blockSpecific: specific, date: Date()))
+        }
```
gastonm5, line 118: Same as above here

VultisigApp/VultisigApp/Services/PendingTransactionManager.swift
```diff
@@ -0,0 +1,335 @@
...
+    static let shared = PendingTransactionManager()
+    
+    @Published private var pendingTransactions: [String: PendingTransaction] = [:]
+    private var pollingTasks: [Chain: Task<Void, Never>] = [:]
+    
+    private init() {
+        // Don't start polling automatically - only when needed
+    }
+    
+    deinit {
```
gastonm5, line 41: Does this get executed at any time? I believe it won't since we're using a singleton

VultisigApp/VultisigApp/Services/PendingTransactionManager.swift
```diff
@@ -0,0 +1,335 @@
...
+        self.txHash = txHash
+        self.address = address
+        self.chain = chain
+        self.sequence = sequence
+        self.timestamp = Date()
+        self.isConfirmed = false
+    }
+}
+
+class PendingTransactionManager: ObservableObject {
```
gastonm5, line 31: Maybe we can improve the logic overall and only keep the newest pending transaction for each chain in DB? 🤔

VultisigApp/VultisigApp/Services/PendingTransactionManager.swift
enriquesouza: They have a different purpose IMO.

VultisigApp/VultisigApp/Views/Send/Screens/SendDetailsScreen.swift
enriquesouza: We are always checking for the specific chain, and in this case for the selected chain. So we only check if the chain supports that.

johnnyluo (approved):
LGTM

PR #3104, opened 2025-10-02
fix: removed nfts tab

gastonm5:
## Description

Please include a summary of the change and which issue is fixed. 

Fixes #<issue-number>

## Which feature is affected?

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

## Summary by CodeRabbit

* **Refactor**
  * Removed the multi-tab header in Vault details — the disabled NFTs placeholder and tab selection were removed so users now see only the Portfolio view.
  * Vault header simplified: the selectable segmented control was replaced by a non-interactive "Portfolio" label with an underline, keeping existing portfolio actions and data.
  * Chain details header simplified: segmented control replaced by a static "Tokens" label and accent line.


gastonm5:
> This one will cause my Phone to freeze when start , something is not right in the SegmentedControlItem, my iPhone 13 on IOS 26.0.1

Just removed segmented control usage from VaultMainScreen, we can add it back in case we use NFTs. Check if it works on chain detail screen please, as it's the same component

johnnyluo (approved):
LGTM

PR #3123, opened 2025-10-04
Dot dynamic fee

enriquesouza:


VultisigApp/VultisigApp.xcodeproj/project.pbxproj
```diff
@@ -6661,7 +6661,7 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/vultisig/commondata";
 			requirement = {
-				branch = main;
+				branch = polkadot-fee;
```
gastonm5, line 6664: should we wait to be merged before merging this PR?

VultisigApp/VultisigApp/Model/Coin.swift
```diff
@@ -190,7 +190,7 @@ class Coin: ObservableObject, Codable, Hashable {
         case .sui:
             return "3000000"
         case .polkadot:
-            return "250000000" // 0.025
+            fatalError("Polkadot uses ONLY dynamic fee calculation - no default fees allowed")
```
gastonm5, line 193: Would it be better to throw error instead of crashing here?

VultisigApp/VultisigApp/Services/BlockChainService.swift
```diff
@@ -362,8 +366,23 @@ private extension BlockChainService {
             
         case .polkadot:
             let gasInfo = try await dot.getGasInfo(fromAddress: coin.address)
-            return .Polkadot(recentBlockHash: gasInfo.recentBlockHash, nonce: UInt64(gasInfo.nonce), currentBlockNumber: gasInfo.currentBlockNumber, specVersion: gasInfo.specVersion, transactionVersion: gasInfo.transactionVersion, genesisHash: gasInfo.genesisHash)
+            let dynamicFee = try await dot.calculateDynamicFee(
+                fromAddress: coin.address,
+                toAddress: toAddress ?? "",
+                amount: amount ?? BigInt.zero,
```
gastonm5, line 372: Should we skip fee calculation if these values are nil?

gastonm5 (approved):
LGTM, just small comments

VultisigApp/VultisigApp/Services/BlockChainService.swift
enriquesouza: Well, the amount should never be zero if DOT, but can be nil for other chains.

VultisigApp/VultisigApp/View Models/SendCryptoViewModel.swift
```diff
@@ -220,30 +220,15 @@ class SendCryptoViewModel: ObservableObject {
...
-                    if percentage == 100 {
-                        gas = tx.coin.feeDefault.toBigInt()
-                    }
-                    
-                    tx.amount = "\(tx.coin.getMaxValue(gas).formatToDecimal(digits: tx.coin.decimals))"
-                    setPercentageAmount(tx: tx, for: percentage)
-                    
-                    convertToFiat(newValue: tx.amount, tx: tx, setMaxValue: tx.sendMaxAmount)
-                    print("Failed to get DOT gas estimation, using default fee: \(error.localizedDescription)")
+                    throw error
```
johnnyluo, line 231: This might cause the app to crash , because `setMaxValues` doesn't have any other try/catch,

johnnyluo (changes requested):
see the comments

johnnyluo:
Let's hold off this one after we ship 1.26 , this change will cause cross device keysign issues

johnnyluo (approved):
LGTM

Issue #3205, opened 2025-10-16
[BUG Rel v1.27.17] - Unable to perform cross-chain swaps from Ethereum to any chain


braja555:
**Summary:**

Cross-chain swaps initiated from the **Ethereum (EVM L1)** network are consistently failing on both** Fast Vault and secure vault v1.27.17**, regardless of the destination chain.,  resulting in **“transaction hash not found”** or failure in Li.Fi scan.

### **Tested Transactions:**

1. https://scan.li.fi/tx/0xc606cff4f868267ff128c617503c864d5984295e151f0851cb2e8bc5b87fa3c7
2. https://scan.li.fi/tx/0x736a990d024aba81dd3a8ba275f27d896d3dbc5977b4d538b0d1c4f5d1b8a5a4
3. https://scan.li.fi/tx/0x2760c863e4fb37b19a7d3cf8abf2883a8750e784af35da5b82147f6c77b41d37 
4. https://scan.li.fi/tx/0xa02e9b73adcb5a336501c8edfacda984e11f1db87ef72035410c16072c042682
5. https://scan.li.fi/tx/0x0b93a6fd85a0702503d0c867af63d9d524bcf56f525e7b7b06e040a04b175a3d

enriquesouza:
> I checked sending is fine , so not sure this is a li.fi issue

I could swap from SOL to ETH, from THOR to ETH and from ETH to ZEC

https://www.mayascan.org/tx/0b1f4fcbff85dc10b7415f8b5ac6e6c023b27a2d68dabbe831b817091b365e5a

https://thorchain.net/tx/27359b041e383687d1f11fed36beb82f661b010ac463c0c79d9660fe940c6985

The issue is on LIFI

gastonm5:
> > Somehow transaction stays on pending and gets dropped.
> 
> Funny it works with MAYA and THOR. I will check the fees for LI.FI, probably something like Kyber, they send a lower fee that it should so the tx does not go

I've checked on previous version and extension also, and they're getting dropped, maybe increasing the fee would work

enriquesouza:
I already fixed that

PR #3213, opened 2025-10-16
fix: improve gas price handling in LIFI

enriquesouza:
FIx
https://github.com/vultisig/vultisig-ios/issues/3205

Test
https://scan.li.fi/tx/0xe304bd002266b9b7bb07d8a136e1a668e01819003c31681e8a0043e2c90d75fc

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gas price calculation for swaps. The system now ensures gas fees are not set too low by using the higher value between provider-supplied and calculated network fees, reducing the risk of transaction failures on Ethereum.


johnnyluo (approved):
LGTM

Issue #3309, opened 2025-11-03
[ADD] Remove Thorchain stagenet from production


realpaaao:
We should not show Thorchain stagenet in the production version for the user:

<img width="386" height="239" alt="Image" src="https://github.com/user-attachments/assets/ce37b52a-6dbf-4f20-9efa-4cea18f0a0df" />

Please remove and let us enable it in the advance settings

PR #3322, opened 2025-11-06
THORChain Stagenet hidden

enriquesouza:
FIX https://github.com/vultisig/vultisig-ios/issues/3309

## Summary by CodeRabbit

* **New Features**
  * Added THORChain Stagenet testnet toggle in advanced settings
  * Implemented asset filtering based on testnet enablement, automatically controlling visibility of chain-specific tokens


SxMShaDoW:
Would an end user ever need to use stagenet? Or should this be a debug feature that a developer needs to use it. Are we planning to support more testnets? Just seems a bit weird to have it as a toggle.

PR #3538, opened 2025-12-05
Fix: Random Vault Switching & Persistence Issues

### LINKED ISSUE #3513: [BUG] When in a vault, it randomly goes back to a different vault (does not persist)

enriquesouza:
# Fix: Random Vault Switching & Persistence Issues

## Description
This PR addresses a critical issue where the application would randomly switch to a different vault (often the first one in the list) after sending a transaction or reloading, failing to persist the selected vault correctly.

## The Problem
The issue was caused by unsafe fallback logic in `AppViewModel` and `DeeplinkViewModel`. When the application failed to immediately find the persisted vault (e.g., due to a slight mismatch or race condition) or during a reload, it would default to `vaults.first`. This "fail-open" behavior resulted in the user being silently switched to the wrong vault.

## The Fix
1.  **Removed Unsafe Fallbacks**: Removed `set(selectedVault: vaults.first)` in `AppViewModel.loadSelectedVault` and similar logic in `DeeplinkViewModel`.
2.  **Strict Matching**: Implemented strict matching for the persisted vault using both `name` and `pubKeyECDSA`.
3.  **Fail-Safe State**: If the persisted vault cannot be found, the app now sets `selectedVault = nil` and shows the vault selector, forcing the user to make a conscious choice rather than assuming a default.

## Verification
- **Manual Test**: Verified that sending a transaction from a specific vault keeps that vault, selected.
- **Persistence**: Verified that reloading the app persists the correct vault or shows the selector if the vault is missing.

## Related Issue
Fixes #3513

## Summary by CodeRabbit

* **Changes**
  * Improved vault selection behavior. Previously, the app would automatically select the first available vault when vault information was missing or unavailable. Now, the app displays the vault selector screen instead, giving you explicit control over which vault to use. This change applies to both normal vault operations and deeplink-based transactions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/VultisigApp/VultisigApp/View Models/AppViewModel.swift
+++ b/VultisigApp/VultisigApp/View Models/AppViewModel.swift
@@ -49,18 +49,18 @@ class AppViewModel: ObservableObject {
     
     func loadSelectedVault(for vaults: [Vault]) {
         if vaultName.isEmpty || selectedPubKeyECDSA.isEmpty {
-            set(selectedVault: vaults.first)
+            set(selectedVault: nil, showingVaultSelector: true)
             return
         }
 
         for vault in vaults {
-            if vaultName==vault.name && selectedPubKeyECDSA==vault.pubKeyECDSA {
+            if vaultName == vault.name && selectedPubKeyECDSA == vault.pubKeyECDSA {
                 set(selectedVault: vault)
                 return
             }
         }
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #3692, opened 2026-01-15
Reduce delay on fetching swaps


realpaaao:
## Summary
Improve the performance of swap fetching by reducing the current delay. This will provide a better user experience by making swap operations more responsive.

## Requirements
TBD

## Acceptance Criteria
- [ ] TBD

## Version
N/A

PR #3763, opened 2026-01-31
feat: improve swap quote fetching performance

enriquesouza:
### Performance Improvements
Fixes https://github.com/vultisig/vultisig-ios/issues/3692
This PR introduces significant performance optimizations for swap quote fetching:

1.  **Parallel Provider Execution with Priority Respect**:
    *   **Old Approach (Sequential)**: We waited for Provider A to finish (or fail/timeout) before even starting Provider B. If Provider A timed out (e.g., 10s), the user waited 10s + Provider B time.
    *   **New Approach (Parallel Start)**: We start Provider A and Provider B simultaneously. We still await them in priority order. If Provider A fails at 10s, Provider B (which likely finished seconds ago) is returned *immediately*.
    *   **Gain**: Eliminates cumulative latency from failed providers. Worst-case scenario becomes best-case execution time of the backup provider.

2.  **Concurrent Quote & Fee Calculation**:
    *   **Old Approach**: `fetchQuote` -> await -> `fetchFees` -> await.
    *   **New Approach**: `fetchQuote` and `fetchFees` run in parallel.
    *   **Gain**: Saves the duration of the shorter request (usually 200-500ms) on every single refresh.

3.  **Responsiveness**:
    *   Reduced debounce delay from 0.3s to 0.2s for snappier UI feedback without overloading the network.

### Changes
- `SwapCryptoViewModel.swift`: Parallelized `updateQuotes` and `updateFees`.
- `SwapService.swift`: Refactored `fetchQuote` to start all providers in parallel but return results based on priority order.

Tests
https://runescan.io/tx/59e11a3b7f560fad5e1686b2075372c571000361536a7e2c5b4aea5893077973
https://scan.li.fi/tx/0x4b46f4afcfa96b1a5067ba8090b8806c6a51b990736defddca256f65c1f0fdad
https://runescan.io/tx/dc8f45a9bb118ed257026e559e5b036fd6e8728ef4d01c9245812a7785efd5ec


johnnyluo (approved):
LGTM

PR #3838, opened 2026-02-07
feat: add macOS-only 'Download via Website' button

### LINKED ISSUE #3439: [BUG] Auto-update removed: App Store now mandatory, no direct download option available on website.
Describe the bug
The iOS/macOS app no longer includes auto-update functionality and now forces users to update exclusively through the Apple App Store. Previously, the app would auto-update itself without requiring App Store interaction. Users who prefer not to use the App Store (for privacy reasons or because they don't maintain an Apple ID login on their devices) must now manually hunt down release files on GitHub, which is significantly less user-friendly than the previous system.
To Reproduce

Attempt to update the app on a macOS device without being logged into the Apple App Store
Check the main website for a direct download link to the latest release
Discover that no auto-update occurs and no direct download link is available on the main site. Links for macos points back to apple store. 
Be forced to either use the App Store or manually search through GitHub repositories for the release file

Expected behavior
One or more of the following should be available:

The app should auto-update itself without requiring App Store access (as it did previously)
OR: A direct, prominent link to the latest release file (.dmg, .ipa, etc.) should be available on the main website
OR: Clear documentation on the main site explaining how to download updates without using the App Store

System:

Device: Any MacOS device
App Version: Latest version

Additional context
This change significantly impacts privacy-conscious users who deliberately avoid logging into Apple services on their devices. The previous auto-update system was ideal for this use case. Forcing App Store usage creates an unnecessary barrier for legitimate users who simply want to keep their software up to date without compromising their privacy preferences.

enriquesouza:
## Summary

Adds a macOS-only **"Download via Website"** button that links to the [GitHub Releases](https://github.com/vultisig/vultisig-ios/releases) page, so macOS users can update without the App Store.

## Changes

- **URL.swift**: Added `GitHubReleasesURL` to `StaticURL`
- **UpdateCheckUpToDateView.swift**: Added macOS-only download button (visible when app is up to date)
- **UpdateCheckUpdateNowView.swift**: Added macOS-only download button (visible when update is available)
- **HomeScreen.swift**: Added macOS-only option in the home screen update alert
- **Localizable.strings** (×7): Added `downloadViaWebsite` key for all languages

All changes wrapped in `#if os(macOS)` — iOS is unaffected.

Closes #3439

image

## Release Notes

* **New Features**
  * Added a "Download via Website" option in the application update dialog for macOS users, providing an alternative download method during app updates.

* **Localization**
  * Added translations for the new download option across multiple languages: English, German, Spanish, Croatian, Italian, Portuguese, and Simplified Chinese.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleDashboardView.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleDashboardView.swift
@@ -21,10 +21,6 @@ struct CircleDashboardView: View {
         !appClosedBanners.contains(circleDashboardBannerId)
     }
 
-    var walletUSDCBalance: Decimal {
-        return CircleViewLogic.getWalletUSDCBalance(vault: vault)
-    }
-
     var body: some View {
         content
     }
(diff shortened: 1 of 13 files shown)


johnnyluo (approved):
LGTM

PR #3846, opened 2026-02-09
Fix False Positive Error Logging (Cancellation Noise)

### LINKED ISSUE #3779: Fix False Positive Error Logging (Cancellation Noise)
Normal user navigation triggers task cancellations, which are currently caught and logged as "Fetch Balances error" or "request cancelled" warnings. These are not errors and pollute the logs.

**Scope:**
- Modify `BalanceService` to catch `CancellationError` and either ignore it or log at `.debug`.
- Modify `HTTPClient` to handle `URLError.cancelled` gracefully (level `.debug`).
- Modify `VaultDetailViewModel` to handle balance update cancellations silently.

enriquesouza:
## Summary
Closes #3779

Silences false-positive error logging and console noise caused by task cancellations, background polling, and verbose HTTP instrumentation.

## Changes

| File | Change |
|---|---|
| `BalanceService.swift` | Catch `CancellationError` / `URLError.cancelled` → log at `.debug` instead of `.warning` |
| `HTTPClient.swift` | Cancellation logs: `.warning` → `.debug`; Request/response logs: `.info` → `.debug` |
| `BackgroundTransactionPoller.swift` | Remove all `print()` statements |
| `PendingTransactionStorage.swift` | Remove cleanup `print()` statement |
| `AppMigrationService.swift` | Remove all migration `print()` statements |
| `SolanaService.swift` | Remove `print()` from `fetchSolanaTokenInfoList` error path |

## Result
The Xcode console is now clean during normal operation. Debug-level events are still traceable via Console.app when needed.

## Summary by CodeRabbit

* **Chores**
  * Reduced log verbosity across the app: network request/response logging was removed, error logging consolidated, and cancellation cases are handled without noisy logs.
  * Silenced many debug/console messages in background polling, pending transaction storage, token/price services, migration routines, and multiple blockchain service handlers.
  * Preserved behavior and control flow; only runtime logging/output was reduced.


VultisigApp/VultisigApp/Services/Network/HTTPClient.swift
```diff
@@ -247,78 +239,23 @@ private extension HTTPClient {
 // MARK: - Logging Methods
 private extension HTTPClient {
 
-    /// Logs the outgoing HTTP request
```
gastonm5: I think it's useful to keep logging request/response for debugging purposes, maybe we can hide it through some Xcode scheme environment variable?

PR #4107, opened 2026-03-30
fix: add block_id fallback filter for Dash Blockchair UTXO path

realpaaao:
## Summary
- Add `blockId` field to `BlockchairUtxo` model
- Filter out unconfirmed UTXOs (`blockId <= 0`) for Dash in the Blockchair path
- Safety net in case Dash RPC (from #4103) is unavailable and code falls back to Blockchair

## Context
PR #4103 switched Dash UTXO fetching to native RPC (`getaddressutxos`) which only returns confirmed UTXOs. This is a defense-in-depth addition: if Dash ever routes through the Blockchair path, unconfirmed UTXOs are filtered out to prevent `tx-txlock-conflict`.

## Changes
- `Blockchair.swift`: Added `blockId: Int?` to `BlockchairUtxo`
- `KeysignPayloadFactory.swift`: Added `!isDash || (item.blockId ?? 0) > 0` guard in Blockchair compactMap

## Test plan
- [ ] Verify Dash sends still work (primary RPC path)
- [ ] Verify other UTXO chain sends unaffected (BTC/LTC/DOGE/BCH/ZEC)

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift
@@ -107,11 +107,13 @@ struct KeysignPayloadFactory {
             )
         } else {
             // Existing Blockchair path for all other UTXO chains
+            let isDash = keysignPayload.coin.chain == .dash
             let info = await utxo.getByKey(key: keysignPayload.coin.blockchairKey)?.utxo?.compactMap { item -> UtxoInfo? in
                 guard
                     let txHash = item.transactionHash, !txHash.isEmpty,
                     let value = item.value,
-                    let index = item.index, index >= 0
+                    let index = item.index, index >= 0,
+                    !isDash || (item.blockId ?? 0) > 0
                 else {
                     return nil
                 }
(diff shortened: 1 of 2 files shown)


VultisigApp/VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift
```diff
@@ -107,11 +107,13 @@ struct KeysignPayloadFactory {
             )
         } else {
             // Existing Blockchair path for all other UTXO chains
+            let isDash = keysignPayload.coin.chain == .dash
```
johnnyluo: `isDash` is always `false` here — the `else` branch is only reached when `chain != .dash`, so this variable is dead code.

VultisigApp/VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift
```diff
@@ -107,11 +107,13 @@ struct KeysignPayloadFactory {
...
             let info = await utxo.getByKey(key: keysignPayload.coin.blockchairKey)?.utxo?.compactMap { item -> UtxoInfo? in
                 guard
                     let txHash = item.transactionHash, !txHash.isEmpty,
                     let value = item.value,
-                    let index = item.index, index >= 0
+                    let index = item.index, index >= 0,
+                    !isDash || (item.blockId ?? 0) > 0
```
johnnyluo: This condition simplifies to `true || (item.blockId ?? 0) > 0` at runtime, which is always `true`. No unconfirmed UTXOs will ever be filtered here.

PR #4131, opened 2026-04-08
[Tron] Fix OUT_OF_ENERGY failures by enforcing minimum feeLimit thresholds

DanialV:
## Description

This PR fixes an issue where TRON smart contract transactions—such as TRC20 token transfers and swaps—would fail with an `OUT_OF_ENERGY` error, even when the user had sufficient delegated energy to cover the execution costs. 

Previously, when the user had enough energy, `TronService` estimated a superficial fee of 1 TRX. This value was directly mapped to the transaction's `fee_limit` parameter. On TRON, `fee_limit` acts as a strict upper bound on execution gas limit, even if it is completely absorbed by free Energy. A 1 TRX limit restricts the smart contract execution to roughly ~2,400 Energy. Because actions like sending USDT require at least ~65,000 Energy, the TVM would reject the execution, mistakenly concluding the transaction did not authorize enough gas, and throw an `OUT_OF_ENERGY` exception.

This PR fixes the issue by replacing the direct `feeLimit` assignments with bounded maximum minimums to ensure the smart contract can be executed successfully using the user's stacked energy:
- **`150_000_000` (150 TRX) minimum** for regular TRC20 transfers, TRC10 asset transfers, and freeze/unfreeze operations.
- **`400_000_000` (400 TRX) minimum** for swaps and complex dApp contract triggers (`triggerSmartContract`).

Brief comments explaining the rationale were also added to `Tron.swift`.

Fixes #<issue-number>

## Which feature is affected?
- [x] Sending  - Please attach a tx link here
https://tronscan.org/#/transaction/48145cf071059937d77bd3cb65e54ee08fe7b14f27192abb7b7f7a48a43e1f5a

## Checklist

- **Agent**: Antigravity
- **Issue**: N/A
- **Confidence**: High
- **Needs human review for**: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TRON transaction fee handling by enforcing minimum fee thresholds consistently for native tokens and contract interactions.
  * Enhanced fee calculation logic to reduce transaction failures when energy coverage is insufficient.


DanialV:
@johnnyluo Hi, could you please take a look at this PR when you have a moment? All checks have passed.

gastonm5 (changes requested):
Thanks @DanialV ! This is a valid fix, but the gasEstimation should happen on `TronService.calculateTronFee` as we calculate it there. Otherwise, the fee we show to the user will differ than the actual used fee.
Could you also add some more information on where you got these numbers for the minimum fees?

DanialV:
> Thanks @DanialV ! This is a valid fix, but the gasEstimation should happen on `TronService.calculateTronFee` as we calculate it there. Otherwise, the fee we show to the user will differ than the actual used fee. Could you also add some more information on where you got these numbers for the minimum fees?

I have updated the code to address your feedback!

Moved Logic to TronService: Instead of overriding the fee right at the last moment in Tron.swift, I reverted all of those assignments back to Int64(gasEstimation). I then migrated the upper-bound limit enforcing directly into the calculateTronFee() and getBlockInfo() pipeline. This perfectly synchronizes what you sign with what is displayed to the user via Vultisig's gasFeeEstimation.
Added Explanation Comments: I added explanatory block comments natively into TronService.swift and heavily updated the pr_description.md artifact to detail exactly why we chose 150 TRX and 400 TRX:
150 TRX (150_000_000) is the standard minimum threshold used across the industry (e.g., Trust Wallet Core and TronWeb) because it's strictly enough to securely absorb standard TRC20 65k–130k energy consumption fluctuations.
400 TRX (400_000_000) ensures safety for cross-chain router transactions and Swaps which notoriously require upwards of 300k+ energy to execute successfully.

VultisigApp/VultisigApp/Blockchain/Tron/Service/TronService.swift
```diff
@@ -119,10 +126,17 @@ class TronService {
                 let isDestinationActive = try await checkIfAccountIsActive(address: to)
                 let energyRequired = isDestinationActive ? 65000 : 130000
 
+                // Minimum safe limit for TRC20 execution. We use 150 TRX (150_000_000 sun) as the recommended 
+                // lower bound for `feeLimit` in standard Wallets to ensure TVM doesn't fail with OUT_OF_ENERGY 
+                // if slightly more energy is consumed than standard.
+                let minimumSafeFeeLimit = BigInt(150_000_000)
```
johnnyluo, line 132: 150 TRX is larger than all the previous hardcode value (18 TRX) and (36 TRX) 

Can you please give more context why and where the number is from?

gastonm5:
Spent some time verifying this against the TRON docs and our Android implementation, and there are three issues that should be addressed before this lands. Sharing the full picture here so the next iteration can take a different shape.

## TRON `fee_limit` mechanics — confirmed

PR's diagnosis is correct: per [TRON docs — set-feelimit](https://developers.tron.network/docs/set-feelimit), `fee_limit` is a **strict cap on the energy budget**, not just a TRX-burn cap:

- `max_energy_for_tx = fee_limit / energy_unit_price` (≈420 sun/energy currently)
- If `actual_energy_consumed > max_energy_for_tx` → `OUT_OF_ENERGY` revert
- The user's staked free energy does **not** extend this cap. So 1 TRX `feeLimit` → ~2,380 energy → USDT transfer (~65k energy) reverts even when the user has plenty of staked energy. ✅ exactly what this PR is targeting.

## Three problems with the fix as it stands

### 1. CodeRabbit's Major on line 61 is valid — 400 TRX floor is bypassed on the exact case it targets

The guard `if coin.isNativeToken && calculatedFee == 0` only fires when bandwidth-discounted fee returns 0. But a **native-TRX swap from a user without bandwidth** gets a non-zero but tiny `calculatedFee` (~300 bytes × `bandwidthFeePrice` ≈ fractions of a TRX), so the `max(...)` never engages. The swap `triggerSmartContract` then gets the tiny `feeLimit` and the TVM still rejects with OUT_OF_ENERGY. This is the exact failure mode the PR is meant to fix.

### 2. `gasFeeEstimation` is dual-purpose — this PR turns the displayed UI fee into the cap, not the actual spend

`BlockChainSpecific.Tron(gasFeeEstimation:)` is consumed twice:
- `Tron.swift` writes it as `$0.feeLimit = Int64(gasEstimation)` for the contract paths.
- The verify-screen UI displays it as the "Fees" line.

After this PR:

| Scenario | Actual TRX burned | Displayed "Fees" |
|---|---|---|
| TRC20 transfer, user has 65k staked energy | **0 TRX** (free) | **150 TRX** |
| TRC20 transfer, no staked energy | ~27 TRX (65k × 420 sun) | 150 TRX |
| Swap, user has 100k staked energy | **0–~5 TRX** | 400 TRX |
| Swap, no staked energy | ~30–150 TRX | 400 TRX |

We'll get support tickets ("why is the fee so high?") and we'll block users whose TRX balance is below 400 TRX from completing legitimate sends/swaps that would actually burn 0–5 TRX.

### 3. The 150 / 400 TRX numbers — answering @johnnyluo's question

`vultisig-android` has a complete, docs-backed implementation in `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronFeeService.kt`. For reference:

- **Android typical path** (`calculateEnergyFee`): calls TRON's `/wallet/triggerconstantcontract` to simulate the actual contract call → returns `energy_used` + `energy_penalty`. Multiplied by the chain's `getMaxFactor` (chain parameter) and the contract's own `energy_factor` → `maxEnergyUnitsRequired`. Then `feeLimit = maxEnergyUnitsRequired × energyPrice`.
- **Android fallback** (`calculateDefaultFees`, used for Swap where contract bytes aren't available at fee time): `DEFAULT_MAX_ENERGY_USED = 50_000_000` energy. At 420 sun/energy that's **21 TRX**. This is the documented default that's been in prod on Android for months.
- **PR #4131**: 150 TRX (≈357k energy) and 400 TRX (≈952k energy) — between 7× and 19× over Android's fallback default. No primary source.

TRON docs cited by the Android impl:
- https://developers.tron.network/docs/set-feelimit
- https://developers.tron.network/docs/resource-model#dynamic-energy-model
- https://developers.tron.network/docs/resource-model#principle

## Recommended direction

Port the Android approach instead of raising hardcoded floors:

1. **Add to `TronAPIService`**: `triggerConstantContract(owner, contract, functionSelector, parameter)` + `getContractMetadata(contract)`. iOS currently has neither — that's why we can't do simulation-based estimation today.
2. **Refactor `TronService.calculateTronFee`** for TRC20 / `triggerSmartContract` paths:
   - Simulate → `energy_used` + `energy_penalty`.
   - Apply `contractEnergyFactor` (per-contract) and `maxFactor` (chain-wide) per the dynamic energy model docs.
   - Subtract user's staked available energy → `energyToPay`.
   - `feeLimit = maxEnergyUnitsRequired × energyPrice` (the cap, written into the tx).
   - `displayedFee = energyToPay × energyPrice` (the actual burn, shown in UI).
3. **Fallback for Swap path** (contract bytes not available at fee time): mirror Android's `DEFAULT_MAX_ENERGY_USED = 50_000_000` ≈ 21 TRX, not 400 TRX.
4. **Fix `calculateTronFee`'s error catch** — CodeRabbit also caught this (the 0.345 TRX fallback). Same minimum-safe value as the swap fallback.
5. **UX split**: keep `gasFeeEstimation` wire-format as the cap (signed into the tx), derive the UI-displayed value separately from the simulated post-discount burn so users don't see a 400 TRX phantom fee.

Internal plan with the full step-by-step is filed under `wiki/pages/projects/vultisig/tron-fee-limit-4131/plan.md` if you want the longer write-up.

Happy to pair on the port — let's pause this PR and reshape it around the simulation approach. Thanks for the dig on this @DanialV!

VultisigApp/VultisigApp/Blockchain/Tron/Service/TronService.swift
```diff
@@ -50,7 +50,14 @@ class TronService {
...
-        let finalFee = calculatedFee == 0 ? coin.feeDefault.toBigInt() : calculatedFee
+        var finalFee = calculatedFee == 0 ? coin.feeDefault.toBigInt() : calculatedFee
+        
+        // Swaps (triggerSmartContract) often originate as Native TRX with 0 calculated fee limit in this context.
+        // Tron sets upper thresholds to up to 400 TRX for complex operations like cross-chain swaps to prevent OUT_OF_ENERGY.
+        if coin.isNativeToken && calculatedFee == 0 {
+             finalFee = max(finalFee, BigInt(400_000_000))
+        }
+        
         let estimation = String(finalFee)
```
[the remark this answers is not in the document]
gastonm5: Confirmed valid — this is the exact failure mode the PR is meant to address. For a native-TRX swap from a user without bandwidth, `calculateTronFee` returns the bandwidth fee (~300 bytes × `bandwidthFeePrice` = a tiny non-zero value), so `calculatedFee == 0` is false and the 400 TRX floor never engages. The swap `triggerSmartContract` then gets that tiny `feeLimit` and the TVM still rejects with OUT_OF_ENERGY. See the top-level review comment for the full picture — the underlying issue is that `feeLimit` shouldnt be set from a static floor; it should come from a `/wallet/triggerconstantcontract` simulation (`vultisig-android` already does this).

VultisigApp/VultisigApp/Blockchain/Tron/Service/TronService.swift
gastonm5: Spent some time digging this. The 150 / 400 TRX numbers dont have a docs-backed origin — they look like generous safety margins picked ad-hoc, not derived from TRON parameters.

For reference, `vultisig-android` has a complete simulation-based implementation in `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronFeeService.kt`:

- **Typical path** (`calculateEnergyFee`): calls TRON `/wallet/triggerconstantcontract` to simulate the actual call → returns `energy_used` + `energy_penalty`. Multiplied by the chain `getMaxFactor` parameter and the contract `energy_factor` (per the [dynamic energy model docs](https://developers.tron.network/docs/resource-model#dynamic-energy-model)) → `maxEnergyUnitsRequired`. Then `feeLimit = maxEnergyUnitsRequired × energyPrice`.
- **Fallback** (`calculateDefaultFees`, used when contract bytes arent available — e.g. Swap path): `DEFAULT_MAX_ENERGY_USED = 50_000_000` energy. At 420 sun/energy = **21 TRX**.

So the docs-backed defaults are ~21 TRX for the fallback case, and a *computed* per-tx number otherwise. PR #4131s 150 TRX is ~7× the Android fallback; 400 TRX is ~19×. Neither maps to a number in the TRON docs.

iOS `TronAPIService` doesnt yet have the `triggerconstantcontract` RPC at all — that needs to be added before we can do the proper computation. The top-level review comment has the recommended port path.

gastonm5:
Closing in favor of the simulation-based fix in **#4392**, which addresses the same `OUT_OF_ENERGY` bug but derives `fee_limit` per-tx from TRON's `/wallet/triggerconstantcontract` simulation (matching the existing `vultisig-android` `TronFeeService.calculateEnergyFee` implementation) instead of hardcoded 150 / 400 TRX floors. See the [PR review here](https://github.com/vultisig/vultisig-ios/pull/4131#issuecomment-4491122280) for the rationale — short version: hardcoded floors leave the native-swap path uncovered when the user lacks bandwidth (your line-61 thread), turn the displayed UI fee into a phantom 150–400 TRX, and don't have a docs-backed basis (your line-132 thread).

Bug now tracked as **#4393**; #4392 closes it on merge.

Thanks again for the dig on this @DanialV — the energy-budget mechanics in your PR description was the right diagnosis, and your CodeRabbit's outside-diff Major (the 0.345 TRX error fallback) is also addressed in #4392.

PR #4155, opened 2026-04-15
fix: Circle DeFi deposited USDC not showing + UI polish

### LINKED ISSUE #3857: [BUG] - Circle DeFi Page Does Not Display Deposited USDC (iOS/macOS)
**Summary:**

On **iOS/macOS**, within **DeFi → Circle**, the page **does not show previously deposited USDC**, even though the deposit exists on-chain.

**Expected Behavior:**

* Deposited USDC should be visible in the **Circle DeFi page**.
* DeFi balances should match on-chain state.

image

gastonm5:
## Summary

Fixes #3857 — the Circle DeFi page not displaying previously deposited USDC — and polishes surrounding UI along the way.

- **Cache Circle positions in SwiftData.** Added `CirclePosition` `@Model` with a cascade `Vault.circlePosition` relationship, an upsert storage service, and a `CircleViewLogic.refresh(vault:)` entry point that fetches fresh balances and writes them through. `CircleSetupView` and `DefiCircleRow` now seed from `vault.circlePosition` on appear (instant render) and always re-fetch in the background, so deposited USDC shows immediately even on a cold open. `CircleView.checkExistingWallet` also skips the Circle REST round-trip when the wallet address is already persisted.
- **Format balances with the shared helpers.** `CircleSetupView` was calling `Decimal.formatted()` and hardcoding `$`, so balances rendered without thousands separators and ignored user currency. It now uses `AmountFormatter.formatCryptoAmount` for USDC amounts and `RateProvider.shared.fiatBalanceString(value:coin:)` for fiat, matching `DefiChainStakedPositionView` and respecting `SettingsCurrency`.
- **Reusable tooltip component.** Extracted `TooltipShape` out of `SwapErrorTooltipView` into `Components/Tooltip/` and added a new `InfoTooltip` (title + description + dismiss) that powers both the swap error tooltip and the Circle rewards tooltip. `TooltipShape` now takes an `arrowDirection: .up | .down`; the `.down` variant uses an explicit path (not a vertical flip) so the distinctive 4pt small corner stays at top-right in both orientations. The Circle rewards tooltip is now positioned above its info icon via an `alignmentGuide` that works regardless of dynamic tooltip height.

## Test plan

- [ ] Open a vault with an existing Circle wallet and a non-zero USDC deposit → balance shows immediately on the Circle dashboard and in the DeFi list row, even before the network fetch resolves
- [ ] Force-quit and relaunch → cached position renders on first paint; pull-to-refresh still updates the balance
- [ ] Create a new Circle wallet → setup flow works; cache populates after first successful fetch
- [ ] Deposit USDC → after returning to the dashboard, balance and fiat value update (re-fetch path)
- [ ] Withdraw USDC → same as above
- [ ] Switch `SettingsCurrency` to EUR/GBP → fiat value renders in the selected currency, not `$`
- [ ] Tap the rewards info icon → tooltip appears above the icon with the arrow pointing down at it, small corner at top-right; dismiss X works
- [ ] Trigger a swap error → `SwapErrorTooltipView` still renders correctly below its icon (same shape, up arrow)
- [ ] Delete a vault with a `CirclePosition` → cascade delete removes the cached row (no orphan)

## Summary by CodeRabbit

* **New Features**
  * Added a reusable info tooltip component and integrated it across Circle and swap UIs
  * Persisted Circle position data for faster cached balance display

* **Design/Styling**
  * New tooltip shape with directional arrow and tooltip background color
  * Reduced Circle card padding for tighter layout

* **Localization**
  * Added Circle rewards title/description in all locales
  * Renamed USDC deposited key for consistency

* **Improvements**
  * Consolidated Circle dashboard into the setup flow with refreshed loading behavior


johnnyluo:
@coderabbitai review please

VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/DefiCircleRow.swift
```diff
@@ -108,25 +107,27 @@ struct DefiCircleRow: View {
...
-                circleBalance = usdcBalance
-                isLoading = false
-                hasError = false
-            }
+            let (usdcBalance, _) = try await logic.refresh(vault: vault)
+            circleBalance = usdcBalance
+            isLoading = false
+            hasError = false
         } catch {
             print("DefiCircleRow: Error loading balance: \(error.localizedDescription)")
```
johnnyluo: **Logging violation**: `print()` is prohibited by the project logging rules — all log output must use `OSLog`. Replace with:
```swift
logger.error("Error loading balance: \(error.localizedDescription)")
```
Add a file-level logger above the struct:
```swift
private let logger = Logger(subsystem: "com.vultisig.app", category: "defi-circle-row")
```

VultisigApp/VultisigApp/Components/Tooltip/InfoTooltip.swift
```diff
@@ -0,0 +1,74 @@
...
+            Text(description)
+                .font(Theme.fonts.footnote)
+                .foregroundStyle(Theme.colors.textTertiary)
+                .fixedSize(horizontal: false, vertical: true)
+        }
+        .padding(.leading, 16)
+        .padding(.trailing, 10)
+        .padding(.top, topPadding)
+        .padding(.bottom, bottomPadding)
+        .background(Color(hex: "F5F5F5"))
```
johnnyluo: **Design system violation**: Hardcoded `Color(hex: "F5F5F5")` breaks the design system contract. Replace with a semantic token from `Theme.colors.*`, e.g.:
```swift
.background(Theme.colors.bgSurface1)
```
If the exact shade is new, add it to the design system rather than inlining it.

VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleSetupView.swift
```diff
@@ -74,53 +110,211 @@ struct CircleSetupView: View {
...
+                    arrowDirection: .down,
+                    onDismiss: {
+                        withAnimation(.spring(response: 0.3, dampingFraction: 0.8)) {
+                            showRewardsTooltip = false
+                        }
+                    }
+                )
+                .fixedSize(horizontal: true, vertical: true)
+                .transition(.opacity.combined(with: .scale(scale: 0.95, anchor: .bottom)))
+                .offset(y: -90)
```
johnnyluo: **Magic number**: `.offset(y: -90)` is fragile — it will break if the tooltip's height changes. Prefer positioning via `alignmentGuide` or `GeometryReader` so the offset is derived from actual size. At minimum extract it to a named constant:
```swift
private enum Layout {
    static let rewardsTooltipOffsetY: CGFloat = -90
}
```

VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleDashboardView.swift
```diff
@@ -2,232 +2,7 @@
 //  CircleDashboardView.swift
 //  VultisigApp
 //
-//  Created by Enrique Souza on 2025-12-11.
+//  Merged into CircleSetupView.swift
```
johnnyluo: **Dead file**: This file now contains only a comment and an unused `import SwiftUI`. It should be deleted from the project rather than left as an empty shell — the `project.pbxproj` reference will need removing too (use the `/add-xcode-files` skill in reverse). An empty file with a forwarding comment is noisy and will confuse future contributors.

VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
```diff
@@ -211,6 +213,7 @@
 "circleSetupOpenAccount" = "Open Account";
 "circleSetupUSDCDeposited" = "USDC deposited";
 "circleTitle" = "Circle";
+"circleUSDCDeposited" = "USDC deposited";
```
johnnyluo, line 214: **Duplicate localization key**: `circleUSDCDeposited` (line 216) and `circleSetupUSDCDeposited` (line 214) both map to `"USDC deposited"`. Decide which key is canonical, remove the other, and update all 7 locale files. Having two keys with identical values creates a maintenance hazard — they will inevitably diverge in future translations.

VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleViewLogic.swift
```diff
@@ -37,6 +37,21 @@ struct CircleViewLogic {
...
     }
 
+    /// Fetches fresh balances from chain and upserts them into the vault's cached `CirclePosition`.
+    @MainActor
+    func refresh(vault: Vault) async throws -> (usdcBalance: Decimal, ethBalance: Decimal) {
+        guard let mscaAddress = vault.circleWalletAddress else {
+            return (.zero, .zero)
+        }
+        let (usdcBalance, ethBalance) = try await fetchData(address: mscaAddress, vault: vault)
+        try CirclePositionStorageService().upsert(
```
johnnyluo, line 50: **Ad-hoc service instantiation**: `CirclePositionStorageService()` is constructed inline here but is used in multiple call sites. This creates hidden allocations and makes it impossible to inject a mock in tests. Inject the service as a dependency or use a shared singleton — consistent with how `BalanceService.shared` and `RateProvider.shared` are used elsewhere.

johnnyluo:
@gastonm5 could you please address the comments

VultisigApp/VultisigApp/Features/Defi/Protocols/Circle/CircleViewLogic.swift
gastonm5: `CirclePositionStorageService` is a stateless struct with no stored properties, used at exactly one call site. Promoting it to a singleton or injecting it adds ceremony without benefit here — no allocation cost beyond the struct value, no hidden state, and no test mockability gain without first extracting a protocol (which is a bigger refactor than the comment scope). Leaving as-is; happy to revisit if a second call site appears.

gastonm5:
> @gastonm5 could you please address the comments

Already addressed them! @johnnyluo

VultisigApp/VultisigApp/Core/Localizables/ko.lproj/Localizable.strings
```diff
@@ -209,8 +209,8 @@
 "circleSetupDeposited" = "입금됨";
 "circleSetupInfoText" = "자금은 볼트의 완전한 통제 하에 있습니다. Circle 계정은 자체 보관이며 수익은 안전한 오프체인 금고를 통해 생성됩니다.";
 "circleSetupOpenAccount" = "계정 개설";
-"circleSetupUSDCDeposited" = "USDC 입금됨";
 "circleTitle" = "Circle";
+"circleUSDCDeposited" = "USDC 입금됨";
```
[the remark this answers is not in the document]
johnnyluo: Fixed. Added `circleRewardsDescription` and `circleRewardsTitle` to `ko.lproj/Localizable.strings` in the correct alphabetical position (before `circleRow*` keys).

johnnyluo (approved):
LGTM

PR #4159, opened 2026-04-16
fix(send): refresh balance after successful broadcast

### LINKED ISSUE #4078: [Fix] Balance not refreshed after successful send — user sees stale balance [Send]
# [Fix] Balance not refreshed after send [Send]

## Problem
After successful broadcast, balance cache is not invalidated. Blockchair UTXO cache has 5-minute TTL. User sees pre-send balance until cache expires or manual refresh.

## Expected Behavior
Balance refreshes within ~5 seconds after returning to vault screen post-send.

## Steps to Reproduce
1. Send BTC (or any chain)
2. See "Transaction Complete" screen
3. Go back to vault
4. Balance still shows pre-send amount
5. Wait 1-5 minutes or pull to refresh — balance updates

## Solution
After successful broadcast, invalidate balance cache for the sender's address and trigger re-fetch. Small delay (2-3s) acceptable to let the node register the pending tx.

### Must Do
- [ ] Invalidate balance cache after successful broadcast in keysign completion flow
- [ ] Trigger balance re-fetch for affected coin/address

### Must NOT Do
- Don't add polling — single refresh after broadcast is sufficient
- Don't change broadcast logic

## Acceptance Criteria
- [ ] Build succeeds
- [ ] After successful send, balance updates within ~5 seconds on vault screen
- [ ] No double-refresh or unnecessary API calls

0x-suzume:
## Summary
- Trigger a delayed (3s) balance refresh after any transaction completion (send, swap, join keysign, referral)
- Observes `appViewModel.restartNavigation` — the universal exit point all completion screens use via `appViewModel.restart()`
- Allows the blockchain node time to register the pending tx before re-fetching balances

## Test Plan
- [ ] Send any token, tap "Done" → balance updates within ~5 seconds
- [ ] Complete a swap, tap "Done" → balance updates within ~5 seconds
- [ ] Join keysign flow → balance updates after completion
- [ ] No double-refresh or unnecessary API calls
- [ ] SwiftLint passes

## Related
- Closes #4078

## Summary by CodeRabbit

* **New Features**
  * Background polling for pending transactions and automatic UI refresh when transactions complete or are detected.
* **Updates**
  * Improved app navigation restart behavior with enhanced refresh timing to ensure smoother transitions when returning to the initial state.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/TransactionStatus/TransactionStatusPoller.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TransactionStatus/TransactionStatusPoller.swift
@@ -3,16 +3,21 @@
 //  VultisigApp
 //
 
+import Combine
 import Foundation
+import OSLog
 
 @MainActor
-final class TransactionStatusPoller {
+final class TransactionStatusPoller: ObservableObject {
     static let shared = TransactionStatusPoller()
+
+    @Published private(set) var completedTransactionCount: Int = 0
 
     private let service = TransactionStatusService.shared
     private let recorder = TransactionHistoryRecorder.shared
     private var activeTasks: [String: Task<Void, Never>] = [:]
     private var taskTokens: [String: UUID] = [:]
+    private let logger = Logger(subsystem: "com.vultisig.app", category: "tx-status-poller")
 
     private init() {}
 
@@ -43,6 +48,7 @@ final class TransactionStatusPoller {
                             errorMessage: timeoutMessage
                         )
                         onUpdate(.error, timeoutMessage)
+                        self?.completedTransactionCount += 1
                         break
                     }
 
@@ -63,6 +69,7 @@ final class TransactionStatusPoller {
                             errorMessage: errorMessage
                         )
                         onUpdate(historyStatus, errorMessage)
+                        self?.completedTransactionCount += 1
                         break
                     }
 
@@ -95,6 +102,18 @@ final class TransactionStatusPoller {
         guard taskTokens[txHash] == token else { return }
         activeTasks.removeValue(forKey: txHash)
         taskTokens.removeValue(forKey: txHash)
+    }
+
+    /// Start polling all pending transactions for a vault.
+    func pollPendingTransactions(pubKeyECDSA: String) {
+        do {
+            let pending = try StoredPendingTransactionStorage.shared.getAllPending()
+            for tx in pending where tx.pubKeyECDSA == pubKeyECDSA {
+                poll(txHash: tx.txHash, chain: tx.chain, createdAt: tx.createdAt, pubKeyECDSA: pubKeyECDSA) { _, _ in }
+            }
+        } catch {
+            logger.error("Failed to fetch pending transactions: \(error)")
+        }
     }
 
     /// Returns a terminal TransactionHistoryStatus if the result is terminal, nil if still pending.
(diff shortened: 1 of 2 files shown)


johnnyluo (commented):
Overall approach is solid — polling pending txs on nav restart and vault selection makes sense. A few issues to address before merge.

VultisigApp/VultisigApp/Features/Home/HomeScreen.swift
```diff
@@ -82,6 +83,21 @@ struct HomeScreen: View {
                 presetValuesForDeeplink()
             }
         }
+        .onChange(of: appViewModel.restartNavigation) { _, newValue in
+            guard newValue else { return }
+            if let vault = appViewModel.selectedVault {
+                transactionPoller.pollPendingTransactions(pubKeyECDSA: vault.pubKeyECDSA)
+            }
+            Task {
+                try? await Task.sleep(for: .seconds(3))
```
johnnyluo, line 92: **Architecture violation — business logic in View**

Spawning a raw `Task` with a delay here violates MVVM. This timer/trigger logic should live in a ViewModel method, not the view. The task is also fire-and-forget with no cancellation if the view disappears.

Suggested approach — expose a method on `HomeViewModel` (or a small coordinator) that does the delay + publishes a \"should refresh\" signal, and have the view observe that instead.

```swift
// In HomeViewModel:
func schedulePostTransactionRefresh() {
    refreshTask?.cancel()
    refreshTask = Task { @MainActor in
        try? await Task.sleep(for: .seconds(3))
        guard !Task.isCancelled else { return }
        shouldRefresh = true
    }
}
```

VultisigApp/VultisigApp/Features/Home/HomeScreen.swift
```diff
@@ -82,6 +83,21 @@ struct HomeScreen: View {
...
+                await MainActor.run {
+                    shouldRefresh = true
+                }
```
johnnyluo, line 95: **`await MainActor.run` is redundant**

`HomeScreen` is a SwiftUI view — its `onChange` closure already executes on the main actor. The `await MainActor.run { ... }` wrapper does nothing here and can be removed:

```swift
Task {
    try? await Task.sleep(for: .seconds(3))
    shouldRefresh = true
}
```

VultisigApp/VultisigApp/Core/Services/TransactionStatus/TransactionStatusPoller.swift
```diff
@@ -3,16 +3,21 @@
...
 import Foundation
+import OSLog
 
 @MainActor
-final class TransactionStatusPoller {
+final class TransactionStatusPoller: ObservableObject {
     static let shared = TransactionStatusPoller()
 
+    @Published private(set) var completedTransactionCount: Int = 0
+
```
johnnyluo, line 15: **Increment only on confirmed, not on error/timeout**

Incrementing `completedTransactionCount` on timeout and error triggers a balance refresh even when the tx never confirmed — causing unnecessary API calls with no balance change. Move the increment to only the confirmed branch:

```swift
// Only here (confirmed/failed terminal state), not in the timeout block
self?.completedTransactionCount += 1
```

Consider renaming to `confirmedTransactionCount` to make the intent clearer, or only incrementing when `historyStatus == .successful`.

johnnyluo (approved):
LGTM

PR #4160, opened 2026-04-16
fix(data): prevent duplicate coins from SwiftData inverse relationship sync

### LINKED ISSUE #4090: Duplicate coins in vault.coins caused by SwiftData inverse relationship sync
## Context

PR #4089 fixes the immediate crash loop (replacing `Dictionary(uniqueKeysWithValues:)` with `uniquingKeysWith:` in `BalanceService`). This issue tracks the **underlying data bug** — how duplicate `Coin` entries get into `vault.coins` in the first place.

### The crash
User reports crash loop after switching wallets. Both crash logs show:
```
_assertionFailure → _NativeDictionary.merge → Dictionary.init(uniqueKeysWithValues:)
  → BalanceService.applyBalanceUpdates → updateBalances → VaultDetailViewModel
```

`vault.coins` contains multiple `Coin` objects with the same `.id`, which causes `Dictionary(uniqueKeysWithValues:)` to fatally crash.

### How duplicates likely get created

**`CoinService.addToChain`** (`CoinService.swift:141-173`) does:

```swift
// 1. Check if coin exists (line 154)
if vault.coins.contains(where: { $0.id == newCoin.id }) {
    return vault.coins.first(where: { $0.id == newCoin.id })
}

// 2. Insert into SwiftData and save (line 169-170)
Storage.shared.insert([newCoin])
try Storage.shared.save()

// 3. Append to vault.coins (line 171)
vault.coins.append(newCoin)
```

`Coin` declares an inverse relationship:
```swift
@Relationship(inverse: \Vault.coins) var vault: Vault?
```

After step 2 (`insert` + `save`), **SwiftData may auto-populate `vault.coins` through the inverse relationship sync**. Step 3 then appends the coin a second time — creating a duplicate.

This is a known SwiftData behavior: when a model with an inverse relationship is inserted and saved, the framework can automatically update the parent's relationship array.

### Additional risk: `VaultDefaultCoinService.setDefaultCoins`

`VaultDefaultCoinService` (`VaultDefaultCoinService.swift:50-58`) has the same pattern:
```swift
self.context.insert(coin)
vault.coins.append(coin)
```

This class is **not `@MainActor`** but mutates `vault.coins` (a SwiftData model property), which is also a potential data race with MainActor-isolated code in `CoinService`.

### Evidence of prior duplicates

`THORChainDuplicateTokensMigration.swift` already exists — the team has encountered duplicate coins before and wrote a migration to clean specific cases. The general-purpose fix was never applied.

## Proposed Fix

In `CoinService.addToChain`, add a guard after `save()` and before `append`:

```swift
Storage.shared.insert([newCoin])
try Storage.shared.save()

// Re-check: SwiftData may have auto-populated vault.coins via inverse relationship
guard !vault.coins.contains(where: { $0.id == newCoin.id }) else {
    return vault.coins.first(where: { $0.id == newCoin.id })
}

vault.coins.append(newCoin)
```

Same pattern should be applied in `VaultDefaultCoinService.setDefaultCoins` (line 52-53).

### Call sites that trigger `addDiscoveredTokens` (for reference)
- `VaultDefaultCoinService.setDefaultCoins` — vault creation/restore (fire-and-forget `Task {}`)
- `ChainDetailScreen.updateBalances` — user views chain detail
- `CoinService.addToChain` — when adding a native token
- `AgentToolExecutor` — agent adding coins

## Related
- #4089 — crash fix (defensive `uniquingKeysWith`)

0x-suzume:
## Summary
- After `insert()` + `save()`, SwiftData may auto-populate `vault.coins` via the inverse relationship on `Coin`
- The subsequent `append()` then creates a duplicate entry
- Added guard checks in `CoinService.addToChain` and `VaultDefaultCoinService.setDefaultCoins` to skip the append if the coin is already present

## Test Plan
- [ ] Add a new chain/token to vault — no duplicate coins appear
- [ ] Create a new vault — default coins don't duplicate
- [ ] SwiftLint passes

## Related
- Closes #4090
- Related to #4089 (crash fix for duplicate coins)

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/CoinService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/CoinService.swift
@@ -168,6 +168,12 @@ struct CoinService {
         // otherwise it report an error "Illegal attempt to map a relationship containing temporary objects to its identifiers."
         Storage.shared.insert([newCoin])
         try Storage.shared.save()
+
+        // Guard against SwiftData inverse relationship auto-populating vault.coins
+        guard !vault.coins.contains(where: { $0.id == newCoin.id }) else {
+            return vault.coins.first(where: { $0.id == newCoin.id })
+        }
+
         vault.coins.append(newCoin)
         return newCoin
     }
--- a/VultisigApp/VultisigApp/Core/Services/VaultDefaultCoinService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/VaultDefaultCoinService.swift
@@ -50,7 +50,11 @@ class VaultDefaultCoinService {
             for coin in coins {
                 if coin.isNativeToken {
                     self.context.insert(coin)
-                    vault.coins.append(coin)
+
+                    // Only append if SwiftData inverse relationship didn't auto-populate
+                    if !vault.coins.contains(where: { $0.id == coin.id }) {
+                        vault.coins.append(coin)
+                    }
 
                     Task {
                         await CoinService.addDiscoveredTokens(nativeToken: coin, to: vault)


johnnyluo (approved):
LGTM

PR #4191, opened 2026-04-20
fix(keygen): pop to email entry when switching email in fast vault OTP

### LINKED ISSUE #4184: [BUG] Fast vault creation — back navigation from OTP loops back to OTP after changing email
## Summary

Navigation stack is polluted during fast vault creation — after the user changes their email mid-flow, pressing back on the OTP screen pushes OTP again instead of popping out of the flow.

## Repro

1. Start Fast Vault creation
2. Enter email, proceed to OTP screen ("4-digit code received via email")
3. Tap **Use a different email** (or navigate back)
4. Change the email, proceed to OTP again
5. Decide to abort → tap the navigation **back** arrow

## Expected

Back navigation pops the screen, user exits fast vault creation (or returns to previous step in the onboarding stack).

## Actual

User is routed **back to the OTP input screen** instead of exiting. The flow is inescapable without killing the app.

## Hypothesis

The "Use a different email" action pushes a new email-entry screen onto the stack rather than popping the OTP screen. When the user advances again, OTP gets re-pushed on top of the stale OTP, so the back action pops one OTP off but lands on the previous OTP. Fix: either pop OTP when "Use a different email" is tapped, or replace (not push) the OTP screen on re-submit.

## Acceptance criteria

- [ ] Tapping **Use a different email** removes OTP from the stack (or replaces it)
- [ ] Back arrow from OTP always exits the fast vault flow in one tap
- [ ] No duplicate OTP screens in the nav stack
- [ ] Verified on iOS and macOS

## Platform

iOS + macOS (universal surface)

gastonm5:
## Summary

- In fast-vault keygen, the OTP verification sheet's **"Use a different email"** button invoked `onBackToEmailSetup`, which incorrectly **pushed** a new `fastVaultPassword` route onto the navigation stack. This caused a loop where the user kept landing back on the OTP screen instead of the email entry screen.
- Replace the push with `router.navigateBack()` so that PeerDiscovery (which hosts `KeygenView` inline) is popped off the stack, leaving the user on the email entry screen that is already on the stack in both entry paths (`FastVaultPasswordScreen` and `VaultSetupScreen`).
- The OTP sheet itself is already dismissed by `ServerBackupVerificationScreen.deleteVault()` (which sets `isPresented = false` before firing the callback), so no extra sheet state reset is needed.

## Test plan

- [ ] Fast vault flow (new vault) via `VaultSetupScreen`: enter name/email/password, continue through keygen until OTP appears. Tap "Use a different email". Verify the app returns to the email step of VaultSetupScreen (not OTP, not a new blank screen).
- [ ] Fast vault flow (existing vault) via `FastVaultPasswordScreen` (single keygen / reshare path): reach OTP, tap "Use a different email", verify user returns to the email step of FastVaultPasswordScreen.
- [ ] Verify no duplicated routes in the back stack after pressing back from the email screen.

Closes #4184

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced back navigation consistency across key-generation and vault setup flows; back buttons now reliably return to the previous screen in the navigation stack.
  * Updated back button visibility to display consistently during wallet setup process.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Navigation/NavigationRouter.swift
+++ b/VultisigApp/VultisigApp/Core/Navigation/NavigationRouter.swift
@@ -11,6 +11,7 @@ typealias NavPath = Hashable
 
 final class NavigationRouter: ObservableObject {
     @Published var navPath: NavigationPath
+    private var history: [any NavPath] = []
 
     init(navPath: NavigationPath = NavigationPath()) {
         self.navPath = navPath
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4196, opened 2026-04-20
fix(send): retry stale Solana blockhash broadcasts

### LINKED ISSUE #4080: [Fix] Solana transactions fail when MPC signing takes >60s — blockhash expires [Keysign]
# [Fix] Solana blockhash expires during long keysign [Keysign]

## Problem
Solana blockhashes expire ~60 seconds after fetch. The blockhash is embedded in the signed transaction — changing it requires re-signing. If MPC ceremony (device coordination, network latency) takes >60s, broadcast fails with "blockhash expired." User must manually retry.

## Expected Behavior
Either prevent expiry (durable nonces) or detect and auto-retry the signing flow.

## Steps to Reproduce
1. Initiate Solana send in 2-of-3 vault
2. Take >60 seconds coordinating devices
3. Broadcast fails: "blockhash not found"
4. User must restart entire flow manually

## Solution
**Option A (recommended):** On "blockhash expired" broadcast error, auto-retry: fetch new blockhash → rebuild payload → restart keysign. Show user "Re-signing with fresh blockhash..." instead of error.

**Option B (future):** Implement Solana durable transaction nonces — requires nonce account creation per vault. Eliminates expiry entirely but is a larger feature.

Note: Simply refreshing blockhash before broadcast won't work — the blockhash is part of the signed data. Re-signing is required.

### Must Do
- [ ] Detect "blockhash expired/not found" error on Solana broadcast
- [ ] Auto-retry keysign flow with fresh blockhash (max 1 retry)
- [ ] Show user-friendly message during retry

### Must NOT Do
- Don't implement durable nonces (separate feature)
- Don't change MPC protocol

## Acceptance Criteria
- [ ] Build succeeds
- [ ] Solana send succeeds even when first attempt takes >60s
- [ ] Quick sends (<60s) unaffected
- [ ] No infinite retry loop (max 1 retry)

gastonm5:
## Summary

- Adds a generic `RetryableBroadcastError` + `BroadcastRetryReason` abstraction so any chain service can flag broadcast failures as recoverable (stale blockhash, stale nonce, stale quote, other).
- Solana adopts it for expired-blockhash errors when MPC signing takes longer than the blockhash validity window.
- Keysign now surfaces a `KeysignRetryRequested` state with a "Try again with fresh data" CTA. Tapping it pops back to the verify screen, re-fetches fresh chain data, and shows a `PopupCapsule` toast explaining the reason. Retries are capped at 1 per keysign.
- Refactors the keysign error layout: centers content vertically and adds horizontal padding (previously top-aligned with a full-bleed button).

Closes #4080.

## Test plan

- [ ] Send a Solana tx with MPC signing delayed past blockhash validity — verify the retry CTA appears, tapping it returns to verify with a toast, fees refresh, re-sign succeeds.
- [ ] Confirm the retry is capped at 1 (second failure falls through to the normal signing error view).
- [ ] Confirm non-Solana chains are unaffected (no retry path taken).
- [ ] Visually confirm the error view is centered with side padding on both iOS and macOS.
- [ ] Confirm the "Cancel"/back flow from the retry view still works without leaking state (`tx.pendingRetryReason` cleared).

## Summary by CodeRabbit

* **New Features**
  * Automatic retry for failed transaction broadcasts when data is stale (one retry) and a retry UI prompting the user to try again.
  * Retry reasons surfaced to users via contextual messages and an error button that can trigger a retry.

* **Improvements**
  * Better error detection for broadcast failures; clearer user-facing messages and structured logging.

* **Localization**
  * Added retry/error message translations in English, German, Spanish, Croatian, Italian, Korean, Portuguese, and Simplified Chinese.


VultisigApp/VultisigApp/Core/Models/RetryableBroadcastError.swift
```diff
@@ -0,0 +1,29 @@
...
+
+protocol RetryableBroadcastError: Error {
+    var retryReason: BroadcastRetryReason { get }
+}
+
+enum BroadcastRetryReason: Equatable {
+    case staleBlockhash
+    case staleNonce
+    case staleQuote
+    case other(String)
```
johnnyluo, line 11: The associated `String` is captured here but `toastKey` for `.other` always returns `"broadcastRetryGeneric"` — the string is never surfaced. Either drop it (`case other` with no payload) or thread it through to the user-facing message, since right now callers have no way to know their diagnostic context is silently discarded.

VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
```diff
@@ -33,6 +34,10 @@ class KeysignViewModel: ObservableObject {
     @Published var txid: String = .empty
     @Published var approveTxid: String?
     @Published var decodedMemo: String?
+    @Published var retryReason: BroadcastRetryReason?
+
+    private var broadcastRetryCount = 0
+    private static let maxBroadcastRetries = 1
```
johnnyluo, line 40: `broadcastRetryCount` is never reset. If this `KeysignViewModel` instance is reused for a second transaction after one retry was already consumed, the counter is already at `maxBroadcastRetries` and the second transaction silently gets zero retries. Reset this alongside the rest of the signing state.

VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCryptoViewModel.swift
```diff
@@ -160,6 +161,15 @@ class SwapCryptoViewModel: ObservableObject, TransferViewModel {
         currentTitle = titles[currentIndex-1]
     }
 
+    func retryBroadcast(reason: BroadcastRetryReason) {
+        pendingRetryReason = reason
+        keysignPayload = nil
+        hash = nil
+        approveHash = nil
```
johnnyluo, line 168: Magic index `2` — this is a fragile coupling to the position in `titles`. If the swap flow gains a new step this silently navigates to the wrong screen without a compiler error. A named constant or index derived from a step enum would be safer.

johnnyluo (commented):
## Overview

Well-designed fix that introduces a reusable `RetryableBroadcastError` + `BroadcastRetryReason` abstraction for recoverable broadcast failures. Solana adopts it for stale blockhash detection; the UX pops back to verify, refreshes gas/blockhash, and shows a banner. Both Send and Swap flows implement the retry path. The 1-retry cap prevents loops.

### Strengths
- Clean protocol/enum abstraction — `RetryableBroadcastError` is ready for future `staleNonce`/`staleQuote` cases without touching `KeysignViewModel`
- Blockhash detection improved: case-insensitive `lowercased()` + `"block height exceeded"` + error code `-32002` covers more RPC variants than the old case-sensitive single-string check
- OSLog migration for all `print()` calls in `KeysignViewModel` and `SolanaService`
- All 8 locales updated correctly
- `sendCryptoKeysignView` `errorButtonTitle`/`errorAction` extension is a clean way to reuse the error UI without forking it

### Issues

**1. `broadcastRetryCount` never resets between transactions** (`KeysignViewModel`)
`broadcastRetryCount` is an instance variable but isn't reset when a new signing session starts. If a `KeysignViewModel` is reused after consuming its one retry, the counter is already at `maxBroadcastRetries` and the next transaction silently gets zero retries. This counter should be zeroed wherever the rest of the keysign state is reset.

**2. `BroadcastRetryReason.other(String)` associated value is silently dropped** (`RetryableBroadcastError.swift`)
`toastKey` for `.other` ignores its associated `String` entirely — the caller-provided diagnostic string is captured and never shown. Either drop the payload (`case other` with no associated value) or surface it, since currently it misleads callers into thinking they can pass context that matters.

**3. `currentIndex = 2` hardcoded in `SwapCryptoViewModel.retryBroadcast`**
Fragile coupling to a position in the `titles` array. If the swap flow gains a new step, this silently lands on the wrong screen with no compiler warning. Consider a named constant or deriving the index from a step enum.

**4. `SwapVerifyView.onAppear` vs `SendVerifyScreen.onNavigationStackChange` inconsistency**
`SendVerifyScreen` uses `onNavigationStackChange` (fires correctly when popping back into view); `SwapVerifyView` uses `onAppear`. For Swap this works today because `keysignView = nil` dismisses the overlay and triggers `onAppear`, but it also means `consumePendingRetry()` runs on first launch — only guarded by the `guard let reason` nil-check. Worth a comment explaining why `onAppear` is safe here, or unifying to `onNavigationStackChange` if available.

**5. `.scrollBounceBehavior(.basedOnSize)` requires iOS 16.4+** (`ErrorView.swift`)
This API was introduced in iOS 16.4. If the deployment target is below 16.4 this will fail at runtime. Needs `if #available(iOS 16.4, *) { ... }` or a version availability guard.

**6. No unit tests for retry-detection logic**
The three-condition blockhash check in `SolanaService` (`"blockhash not found"`, `"block height exceeded"`, `errorCode == -32002`) is now the critical gate for the whole retry path. A unit test asserting each condition maps to `SolanaRetryableError.blockhashExpired` (and non-matching strings fall through to `rpcError`) would prevent a regression here.

### Nits
- `RetryableBroadcastError.swift` lands in `Core/Models/` — it's a protocol+error definition rather than a model; `Core/Errors/` or alongside the keysign types might be a better home (minor)
- `guard reason != nil else { return }` in `SwapCryptoView.onChange` reads a little awkward; `guard swapViewModel.pendingRetryReason != nil else { return }` or pattern-matching would be clearer

johnnyluo (approved):
LGTM

PR #4198, opened 2026-04-20
fix(fastvault): surface registration errors instead of silent failure

### LINKED ISSUE #4140: FastVaultService fire-and-forget methods silently swallow server registration failures
## Problem

All `FastVaultService` methods that register the server for keygen ceremonies use a fire-and-forget pattern via `Utils.sendRequest` with a completion handler that discards errors:

- `create()` — `POST /vault/create`
- `batchCreate()` — `POST /vault/batch/keygen`
- `keyImport()` — `POST /vault/import`
- `reshare()` — `POST /vault/reshare`
- `migrate()` — `POST /vault/migrate`

If any of these HTTP calls fail (network error, 4xx/5xx, timeout), the client proceeds into the keygen ceremony without the server party being registered. The ceremony then times out or fails with no indication of what went wrong.

## Impact

- User sees a spinning keygen screen for 60+ seconds before a generic timeout error
- No way to distinguish "server registration failed" from "TSS protocol failure"
- Particularly risky for `batchCreate` where the server must run parallel goroutines — a silent failure here means the client uses `p-ecdsa`/`p-eddsa` relay namespaces that nobody is listening on

## Expected behavior

Server registration methods should be `async throws` so callers can abort the flow early with a clear error message if the server doesn't accept the registration.

## Affected files

- `VultisigApp/Core/Services/FastVault/FastVaultService.swift` — all 5 methods listed above
- `VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift` — call sites

## Context

Identified during review of PR #4139 (parallel keygen). The pattern is pre-existing across all methods, not specific to `batchCreate`.

gastonm5:
## Summary
- Refactor `FastVaultService` to use the `HTTPClient` + `TargetType` pattern (new `FastVaultAPI.swift`) so all Vultiserver calls go through the project's standard networking layer.
- Convert fire-and-forget registration methods (`create`, `batchCreate`, `keyImport`, `batchKeyImport`, `reshare`, `batchReshare`, `migrate`, `singleKeygen`) to `async throws` and map HTTP failures to a typed `FastVaultServiceError`.
- `KeygenPeerDiscoveryViewModel` now awaits registration inside a single `do/catch`; on failure it logs via `OSLog`, sets `status = .Failure`, and surfaces `fastVaultRegistrationFailed` (localized in all 8 locales) instead of letting users wait for a generic ceremony timeout.

Closes #4140

> [!NOTE]
> Stacked on top of #4139 (`roman/parallel-keygen-execution`). Rebase onto `main` after that PR merges.

## Test plan
- [ ] Keygen with valid Vultiserver credentials completes successfully (single + batch paths).
- [ ] Simulate a Vultiserver 5xx/4xx response and confirm the UI shows the localized failure message immediately rather than hanging until timeout.
- [ ] `Reshare` / `Migrate` / `SingleKeygen` flows all surface the same error UX on server rejection.
- [ ] `validateAccess` still differentiates `.valid`, `.invalidPassword`, and `.vaultNotFound` (custom status codes preserved).

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultAPI.swift
+++ b/VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultAPI.swift
@@ -1,0 +1,113 @@
+//
+//  FastVaultAPI.swift
+//  VultisigApp
+//
+
+import Foundation
+
+enum FastVaultAPI: TargetType {
+    case validateAccess(pubKeyECDSA: String, base64Password: String)
+    case exists(pubKeyECDSA: String)
+    case create(VaultCreateRequest)
+    case batchCreate(BatchKeygenRequest)
+    case keyImport(KeyImportRequest)
+    case batchKeyImport(BatchKeyImportRequest)
+    case reshare(ReshareRequest)
+    case batchReshare(BatchReshareRequest)
+    case sign(KeysignRequest)
+    case migrate(MigrationRequest)
+    case singleKeygen(CreateMldsaRequest)
+    case verifyBackupOTP(pubKeyECDSA: String, code: String)
+
+    var baseURL: URL {
+        guard let url = URL(string: Endpoint.vultisigApiProxy) else {
+            fatalError("Invalid FastVault base URL")
+        }
+        return url
+    }
+
+    var path: String {
+        switch self {
+        case .validateAccess(let pubKey, _):
+            return "/vault/get/\(pubKey)"
+        case .exists(let pubKey):
+            return "/vault/exist/\(pubKey)"
+        case .create:
+            return "/vault/create"
+        case .batchCreate:
+            return "/vault/batch/keygen"
+        case .keyImport:
+            return "/vault/import"
+        case .batchKeyImport:
+            return "/vault/batch/import"
+        case .reshare:
+            return "/vault/reshare"
+        case .batchReshare:
+            return "/vault/batch/reshare"
+        case .sign:
+            return "/vault/sign"
+        case .migrate:
+            return "/vault/migrate"
+        case .singleKeygen:
+            return "/vault/mldsa"
+        case .verifyBackupOTP(let pubKey, let code):
+            return "/vault/verify/\(pubKey)/\(code)"
+        }
+    }
+
+    var method: HTTPMethod {
+        switch self {
+        case .validateAccess, .exists, .verifyBackupOTP:
+            return .get
+        case .create, .batchCreate, .keyImport, .batchKeyImport,
+             .reshare, .batchReshare, .sign, .migrate, .singleKeygen:
+            return .post
+        }
+    }
+
+    var task: HTTPTask {
+        switch self {
+        case .validateAccess, .exists, .verifyBackupOTP:
+            return .requestPlain
+        case .create(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .batchCreate(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .keyImport(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .batchKeyImport(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .reshare(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .batchReshare(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .sign(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .migrate(let req):
+            return .requestCodable(req, .jsonEncoding)
+        case .singleKeygen(let req):
+            return .requestCodable(req, .jsonEncoding)
+        }
+    }
+
+    var headers: [String: String]? {
+        switch self {
+        case .validateAccess(_, let base64Password):
+            return [
+                "Content-Type": "application/json",
+                "x-password": base64Password
+            ]
+        default:
+            return ["Content-Type": "application/json"]
+        }
+    }
+
+    var validationType: ValidationType {
+        switch self {
+        case .validateAccess:
+            return .customCodes([200, 401, 403, 404])
+        default:
+            return .successCodes
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultAPI.swift
```diff
@@ -0,0 +1,113 @@
...
+    case reshare(ReshareRequest)
+    case batchReshare(BatchReshareRequest)
+    case sign(KeysignRequest)
+    case migrate(MigrationRequest)
+    case singleKeygen(CreateMldsaRequest)
+    case verifyBackupOTP(pubKeyECDSA: String, code: String)
+
+    var baseURL: URL {
+        guard let url = URL(string: Endpoint.vultisigApiProxy) else {
+            fatalError("Invalid FastVault base URL")
```
johnnyluo, line 24: `fatalError` here is fine since `Endpoint.vultisigApiProxy` is a compile-time constant, but consider `preconditionFailure` instead — it is more idiomatic for invariant violations. More importantly, any unit test that exercises `FastVaultAPI` will crash rather than fail gracefully if this constant is ever malformed.

VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultService.swift
```diff
@@ -384,6 +378,8 @@ enum FastVaultServiceError: Error, LocalizedError {
     case invalidSignURL
     case invalidResponse
```
johnnyluo: `invalidSignURL` and `invalidResponse` are now dead code. `invalidSignURL` was only thrown when `URL(string: endpoint + "/sign")` failed — that construction now lives inside `FastVaultAPI` and is validated at startup. `invalidResponse` is misused in the ViewModel (see comment there) and is no longer thrown from `sign`. Consider removing both cases so they don't accumulate as noise.

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
```diff
@@ -113,106 +113,118 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
...
-                fastVaultService.migrate(
-                    publicKeyECDSA: vault.pubKeyECDSA,
+            }
+        }
+        self.isLoading = false
+    }
+
+    private func registerFastVaultServer(tssType: TssType, config: FastSignConfig) async throws {
+        guard let encryptionKeyHex else {
+            throw FastVaultServiceError.invalidResponse
```
johnnyluo: Throwing `FastVaultServiceError.invalidResponse` here is a semantic mismatch — its `errorDescription` reads as `"FastVault sign returned an invalid response"`, which is confusing when the real cause is a nil encryption key during keygen. Consider a dedicated `FastVaultServiceError.missingEncryptionKey` case, or use `registrationFailed(operation: "keygen", statusCode: 0, responseBody: "missing encryption key")` as a stopgap.

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
```diff
@@ -113,106 +113,118 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
...
-                            sessionID: sessionID,
-                            hexEncryptionKey: encryptionKeyHex!,
-                            hexChainCode: vault.hexChainCode,
-                            encryptionPassword: config.password,
-                            email: config.email,
-                            oldParties: vault.signers,
-                            oldResharePrefix: vault.resharePrefix ?? "",
-                            lib_type: vault.libType == .DKLS ? 1 : 0
-                        )
+            Task { [weak self] in
```
johnnyluo, line 133: This `Task` accesses `vault.name`, `vault.pubKeyECDSA`, `vault.hexChainCode`, `vault.signers`, `vault.resharePrefix`, and `vault.libType` from inside an unstructured `Task` with no actor annotation. `Vault` is a SwiftData `@Model` that must only be read on `@MainActor`, and `KeygenPeerDiscoveryViewModel` is not `@MainActor`-annotated. Under Swift 6 strict concurrency this is a data race. The safe fix is to capture all needed value types before the Task starts:

```swift
let vaultName = vault.name
let vaultPubKey = vault.pubKeyECDSA
// ...
Task { [weak self] in
    try await self?.registerFastVaultServer(..., vaultName: vaultName, ...)
}
```

This was a pre-existing pattern in the old code, but `registerFastVaultServer` now centralises the access, making it easier to audit and fix here.

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
```diff
@@ -113,106 +113,118 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
...
+                        self.errorMessage = "fastVaultRegistrationFailed".localized
+                        self.status = .Failure
                     }
                 }
-            case .Migrate:
-                fastVaultService.migrate(
-                    publicKeyECDSA: vault.pubKeyECDSA,
+            }
+        }
+        self.isLoading = false
```
johnnyluo, line 147: `isLoading` is set to `false` immediately after the registration `Task` is spawned, before registration completes. This preserves the original behaviour (the old fire-and-forget Tasks also didn't block `isLoading`), but it looks like a bug on first read. A brief comment would help future readers:

```swift
// isLoading reflects peer-discovery state, not registration — registration failure is surfaced via status = .Failure
self.isLoading = false
```

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
johnnyluo: or we can mark `registerFastVaultServer` with `@MainActor`

johnnyluo:
Also , are we able to merge this one to `main` branch?

gastonm5:
> Also , are we able to merge this one to `main` branch?

We would need to rebase it to main, bc it contains all the parallel keygen changes. We could merge parallel keygen and keep the flags false to include it. @johnnyluo

PR #4202, opened 2026-04-21
fix(keygen): resolve tss-batch feature flag once per flow

### LINKED ISSUE #4141: Feature flag fetched independently in PeerDiscovery and KeygenViewModel — potential mode split
## Problem

The `tss-batch` feature flag is fetched twice during a single keygen flow from two separate `FeatureFlagService()` instances:

1. **KeygenPeerDiscoveryViewModel.setData()** — decides whether to call `batchCreate()` or `create()` on the server
2. **KeygenViewModel.startKeygenDKLS()** — decides whether to run ECDSA+EdDSA in parallel or sequentially

Each fetch is an independent HTTP call to `api.vultisig.com/feature/release.json`. If one succeeds and the other fails (or the flag changes between the two calls), the client ends up in a split state:

- Server registered with batch mode, but client runs sequential keygen (or vice versa)
- Relay message routing uses `p-ecdsa`/`p-eddsa` while server expects default namespace (or vice versa)

The current fallback (`false` on failure) makes both default to sequential, which is consistent — but a transient network glitch where only one call succeeds could still produce a mismatch.

## Suggested fix

Resolve the flag once at the start of the flow and pass the resolved boolean through the navigation route or as a shared property on the ViewModel, so both the server registration and the keygen execution see the same value.

## Affected files

- `VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift`
- `VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift`

## Context

Same pattern exists on Android (PR #3980). Identified during review of PR #4139.

gastonm5:
## Summary

Fixes #4141.

The `tss-batch` feature flag was being fetched independently by `KeygenPeerDiscoveryViewModel` / `JoinKeygenViewModel` (which decides whether to hit the batch or non-batch FastVault registration endpoint) and by `KeygenViewModel` (which decides whether to run ECDSA/EdDSA in parallel or sequentially). The two fetches could diverge and leave the client in a split state — registered with one endpoint variant but executing the other.

- Resolve the flag once on `KeygenPeerDiscoveryViewModel` / `JoinKeygenViewModel` and expose it as `@Published var isTssBatchEnabled`
- Thread the resolved boolean through `KeygenView` into `KeygenViewModel.setData(...)`
- `KeygenViewModel.startKeygenDKLS` and `startKeyImportKeygen` now read `self.isTssBatchEnabled` instead of fetching again
- Peer-discovery's consolidated task also drives the `batchCreate` / `batchKeyImport` / `batchReshare` vs non-batch FastVault calls from the same resolved value

## Test plan

- [ ] Keygen (fast vault) — verify server registration and TSS execution agree
- [ ] Reshare (fast vault) — same
- [ ] KeyImport (fast vault) — same
- [ ] Regular (non-fast) Keygen still works
- [ ] Joining device (`JoinKeygenView`) picks up the resolved flag

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/JoinKeygenViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/JoinKeygenViewModel.swift
@@ -53,6 +53,7 @@ class JoinKeygenViewModel: ObservableObject {
     @Published var error: JoinKeygenError? = nil
     @Published var serverAddress: String? = nil
     @Published var oldResharePrefix: String = ""
+    @Published var isTssBatchEnabled: Bool = false
 
     var encryptionKeyHex: String = ""
     var singleKeygenType: SingleKeygenType = .MLDSA
@@ -84,6 +85,10 @@ class JoinKeygenViewModel: ObservableObject {
 
         if let isAllowed = self.isCameraPermissionGranted, !isAllowed {
             status = .NoCameraAccess
+        }
+
+        Task { @MainActor in
+            self.isTssBatchEnabled = await FeatureFlagService().isFeatureEnabled(feature: .TssBatch)
         }
     }
 
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
@@ -40,6 +40,7 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
         }
     }
     @Published var isLoading: Bool = false
+    @Published var isTssBatchEnabled: Bool = false
 
     private var peersFoundCancellable: AnyCancellable?
     private let mediator = Mediator.shared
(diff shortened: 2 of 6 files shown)


gastonm5:
@coderabbitai review

VultisigApp/VultisigApp/Features/Keygen/ViewModels/JoinKeygenViewModel.swift
```diff
@@ -85,6 +86,10 @@ class JoinKeygenViewModel: ObservableObject {
         if let isAllowed = self.isCameraPermissionGranted, !isAllowed {
             status = .NoCameraAccess
         }
+
+        Task { @MainActor in
+            self.isTssBatchEnabled = await FeatureFlagService().isFeatureEnabled(feature: .TssBatch)
```
johnnyluo, line 91: @gastonm5 I think we should add this flag into the QR code

gastonm5: I've created the PRs for it. We can close this if we merge the other one

PR #4206, opened 2026-04-22
feat(keygen): carry is_tss_batch on message instead of feature flag

gastonm5:
## Summary

Stacks on top of #4139. Moves the parallel-keygen decision off the remote \`tss-batch\` feature flag / local toggle and onto the KeygenMessage / ReshareMessage proto payload that's already encoded in the QR code.

**Why:** every device (initiator + joiners) currently resolves \`tss-batch\` independently via \`FeatureFlagService\`. If the flag flips between lookups — or the local Advanced Settings toggle is set on only one device — peers desync: initiator uploads parallel setup messages to isolated relay namespaces, joiners wait on the sequential channels, ceremony hangs. The fix is to put the bool on the wire: initiator picks once, stamps it into the proto, joiners read the same value.

**Initiator value:** \`vault.libType == .DKLS\` — DKLS is the only lib where parallel ever applied. GG20 stays sequential. This also removes the need for an ops-controlled rollout.

## Depends on

- vultisig/commondata#81 — adds \`is_tss_batch\` to \`KeygenMessage\` (field 9) and \`ReshareMessage\` (field 10). Must merge first. Once merged, the \`Package.resolved\` pin will be bumped in a follow-up commit on this branch before merging.

### Swift model + proto mapping
- \`KeygenMessage\` and \`ReshareMessage\` gain \`isTssBatch: Bool\`
- Both \`+ProtoMappable\` extensions read/write the new field

### Initiator (\`KeygenPeerDiscoveryViewModel\`)
- New computed \`var isTssBatch: Bool { vault.libType == .DKLS }\`
- \`generateQRdata()\` stamps \`isTssBatch\` into KeygenMessage / ReshareMessage
- The three \`FastVaultService\` dispatch sites drop their \`Task { await FeatureFlagService() ... }\` and use \`self.isTssBatch\` directly
- Passes \`isTssBatch\` to \`KeygenView\` via \`PeerDiscoveryScreen\`

### Joiner (\`JoinKeygenViewModel\`)
- New \`@Published var isTssBatch: Bool\`
- \`handleQrCodeSuccessResult\` populates it from \`keygenMsg.isTssBatch\` / \`reshareMsg.isTssBatch\`
- Passes it to \`KeygenView\` via \`JoinKeygenView\`

### \`KeygenView\` + \`KeygenViewModel\`
- \`KeygenView\` takes a new \`let isTssBatch: Bool\` and forwards into \`viewModel.setData(...)\`
- \`KeygenViewModel\` stores it and uses \`self.isTssBatch\` in \`startKeygenDKLS\` and \`startKeyImportKeygen\` instead of looking up \`FeatureFlagService\`

### Removals
- \`FeatureFlag.TssBatch\` case deleted (and the \`localOverrideKeys\` machinery it motivated)
- \`@AppStorage("tssBatchEnabled")\` removed from \`SettingsViewModel\`
- "TSS Batching" toggle block removed from \`SettingsAdvancedView\`

## Test plan

After bumping the Package.resolved pin to the commondata merge commit:

- [ ] \`xcodebuild build -scheme VultisigApp\` succeeds
- [ ] Two-device DKLS FastVault keygen: both devices log \`execution=parallel\`, ceremony completes
- [ ] Two-device DKLS reshare + migrate: both devices run parallel path
- [ ] Key Import (DKLS): both devices run parallel path
- [ ] GG20 regression: \`is_tss_batch\` is \`false\` on wire, both sides log \`execution=sequential\`
- [ ] Advanced Settings → the "TSS Batching" toggle is gone
- [ ] Coordinate app release with Android (#3980) and Windows — older joiners read zero-value \`false\` and will desync from a parallel initiator

## Summary by CodeRabbit

* **New Features**
  * Added configurable TSS batch mode across key-generation flows and QR/deeplink payloads.
  * Exposed TSS batch option to keygen UI paths so joins and keygen screens can opt into batch behavior.
  * Added new persistent settings for agent and transaction history.

* **Refactor**
  * Migrated TSS batch control from a remote feature flag to a local persisted setting.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
@@ -438,7 +438,8 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
                     useVultisigRelay: VultisigRelay.IsRelayEnabled,
                     vaultName: vault.name,
                     libType: vault.libType ?? .GG20,
-                    chains: chains ?? []
+                    chains: chains ?? [],
+                    isTssBatch: isTssBatch
                 )
                 let data = try ProtoSerializer.serialize(keygenMsg)
                 return "https://vultisig.com?type=NewVault&tssType=\(tssType.rawValue)&jsonData=\(data)"
@@ -453,7 +454,8 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
                     useVultisigRelay: VultisigRelay.IsRelayEnabled,
                     oldResharePrefix: vault.resharePrefix ?? "",
                     vaultName: vault.name,
-                    libType: vault.libType ?? .GG20
+                    libType: vault.libType ?? .GG20,
+                    isTssBatch: isTssBatch
                 )
                 let data = try ProtoSerializer.serialize(reshareMsg)
                 return "https://vultisig.com?type=NewVault&tssType=\(tssType.rawValue)&jsonData=\(data)"
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
@@ -34,6 +34,13 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
     var encryptionKeyHex: String?
     var chains: [Chain]?
 
+    @AppStorage("tssBatchEnabled") private var tssBatchEnabled: Bool = false
+
+    var isTssBatch: Bool {
+        vault.libType == .DKLS && tssBatchEnabled
+    }
+
+
     @Published var status = PeerDiscoveryStatus.WaitingForDevices
     @Published var serviceName = ""
     @Published var errorMessage = ""
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
@@ -37,7 +37,8 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
     @AppStorage("tssBatchEnabled") private var tssBatchEnabled: Bool = false
 
     var isTssBatch: Bool {
-        vault.libType == .DKLS && tssBatchEnabled
+        let supportsBatch = vault.libType == .DKLS || vault.libType == .KeyImport
+        return supportsBatch && tssBatchEnabled
     }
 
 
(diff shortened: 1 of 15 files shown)


VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
```diff
@@ -34,6 +34,14 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
     var encryptionKeyHex: String?
     var chains: [Chain]?
 
+    @AppStorage("tssBatchEnabled") private var tssBatchEnabled: Bool = false
+
+    var isTssBatch: Bool {
+        let supportsBatch = vault.libType == .DKLS || vault.libType == .KeyImport
+        return supportsBatch && tssBatchEnabled
+    }
```
johnnyluo, line 42: **Bug — critical:** `isTssBatch` is computed here but never passed to `KeygenMessage` or `ReshareMessage` in `generateQRdata()` (lines 433 and 446). Both structs now declare `let isTssBatch: Bool` as a required field, so this is a compile error. Joiners reading from the QR code would always see the proto zero-value `false` and never use the parallel path — the desync problem this PR is fixing.

```swift
// generateQRdata() — KeygenMessage init
let keygenMsg = KeygenMessage(
    ...
    chains: chains ?? [],
    isTssBatch: isTssBatch   // ← missing
)

// generateQRdata() — ReshareMessage init
let reshareMsg = ReshareMessage(
    ...
    libType: vault.libType ?? .GG20,
    isTssBatch: isTssBatch   // ← missing
)
```

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
```diff
@@ -155,12 +163,13 @@ class KeygenPeerDiscoveryViewModel: ObservableObject {
         guard let encryptionKeyHex else {
             throw FastVaultServiceError.missingEncryptionKey
         }
-        let isTssBatchEnabled = await FeatureFlagService().isFeatureEnabled(feature: .TssBatch)
+        let supportsBatch = vault.libType == .DKLS || vault.libType == .KeyImport
+        let isTssBatch = supportsBatch && tssBatchEnabled
```
johnnyluo, line 167: `registerFastVaultServer` recomputes `isTssBatch` from scratch instead of using the `self.isTssBatch` computed property defined above. The `vault` snapshot is constructed from `self.vault`, so the result is identical — but if the definition of `isTssBatch` ever changes (e.g. adding another libType), this local copy will silently diverge.

```swift
// instead of:
let supportsBatch = vault.libType == .DKLS || vault.libType == .KeyImport
let isTssBatch = supportsBatch && tssBatchEnabled

// prefer:
let isTssBatch = self.isTssBatch
```

VultisigApp/VultisigApp/Features/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -34,6 +34,8 @@ class SettingsViewModel: ObservableObject {
     @AppStorage("thorchainChainnet") var enableThorchainChainnet: Bool = false
     @AppStorage("SellEnabled") var sellEnabled: Bool = false
     @AppStorage("isMLDSAEnabled") var isMLDSAEnabled: Bool = false
+    @AppStorage("agentEnabled") var agentEnabled: Bool = false
+    @AppStorage("txHistoryEnabled") var txHistoryEnabled: Bool = false
```
johnnyluo: `agentEnabled` and `txHistoryEnabled` look unrelated to this PR's feature (`is_tss_batch` on the keygen wire message). If these belong to a stacked PR, they should be on that branch's diff instead.

VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenPeerDiscoveryViewModel.swift
johnnyluo: this one fixed in [ac2d77b](https://github.com/vultisig/vultisig-ios/pull/4206/commits/ac2d77b1011a0e5cd41336369a5bfd7cb059fbe6)

johnnyluo (approved):
LGTM

PR #4203, opened 2026-04-21
feat(keysign): integrate Blockaid simulation into dApp hero (Phase 2A)

0x-suzume:
## Summary

Phase 2A — rebased onto [#4145 (Phase 1)](https://github.com/vultisig/vultisig-ios/pull/4145) which ships the dApp hero infrastructure. Phase 1 deliberately keeps the hero **title-only** for 4byte-decoded contract calls because 4byte is front-runnable. Phase 2A **promotes Blockaid's simulation** to the hero when it returns an authoritative `send` / `swap.fromCoin` / `transfer.fromCoin` branch, and keeps the title-only fallback with a small **"Unverified function"** caption when it doesn't.

Parity with [vultisig/vultisig-windows#3740](https://github.com/vultisig/vultisig-windows/issues/3740) (delivered).

## What this PR does

- Flips Phase 1's hardcoded `heroAmount: nil` / `heroTicker: nil` / `heroImage: nil` in the verify + done call sites to read from new `ViewModel` properties that return values **only when Blockaid simulation succeeded**
- Adds `BlockaidSimulationService` (actor) that caches results by `(chain, memo)` and coalesces in-flight requests so verify → sign → done all share a single API call
- Adds `BlockaidSimulationParser` (mirrors `parseBlockaidEvmSimulation` in the Windows reference)
- Extends `BlockaidRpcClient` with `simulateEVMTransaction` (POST with `options: ["simulation"]`) and a new response model
- Threads `heroCaption` through `SendCryptoVerifySummary` / `SendCryptoContent` / `SendCryptoDoneHeaderView` so both screens render the "Unverified function" label below Phase 1's title when simulation is absent
- Parallelizes `loadThorchainID` / `loadFunctionName` / `loadSimulation` via `async let` so Blockaid doesn't sit behind 4byte
- Replaces a stray `print()` with `logger.error()` (rule: `.claude/rules/logging.md`)
- Adds `unverifiedFunction` to all 8 locales

## Integration points with #4145

- `JoinKeysignViewModel` / `KeysignViewModel` gain `heroAmount` / `heroTicker` / `heroImage` / `unverifiedFunctionCaption` computed properties. Non-nil **iff** `BlockaidSimulationService.simulate()` returned a `transfer` or `swap` branch.
[...]

PR #4207, opened 2026-04-22
feat(blockaid): Solana simulation for dApp keysign hero (Phase 4)

gastonm5:
## Summary

Phase 4 of the Blockaid parity plan — extends the dApp keysign hero's Blockaid integration to cover Solana, matching the browser extension.

**Stacked on #4203** (Phase 2A-3-5 + validation-header wiring). Review after that one lands; rebase to `main` once it merges.

## What this PR does

- Ports `parseBlockaidSolanaSimulation` from `vultisig-windows/core/chain/security/blockaid/tx/simulation/api/core.ts`.
- Adds `BlockaidRpcClient.simulateSolanaTransaction(address:rawTransactions:)` which hits the same `/solana/message/scan` endpoint as validation but with `options: ["simulation"]`.
- New `BlockaidSolanaSimulationResponseJson` response model (note: Blockaid's Solana shape uses `account_assets_diff` with singular-object `in`/`out`, unlike the EVM `assets_diffs` with arrays).
- Extends `BlockaidSimulationService.scan(...)` with a Solana branch:
  - Reads `KeysignPayload.signSolana.rawTransactions` (base64 from the dApp-supplied sign payload).
  - Converts base64 → base58 for the Blockaid request, matching `getBlockaidPayloadSimulationInput` in the extension.
  - Cache key becomes per-chain-kind: `.evm(chain, memo)` (existing) vs `.solana(rawTransactions digest)`.
- Parser wraps native SOL in the SPL wrapped-SOL mint sentinel (`So11111111111111111111111111111111111111112`) so `TokensStore` lookups work uniformly; for 3-diff responses with a native-SOL leg, filters that diff (the extension's "likely the tx fee" heuristic).
- Coin enrichment prefers Blockaid-provided metadata, falls back to `TokensStore.findTokenMeta(chain: .solana, ...)`, and finally to a truncated-mint ticker for unknown tokens.

## What this PR does NOT do

- Solana validation is still served by the existing `scanSolanaTransaction` path (unchanged). The extension also requests Solana simulation-only, so we match.
- No UI changes — the existing hero reads `BlockaidSimulationInfo` chain-agnostic, so Solana payloads just populate the same `.transfer` / `.swap` cases the EVM path already drives.

## Risk item (4h in the plan) — resolved

Pre-sign raw-tx availability was the biggest risk. Confirmed: `KeysignPayload.signSolana?.rawTransactions` carries the base64-encoded unsigned bytes in the dApp keysign flow (`SignSolana.swift:11`), so we can simulate before the user signs.

## Tests

15 new XCTest cases (**36 total** when run with the existing EVM suite):

### `BlockaidSolanaSimulationParserTests` (10 new)
- Short-circuit paths: nil result, nil simulation, empty diffs.
- Transfer: happy path, native-SOL uses wrapped-SOL mint, returns nil when `out` missing, returns nil when decimals missing and mint is unknown to `TokensStore`.
- Swap: happy path, native-SOL-fee filter (3 diffs → swap on remaining 2), fallback to `.transfer` when `in` side missing.

### `BlockaidSimulationServiceTests` (5 new)
- Dispatches to Solana RPC (and not the EVM RPC) for Solana payloads.
- Decodes `rawTransactionsBase64 → base58` end-to-end (asserts against `base58("hello") == "Cn8eVZg"`).
- Returns `.empty` when `signSolana` is missing.
- Caches success; different raw txs cache independently; does not cache failures.

### Mock
- `MockBlockaidRpcClient` gains `simulateSolanaResult`, `simulateSolanaCallCount`, `simulatedSolanaRawTransactions` tracking.

## Test plan

- [x] `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/ VultisigApp/VultisigAppTests/` — clean (2 pre-existing warnings in `KeysignPayloadCodable.swift` are unrelated)
- [x] `xcodebuild build` — **BUILD SUCCEEDED**
- [x] `xcodebuild test -only-testing:VultisigAppTests/BlockaidSolanaSimulationParserTests …` — 36/36 passed
- [ ] Manual: dApp cosign on Solana (1inch SOL swap / Jupiter swap / SPL transfer) — verify hero shows correct balance change and no duplicate network calls between verify/sign/done

## Related

- Stacked on: #4203
- Follows: Phase 2A-3-5 work in #4203
- Plan: `BLOCKAID_PARITY_PLAN.md`

## Release Notes

* **New Features**
  * Added Solana transaction simulation and security scanning via Blockaid integration, enabling automatic transfer and swap detection with detailed risk levels, comprehensive security warnings, and full transaction outcome analysis prior to signing operations.

* **Tests**
  * Added test coverage for Solana simulation parsing, response conversion to security results, and complete service layer integration validation including caching behavior.


gastonm5:
@coderabbitai full review

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
```diff
@@ -93,4 +99,126 @@ enum BlockaidSimulationParser {
...
+    /// `.transfer`, two remaining diffs map to `.swap` (or `.transfer` if only
+    /// the out side has a value).
+    static func parseSolana(
+        response: BlockaidSolanaSimulationResponseJson
+    ) -> BlockaidSimulationInfo? {
+        guard let diffs = response.result?.simulation?.accountSummary?.accountAssetsDiff,
+              !diffs.isEmpty else {
+            return nil
+        }
+
```
johnnyluo, line 117: **SOL fee filter can misfire for incoming SOL.**

This filters out native SOL whenever exactly 3 diffs are present, regardless of direction. If the user is *receiving* SOL (e.g. a contract crediting them), the SOL leg is silently dropped and the hero shows a lopsided swap.

Consider only filtering when the SOL diff has an `out` and no `in` (i.e. it's a pure debit — the fee signature):
```swift
if diffs.count == 3,
   let solDiff = diffs.first(where: { $0.asset.type == "SOL" || $0.assetType == "SOL" }),
   solDiff.`in` == nil, solDiff.out != nil {
    relevant = diffs.filter { $0.asset.type != "SOL" && $0.assetType != "SOL" }
} else {
    relevant = diffs
}
```
The existing test covers the fee case; a companion test for the incoming-SOL case would lock in the correct behaviour.

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationService.swift
```diff
@@ -5,10 +5,13 @@
...
+import WalletCore
 
-/// Combined output of a Blockaid `/evm/json-rpc/scan` call for a keysign
-/// payload: the parsed balance-change simulation plus the risk validation
-/// result that drives the `SecurityScannerHeaderView` above the hero.
+/// Combined output of a Blockaid scan call for a keysign payload: the parsed
+/// balance-change simulation plus the risk validation result that drives the
+/// `SecurityScannerHeaderView` above the hero. Solana scans populate only the
+/// `simulation` field today (validation for Solana is served by a separate
+/// `/solana/message/scan` call with `options: ["validation"]`).
```
johnnyluo, line 14: **Stale doc comment.** This line says validation is served by a *separate* `/solana/message/scan` call, but `buildSolanaSimulateRequest` already requests `["simulation", "validation"]` in a single combined call. The comment should say something like: *"A single call requests both simulation and validation so the hero and the Scanned-by-Blockaid header are populated without a second round-trip."*

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidExtensions.swift
```diff
@@ -100,6 +100,53 @@ extension BlockaidTransactionScanResponseJson {
...
 }
 
+// MARK: - BlockaidSolanaSimulationResponseJson Extensions
+
+extension BlockaidSolanaSimulationResponseJson {
+
+    /// Mirrors `toKeysignScannerResult()` on the EVM response but reads from
+    /// the Solana-shape validation nested under `result.validation`. Returns
+    /// `nil` when validation is absent so callers can fall back to `.idle`.
+    func toKeysignScannerResult() -> SecurityScannerResult? {
```
johnnyluo, line 110: **Degradation surface when `validation` is absent.**

Returning `nil` here is consistent with the EVM path, but it means a Blockaid API change that omits `validation` from the Solana response silently disables the entire feature. A future-hardening option would be to return a fallback `SecurityScannerResult` with `isSecure: true` and no warnings when `simulation` succeeded but `validation` is missing — similar to how an empty `extendedFeatures` array is handled today. Not blocking, but worth a `// TODO` if intentional.

VultisigApp/VultisigAppTests/Services/BlockaidSolanaSimulationParserTests.swift
```diff
@@ -0,0 +1,324 @@
...
+        XCTAssertTrue(scannerResult?.isSecure ?? false)
+    }
+
+    func test_toKeysignScannerResult_returnsNil_whenValidationMissing() {
+        let response = BlockaidSolanaSimulationResponseJson(
+            result: BlockaidSolanaSimulationResponseJson.BlockaidSolanaSimulationResultJson(
+                simulation: nil,
+                validation: nil
+            ),
+            status: "SUCCESS",
```
johnnyluo: **`status` casing inconsistency.** Every other fixture in this file uses `"Success"` (mixed-case), but this one uses `"SUCCESS"` (all-caps). If `status` is ever used in a case-sensitive comparison this becomes a silent test blind-spot. Please standardise to `"Success"` to match the rest of the file.

johnnyluo (commented):
Strong implementation with thorough test coverage — the fixture helpers and per-case tests are exactly the right level of detail. Four inline comments below; the SOL-fee filter (first comment) is the one change I'd want before merge.

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationService.swift
```diff
@@ -77,32 +103,78 @@ actor BlockaidSimulationService {
...
+            // getBlockaidPayloadSimulationInput resolver.
+            let rawTxsBase64 = keysignPayload.signSolana?.rawTransactions ?? []
+            let rawTxsBase58: [String] = rawTxsBase64.compactMap { base64 -> String? in
+                guard let data = Data(base64Encoded: base64) else { return nil }
+                return Base58.encodeNoCheck(data: data)
+            }
+            guard !rawTxsBase58.isEmpty else {
+                logger.warning("solana scan aborted: rawTransactions base64 decode produced empty list (input count=\(rawTxsBase64.count))")
+                return .empty
+            }
```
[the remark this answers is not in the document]
johnnyluo: @gastonm5 maybe we need to think about this one , is it possible that tx in rawTransactions will not in base64 format?  

If it can , whether we are still able to scan it?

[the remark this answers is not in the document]
gastonm5: Good question — I traced it. `rawTransactions` is base64 by contract, enforced upstream:

- `SignSolana.swift:10-14` comment declares `[String] // base64 encoded`.
- In the browser extension (`vultisig-windows/core/inpage-provider/.../solana/parser.ts:43`), the raw tx is decoded as base64 (`Buffer.from(data[0], 'base64')`) and run through `TransactionDecoder.decode` **before** the keysign payload is constructed. A non-base64 string would fail there and the payload would never reach iOS.
- The Windows Blockaid input resolver does the same base64→base58 conversion we do.

So by the time we see `rawTransactions`, base64 is a precondition — a non-base64 entry would mean a truly corrupted payload, not a format mismatch. In that case simulating on the remaining subset would be worse than not simulating, since the hero would show balance changes that don't match what the user signs.

Fixed in ad229bb63: on any decode failure we now log and return `.empty` instead of silently dropping entries. This also addresses CodeRabbit's comment above.

VultisigApp/VultisigAppTests/Services/BlockaidSolanaSimulationParserTests.swift
gastonm5: Fixed in 0a07a0d7d — standardised all three `"SUCCESS"` occurrences (lines 22, 73, 92) to `"Success"` to match the rest of the file.

PR #4252, opened 2026-04-27
fix(keysign): hash-verify fallback when peer wins broadcast race

### LINKED ISSUE #4170: iOS: Keysign shows error when peer already broadcast the tx (hash-verify fallback)
## Summary

When co-signing a transaction across devices (any combination of iOS/macOS/Windows/Android), every device broadcasts the signed tx independently. When a peer wins the RPC race, iOS sees "tx already in mempool / already exists / sequence mismatch" style errors and flips the signing screen to failure — even though the transaction is on-chain and the user's funds moved correctly.

iOS has this pattern-matched only for **EVM** and **Cardano**. Every other chain falls through to the failure branch.

## Why it happens

`VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:526-700` — `broadcastTransaction()` is non-throwing; errors are caught and routed to `handleBroadcastError` (lines 734-764).

`handleBroadcastError` only short-circuits to success for:
- EVM: `"already known"`, `"replacement transaction underpriced"`, `"This transaction has already been processed"` (lines 741-747)
- Cardano: `"BadInputsUTxO"`, `"timed out"` (lines 751-755)

Everything else falls through to:

```swift
DispatchQueue.main.async {
    self.keysignError = errMessage
    self.status = .KeysignFailed
}
```

Chains that hit the default branch on duplicate-broadcast:
Cosmos (gaia, kujira, osmosis, dydx, terra, noble, akash, qbtc), THORChain, Mayachain, UTXO (Bitcoin, Litecoin, Dogecoin, Dash, Zcash, Bitcoin Cash), Solana, Tron, Sui, Ripple, Polkadot, Bittensor, Ton.

## Proposed fix — hash-verification (always on error)

Since every signed tx has a deterministic hash (already computed into `transactionType.transactionHash`), replace the fragile per-chain error-string matching with an on-chain lookup in `handleBroadcastError`:

```
on any broadcast error:
  if chain.getTxByHash(transactionType.transactionHash) exists (mempool or block):
    self.txid = transactionType.transactionHash
    return   // treat as success
  else:
    existing failure path
```

Benefits:
- Works for every chain; no per-chain string catalog to maintain
- Can't false-positive on "different tx advanced sequence / nonce" scenarios
- Failure path only — no extra RPC on the happy path
- Doesn't touch the post-success pending-tx tracker (`savePendingTransaction` at line 702) — that continues unchanged

Each chain already has a "tx status / get tx" method used by `StoredPendingTransactionStorage` — consolidate those behind a `getTxByHash(chain, hash)` helper and call it from the error handler.

## Audit existing short-circuits (likely tighten)

The current EVM/Cardano shortcuts can themselves false-positive:

- **EVM `"replacement transaction underpriced"`** — fires when a *different* tx occupies the same nonce, not when peer broadcast our tx. If peer broadcast our tx, geth returns `"already known"`, not this. This string-match can claim success for a tx that was never broadcast.
- **Cardano `"timed out"`** — client-side timeout. Doesn't mean tx landed.
- **Cardano `"BadInputsUTxO"`** — UTxOs could have been spent by any other tx, not only ours.

Once hash-verify is in, these string-matches become redundant *and* safer to delete. Initial PR can keep them alongside hash-verify; follow-up removes them.

## Non-goals

- No cross-device coordination / initiator-only broadcast. Every device still broadcasts independently.
- `savePendingTransaction` + pending-tx polling stay untouched.

## Files

- `VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:526-700` — broadcast switch
- `VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:734-764` — `handleBroadcastError` (add hash-verify before default branch)
- `VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:694-722` — existing pending-tx tracker (unchanged)

## Related

Matching issues will be filed on `vultisig-sdk` (Windows) and `vultisig-android` with the same design.

gastonm5:
## Summary

When co-signing a transaction across devices, every device broadcasts independently. When a peer wins the RPC race, iOS would see "tx already in mempool / sequence mismatch / already known" errors and flip the signing screen to **failed** — even though the tx is on-chain and funds moved.

Pre-fix, this was pattern-matched only for **EVM** (`already known`, `replacement transaction underpriced`, `This transaction has already been processed`) and **Cardano** (`BadInputsUTxO`, `timed out`). Every other chain (Cosmos, THORChain, Mayachain, UTXO, Solana, Sui, Tron, Ripple, Polkadot, Bittensor, Ton) fell through to the failure branch.

## Fix — chain-agnostic hash verification

`handleBroadcastError` now runs a hash lookup before falling through to `.KeysignFailed`:

```swift
if await isAlreadyOnChain(transactionType: transactionType) {
    self.txid = transactionType.transactionHash
    return  // treat as success
}
```

`isAlreadyOnChain` calls `TransactionStatusService.shared.checkTransactionStatus(txHash:chain:)` which already has per-chain providers (EVM, UTXO, Cosmos, Solana, THORChain, Cardano, Polkadot, Bittensor, Sui, Ton, Ripple, Tron). Only short-circuits to success on **positive evidence** (`.confirmed` or `.pending`). `.notFound`, `.failed`, or thrown errors fall through so we never claim success on uncertainty.

`handleBroadcastError` becomes `async`; inline `catch` sites add `await`, the two callback-based UTXO sites wrap the call in `Task { ... }`.

The existing EVM/Cardano string short-circuits stay alongside hash-verify per the issue's staged plan — a follow-up can remove them once the hash-verify path has soaked.

## Test plan

- [ ] `make build-check` clean
- [ ] SwiftLint passes on `KeysignViewModel.swift`
- [ ] Two-device keysign on a Cosmos chain (e.g. THORChain): force the second device to lose the race and confirm the signing screen shows success with the correct txid
- [ ] Same scenario on a UTXO chain (Bitcoin or Litecoin)
- [ ] Same scenario on Solana
- [ ] Single-device keysign on every supported chain still succeeds (regression — the new path only fires on broadcast failure)
- [ ] Broadcast a deliberately invalid tx (insufficient funds) and confirm the failure path still shows the error (the hash lookup must return `.notFound` and fall through)

## Files

- `VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift` — new `isAlreadyOnChain(transactionType:)` helper, async `handleBroadcastError`, callers updated

## Related

Closes #4170. Sibling issues will be filed against `vultisig-sdk` (Windows) and `vultisig-android`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction broadcast reliability with an automatic on-chain status fallback: confirmed or pending results are treated as successful and clear UTXO cache where applicable, reducing false failure alerts.
  * More consistent broadcast status handling and retry behavior to avoid race conditions that could overwrite failure or retry states.
  * Better population of approval transaction IDs for previously handled/bad-input cases for clearer tracking.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
@@ -834,6 +834,10 @@ class KeysignViewModel: ObservableObject {
         if await isAlreadyOnChain(transactionType: transactionType) {
             logger.info("transaction already on-chain via peer broadcast — using hash \(transactionType.transactionHash, privacy: .public)")
             self.txid = transactionType.transactionHash
+            self.approveTxid = transactionType.approveTransactionHash
+            if let coin = keysignPayload?.coin, coin.chainType == .UTXO {
+                await BlockchairService.shared.clearUTXOCache(for: coin)
+            }
             return
         }
 
--- a/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
@@ -662,7 +662,7 @@ class KeysignViewModel: ObservableObject {
                                 await BlockchairService.shared.clearUTXOCache(for: keysignPayload.coin)
                             }
                         case .failure(let error):
-                            self.handleBroadcastError(error: error, transactionType: transactionType)
+                            Task { await self.handleBroadcastError(error: error, transactionType: transactionType) }
                         }
                     }
                 case .bitcoinCash, .litecoin, .dogecoin, .dash, .zcash:
@@ -676,14 +676,14 @@ class KeysignViewModel: ObservableObject {
                                 await BlockchairService.shared.clearUTXOCache(for: keysignPayload.coin)
                             }
                         case .failure(let error):
-                            self.handleBroadcastError(error: error, transactionType: transactionType)
+                            Task { await self.handleBroadcastError(error: error, transactionType: transactionType) }
                         }
                     }
                 case .cardano:
                     do {
                         self.txid = try await CardanoService.shared.broadcastTransaction(signedTransaction: tx.rawTransaction)
                     } catch {
-                        self.handleBroadcastError(error: error, transactionType: transactionType)
+                        await self.handleBroadcastError(error: error, transactionType: transactionType)
                     }
                 case .gaiaChain, .kujira, .osmosis, .dydx, .terra, .terraClassic, .noble, .akash, .qbtc:
                     let service = try CosmosService.getService(forChain: keysignPayload.coin.chain)
@@ -745,7 +745,7 @@ class KeysignViewModel: ObservableObject {
                 self.txid = regularTxHash
             }
         } catch {
-            handleBroadcastError(error: error, transactionType: transactionType)
+            await handleBroadcastError(error: error, transactionType: transactionType)
         }
 
         if txid == "Transaction already broadcasted." {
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4270, opened 2026-04-29
fix(swap): rank quotes by net output instead of priority order

### LINKED ISSUE #4268: Swap: same-chain ERC20→ETH picks THORChain over aggregators (no quote ranking)
## Repro

User-reported swap on iOS: **USDC (Ethereum) → ETH (Ethereum)**, amount **$10**.

- Provider chosen: **THORChain**
- Total fee: **$7.88** on a $10 swap (~79%)
  - Network Fee: $7.58 (3.3M gwei — THORChain Router `depositWithExpiry` deposit tx)
  - Outbound Fee: $0.24
  - Swap Fee: $0.02

For a same-chain Ethereum ERC20→ETH route, an aggregator (1inch / LI.FI / KyberSwap) should win on net output. THORChain is a cross-chain router; using it here forces the user to pay mainnet gas for a vault deposit instead of a direct DEX swap.

## Root cause

**`VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift:20-73`** — quote selection returns the **first successful provider in priority order** and cancels the rest. There is no comparison of output amounts or net-of-fees ranking across providers.

```swift
for task in tasks {
    switch await task.value {
    case let .success(quote):
        tasks.forEach { \$0.cancel() }
        return quote
    case let .failure(error):
        lastError = error
        continue
    }
}
```

**`VultisigApp/VultisigApp/Core/Services/Actions/Coin+Swaps.swift:25-101`** — provider list ordering for Ethereum:

```swift
case .ethereum:
    let defaultProviders: [SwapProvider] = [.oneinch(chain), .lifi, .kyberswap(chain)]
    var providers: [SwapProvider] = []
    if thorEthTokens.contains(ticker) { providers.append(.thorchain) }
    if mayaEthTokens.contains(ticker) { providers.append(.mayachain) }
    return providers + defaultProviders   // [.thorchain, .mayachain, .oneinch, .lifi, .kyberswap]
```

Both ETH and USDC are in `thorEthTokens`/`mayaEthTokens`, so the eligible list starts with THORChain. THORChain succeeds on same-chain routes (it routes through a single inbound→outbound vault), so the first task returns and the aggregators are never queried.

## Expected behaviour

For each swap, query all eligible providers in parallel, then pick the quote with the highest **net output** (output amount − network gas − provider fees − outbound fees, all in a common denomination such as fiat).

## Suggested fix direction

1. In `SwapService.fetchQuote`, await **all** tasks instead of returning on first success.
2. Compute net output per quote (each `SwapQuote` already carries fee fields).
3. Return the max-net-output quote; surface losers for telemetry.
4. Add a fallback path so a single provider failure doesn't block selection.

## Cross-platform context

Same architectural issue exists on Android (`vultisig-android`) and SDK (`vultisig-sdk`), filed separately. iOS is the platform reproducing the user-reported $10 USDC→ETH case directly because of the priority order above.

gastonm5:
## Summary

`SwapService.fetchQuote` returned the first successful provider in priority order, so same-chain Ethereum ERC20→ETH swaps picked THORChain even when an aggregator would have produced a much higher net output. The reported case was a \$10 USDC→ETH swap quoted at \$7.88 in fees on THORChain because the Router's `depositWithExpiry` deposit costs ~3.3M gas, while a direct DEX swap via 1inch / LI.FI / KyberSwap would have been ~10× cheaper.

### What changed

- **`SwapService.fetchQuote`** now fans out every eligible provider in parallel via `withTaskGroup`, collects all results, and picks the highest net-output quote.
- **New `SwapQuote+Ranking.swift`** with one helper:
  - `expectedNetToAmount(toCoin:)` — expected output in `toCoin` units net of provider/inbound/outbound fees. THORChain `expectedAmountOut` is already net; LiFi additionally subtracts the integrator fee; aggregators use `dstAmount`.
- **`selectBestQuote(quotes:toCoin:)`** ranks by `expectedNetToAmount` and falls back to the first quote (priority order from `resolveAllProviders`) when no candidate is rankable. Logs every candidate's value next to the final pick (`[swap-rank] candidates=4 [...] → 1Inch`).
- **All-failures path** preserves the prior behaviour of throwing the first encountered error.

### Why this is enough

Every provider in a candidate set swaps to the same `toCoin`, so the destination amount is directly comparable in raw token units — no fiat conversion needed. THORChain's `expectedAmountOut` is already net of inbound + swap + outbound fees, so the comparison naturally penalises its high outbound fees against an aggregator's `dstAmount`. The user-reported case is fixed because THORChain's net output (~0.0029 ETH) is lower than 1inch's `dstAmount` (~0.0030 ETH).

### What was considered and dropped

An earlier draft also subtracted source-chain gas in fiat, with a three-tier ranking (fiat → toCoin amount → priority). Writing the unit tests surfaced a real bug: gas was priced against `fromCoin` (the swap's source token, e.g. USDC) instead of the chain's native gas token (e.g. ETH). Fixing it correctly would require a separate native-token price lookup. The simpler single-tier ranker covers the bug cleanly without that machinery; if the team later wants to break ties between aggregators by gas, the helper can be re-added behind a native-coin price source.

### Out of scope

Android (`vultisig-android`) and SDK (`vultisig-sdk`) have the same architectural issue and are tracked separately, per the issue body. EVM gas accounting in the ranker is a follow-up — file separately if needed.

## Test Plan

- [x] SwiftLint clean (`swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/Blockchain/Swaps/Common/`)
- [x] iOS Simulator build succeeds (`xcodebuild -scheme VultisigApp build`)
- [x] **Unit tests** — `SwapQuoteRankingTests` covers `expectedNetToAmount` per provider (THORChain, Maya, 1inch, KyberSwap, LiFi with/without integrator fee), `selectBestQuote` rank/fallback paths, and a same-chain ERC20→ETH regression case. 15 tests, all passing.
- [ ] Manual: Ethereum vault, USDC→ETH \$10 swap → confirm aggregator (1inch / LI.FI / KyberSwap) is picked, total fee ≪ \$7.88
- [ ] Manual: BTC→ETH (cross-chain) → THORChain still wins (only eligible provider)
- [ ] Manual: same-chain swap with all providers offline except one → fallback path still returns that quote
- [ ] Verify `[swap-rank]` log line appears on each quote refresh

## Related

- Closes #4268

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
@@ -1,0 +1,86 @@
+//
+//  SwapQuote+Ranking.swift
+//  VultisigApp
+//
+
+import Foundation
+import BigInt
+
+extension SwapQuote {
+
+    /// Expected output amount in `toCoin` units, net of provider/inbound/outbound fees.
+    /// THORChain/Maya: `expectedAmountOut` is already net of inbound + swap + outbound fees.
+    /// LiFi/1inch/KyberSwap: `dstAmount` is the swapped amount; LiFi additionally subtracts the
+    /// integrator fee that is charged on the output side.
+    func expectedNetToAmount(toCoin: Coin) -> Decimal? {
+        switch self {
+        case .thorchain(let quote),
+             .thorchainChainnet(let quote),
+             .thorchainStagenet(let quote),
+             .mayachain(let quote):
+            guard let raw = Decimal(string: quote.expectedAmountOut), raw > 0 else { return nil }
+            return raw / toCoin.thorswapMultiplier
+
+        case .oneinch(let quote, _),
+             .kyberswap(let quote, _):
+            guard let raw = BigInt(quote.dstAmount), raw > 0 else { return nil }
+            return toCoin.decimal(for: raw)
+
+        case .lifi(let quote, _, let integratorFee):
+            guard let raw = BigInt(quote.dstAmount), raw > 0 else { return nil }
+            let amount = toCoin.decimal(for: raw)
+            let fee = amount * (integratorFee ?? 0)
+            return amount - fee
+        }
+    }
+
+    /// Source-chain gas cost in `fromCoin` native units, when estimable from the quote.
+    /// Aggregator quotes carry `gas`/`gasPrice` directly. THORChain Router deposits on EVM are
+    /// estimated via the project's standard swap gas constant (matches how 1inch/Kyber/LiFi normalize
+    /// missing gas). For non-EVM source chains the gas is paid in the native coin and is not exposed
+    /// at quote time, so we return `nil` and skip gas in the comparison.
+    func sourceChainGasInNative(fromCoin: Coin) -> Decimal? {
+        switch self {
+        case .oneinch(let quote, _),
+             .kyberswap(let quote, _),
+             .lifi(let quote, _, _):
+            guard fromCoin.chain.chainType == .EVM else { return nil }
+            return evmGasInNative(gas: quote.tx.gas, gasPrice: quote.tx.gasPrice, fromCoin: fromCoin)
+
+        case .thorchain, .thorchainChainnet, .thorchainStagenet, .mayachain:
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4275, opened 2026-04-30
feat(keysign): live ERC-20 metadata fallback for unknown tokens

### LINKED ISSUE #4149: feat: Full ABI parameter decoding for dApp display (Phase 2)
## Context
Phase 1 shows decoded function names and resolved token amounts. Phase 2 adds full parameter decoding with rich labels.

## What to build
- **ABI param decode:** For approve → show spender + amount. For transfer → show recipient + amount. For swaps → show input/output tokens + min receive.
- **Token metadata resolution:** On-chain \`symbol()\` + \`decimals()\` call for unknown tokens
- **Known contract labels:** "Uniswap V2 Router", "1Inch Router", etc.

**Companion issue:** vultisig/vultisig-windows#3738
**PRD:** sections 2.1–2.4

gastonm5:
Closes #4149 (Phase 2 token-metadata RPC piece).

## Summary

Adds `TokenMetadataResolver`, an actor that fetches `symbol()` + `decimals()` from an unknown ERC-20 contract via `eth_call` and caches the result for 24h. Plumbs it into `JoinKeysignViewModel.resolveTokenDisplay(...)` as the third fallback after vault coins and `TokensStore`.

Before this change, dApp transactions involving tokens not in the user's vault and not in `TokensStore` showed the function name with no token context. After this change, the verify/done screens render the resolved symbol + decimals so the user sees ``Approve 100 NEW_TOKEN`` instead of the raw call.

Reuses existing `RpcEvmService.getTokenInfo(...)` plumbing — no new ABI decoder.

## Behavior

* Cache key is ``(chain.rawValue, contractAddress.lowercased())`` so the same address on different chains is a separate slot and casing differences collapse.
* Single-flight: concurrent resolves of the same key dedupe to one in-flight Task. Three parallel callers ⇒ one `eth_call` round trip.
* Failures are NOT cached. Next caller retries.
* TTL defaults to 24h. Token metadata is immutable on-chain so expiry only matters for new deployments.

## DI seam

The resolver takes an injectable ``TokenMetadataFetcher`` closure so tests don't need to mock SwiftData or the HTTP stack. The production default uses ``EvmServiceConfig.configurations[chain]`` to pick the right RPC endpoint per chain.

## Tests

8 new tests in `VultisigAppTests/Services/TokenMetadataResolverTests.swift`:

| Test | Asserts |
|---|---|
| ``testResolveCachesSuccessfulFetch`` | second call hits cache (1 fetch invocation) |
| ``testResolveDifferentChainsCacheIndependently`` | same address, different chains ⇒ separate cache slots |
| ``testResolveAddressIsCaseInsensitive`` | upper/lower casing of the same address share a slot |
| ``testResolveReturnsNilOnFetcherThrow`` | RPC error ⇒ nil, no crash |
| ``testResolveReturnsNilOnEmptySymbol`` | ``RpcEvmService.getTokenInfo`` failure mode (empty tuple) ⇒ nil, not cached |
| ``testResolveDoesNotCacheFailures`` | failed lookup doesn't poison the cache; subsequent success retries |
| ``testResolveSingleFlightsConcurrentCalls`` | 3 parallel resolves of the same key ⇒ 1 fetch |
| ``testResolveRefetchesAfterTTL`` | injected mutable clock; expiry triggers refetch |

## Out of scope (deferred to follow-ups)

- Known-contracts label registry (``"Uniswap V2 Router"``, etc.) — Blockaid already labels suspicious contracts.
- Generic swap-intent inference beyond the existing KyberSwap path — Blockaid surfaces ``"you're swapping X for Y"`` from token-transfer events.

Both noted in the audit at the [internal wiki](file:///Users/gaston/workspace/wiki/pages/projects/vultisig/dapp-signing-decode/overview.md) (not in-repo).

## Test plan

- [x] ``swiftlint`` — 0 violations on the 3 changed files.
- [x] ``xcodebuild test -only-testing VultisigAppTests/TokenMetadataResolverTests`` — 8 tests, 0 failures.
- [ ] Smoke: trigger a dApp transaction (e.g. via WalletConnect with a fresh deployer token), confirm the verify screen shows the resolved symbol + amount instead of the raw function name.

## Summary by CodeRabbit

* **New Features**
  * Token metadata cached with a 24‑hour TTL to improve performance
  * Live fallback resolution for unknown contract tokens to fetch symbols and decimals
  * Chain-aware and case‑insensitive token resolution so identical contracts on different networks are handled separately
  * Concurrent lookups deduplicated to avoid duplicate network requests; invalid or failing fetches are ignored (not cached)

* **Tests**
  * Added comprehensive tests for caching, TTL expiry, single‑flight deduplication, retries, and error handling

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/TokenMetadataResolver.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TokenMetadataResolver.swift
@@ -1,0 +1,104 @@
+//
+//  TokenMetadataResolver.swift
+//  VultisigApp
+//
+//  Live ERC-20 metadata fallback for unknown contracts during dApp tx display.
+//  Used by `JoinKeysignViewModel.resolveTokenDisplay(...)` after vault + TokensStore
+//  lookups miss. Caches per (chain, contractAddress) for 24h since ERC-20 symbol /
+//  decimals are immutable on-chain. Single-flights concurrent resolves of the same
+//  pair so the keysign UI never fires duplicate `eth_call`s.
+//
+
+import Foundation
+import OSLog
+
+private let logger = Logger(subsystem: "com.vultisig.app", category: "token-metadata-resolver")
+
+/// Successful metadata lookup. Empty `symbol` is treated as a failure and not cached.
+struct TokenMetadata: Equatable, Sendable {
+    let symbol: String
+    let decimals: Int
+}
+
+/// Closure-based fetcher so tests can inject a deterministic fake without mocking
+/// SwiftData or the HTTP stack. The production default uses `RpcEvmService.getTokenInfo`.
+typealias TokenMetadataFetcher = @Sendable (_ chain: Chain, _ contractAddress: String) async throws -> TokenMetadata
+
+actor TokenMetadataResolver {
+    static let shared = TokenMetadataResolver()
+
+    private struct CachedEntry {
+        let metadata: TokenMetadata
+        let fetchedAt: Date
+    }
+
+    private var cache: [String: CachedEntry] = [:]
+    private var inFlight: [String: Task<TokenMetadata?, Never>] = [:]
+
+    private let ttl: TimeInterval
+    private let fetcher: TokenMetadataFetcher
+    private let now: @Sendable () -> Date
+
+    init(
+        ttl: TimeInterval = 24 * 60 * 60,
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4276, opened 2026-04-30
fix(solana): thread memo through SPL token signing paths

realpaaao:
## Summary

`SolanaTokenTransfer` and `SolanaCreateAndTransferToken` both expose a `memo` field; without setting it, SPL sends silently drop any memo the user attached. Native SOL already wired memo through, so a THORChain LP add for native SOL works but the same flow on USDC.SOL (or any SPL asset) would broadcast memo-less and get refunded by THORChain.

Two-line fix in `SolanaHelper.swift`: pipe `keysignPayload.memo` into both SPL builder paths, mirroring the existing native-SOL pattern.

## Cross-client coordination

Same fix lands in vultisig-sdk PR #369 (extension/desktop) and vultisig-android (branch `fix/solana-spl-memo`). Native SOL was already correct on all three; only SPL was broken. No proto/wire-format change.

## Test plan

- [x] swiftlint clean on the touched file
- [ ] Build the app and send 0.001 USDC with a memo to a self-address — verify the on-chain tx contains a Memo program instruction with the expected memo string

## Summary by CodeRabbit

* **New Features**
  * Added memo support for Solana transactions, including native SOL transfers, SPL token transfers, and SPL token transfers with automatic account creation.


gastonm5 (approved):

johnnyluo (approved):
LGTM

PR #4289, opened 2026-05-03
fix: keysign discovery success banner shows error icon (#4286)

### LINKED ISSUE #4286: Keysign discovery: 'Notification was successfully sent!' banner shows error (red X) icon — should be success (check)
## Symptom

After kicking off a non-fast-vault keysign, the screen shows the banner *"Notification was successfully sent!"* — but it renders with the **error** style (red circle, red X icon) instead of the **success** style (green circle, check icon). Confusing for users since the message text and the icon disagree.

## Root cause

`withBanner(text:style:)` defaults `style` to `.error`:

`VultisigApp/VultisigApp/Components/NotificationBanner/BannerViewModifier.swift:36`:
```swift
func withBanner(text: Binding<String?>, style: NotificationBannerStyle = .error) -> some View {
    modifier(BannerViewModifier(text: text, style: style))
}
```

`KeysignDiscoveryView` attaches the modifier without specifying a style:

`VultisigApp/VultisigApp/Features/Keysign/Views/KeysignDiscoveryView.swift:73`:
```swift
.withBanner(text: $bannerText)
```

…then writes a *success* message to that binding:

`KeysignDiscoveryView.swift:275`:
```swift
bannerText = "notificationSentSuccessfully".localized
```

Net: success message + error chrome.

In this file `bannerText` is only ever assigned at line 275 (single use), so the screen's banner is always a success message — the `.error` default is unambiguously wrong here.

## Suggested fix

Two options, lowest-friction first:

1. **Pass the right style at the modifier site.**
   ```swift
   .withBanner(text: $bannerText, style: .success)
   ```
   Two-character change. Safe because the screen has no error-banner path.

2. **Replace `String?` with a typed banner state** so any future call site can't get the icon wrong:
   ```swift
   struct BannerMessage { let text: String; let style: NotificationBannerStyle }
   @State var banner: BannerMessage?
   .withBanner(banner: $banner)
   ```
   And update `BannerViewModifier` + `extension View` to take the typed binding. Worth doing if other screens have the same shape (worth a quick audit — search for `withBanner(text:` and check whether any of the bound `String?` properties carry success messages).

I'd ship option 1 now and open a separate cleanup for option 2 if the audit finds more sites.

## Quick audit hook

```bash
grep -rn "withBanner(text:" VultisigApp/VultisigApp
```

Any call site that doesn't pass `style:` AND writes a non-error message to its binding has the same bug.

cc @GaetanoIusto

gastonm5:
## Summary
- Closes #4286
- Flip `withBanner(text:style:)` default from `.error` to `.success`, since the only call sites that omit `style:` (now just `KeysignDiscoveryView`) emit success messages
- Pass `style: .error` explicitly at `SwapVerifyView` and `SendVerifyScreen`, which surface `RetryableBroadcastError.userFacingMessage` and were relying on the old `.error` default — without this, both retry banners would render success/green chrome on a real error

## Test plan
- [ ] Trigger a keysign on a non-fast-vault and confirm the *"Notification was successfully sent!"* banner renders with green/check chrome (not red/X)
- [ ] Force a swap broadcast retry path (e.g. submit a swap that hits a transient retryable error) and confirm `SwapVerifyView` still shows the retry message in red/error chrome
- [ ] Force a send broadcast retry path and confirm `SendVerifyScreen` retry banner still renders red/error chrome
- [ ] No regression on `DefiChainMainScreen` (passes `style: .error` explicitly — unchanged)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected notification banner styling to ensure error notifications consistently display with appropriate visual treatment across send and swap verification screens, improving user experience and feedback clarity.

* **Style**
  * Updated the default notification banner appearance to use success styling instead of the previous error styling, providing better visual consistency throughout the application when banner style is not explicitly specified.

### DIFF
--- a/VultisigApp/VultisigApp/Components/NotificationBanner/BannerViewModifier.swift
+++ b/VultisigApp/VultisigApp/Components/NotificationBanner/BannerViewModifier.swift
@@ -33,7 +33,7 @@ private struct BannerViewModifier: ViewModifier {
 }
 
 extension View {
-    func withBanner(text: Binding<String?>, style: NotificationBannerStyle = .error) -> some View {
+    func withBanner(text: Binding<String?>, style: NotificationBannerStyle = .success) -> some View {
         modifier(BannerViewModifier(text: text, style: style))
     }
 }
--- a/VultisigApp/VultisigApp/Features/Send/Views/Screens/SendVerifyScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Send/Views/Screens/SendVerifyScreen.swift
@@ -32,7 +32,7 @@ struct SendVerifyScreen: View {
             }
         }
         .screenTitle("verify".localized)
-        .withBanner(text: $retryBannerText)
+        .withBanner(text: $retryBannerText, style: .error)
         .alert(item: $error) { error in
             Alert(
                 title: Text(NSLocalizedString("error", comment: "")),
--- a/VultisigApp/VultisigApp/Features/Swap/Views/SwapVerifyView.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/Views/SwapVerifyView.swift
@@ -28,7 +28,7 @@ struct SwapVerifyView: View {
             Background()
             view
         }
-        .withBanner(text: $retryBannerText)
+        .withBanner(text: $retryBannerText, style: .error)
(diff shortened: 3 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4267, opened 2026-04-28
fix: WelcomeView snapshot render drift

gastonm5:
## Summary
- add strict precision and perceptual precision to WelcomeView snapshot assertions
- avoid failures from one-bit simulator renderer drift while preserving visual regression coverage

## Testing
- xcodebuild test -project VultisigApp/VultisigApp.xcodeproj -scheme VultisigApp -destination 'id=07344DB6-5E1F-42C4-832A-E4C9334944E8' -only-testing:VultisigAppTests/CreateVaultSnapshotTests/testWelcomeView_tryAgainState
- xcodebuild test -project VultisigApp/VultisigApp.xcodeproj -scheme VultisigApp -destination 'id=07344DB6-5E1F-42C4-832A-E4C9334944E8' -only-testing:VultisigAppTests/CreateVaultSnapshotTests/testWelcomeView_loading

## Summary by CodeRabbit

* **Tests**
  * Updated snapshot test configurations for enhanced accuracy.

PR #4301, opened 2026-05-04
fix(tests): stabilize testCreateVaultScreen_iPhone16Pro

gastonm5:
## Summary

`testCreateVaultScreen_iPhone16Pro` was flaking on CI from three stacked sources:

1. **`VultisigLogoAnimation` runs a Rive splash on its own clock.** The captured frame depended on wall-clock timing relative to `onLoad`. Frame-by-frame drift exceeded the 0.95 perceptual tolerance the test was using.
2. **The test snapshotted mid-animation.** `setData()` schedules three `asyncAfter` reveals at 0.1/0.2/0.3s, each driving a 0.3s spring. The test waited only 0.5s — capturing the button stack while opacity, offset, scale, and blur were all in flight. Springs asymptote rather than settling exactly, so no precision tolerance could absorb the variance.
3. **The test attached the `UIHostingController` to the live key window.** Racy under parallel test execution, and `UIScreen.main.bounds` couples to whichever simulator scene is foreground.

## Fix

Replace timing-based determinism with structural determinism:

- `CreateVaultView` gains `staticForSnapshot: Bool` (default `false`). When `true`, the three `@State` reveal flags (`showNewVaultButton`, `showSeparator`, `showButtonStack`) are seeded to `true` at `init`. The `.animation(.spring(...), value:)` modifier's `value` never changes, so no animation triggers — the view renders the fully-revealed final frame from the first pass.
- `VultisigLogoAnimation` gains `isStatic: Bool` (default `false`) that swaps the Rive view for a static `Image("vultisig-logo").resizable().scaledToFit()`. The logo region is now byte-stable.
- Test rewritten to mirror the existing `WelcomeView` snapshot pattern: pass the SwiftUI view directly to `.image(layout: .device(config: .iPhone16Pro))` with `0.99999`/`0.99999` precision. No window manipulation, no `asyncAfter`, no `expectation`.
- Reference PNG regenerated against the new deterministic render.

Production behaviour is unchanged — the new flags default to `false`, and any production caller path produces byte-identical output to before.

## Test Plan

- [x] `swiftlint lint` — 0 violations on touched files
- [x] `make build-check` — `BUILD SUCCEEDED`
- [x] `xcodebuild test … -only-testing:VultisigAppTests/CreateVaultSnapshotTests` on iPhone 17 Pro Max simulator (matching CI) — all 3 tests pass
- [x] Re-running the test on a fresh simulator passes deterministically

## Related

- Re-stabilizes the test category that #4267 patched for `WelcomeView` (precision tightening alone wasn't enough for `CreateVaultView` because of the active animations).

## Release Notes

* **Tests**
  * Enhanced snapshot testing for improved consistency and reliability in validating UI rendering.


johnnyluo (approved):
LGTM

PR #4328, opened 2026-05-08
fix(swap): show LI.FI/OneInch/Kyber network fee in native ETH (#4322)

### LINKED ISSUE #4322: Swap: Network Fee displayed in 'Gwei' is incorrect / inconsistent with Android
## Summary

On the Swap screen, the **Network Fee** for a LI.FI Ethereum → Base swap is displayed as a raw Gwei figure (e.g. `121,094 Gwei`). This appears to be either the wrong unit, the wrong value, or both. The number reads more like a gas-limit (units of gas) than a fee total in gwei, and the resulting USD subtotal looks too low for an Ethereum L1 transaction.

The Android app on the same swap route shows the network fee as an ETH amount with a USD equivalent (`0.00086087 ETH (~$2.00)`), which is the standard format used elsewhere in the app and matches realistic L1 gas costs.

## Steps to reproduce

1. Open Swap
2. From: Ethereum / ETH (Native)
3. To: Base / ETH (Native)
4. Get a quote (provider: LI.FI)
5. Expand **Total Fee** and look at **Network Fee**

## Observed (iOS)

- Network Fee: `121,094 Gwei`
- Total Fee: `$0.30`
- No ETH amount, no USD breakdown for the network fee line

iOS

## Expected (matches Android)

- Network Fee: `<amount> ETH (~$<usd>)`
- Total Fee USD reflects realistic L1 gas (typically a few dollars on Ethereum mainnet)

Android

## Why this looks wrong

- `121,094` is in the typical range of a **gas limit** (units of gas) for a swap/bridge call, not a total fee in gwei. If it really is the fee in gwei, then `121,094 gwei = 0.000121094 ETH ≈ \$0.28`, which would imply a sub-1-gwei effective gas price on Ethereum mainnet — not realistic under normal conditions.
- The Android value (`0.00086087 ETH`, ~\$2) is consistent with `gasLimit × gasPrice` at typical mainnet gas prices.
- Showing a network fee as a raw Gwei integer (without an ETH amount or USD equivalent) is also inconsistent with how the rest of the app presents fees.

## Notes

- The two screenshots were captured from different quote refreshes (different input amounts: 0.0018 ETH on iOS vs 0.001648 ETH on Android), so the absolute values are not directly comparable. The issue is the **format and order of magnitude**, not the exact dollar amount.
- Worth checking whether the displayed value is the LI.FI quote's gas limit being rendered with a `Gwei` suffix, vs. the actual fee in wei/gwei.

## Suggested fix

- Display Network Fee in the same format as Android and the rest of iOS: `<amount> ETH (~$<usd>)`.
- If a Gwei representation is desired, ensure the value is the actual fee (`gasLimit × gasPrice`), not the gas limit, and include the ETH/USD equivalent alongside it.

gastonm5:
## Summary

- The Swap verify screen was rendering the LI.FI/OneInch/KyberSwap EVM **Network Fee** as a raw Gwei figure (e.g. `121,094 Gwei`), which reads like a gas-limit. Android shows it correctly as `0.00086087 ETH (~$2.00)`.
- Root cause in `SwapCryptoLogic.swapGasString(tx:)`: when a quote was attached, `tx.fee` is already the *total* network fee in wei (gasPrice × gasLimit), but the EVM branch divided by 1e9 and labelled the result `Gwei`.
- Fix: when `tx.quote != nil`, format `tx.fee` as a native amount (using the existing `coin.decimal(for:)` helper) so the verify row reads `0.000861 ETH (~$2.00)`. The no-quote branch is unchanged — there `tx.gas` is a gas price in wei and the `Gwei` label is correct.
- Added `SwapNetworkFeeTests` to pin the new behaviour for LI.FI, OneInch, and KyberSwap, plus a guard test that the no-quote path still renders Gwei.

## Test plan

- [x] `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — no new warnings
- [x] `make build-check` — BUILD SUCCEEDED
- [x] `xcodebuild test ... -only-testing:VultisigAppTests/SwapNetworkFeeTests` — 4/4 pass
- [ ] Manual on-device verification (LI.FI Ethereum→Base swap) — out of scope for this PR; the fix is unit-tested and the verify view already pairs the crypto string with a fiat string so no UI changes are needed.

Closes #4322

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap network fee display to show accurate native token amounts when a swap quote is present (e.g., shows ETH amount instead of Gwei).
  * Preserved legacy Gwei display for EVM flows when no quote is available.
  * Fixed fee display for non-native source assets to avoid showing incorrect fee text.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCryptoLogic.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCryptoLogic.swift
@@ -150,20 +150,25 @@ struct SwapCryptoLogic {
 
     func swapGasString(tx: SwapTransaction) -> String {
         let coin = feeCoin(tx: tx)
-        let decimals = coin.decimals
 
-        // Use tx.fee for swap quotes (which includes corrected gas price calculations)
-        // Fall back to tx.gas for other transaction types
-        let gasValue = tx.quote != nil ? tx.fee : tx.gas
+        // Quote-driven swaps: `tx.fee` is the total network fee in the chain's
+        // smallest unit (gasPrice × gasLimit for EVM). Format as a native amount
+        // so the row reads "0.000861 ETH (~$2.00)" rather than a raw Gwei figure.
+        if tx.quote != nil {
+            let amount = coin.decimal(for: tx.fee)
+            return "\(amount.formatToDecimal(digits: coin.decimals).description) \(coin.ticker)"
+        }
 
+        // No quote: `tx.gas` is a gas price in wei for EVM chains, so display Gwei.
         if coin.chain.chainType == .EVM {
             guard let weiPerGWeiDecimal = Decimal(string: EVMHelper.weiPerGWei.description) else {
                 return .empty
             }
-            return "\((Decimal(gasValue) / weiPerGWeiDecimal).formatToDecimal(digits: 0).description) \(coin.chain.feeUnit)"
-        } else {
-            return "\((Decimal(gasValue) / pow(10, decimals)).formatToDecimal(digits: decimals).description) \(coin.ticker)"
+            return "\((Decimal(tx.gas) / weiPerGWeiDecimal).formatToDecimal(digits: 0).description) \(coin.chain.feeUnit)"
         }
+
+        let amount = coin.decimal(for: tx.gas)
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4331, opened 2026-05-08
refactor(swap): split monolithic flow into router-managed screens

### LINKED ISSUE #4330: Refactor Swap flow to use NavigationRouter (one screen per step)
## Problem

The Swap flow lives in a single monolithic view (\`SwapCryptoView\`) that switches its content via an internal \`currentIndex\` state machine. The \`SwapCryptoViewModel\` is a god-object that owns step state, quote/fees state, keysign continuity, and \`TransferViewModel\` conformance all at once.

This causes:

- **Layout coupling with Send.** \`SendCryptoDoneView\` is shared between flows, but Swap renders it raw with manual padding while Send wraps it in \`Screen { … }\`. Same component, two different parent contexts → inconsistent spacing/safe-area behavior.
- **No screen-level separation of concerns.** Back button, title, and toolbar are all conditional on \`currentIndex\` rather than per-screen.
- **macOS divergence.** A separate \`SwapCryptoHeader\` reimplements the title bar for macOS because the monolith doesn't compose with the cross-platform \`Screen\` chrome.
- **Hard to test individual steps in isolation** — every step depends on the same VM instance.

The Send and Function-call flows already follow the per-screen-router pattern (\`SendRoute\`, \`SendRouter\`, \`SendRouteBuilder\`, \`Send*Screen\` files). Swap should match.

## Proposed solution

Introduce \`SwapRoute\` / \`SwapRouter\` / \`SwapRouteBuilder\` mirroring \`SendRoute\` 1:1. Split the monolith into five \`Screen\`-wrapped views: \`SwapDetailsScreen\`, \`SwapVerifyScreen\`, \`SwapPairScreen\`, \`SwapKeysignScreen\`, \`SwapDoneScreen\`. Split the god-VM into per-screen VMs. Continuity data (\`KeysignPayload\`, \`KeysignInput\`, \`hash\`, \`approveHash\`) rides in route associated values. Retry path matches Send (keysign → \`tx.pendingRetryReason\` → pop 2 → verify banner).

\`SwapCryptoLogic\` becomes a stateless namespace (caseless \`enum\` with static methods) so any screen can call \`SwapCryptoLogic.swapFeeString(tx:)\` directly without per-VM delegation boilerplate.

## Acceptance criteria

- \`SwapCryptoView\`, \`SwapCryptoViewModel\`, \`SwapCryptoVerifyViewModel\`, \`SwapCryptoHeader\` removed
- Both Swap entry points (\`HomeRouteBuilder\`, \`VaultRouteBuilder\`) push \`SwapRoute.root\` instead of returning a view directly
- Single \`Screen\`-based layout works for both iOS and macOS — no \`#if os\` divergence at the screen level
- Done screen has no back button; Keysign screen has no nav bar
- Done screen scrolls; status animation renders correctly
- macOS + iOS builds clean; SwiftLint clean

gastonm5:
Closes #4330

## Summary

- Replaces the `SwapCryptoView` state-machine monolith with five `Screen`-wrapped views managed by `NavigationRouter`, mirroring the existing Send / FunctionCall pattern
- Splits the `SwapCryptoViewModel` god-object into per-screen view models: `SwapDetailsViewModel`, `SwapVerifyViewModel` (absorbs the old `SwapCryptoVerifyViewModel`), and a tiny `SwapKeysignViewModel` that mirrors `SendKeysignViewModel`
- Converts `SwapCryptoLogic` to a stateless namespace (caseless `enum` with static methods) so any screen calls `SwapCryptoLogic.swapFeeString(tx:)` directly — no per-VM delegation boilerplate
- Unifies iOS + macOS at the screen level: removes the macOS-only `SwapCryptoHeader`; each screen sets its own title/toolbar via `Screen` modifiers
- Fixes a latent bug where `screenBackButtonHidden` / `screenNavigationBarHidden` were silently ignored on iOS (Done now correctly hides its back button; Keysign hides the whole bar)

## What changed

**New files**
- `Features/Swap/Views/Navigation/{SwapRoute, SwapRouter, SwapRouteBuilder}.swift`
- `Features/Swap/Views/Screens/{SwapDetailsScreen, SwapVerifyScreen, SwapPairScreen, SwapKeysignScreen, SwapDoneScreen}.swift`
- `Features/Swap/ViewModels/{SwapDetailsViewModel, SwapVerifyViewModel, SwapKeysignViewModel}.swift`

**Removed**
- `SwapCryptoView.swift`, `SwapCryptoViewModel.swift`, `SwapCryptoVerifyViewModel.swift`, `SwapCryptoDetailsView.swift`, `SwapVerifyView.swift`, `SwapCryptoHeader.swift`

**Modified**
- `SwapTransaction` — `Hashable` conformance + `pendingRetryReason` for retry continuity
- `Components/Screen/Screen.swift` — iOS now honors `screenBackButtonHidden` (`.navigationBarBackButtonHidden`) and `screenNavigationBarHidden` (`.toolbar(.hidden, for: .navigationBar)`)
- `Core/Navigation/VultisigRouter.swift` + `App/ContentView.swift` — wire `SwapRouter` and `.navigationDestination(for: SwapRoute.self)`
- `Features/Home/Navigation/HomeRouteBuilder.swift` + `Features/Wallet/Views/Navigation/VaultRouteBuilder.swift` — push `SwapRoute.root` instead of returning the view directly
- `SwapFromToField`, `SwapDetailsSummary`, `SendCryptoDoneSummary`, `SendCryptoDoneView`, `SwapCryptoDoneView` — drop `SwapCryptoViewModel` references; use `SwapCryptoLogic.X(tx:)` static calls and the new VMs
- `SwapCryptoDoneView` — `cards` block wrapped in `ScrollView`; `TransactionStatusHeaderView` clamped to `.frame(minHeight: 150, maxHeight: 200)` so the Rive animation renders inside the scroll view (matches `SendCryptoDoneHeaderView`)

## Retry flow

Stack: `details → verify → pair → keysign`. On broadcast failure:

1. `KeysignView` reports failure → `SwapKeysignViewModel.retryBroadcast(reason:)`
2. `SwapKeysignScreen.onChange(viewModel.pendingRetryReason)` writes to `tx.pendingRetryReason` and pops 2 stack entries
3. `SwapVerifyScreen.onAppear` consumes `tx.pendingRetryReason`, shows banner, refreshes quotes

Same pattern as `SendKeysignScreen` / `SendVerifyScreen`.

## Test plan

- [ ] Golden path: details → verify → pair → keysign → done, on a non-fast vault
- [ ] Golden path on a fast vault (long-press flow + password entry)
- [ ] ERC20 swap that requires approval (verify both `approveHash` and `hash` land on Done)
- [ ] Native cross-chain swap (THORChain / Maya path)
- [ ] EVM swap via OneInch / Kyber / LiFi
- [ ] Back-navigate from Verify → Details (form state intact)
- [ ] Forced retry: trigger a broadcast failure during keysign and confirm the banner shows on Verify
- [ ] Done screen: back button absent, content scrolls, status animation renders, "track" + "done" buttons work
- [ ] Keysign screen: nav bar fully hidden (no back chevron leak)
- [ ] macOS build: title bar shows the right title per step; refresh counter visible on details/verify; QR-share visible on pair
- [ ] SwiftLint clean (`make lint`); `xcodebuild` succeeds for iOS + macOS

## Summary by CodeRabbit

* **Refactor**
  * Reorganized swap workflow navigation and improved transaction flow patterns
  * Updated swap interface components and layout structure
  * Streamlined state management and view composition for swap transactions

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4331)


VultisigApp/VultisigApp/Features/Swap/Views/Navigation/SwapRoute.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  SwapRoute.swift
+//  VultisigApp
+//
+
+enum SwapRoute: Hashable {
+    case root(fromCoin: Coin?, toCoin: Coin?, vault: Vault)
```
johnnyluo: **[Issue 1] SwiftData `@Model` objects stored as route associated values — actor safety risk**

`Vault` and `Coin` are SwiftData `@Model` classes that must never escape their `ModelContext`'s actor. Wrapping them in a `Hashable` enum that flows through `NavigationPath` may cross actor boundaries.

Consider passing `vault` by `PersistentIdentifier` in the route and re-fetching in each builder, mirroring how the Send flow avoids this.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapDetailsScreen.swift
```diff
@@ -0,0 +1,344 @@
...
+            UIApplication.shared.isIdleTimerDisabled = true
+            #endif
+            detailsViewModel.load(
+                initialFromCoin: fromCoin,
+                initialToCoin: toCoin,
+                vault: vault,
+                tx: tx
+            )
+            if let fromCoin {
+                tx.fromCoin = fromCoin
```
johnnyluo: **[Issue 2] Double-trigger race on load — redundant `tx.fromCoin` assignment**

`detailsViewModel.load(initialFromCoin:...)` already sets `tx.fromCoin` internally. The subsequent `tx.fromCoin = fromCoin` immediately after will fire `onChange(of: tx.fromCoin)` → `updateFromCoin()` → `fetchQuotes()`, producing a redundant racing quote request on every screen appearance.

Remove this manual assignment and rely solely on `load()` to initialise the coin.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapKeysignScreen.swift
```diff
@@ -0,0 +1,59 @@
...
+                progressLink: SwapCryptoLogic.progressLink(tx: tx, hash: hash)
+            ))
+        }
+        .onChange(of: viewModel.pendingRetryReason) { _, reason in
+            guard let reason else { return }
+            tx.pendingRetryReason = reason
+            viewModel.pendingRetryReason = nil
+            // Stack: details -> verify -> pair -> keysign. Pop pair + keysign.
+            let popCount = min(2, router.navPath.count)
+            router.navPath.removeLast(popCount)
```
johnnyluo: **[Issue 3] Fragile hard-coded pop count**

This assumes the verify screen is always exactly 2 levels back in the nav stack. If the entry point ever changes (e.g. a deep-link adds an extra route before `.root`), this silently pops to the wrong screen.

Consider using a named-route pop (`router.popToRoute(.verify(...))`) or at minimum asserting the route type before removing entries.

VultisigApp/VultisigApp/Features/Swap/Views/Navigation/SwapRouteBuilder.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  SwapRouteBuilder.swift
+//  VultisigApp
+//
+
+import SwiftUI
+
+struct SwapRouteBuilder {
```
johnnyluo: **[Issue 4] `SwapRouteBuilder` is a redundant indirection layer**

`SwapRouter.build(_:)` does nothing but instantiate a `SwapRouteBuilder()` and call one of its five methods, adding an allocation and an extra hop with no observable benefit. The five `@ViewBuilder` methods could live directly on `SwapRouter`, matching the pattern used by `HomeRouteBuilder` and `VaultRouteBuilder`.

Consider removing `SwapRouteBuilder` unless there's a concrete reason to swap implementations.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapDetailsScreen.swift
```diff
@@ -0,0 +1,344 @@
...
+    @StateObject var referredViewModel = ReferredViewModel()
+    @StateObject var keyboardObserver = KeyboardObserver()
+
+    @State var buttonRotated = false
+    @State var showErrorTooltip = false
+
+    @EnvironmentObject var coinSelectionViewModel: CoinSelectionViewModel
+    @Environment(\.router) var router
+
+    let timer = Timer.publish(every: 1, on: .main, in: .common).autoconnect()
```
johnnyluo: **[Issue 5] Timer and `refreshData` logic duplicated across two screens**

Both `SwapDetailsScreen` and `SwapVerifyScreen` declare an identical `timer` and call near-identical `updateTimer` / `refreshData` paths. If the quote refresh interval or logic changes, it must be updated in two places.

Consider extracting a shared view modifier or moving the refresh logic entirely into a ViewModel helper to avoid the duplication.

VultisigApp/VultisigApp/Features/Swap/Views/Navigation/SwapRoute.swift
johnnyluo: @gastonm5 , instead of passing `Vault` , we probably should pass vault.publicKeyECDSA , and then we fetch it from datastore on the new screen

By Passing the @Model object around , which expose us to some random swiftdata issues , which usually cause the app to crash 

For `Coin` we can pass `Coin.id`  , can you please explore this direction?

josephjamesmorris700-hub (approved):

VultisigApp/VultisigApp/Features/Swap/Views/Navigation/SwapRoute.swift
gastonm5: Addressed in 9fe4bb2 — `SwapRoute` now carries identifiers (`vaultPubKeyECDSA: String`, `fromCoinID/toCoinID: String?`) instead of `Vault`/`Coin` (`SwapRoute.swift` L11-17). The live `@Model` objects are re-fetched from `SwiftData` inside `SwapRouter.build(_:)` before being handed to each screen, mirroring the Send flow.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapDetailsScreen.swift
gastonm5: Addressed in 6a08656 — the redundant `tx.fromCoin = fromCoin` after `detailsViewModel.load(...)` is gone. `SwapDetailsScreen.swift` L77-86 now relies solely on `load(...)` to seed `tx.fromCoin`, with an inline comment documenting why re-assigning would trigger the racing quote fetch.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapKeysignScreen.swift
gastonm5: Addressed in 6a08656 — `SwapKeysignScreen.swift` L47-57 no longer pops a fixed count. The retry handler now calls `router.navigateBack { destination in if case .verify = destination as? SwapRoute { return true } else { return false } }`, which pops to the verify screen by route identity and is safe against deep-link routes inserted before `.root`.

VultisigApp/VultisigApp/Features/Swap/Views/Navigation/SwapRouteBuilder.swift
gastonm5: Addressed in 6a08656 — `SwapRouteBuilder.swift` was deleted. The five `@ViewBuilder` methods now live directly on `SwapRouter` (`SwapRouter.swift`), matching the `HomeRouteBuilder` / `VaultRouteBuilder` pattern referenced in the comment.

VultisigApp/VultisigApp/Features/Swap/Views/Screens/SwapDetailsScreen.swift
gastonm5: Addressed in 6a08656 — the timer/refresh orchestration is extracted into `SwapRefreshTickModifier.swift` and exposed as `View.swapRefreshTick { ... }`. `SwapDetailsScreen.swift` L95-97 and `SwapVerifyScreen.swift` L35-39 both consume the modifier, so the refresh cadence has a single source of truth now.

johnnyluo (approved):
LGTM

PR #4342, opened 2026-05-12
fix(thorchain): restore sRUJI to its on-chain denom

gastonm5:
## Summary
- Reverts `TokensStore.sruji.contractAddress` to `x/staking-x/ruji` (the actual on-chain denom — sRUJI never moved, despite PR #3837 renaming it locally to mirror sTCY).
- Drops the `apiDenom(for:)` redirect in `THORBalanceExtension` (introduced by PR #4154 as a symptom-level patch) — contract address now equals denom again, no special-case needed.
- Flips the RUJI entry in `CoinService.migrateOldContractAddresses` + `HiddenToken.migratedContractAddresses` so vaults that already absorbed the stale rename rewrite back to the on-chain denom on next launch. sTCY entries untouched (that one is a real chain-side move).
- Updates the `THORChainTokenMetadataFactory.create(asset:)` dead branch to key on the real denom, so a freshly-discovered sRUJI balance maps to `sRUJI` / `sruji` instead of falling through to the raw denom string.
- All non-source-of-truth references go through `TokensStore.sruji.contractAddress` — the literal `"x/staking-x/ruji"` exists in exactly one place now.

## Why the previous fix wasn't enough
The `apiDenom` redirect made the user-toggled `TokensStore.sruji` row resolve its balance, but auto-discovery (`ThorchainService.fetchTokens` → `CoinService.addDiscoveredTokens`) builds a fresh `CoinMeta(contractAddress: "x/staking-x/ruji")` on every refresh. `Vault.coin(for:)` keys on contract address, so the discovered CoinMeta didn't match the user's row → `addToChain` inserted a second sRUJI Coin → the balance row visibly clobbers itself on each refresh (issue #4318's "appears briefly, disappears within 1-2 seconds").

## Verification against the chain
- `gateway.liquify.com/chain/thorchain_api/cosmos/bank/v1beta1/denoms_metadata` returns `base: "x/staking-x/ruji", symbol: "sRUJI"`.
- `thor1n7q4ztkmqd7u89g0mxs68eg2uvasa76459kphz` (88.33 sRUJI) reports the balance under the same denom — useful test address.

## Test plan
- [x] `ThorchainSRujiTests` — 7 unit cases (TokensStore pin, balance lookup including a stale-denom negative case, metadata factory mapping, hidden-token matches against both current and stale contracts).
- [ ] Manual: toggle sRUJI off then on from THORChain detail's "Manage tokens" — single row, balance persists across pull-to-refresh.
[...]

PR #4344, opened 2026-05-12
fix(wallet): reduce token list flicker post-swap

### LINKED ISSUE #4337: [BUG] Tokens flickering on and off
# [Fix] Token list flickering display issue [Wallet]

## Problem
Tokens in the wallet are intermittently flickering on and off the screen, creating a poor user experience and making it difficult to view or interact with token balances. This affects users on version 1.36.54 running on macOS.

## Expected Behavior
Token list should display consistently without flickering or disappearing elements.

## Steps to Reproduce
1. Open the Vultisig app
2. Navigate to the wallet/token list view
3. Observe the token display behavior
4. Notice tokens flickering on and off

### Must Do
- [ ] Investigate the token list rendering logic for race conditions or state management issues
- [ ] Fix the underlying cause of the flickering behavior
- [ ] Test token display stability across different wallet states

### Must NOT Do
- Don't change unrelated wallet functionality
- Don't refactor the entire token management system

## Acceptance Criteria
- [ ] `xcodebuild -scheme Vultisig build` succeeds
- [ ] Token list displays consistently without flickering
- [ ] All tokens remain visible and stable during normal wallet operations

## Context
- Reported by **SamYap** (@samyap on Discord) on Discord

cc @realpaaao

gastonm5:
## Summary

Three independent guardrails on the wallet/chain-detail render path, each addressing one hypothesis from the [#4337](https://github.com/vultisig/vultisig-ios/issues/4337) investigation. None changes the data model or rendering layer; they collectively cut redundant work along the post-swap refresh path.

1. **Debounce `VaultDetailViewModel.updateBalance`** — 250 ms leading sleep before the heavy fetch. `VaultMainScreen.refresh()` is wired to six triggers and post-swap they stack 2–3 deep within a second; the existing `updateBalanceTask?.cancel()` pattern only cancels the prior task — without a leading sleep each call still fired a full price + per-coin balance refresh.
2. **Drop sync `groupChains()` from `refresh()`** — the synchronous call re-sorted on stale cached balances and produced a visible stale-then-fresh reorder when the async update landed. `updateBalance` now also seeds `chains` synchronously on first call so the initial render isn't blank through the debounce window.
3. **Cache `ChainDetailViewModel.tokens`** — convert from a computed property to a stored `@Published private(set) var tokens: [Coin]`, recomputed on a debounced (100 ms) `vault.objectWillChange`. `BalanceService.applyBalanceUpdates` writes N coins on MainActor per cycle; the prior computed property re-filtered + re-sorted on every property publish, swapping two non-native tokens with similar fiat values on every tick.

## Why these three

Investigation report at [`pages/projects/vultisig/wallet-list-flicker/investigation-4337.md`](https://github.com/vultisig/vultisig-ios/issues/4337) walks the render pipeline + ranks hypotheses. The two highest-confidence causes (refresh fan-out + per-property cascade through `@ObservedObject Vault`) are addressed by (1)+(2) and (3) respectively. Architectural alternatives (e.g. value-type `ChainBalanceSnapshot` projection) are recorded as follow-ups; this PR stays surgical.

## Test plan

- [x] `xcodebuild build` clean
- [x] `VultisigAppTests` — 408/408 pass
- [x] SwiftLint clean for the 3 modified files
- [ ] **Manual repro** of the reporter's scenario: complete a swap, watch the wallet list during the post-tx refresh window. List shouldn't reshuffle.
- [ ] First-launch render: app opens to a non-empty wallet list (the sync seed in `updateBalance` covers this).
- [ ] Pull-to-refresh after a balance change: list reorders once when the fresh data lands, not twice.
- [ ] Chain detail with multiple tokens: opening, refreshing, no visible token reordering between balance ticks.

## Out of scope

- Migrate cells to a narrow value-type projection instead of `@ObservedObject Vault` — recorded in the investigation as a Medium-risk architectural follow-up.
- Audit `TokensStore` entries for chain-vs-store denom mismatches that produce sRUJI-style duplicate row insertion. iOS [#4342](https://github.com/vultisig/vultisig-ios/pull/4342) covers sRUJI specifically.

Fixes #4337.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved balance and token refresh operations to eliminate flickering during updates.
  * Reduced visual reordering that occurred when refreshing wallet data.
  * Optimized handling of rapid refresh requests to prevent unstable list behavior.
  * Enhanced token sorting to maintain consistent ordering during background updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4344)


johnnyluo (approved):
LGTM

PR #4370, opened 2026-05-15
refactor(function-call): contain FunctionCallForm inside FunctionCallDetailsScreen

gastonm5:
## Summary

Drops `@StateObject FunctionCallForm()` from every entry-point screen, route enum, and scanner view. The class survives as a private detail of `FunctionCallDetailsScreen`; nothing outside `Features/FunctionCall/` (or the explicitly out-of-scope `Features/Referral/`) holds an instance.

This is PR-B of the FunctionCallForm elimination plan — a containment refactor, not the full kill. Sub-model rewrites + class deletion are deferred to PR-C.

### Key shape changes

- `FunctionCallDetailsScreen` now owns its own `@StateObject` tx; init drops the `tx` param and falls back to vault native coin / `Coin.example` for `defaultCoin`.
- `FunctionCallRoute.details` drops `sendTx` from its associated values.
- `HomeRoute.vaultAction` drops `sendTx`; `VaultAction.send` gains optional `prefilledToAddress` / `prefilledAmount` / `prefilledMemo` fields so the deeplink + scanner paths carry prefill on the route value instead of mutating a long-lived shared form.
- `HomeScreen`, `ChainDetailScreen`, `CoinDetailScreen`, `DefiChainMainScreen` each drop their `@StateObject FunctionCallForm`.
- `GeneralCodeScannerView` + `MacScannerView` drop the FunctionCallForm parameter entirely (it was already unused inside both views — deeplink prefill flows through `DeeplinkViewModel` -> `HomeScreen.handleSendDeeplinkAfterVaultSelection` -> the new prefilled-route fields).
- `KeygenRoute.macScanner` / `.generalQRImport` drop their `sendTx` parameters; `CreateVaultView` + `ReshareView` (iOS + macOS) drop the placeholder `FunctionCallForm()` threading.
- `UTXOTransactionCell` (currently unused but still in target) switches to a plain `Chain` parameter — it only ever read `tx.coin.chain`.

### Containment check

```
grep -rn FunctionCallForm Features/ \
  | grep -v Features/FunctionCall/ \
  | grep -v Features/Referral/ \
  | grep -v Features/FunctionTransaction/ \
  | grep -v Features/Send/
```
Returns one hit — a comment in `KeygenRouteBuilder.swift`. No type references remain outside the FunctionCall family + Referral (out of scope) + Send (boundary converter).

[...]

PR #4371, opened 2026-05-15
feat(fast-vault): cache eligibility on Vault + drop loadFastVault indirection

gastonm5:
## Summary

Fast-vault eligibility (`FastVaultService.isEligibleForFastSign(vault:)`) was hit per screen mount in Send, Swap, FunctionCall, Referral, and Settings flows — a network round-trip every time. This PR caches the result on the `Vault` SwiftData model and refreshes it only at two trigger points:

1. **App foreground** (`scenePhase` becomes `.active`) — refreshes the currently selected vault if its cache is stale (>24h or never checked).
2. **Vault switch** (`AppViewModel.set(selectedVault:)`) — same.

All consumer reads become synchronous (`vault.fastVaultEligibility`); no hot-path network calls remain.

## Changes

- **New**: `Core/Services/FastVault/FastVaultEligibilityRefresher.swift` — `@MainActor` service, closure-injected (`checkEligibility`, `saveStorage`, `now`), 24h staleness threshold. Two entry points: `refresh(_:)` (force) and `refreshIfStale(_:)` (skip if fresh).
- **`Vault` SwiftData `@Model`** — adds `fastVaultEligibility: Bool` + `fastVaultEligibilityCheckedAt: Date?`. Not in `CodingKeys` (local-only state; doesn't roundtrip with vault import/export). Lightweight SwiftData migration.
- **`AppViewModel.set(selectedVault:)`** — fires async background refresh for the new vault.
- **`AppViewModel.refreshFastVaultEligibilityIfNeeded()`** — called from `VultisigApp.swift` on `scenePhase .active`.
- **`DefaultSendInteractor.loadFastVault`** / **`DefaultSwapInteractor.loadFastVault`** / **`FunctionCallViewModel.loadFastVault`** — return `vault.fastVaultEligibility` (sync wrapped in async to preserve protocol contracts; `async_without_await` suppressed per-line).
- **`SettingsCustomMessageView.onLoad`** — reads `vault.fastVaultEligibility` directly (no more `await fastVaultService.isEligibleForFastSign`).

## Test plan

- [x] 7 new `FastVaultEligibilityRefresherTests`: refresh updates cache + timestamp; can flip true→false; passes vault to check closure; refreshIfStale skips when fresh; runs when expired / never-checked; threshold boundary (>=).
- [x] Full Send + SendPhaseD + Referral regression suite (75 tests total) green.
- [x] SwiftLint clean on all 9 changed files.
- [ ] **Manual smoke required** (the cache's behavior changes user-visible state on first-run / stale-cache edge cases):
  - [ ] Cold start with selected Fast Vault → Send screen loads, `isFastVault` becomes true within seconds (cache populates async).
  - [ ] Cold start with selected Secure Vault → Send screen loads, `isFastVault` stays false.
  - [ ] Background → 24h → foreground → cache refreshes and reflects current server state.
  - [ ] Switch from Fast → Secure vault → next Send screen shows correct fast-vault routing.
  - [ ] First-ever launch with newly imported Fast Vault → first Send screen mount may show `isFastVault = false` until refresh completes (acceptable race; UI re-renders via SwiftData observation when cache flips).

## Known limitations

- `FastVaultService.isEligibleForFastSign(vault:)` swallows network errors and returns `false`. Transient network failures during refresh will cache a `false` value until the next trigger (24h or vault switch). Future work could refactor `exist(pubKeyECDSA:)` to throw on transient errors so the refresher can keep the previous cached value — out of scope here.
- The cache lives on the Vault `@Model`, so SwiftData persists it. Cross-device vault import (via `.vult` file or re-encoded backup) starts with `fastVaultEligibility: false` and re-populates on next refresh trigger — that's by design (the field is local-only state, not a vault property).

## Stack

- This PR branches off **`refactor/referral-form-vm` (PR-A, #4369)**. Inherits PR-A's clean Referral surface so `ReferralDetailsViewModel.loadFastVault` is migrated as part of this PR's `interactor.loadFastVault` change.
- Once PR-A and PR-B (#4370) land, this rebases onto whichever is base at the time.
- Tracked in workstream [[fast-vault-cache/overview]] (wiki).

### DIFF
--- a/VultisigApp/VultisigApp/App/VultisigApp.swift
+++ b/VultisigApp/VultisigApp/App/VultisigApp.swift
@@ -159,6 +159,7 @@ extension VultisigApp {
                 switch scenePhase {
                 case .active:
                     continueLogin()
+                    appViewModel.refreshFastVaultEligibilityIfNeeded()
                 case .background:
                     resetLogin()
                 default:
--- a/VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultEligibilityRefresher.swift
+++ b/VultisigApp/VultisigApp/Core/Services/FastVault/FastVaultEligibilityRefresher.swift
@@ -1,0 +1,72 @@
+//
+//  FastVaultEligibilityRefresher.swift
+//  VultisigApp
+//
+//  Refreshes the cached `fastVaultEligibility` field on the `Vault` model.
+//  Reads are sync (`vault.fastVaultEligibility`); refreshes happen only at
+//  planned trigger points (app foreground, vault switch). Replaces the
+//  pattern where every Send / FunctionCall / Referral / QBTC screen called
+//  `FastVaultService.isEligibleForFastSign(vault:)` on mount.
+//
+
+import Foundation
+import OSLog
+
+@MainActor
+final class FastVaultEligibilityRefresher {
+
+    static let shared = FastVaultEligibilityRefresher()
+
+    private let logger = Logger(subsystem: "com.vultisig.app", category: "fast-vault-cache")
+    private let checkEligibility: @MainActor (Vault) async -> Bool
+    private let saveStorage: @MainActor () -> Void
+    private let now: @MainActor () -> Date
+    private let stalenessThreshold: TimeInterval
+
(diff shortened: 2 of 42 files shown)


johnnyluo:
do you want to update this PR to target for `main` branch?

Issue #4327, opened 2026-05-08
Keysign screen stays stuck after another device signs and broadcasts successfully


realpaaao:
## Problem
A user reports that iOS can remain stuck on the Keysign screen after the transaction was successfully signed and broadcast on another device.

Observed state from the report:
- Screen: `Keysign`
- Status text: `Signing`
- Transaction was already successfully signed and broadcast by another device
- Initiator/device shown in the screenshot does not transition out of signing state

## Expected
When another device completes signing and broadcast for the same keysign session, the iOS device should leave the signing progress state and transition to the appropriate completion/result state, or at minimum detect that the session has completed and stop waiting indefinitely.

## Actual
The iOS device remains stuck on the Keysign `Signing` screen after successful sign/broadcast elsewhere.

## Notes
This looks related to cross-device keysign session completion propagation rather than transaction broadcast itself. Please check whether the initiator device listens for:

- remote signing completion
- successful broadcast by another participant
- keysign session terminal state
- timeout/recovery path when the local device is no longer the broadcaster

## Suggested scope
- Reproduce with multi-device keysign where a non-initiator/cosigner performs successful broadcast.
- Verify the initiator receives and handles the terminal session event.
- Ensure UI state exits `Signing` once another participant completes broadcast.
- Add regression coverage around multi-device completion state handling if practical.

realpaaao:
Additional user detail: switching away from the iOS app and returning resolves the stuck Keysign screen.

That suggests the terminal session/broadcast state may already exist, but the active Keysign view is not refreshing or consuming it until an app lifecycle/foreground event forces a state refresh. Please check foreground refresh handlers versus live in-session state updates.

johnnyluo:
can't reproduce , closing , if we see this again , let raise another issue

PR #4379, opened 2026-05-18
fix(keysign): stage timeout + handleHelperError race + DKLS rethrow + instrumentation (#4327)

gastonm5:
## Summary
Bundles all five recommendations from the [#4327 investigation](https://github.com/vultisig/vultisig-ios/issues/4327) into one PR. Together they remove the "stuck on Signing" failure mode and give us the `os_log` trail to localise the remaining background-suspension stall in a TestFlight repro.

### 1. `handleHelperError` deferred-dispatch race (`KeysignViewModel`)
Drops the `DispatchQueue.main.async { … }` wrapper around the terminal status write. The class is `@MainActor`; the wrapper deferred `status = .KeysignFailed` to the next runloop tick, which lost the race against `tssKeysign`'s post-broadcast `status = .KeysignFinished` guard and left the screen stuck on the "Signing" animation.

### 2. Overall keysign-stage timeout (`KeysignViewModel.startKeysign`)
Wraps the keysign body in a `withThrowingTaskGroup` race against a 90s sleeper. On timeout — and only when status hasn't already reached a terminal state — sets `retryReason = .other("KeysignStalled")` and flips `status` to `KeysignRetryRequested`. This surfaces the existing "Try again" UI and pops the user back to verify. Removes the failure mode where the screen sits on the "Signing" animation indefinitely.

### 3. DKLS retry-exhaustion rethrow (`DKLSKeysign.DKLSKeysignOneMessageWithRetry`)
Replaces the `attempt >= 3` silent return with `throw HelperError.runtimeError(...)`, preserving the underlying error message. Swaps `print` for `logger.error` per `.claude/rules/logging.md` and `.claude/rules/error-handling.md`. The caller's empty-signatures check previously collapsed every cause into a generic "fail to sign transaction" message; it now propagates the real cause.

### 4. Status / txid instrumentation (`KeysignViewModel.setStatus/setTxid`)
Single-funnel helpers that log every transition at `info` level with `#fileID:#line` context. Wraps the terminal-state writes (`KeysignFinished` / `KeysignFailed` / `KeysignRetryRequested`) and the peer-broadcast-race `txid` recovery sites in `handleBroadcastError`. Gives a TestFlight `os_log` capture enough trail to pin which poll loop stalls.

### 5. ScenePhase observability (`KeysignView`)
Logs scene transitions while status is mid-signing. The reporter's description ("switch away and back unblocks the screen") fingerprints iOS auto-resuming a suspended network session; correlating phase transitions with the status trail in TestFlight logs will localise the offending poll loop.

## Why one PR

The five items address different defects in the same code path; shipping them together gives the reporter a build where:
- the screen no longer stays on "Signing" forever (timeout flips into the retry UI),
- a DKLS retry exhaustion no longer silently swallows the cause,
- the deferred-dispatch race can't put the status into an indeterminate state, and
- if the bug still repros somehow, the `os_log` capture will tell us exactly which signal stalled.

Each item is independently reviewable inside the commit pair (`94e517472` = #1, `91217771b` = #2–#5).

## Scope

- Touches three files in `Features/Keysign/` and `Blockchain/States/Keysign/` — does **not** touch the `Blockchain/Tss/` critical boundary. `DKLSKeysignOneMessageWithRetry` is the Swift retry wrapper around the TSS call, not the TSS binding itself.
- 0 SwiftLint violations on changed files. Pre-existing custom `no_raw_urlsession` rule text-matches comments, so the network-session phrasing avoids the word verbatim.
- No localization changes — the timeout path uses the existing generic `broadcastRetryGeneric` key via `BroadcastRetryReason.other("KeysignStalled")`.

## Test plan
- [ ] Build + run on iOS / macOS.
- [ ] Happy-path keysign: signature flows through, `setStatus(.KeysignFinished)` fires, navigation to Done works. Verify `os_log` shows the status trail.
- [ ] Force a `getSignedTransaction` failure (e.g., inject in chain-specific signing): screen transitions to `KeysignFailed` immediately without flashing through `KeysignFinished` or stalling on "Signing".
- [ ] Force a peer-broadcast race (start a co-signed tx, let the peer win): `handleBroadcastError`'s `isAlreadyOnChain` fallback still sets `txid` and navigates to Done. The new `setTxid` log line marks the recovery.
- [ ] Force a keysign stall (interrupt one device's network mid-DKLS, or kill the relay): after ~90s the screen should flip into the retry view, *not* sit on "Signing" indefinitely.
- [ ] Force a DKLS retry exhaustion (block the relay POST so attempts 1-3 all fail): the screen now surfaces a `HelperError.runtimeError` with the underlying cause rather than the generic "fail to sign transaction".
- [ ] Background the app mid-signing and return: `os_log` shows a `scenePhase: ... → background → active` line correlated with the status trail.

## Summary by CodeRabbit

* **New Features**
  * Per-stage timeout for key-signing runs that triggers the retry UI when a stage stalls.
  * Scene-phase logging to correlate app background/foreground transitions with ongoing signing.

* **Bug Fixes**
  * More robust error handling to avoid stale or overwritten signing states and to surface retryable vs final failures.

* **Style / Logging**
  * Improved structured logging for signing attempts and failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4379?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/States/Keysign/DKLSKeysign.swift
+++ b/VultisigApp/VultisigApp/Blockchain/States/Keysign/DKLSKeysign.swift
@@ -387,9 +387,17 @@ final class DKLSKeysign {
                 try await Task.sleep(for: .milliseconds(500))
             }
         } catch {
-            print("Failed to sign message (\(messageToSign)), error: \(error.localizedDescription)")
+            logger.error("Failed to sign message (\(messageToSign, privacy: .public)) on attempt \(attempt, privacy: .public): \(error.localizedDescription, privacy: .public)")
             if attempt < 3 {
                 try await DKLSKeysignOneMessageWithRetry(attempt: attempt+1, messageToSign: messageToSign)
+            } else {
+                // Surface retry-exhaustion as a typed throw rather than silently
+                // returning. Previously the caller only learned about the failure
+                // indirectly via `signatures.isEmpty` further up the stack, which
+                // collapsed every cause into a generic "fail to sign transaction"
+                // message and risked leaving the UI in a half-finished state.
+                // See vultisig-ios#4327.
+                throw HelperError.runtimeError("DKLS sign message exhausted retries: \(error.localizedDescription)")
             }
         }
     }
--- a/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
@@ -884,11 +884,12 @@ class KeysignViewModel: ObservableObject {
         default:
             errMessage = "Failed to get signed transaction,error:\(err.localizedDescription)"
         }
-        // since it failed to get transaction or failed to broadcast , go to failed page
-        DispatchQueue.main.async {
-            self.status = .KeysignFailed
-            self.keysignError = errMessage
-        }
+        // Class is @MainActor — assign directly. A previous `DispatchQueue.main.async`
+        // wrapper raced with `tssKeysign`'s post-broadcast `status = .KeysignFinished`
+        // guard: the dispatched block ran on the next runloop tick, after the guard
+        // had already overwritten the status. See vultisig-ios#4327.
+        self.status = .KeysignFailed
+        self.keysignError = errMessage
     }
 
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4380, opened 2026-05-18
fix(swap): surface vault-resident coins in the picker (#4321)

gastonm5:
## Summary
- `SwapCoinSelectionLogic.fetchCoins(chain:)` previously built its candidate list from the static `TokensStore` native entry plus the external token service. Coins the user had already added to the vault but that the external service didn't know about — TC Secured Assets being the motivating case — never appeared in the picker, so the user couldn't select them as a swap side.
- This PR merges `vault.coins.filter { \$0.chain == chain }` into the candidate list. `uniqueBy { \$0.ticker.lowercased() }` already runs after the merge, so external entries continue to take precedence on ticker collision (external metadata stays canonical).

## Why this is the right scope for #4321

[Issue #4321](https://github.com/vultisig/vultisig-ios/issues/4321) reports that swap **FROM** TC Secured ETH-USDC works but swap **TO** TC Secured ETH-USDC fails. The full investigation lives in the wiki under `pages/projects/vultisig/secured-asset-swap-4321/investigation.md` and concludes that the swap-to-secured direction is most likely a server-side question (THORChain may not honour a swap memo destined for a secured-asset denom — the canonical inbound path is the `SECURE+:<thoraddr>` mint flow, which iOS already has in `FunctionCallSecuredAsset`).

This PR ships the smallest correct iOS-side change for users who already hold the secured asset: surface it in the picker so they can pick it as a swap side. Server-side direction support is tracked separately.

## Scope

- One file touched: `Features/Swap/ViewModels/SwapCoinSelectionViewModel.swift`.
- Two-line addition; no changes to ranking, ticker collision behaviour, or the underlying `TokenSearchService`.
- Does **not** touch the TSS critical boundary or any chain-specific signing path.

## Test plan
- [ ] Build + run on iOS / macOS.
- [ ] Confirm a vault that holds a TC Secured Asset (e.g., ETH-USDC enabled via the withdraw flow) sees it offered in the swap coin selection for the THORChain chain.
- [ ] Confirm swap **FROM** TC Secured ETH-USDC still works end-to-end.
- [ ] Confirm regular ETH-chain ERC20s the user has imported but that aren't in the external token service now appear in the picker (positive side effect).
- [ ] Confirm ticker collisions resolve in favour of the external metadata (existing behaviour preserved).
- [ ] No SwiftLint regressions on the touched file.

## Summary by CodeRabbit

* **New Features**
  * Token selection in Swap now includes all tokens in your vault, providing a more complete list of available trading options.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4380?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCoinSelectionViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCoinSelectionViewModel.swift
@@ -81,9 +81,10 @@ struct SwapCoinSelectionLogic {
     func fetchCoins(chain: Chain) async throws -> [CoinMeta] {
         let nativeToken = TokensStore.TokenSelectionAssets.first { $0.chain == chain && $0.isNativeToken }
 
+        let localTokens = vault.coins.filter { $0.chain == chain }.map { $0.toCoinMeta() }
         // Propagate errors instead of swallowing with try?
         let externalTokens = try await service.loadTokens(for: chain)
-        let tokens = ([nativeToken] + externalTokens).compactMap { $0 }
+        let tokens = ([nativeToken] + externalTokens + localTokens).compactMap { $0 }
         let uniqueTokens = tokens.uniqueBy { $0.ticker.lowercased() }
         return sort(tokens: uniqueTokens)
     }


gomesalexandre (commented):
CI: green. Diff: 1 file, 2 lines added.

The change is exactly what it claims. `fetchCoins(chain:)` previously
built its candidate list from the static `TokensStore` entry plus
external service tokens. Vault-resident coins that the external service
didn't know about (TC Secured Assets being the motivating case) were
invisible in the picker. This PR folds `vault.coins.filter { $0.chain == chain }` into the merge and lets `uniqueBy { $0.ticker.lowercased() }` handle ticker collisions, with external metadata winning on collision since it's prepended before `localTokens`.

**Fund-safety / correctness**

The "can this show coins the user doesn't own?" threat: nope. The filter
is `vault.coins.filter { $0.chain == chain }` — `vault.coins` is the
set of coins the user explicitly added to their vault. If it's in the
vault, the user already holds it or has chosen to track it. There's no
path here to phantom coins appearing.

The "external takes precedence on ticker collision" invariant is
preserved — the local tokens are appended *after* the external list, so
`uniqueBy` keeps the external entry. Good.

The `toCoinMeta()` conversion: worth knowing what that does to the
metadata (icon, contract address) for vault-resident secured assets vs
what the external service would have returned, but this is a display
concern, not a safety one.

**Scope**

The PR description says this is the smallest correct iOS-side fix for
#4321: surface the asset in the picker. The harder question — whether
THORChain will actually honour a `=:ETH.USDC-EC7:<secured-asset-addr>`
memo on the swap-to-secured direction — is correctly deferred as a
server-side open question. Honest scoping, I like it.

**Test plan**

No unit tests touched, but the logic is simple enough that a targeted
integration test would be overkill for a 2-line PR. The manual checklist
in the description covers the happy path (secured asset appears), the
existing FROM path (still works), a new token the user imported (also
appears), and ticker-collision deduplication. That's all you need here.

---

`praise:` clean scope discipline — one file, two lines, well-understood
blast radius. The ticker-collision ordering reasoning in the PR
description is thorough and correct. This is exactly the right size fix
for the issue.

`suggestion:` worth a unit test for `fetchCoins` that seeds a vault with
a coin unknown to the mock external service and asserts it appears in the
result. Not blocking for a 2-liner but would catch regressions if the
merge order ever flips. Could land as a follow-up.

`q:` what does `toCoinMeta()` produce for the `logo`/`icon` URL on a
TC Secured Asset that isn't in the external token service? If the URL is
empty/nil the coin will show up with a blank icon in the picker — minor
UX thing, but worth a glance.

LGTM otherwise. Ship it.

gomesalexandre (approved):
Clean 2-line fix. No phantom coin risk —  is strictly what the user already holds. External metadata ordering is correct (external wins uniqueBy). CI green. Approved.

gomesalexandre (approved):
Clean 2-line fix — vault.coins is strictly what the user already holds. External metadata ordering correct (external wins uniqueBy). CI green. Approved.

PR #4411, opened 2026-05-25
refactor(function-call): C-2a foundation — hoist screen state + add dispatch shell

gastonm5:
## Summary

C-2a is the **foundation** sub-PR of the function-call-submodel-rewrite workstream (1 of 7 stacked sub-PRs: C-2a → C-2g). It is deliberately narrow: hoists shared state out of sub-models into `FunctionCallDetailsScreen` and adds an empty dispatch shell that C-2b → C-2g will populate case-by-case. **No sub-model is touched here.**

The goal is to lay the surface for the per-sub-model `@Observable` rewrites without changing user-visible behaviour. Memo encoding is untouched; every sub-flow continues to render via the legacy `fnCallInstance?.view` dispatch.

## What's in this PR

Per the C-2a tasks spec (steps 1-8 + addendum steps 9-10):

1. Adds `@State private var selectedCoin: Coin`, `@State private var gas: BigInt`, `@State private var isFastVault: Bool` to `FunctionCallDetailsScreen`.
2. Initialises `selectedCoin = defaultCoin` and `isFastVault = vault.isFastVault` in `setData()`.
3. Wires `.onChange(of: selectedCoin)` → `loadGasInfo()` (replaces `.onChange(of: tx.coin)`).
4. Rewrites `loadGasInfo()` to call `BlockChainService.fetchSpecific(tx:)` directly, dual-writing into both the new `@State gas` and `tx.gas` so un-migrated sub-models keep seeing gas through the legacy path.
5. Reads `vault.isFastVault` into the screen-owned `@State isFastVault` (no async loader is needed today; the spec's `interactor.loadFastVault(vault:)` is forward-looking — see "decision calls" below).
6. Adds a **transitional shim** `.onChange(of: tx.coin) { selectedCoin = tx.coin }` so sub-models that still mutate `tx.coin` from internal dropdowns keep the screen's `selectedCoin` in sync until C-2b → C-2e migrate per sub-model.
7. Gas-refresh is verified by the targeted regression suite below.
8. `SendDetailsViewModelTests` + `SendPhaseDRegressionTests` regress green.
9. Adds an **empty `FunctionCallContentView`** dispatch shell at `Features/FunctionCall/Views/FunctionCallContentView.swift` with an exhaustive `switch` over every `FunctionCallInstance` case returning `EmptyView()`. The legacy dispatch still drives the UI; C-2b → C-2e will replace the `EmptyView()` placeholders case-by-case. Adding a new enum case is now a hard compile error here.
10. **No `FunctionCallAddressTextField` touches.** Legacy component stays alive for un-migrated sub-models; C-2g deletes it whole.

## Out of scope (deferred to later sub-PRs)

- No sub-model conversions to `@Observable @MainActor final class` — that's C-2b → C-2e per phase.
- No `FunctionCallAddressable` deletion — C-2g.
- No `getView()` removal — C-2g (and per-sub-model in C-2b → C-2e as views move to typed `XxxFormView` structs).
- No `FunctionCallAddressTextField` migration to canonical `AddressTextField` — C-2b → C-2e per sub-model.
- No `FunctionCallInstance.view: AnyView` deletion — C-2g.
- No `FunctionCallForm.swift` touches — survives until C-2g.

## Decision calls (spec was silent / mistaken)

The spec's step 4 references `interactor.fetchChainSpecific(...)` and step 5 references `interactor.loadFastVault(vault:)` "replacing" `functionCallViewModel.loadFastVault(tx:vault:)`. After audit: no `FunctionCallInteractor` exists yet, `loadFastVault(tx:vault:)` never existed on `FunctionCallViewModel`, and `FunctionCallForm` has no `isFastVault` field. The narrowest reasonable in-scope move:
- Route `loadGasInfo()` through the existing `BlockChainService.fetchSpecific(tx:)` (keeping `tx.coin` mirrored to `selectedCoin` via the dual-write shim so the legacy method works against the new state).
- Initialise `isFastVault` from `vault.isFastVault` directly — no loader needed today; an interactor abstraction can come in a later sub-PR if the spec calls for one.

Both are explicit one-liners in the code comments; happy to revisit if reviewers want a `FunctionCallInteractor` introduced in this PR instead.

## Sequence

This is **1 of 7** stacked sub-PRs in the function-call-submodel-rewrite workstream:
- **C-2a (this PR)** — foundation: hoist screen state + add dispatch shell.
- C-2b — migrate Phase 1 sub-models (Leave, Vote, Unstake, UnbondMaya).
- C-2c — migrate Phase 2 (Custom, Stake, BondMaya) — token-dropdown cross-mutators.
- C-2d — migrate Phase 3 (CosmosIBC, Merge, Unmerge, Switch, ReBond).
- C-2e — migrate Phase 4 (AddThorLP, SecuredAsset, WithdrawSecuredAsset) — heavy.
- C-2f — `TransactionBuilder.buildTransaction()` deletion + `FunctionTransactionScreen` migration.
- C-2g — final deletion (`FunctionCallForm`, `FunctionCallAddressable`, `FunctionCallAddressTextField`, `FunctionCallInstance.view: AnyView`).

## Test plan

- [x] SwiftLint clean on `VultisigApp/Features/FunctionCall/` (0 violations).
- [x] iOS build succeeds (`xcodebuild build -scheme VultisigApp -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max'`).
- [x] Regression: `VultisigAppTests/SendDetailsViewModelTests` + `VultisigAppTests/SendPhaseDRegressionTests` — 40 tests, 0 failures.
- [x] Full `VultisigAppTests` suite — 1064 tests, 0 failures.
- [ ] Manual smoke: open FunctionCallDetailsScreen for each chain, change the token in any sub-flow with a token dropdown (Custom, Stake, AddThorLP), confirm gas refreshes off the new screen state.

## Summary by CodeRabbit

* **New Features**
  * Updated function-call forms with native SwiftUI views for many actions, improving form UX and validation.
  * Added robust address-validation for supported chains.

* **Refactor**
  * Improved chain-specific gas loading and state handling for transaction screens.

* **Localization**
  * Added "Destination Address" and "THORChain Address" translations for multiple languages.

* **Tests**
  * New unit and UI tests covering function-call forms, address validation, and transaction serialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4411?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Features/FunctionCall/Views/FunctionCallDetailsScreen.swift
+++ b/VultisigApp/VultisigApp/Features/FunctionCall/Views/FunctionCallDetailsScreen.swift
@@ -1,6 +1,9 @@
+import BigInt
 import Foundation
 import OSLog
 import SwiftUI
+
+private let logger = Logger(subsystem: "com.vultisig.app", category: "function-call-details-screen")
 
 struct FunctionCallDetailsScreen: View {
     @Environment(\.router) var router
@@ -12,6 +15,14 @@ struct FunctionCallDetailsScreen: View {
     @State private var selectedContractMemoType: FunctionCallContractType = .thorChainMessageDeposit
     @State private var showInvalidFormAlert = false
     @State private var hasCompletedInitialSetup = false
+
+    // C-2a foundation — screen owns active coin / gas / fast-vault flag.
+    // Sub-models still read/write `tx.coin` during the transitional period;
+    // `.onChange(of: tx.coin)` dual-writes `selectedCoin` so the new shape
+    // shadows the legacy form-state until C-2b → C-2e migrate per sub-model.
+    @State private var selectedCoin: Coin = .example
+    @State private var gas: BigInt = .zero
+    @State private var isFastVault: Bool = false
 
     @State var fnCallInstance: FunctionCallInstance?
     let defaultCoin: Coin
@@ -60,9 +71,18 @@ struct FunctionCallDetailsScreen: View {
                 await loadGasInfo()
             }
         }
-        .onChange(of: tx.coin) {
+        .onChange(of: selectedCoin) {
             Task {
                 await loadGasInfo()
+            }
+        }
+        // Transitional shim (C-2a): sub-models still mutate `tx.coin` from
+        // dropdowns. Mirror those writes into the screen-owned
+        // `selectedCoin` so the new state shadows the legacy form-state
+        // until C-2b → C-2e migrate each sub-model to a `@Binding<Coin>`.
+        .onChange(of: tx.coin) {
+            if selectedCoin != tx.coin {
+                selectedCoin = tx.coin
             }
         }
         .onChange(of: selectedFunctionMemoType) {
@@ -251,6 +271,8 @@ private extension FunctionCallDetailsScreen {
         setupForm()
         tx.vault = vault
         tx.coin = defaultCoin
+        selectedCoin = defaultCoin
+        isFastVault = vault.isFastVault
     }
 
     func setupForm() {
@@ -297,6 +319,18 @@ private extension FunctionCallDetailsScreen {
     }
 
     func loadGasInfo() async {
-        await functionCallViewModel.loadGasInfoForSending(tx: tx)
+        // C-2a: screen-owned `selectedCoin` drives the chain-specific fetch.
+        // `tx.coin` is dual-written from `selectedCoin` to keep un-migrated
+        // sub-models in sync until C-2b → C-2e migrate each one.
+        if tx.coin != selectedCoin {
+            tx.coin = selectedCoin
+        }
+        do {
+            let chainSpecific = try await BlockChainService.shared.fetchSpecific(tx: tx)
+            gas = chainSpecific.gas
+            tx.gas = chainSpecific.gas
+        } catch {
+            logger.error("failed to fetch chain-specific data: \(error.localizedDescription, privacy: .public)")
+        }
     }
 }
--- a/VultisigApp/VultisigApp/Features/FunctionCall/Views/FunctionCallDetailsScreen.swift
+++ b/VultisigApp/VultisigApp/Features/FunctionCall/Views/FunctionCallDetailsScreen.swift
@@ -50,8 +50,8 @@ struct FunctionCallDetailsScreen: View {
                     VStack(spacing: 16) {
                         contractSelector
                         functionSelector
-                        if let fnView = fnCallInstance?.view {
-                            fnView
+                        if let instance = fnCallInstance {
+                            FunctionCallContentView(instance: instance, selectedCoin: $selectedCoin)
                         }
                     }
                 }
@@ -91,22 +91,22 @@ struct FunctionCallDetailsScreen: View {
             let currentNodeAddress = extractNodeAddress(from: fnInstance)
             switch selectedFunctionMemoType {
             case .rebond:
-                // Ensure RUNE token is selected for REBOND operations on THORChain
+                // Ensure RUNE token is selected for REBOND operations on THORChain.
+                // Hoisted here per the FunctionCall sub-model rewrite —
+                // ReBond is a pure value-reader, the screen owns the
+                // RUNE-pin so the sub-model can drop its init-time write.
                 ensureRuneCoin()
-                let rebondInstance = FunctionCallReBond(tx: tx, vault: vault)
+                let rebondInstance = FunctionCallReBond()
 
                 if let nodeAddress = currentNodeAddress, !nodeAddress.isEmpty {
                     rebondInstance.nodeAddress = nodeAddress
-                    rebondInstance.nodeAddressValid = Self.validateNodeAddress(nodeAddress)
                 }
 
                 fnCallInstance = .rebond(rebondInstance)
             case .bondMaya:
                 DispatchQueue.main.async {
-                    MayachainService.shared.getDepositAssets {assetsResponse in
-                        let assets = assetsResponse.map {
-                            IdentifiableString(value: $0)
-                        }
+                    MayachainService.shared.getDepositAssets { assetsResponse in
+                        let assets = assetsResponse.map { IdentifiableString(value: $0) }
                         DispatchQueue.main.async {
                             fnCallInstance = .bondMaya(
                                 FunctionCallBondMayaChain(assets: assets)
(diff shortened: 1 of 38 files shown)


gomesalexandre (commented):
> draft PR - posting as COMMENT per protocol (drafts are comment-only)

## overview

C-2a is a narrow, well-scoped foundation slice. The intent is sound: hoist `selectedCoin`/`gas`/`isFastVault` to the screen level, add the dual-write shim so legacy sub-models keep working, and drop in the exhaustive `FunctionCallContentView` switch shell as a compile-time guard against missing cases. The PR description is among the most thorough I've seen in this codebase - kudos on documenting the "spec was wrong / here's what we did instead" decision calls explicitly.

CI: green (build + 1064 tests). CodeRabbit left 2 inline threads that need a response before this exits draft.

---

## correctness walk

**`.onChange` ordering is load-bearing and correct.** The two handlers -
1. `.onChange(of: selectedCoin)` -> triggers `loadGasInfo()`
2. `.onChange(of: tx.coin)` -> mirrors into `selectedCoin` if different

order in `body` matters in SwiftUI (modifiers are applied top-to-bottom, observation order can differ per iOS version). The way they're stacked right now, a sub-model mutating `tx.coin` fires handler (2), which sets `selectedCoin`, which fires handler (1), which re-fetches gas. That's the correct chain. Good.

**`setData()` initialises both `tx.coin` and `selectedCoin` to `defaultCoin`** before the `.onLoad` task fires. Clean. No race between init and first gas fetch.

**`ensureRuneCoin()` / `ensureTCYCoin()` write to `tx.coin` directly.** They're called inside `onChange(of: selectedFunctionMemoType)`. This goes `tx.coin = runeCoin` -> shim fires -> `selectedCoin = runeCoin` -> gas fetch. Works. But there's a subtle wrinkle: `ensureRuneCoin()` writes `tx.coin` unconditionally, which means if `selectedCoin` already equals `runeCoin`, the shim's `if selectedCoin != tx.coin` guard correctly short-circuits the redundant gas fetch. Good defensive coding.

**`FunctionCallContentView` is referenced in the file but unused in the actual rendered `body`.** The screen still renders via `fnCallInstance?.view`. The new view is a dead stub right now - no UI regression risk. The only risk is if someone wires it up prematurely in C-2b without removing the legacy dispatch, but that's a C-2b concern.

**Exhaustive switch over `FunctionCallInstance`.** Correct - 15 cases in the enum, 15 cases in the shell. Adding a 16th case will fail to compile without a corresponding branch here. This is the key compile-time safety win.

---

## issues

**should-fix: unstructured Task dispatch - no cancellation on rapid coin changes**

This is CodeRabbit's inline comment at line 77 and it's valid. Both `.onLoad` and `.onChange(of: selectedCoin)` spawn unstructured `Task { await loadGasInfo() }` with no cancellation. If a user taps a dropdown and changes coin twice quickly, two tasks race to write `gas` and `tx.gas`. Whichever finishes last wins, regardless of which coin is actually selected. The fix is straightforward - store the task in `@State private var gasLoadTask: Task<Void, Never>?`, cancel before spawning, and guard stale completion with `Task.isCancelled` + a snapshot check inside `loadGasInfo`. This is a real correctness bug (stale gas displayed for wrong coin), not just style. Should be fixed before merging even though this is a foundation PR - C-2b and beyond will inherit this pattern.

**q: `@Binding var selectedCoin: Coin` on `FunctionCallContentView` - is it needed today?**

The shell view takes `@Binding var selectedCoin: Coin` but every branch is `EmptyView()`. Nothing reads or writes the binding. If the intent is that future branches will need it, keeping it as a forward-looking API contract is fine - but it should at minimum be documented in a comment, otherwise reviewers in C-2b will wonder why the binding is there. If it turns out some sub-models don't need it, we'll end up threading it through unnecessarily. `suggestion:` either add a comment explaining why the binding lives here now, or defer it until C-2b when the first real branch needs it.

**q: `@State private var gas: BigInt = .zero` is owned by the screen but never read in `body`**

`isFastVault` is also stored but not consumed anywhere in `body` today. These are forward-looking state holders for C-2b+. That's totally fine for a foundation PR, but worth a comment to that effect (or a `// C-2a: consumed in C-2b+` note) so the next reviewer doesn't think it's dead state. The PR description covers this but in-code signal helps.

**suggestion: `private let logger` at file scope vs. inside the type**

Minor style nit - the `logger` is declared as a `private let` at module scope. SwiftUI `View` structs can't have stored instance properties that aren't `@State`/`@Binding`/etc., so putting it at file scope is technically necessary here, but a `// file-scope because View structs can't store logger` comment would preempt the "why is this not on the struct?" question.

---

## CodeRabbit threads - needs author response

Two open inline threads from Mr Rabbit:

1. **line 77 - stale gas race** - valid, see my `should-fix` above. Author should either fix in this PR or explicitly defer with rationale.
2. **line 334 - gas loading in the view** - disagree with CodeRabbit here. The PR description explains clearly that no `FunctionCallInteractor` exists yet and routing through the VM would just be indirection for its own sake. The `loadGasInfo()` on the VM (`loadGasInfoForSending`) still exists for `FunctionTransactionScreen` which still uses it. Moving the new call back through the VM would couple two different screens' gas logic into one method. This architecture makes sense for a transitional PR. `@coderabbitai dis tackled` is the right response here with the rationale from the PR description.

---

## verdict

Solid foundational work. The dual-write shim is well-reasoned, the exhaustive switch is a genuine improvement to maintainability, and the decision call documentation is excellent. The race condition in gas loading is a real bug that should get a fix or at minimum an explicit deferral comment before this exits draft.

Not approving yet (draft + race condition should-fix open), but structure is clean and I don't see anything blocking the overall C-2 workstream direction.

VultisigApp/VultisigApp/Features/FunctionCall/Views/FunctionCallDetailsScreen.swift
```diff
@@ -297,6 +319,18 @@ private extension FunctionCallDetailsScreen {
...
+        if tx.coin != selectedCoin {
+            tx.coin = selectedCoin
+        }
+        do {
+            let chainSpecific = try await BlockChainService.shared.fetchSpecific(tx: tx)
+            gas = chainSpecific.gas
+            tx.gas = chainSpecific.gas
+        } catch {
+            logger.error("failed to fetch chain-specific data: \(error.localizedDescription, privacy: .public)")
+        }
```
gomesalexandre, line 347: @coderabbitai this is a transitional PR - no FunctionCallInteractor exists yet. Moving loadGasInfo() back through FunctionCallViewModel would just be hollow indirection since the VM's existing loadGasInfoForSending is already used by FunctionTransactionScreen. Coupling both screens' gas logic into one VM method is worse than the current split. Deferring the interactor abstraction to a later sub-PR as documented. @coderabbitai dis tackled

gastonm5:
Consolidated into PR #4437.

PR #4413, opened 2026-05-25
fix(polkadot): align methodindex with sdk to 3 (transfer_keep_alive) for paired-vault parity

gomesalexandre:
Closes vultisig/vultiagent-app#851

align iOS polkadot signing to use `methodIndex: 3` (`transfer_keep_alive`) instead of `methodIndex: 0` (`transfer_allow_death`) - matching what the SDK does since sdk#548 landed today.

## what

`Polkadot.swift:74` had `methodIndex: 0` pinned in the custom `callIndices` for native DOT sends on Asset Hub. sdk#548 switched the SDK to method 3. In a 2-of-2 paired vault where iOS is co-signer:

- iOS builds SCALE bytes with method 0
- SDK/RN builds SCALE bytes with method 3
- MPC pre-image hashes diverge -> ceremony aborts (no fund loss, just broken UX)

the fix is a single byte constant: 0 -> 3.

also fixes the stale comment claiming "WalletCore encodes as TransferAllowDeath when assetId=0" - TW actually respects custom callIndices when set (verified against TW 4.6.9 WASM in sdk#548).

## why transfer_keep_alive

`transfer_keep_alive` reverts the tx when sender remaining balance drops below existential deposit (0.01 DOT on Asset Hub). `transfer_allow_death` reaps the account instead. keep_alive is the safer default - no accidental account deletion.

## receipts

build verification deferred to NeO - xcodeproj is XcodeGen-generated and pbxproj is not in the clone. change is a single byte constant at `Polkadot.swift:74`, aligning with the canonical SDK change at vultisig/vultisig-sdk#548 (merged 2026-05-25).

companion PR: vultisig-android#fix_polkadot_method_index_3_keep_alive (same change in `PolkadotHelper.kt:46`)

🤖 Agent-generated


johnnyluo (approved):
LGTM

PR #4418, opened 2026-05-26
fix(swap): SwapKit "no routes" → "amount too small" when pair is supported (#4408)

### LINKED ISSUE #4408: Swap: low amount on SwapKit route shows "No routes available" instead of "amount too small"
### Summary
When a SwapKit-routed swap is below the route's minimum amount, the error toast reads **"Unexpected Error / No routes available for this pair"** — but routes for the pair clearly exist; the amount is just too low.

### Repro
1. Swap **BCH → ETH**
2. Enter a small amount (`0.0115 BCH`, ~$3.98)
3. Toast: title **"Unexpected Error"**, body **"No routes available for this pair"**

BCH→ETH is a common SwapKit pair — routes exist. A larger amount quotes fine.

### Expected
An "amount too small / below minimum" message, not a routing/pair error. The app already has the right copy and enum cases for this — they're just not reached on the SwapKit path:
- `SwapCryptoLogic.Errors.swapAmountTooSmall` (title `swapErrorAmountTooSmallTitle`, desc `swapErrorAmountTooSmallDescription`) — `Features/Swap/ViewModels/SwapCryptoLogic.swift:26`
- `SwapError.swapAmountTooSmall` / `.lessThenMinSwapAmount(amount:)` — `Blockchain/Swaps/Common/SwapError.swift:13-14`

### Probable area
SwapKit appears to return `noRoutesFound` for below-minimum amounts, and the app maps it straight to the route error:
- `Blockchain/Swaps/SwapKit/SwapKitError.swift:60` (`noRoutesFound` case) → `:102` (`swapKitErrorNoRoutesFound` = "No routes available for this pair")
- The thrown `SwapKitError` also isn't classified into `SwapCryptoLogic.Errors`, so the toast title falls back to the generic "Unexpected Error" (`SwapCryptoLogic.swift:32`) — secondary, but worth fixing alongside.

Needs a look at the actual SwapKit `/quote` response for the low-amount case to confirm whether it carries a distinct min-amount signal that should map to `swapAmountTooSmall`.

*(Screenshot of the BCH→ETH toast available — attach to this issue.)*

gastonm5:
## Summary

iOS issue #4408: when a SwapKit-routed swap is below the route's minimum amount, users see a "No routes available for this pair" tooltip wrapped in an "Unexpected Error" title. The pair clearly has routes — the amount is just too low. SwapKit's `/v3/quote` 404 collapses "amount below provider minimum" and "pair not supported by any provider" into one byte-identical envelope, so iOS has been unable to disambiguate.

This PR closes that gap by cross-checking the `/v3/providers` snapshot we already cache: if a non-filtered provider lists both source and destination chains in its `enabledChainIds`, the pair IS structurally supported and the failure must be amount-related. We re-classify the error and let the existing "Amount Too Small" copy (already used by the THORChain path) render.

Full empirical reasoning, the live SwapKit API probe, and the alternatives considered are in the wiki plan: `wiki/pages/projects/vultisig/swapkit-integration/no-routes-disambiguation-plan.md`. The plan was reviewed and the false-positive risk explicitly accepted ("ok if we have some false positives, this is only to improve the UX in most of the cases").

## What's in this PR

- **`SwapKitProviderCache.pairEnabled` / `isPairSupported`** — pure-logic per-pair predicate over the cached providers snapshot. No new wire calls.
- **`SwapKitError.amountBelowProviderMinimum`** — new synthetic case raised only by `SwapKitService.fetchBestRoute` after the cache cross-check. Never decoded from the envelope. Reuses the existing `swapErrorAmountTooSmallDescription` localized string — no new translation cycle.
- **`SwapKitService.fetchBestRoute` interception** — when `/v3/quote` returns `noRoutesFound`, ask the cache whether the pair is supported. If yes, throw `amountBelowProviderMinimum`. If no, throw `noRoutesFound` verbatim. Other documented error codes flow through unchanged.
- **`SwapErrorTooltipView` normalization** — map `SwapKitError.amountBelowProviderMinimum` to `SwapCryptoLogic.Errors.swapAmountTooSmall` so the tooltip title flips from "Unexpected Error" to "Amount Too Small" and the description from "No routes available for this pair" to "The swap amount is too small. Please increase the amount."
- **11 new tests** covering the pair predicate (intersection / split / filtered-providers / unmapped chains / empty-snapshot fail-open) and the service-level reclassification (supported pair → amountBelowProviderMinimum, unsupported pair → noRoutesFound verbatim, other codes → pass-through).

## Why this design

- The disambiguation runs only on the unhappy path — the providers cache is already loaded before `/v3/quote` can return `noRoutesFound`, so the heuristic is zero-cost in the steady state.
- Intersection on a SINGLE non-filtered provider entry, not a union across providers. A provider that enables both chains is far more likely to actually route between them than two providers each handling one side. This narrows the false-positive surface to the rare "every chain individually supported but no single provider connects them" case.
- The new case lands at the SwapKitError layer (not directly at `SwapCryptoLogic.Errors`) so the SwapKit layer stays unaware of the swap feature's error vocabulary. The view layer is where errors are normalized for display today, mirroring the existing pattern.
- Reuses `swapErrorAmountTooSmallTitle` / `swapErrorAmountTooSmallDescription` — same copy THORChain already produces — so no translation work is required.

## False positive scope

If every provider enables both `fromChain` and `toChain` individually but NO single provider has both in its `enabledChainIds`, this fix would falsely classify a genuinely-unsupported pair as "amount too small." For NEAR-Intents' 24-chain coverage today (every Vultisig-supported UTXO + every EVM L2 + Solana + TON + Cardano + Sui + Ripple + Tron + ZCash), this is essentially impossible to hit. The risk has been explicitly accepted by the user.

If the cache is empty and the provider list can't be fetched, the predicate fails open and the reclassification fires — degrading to "Amount Too Small" rather than today's "Unexpected Error" tooltip is still strictly better UX.

## Test plan

- [x] `swiftlint` clean on every changed file
- [x] `xcodebuild build -scheme VultisigApp -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max'` — **BUILD SUCCEEDED**
- [x] New tests pass: `xcodebuild test -only-testing:VultisigAppTests/SwapKitServiceTests -only-testing:VultisigAppTests/SwapKitProviderCacheTests -only-testing:VultisigAppTests/SwapKitErrorTests`
- [x] Full unit suite: 1108 tests pass, 0 failures (baseline 1097 + 11 new)
- [ ] Manual: in a debug build, attempt BCH→ETH at 0.0115 BCH — confirm tooltip title reads "Amount Too Small" with the THORChain-style description, not "Unexpected Error" / "No routes available"
- [ ] Manual: attempt a genuinely unsupported pair (e.g. GAIA.ATOM → DASH.DASH) — confirm the tooltip still reads "No routes available for this pair"

Closes #4408

## Release Notes

* **Bug Fixes**
  * Improved error messaging when swap amounts fall below provider minimums
  * Enhanced detection to distinguish between unsupported chain pairs and insufficient amounts
  * Clearer error messages for swap amount constraints

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4418?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
@@ -23,6 +23,14 @@ enum SwapKitError: Error, LocalizedError, Equatable {
     case swapRouteNotFound
     case outputAmountDeviationTooHigh
     case noRoutesFound
+    /// Disambiguated form of `noRoutesFound` thrown when the cached
+    /// `/v3/providers` snapshot shows at least one non-filtered provider
+    /// enables both source and destination chains — meaning the pair is
+    /// structurally supported, so the 404 must be amount-related rather
+    /// than a pair-coverage gap. The view layer normalizes this to
+    /// `SwapCryptoLogic.Errors.swapAmountTooSmall` so users see the same
+    /// "Amount Too Small" tooltip the THORChain path already produces.
+    case amountBelowProviderMinimum
     case blackListAsset
     case invalidSourceAddress
     case invalidDestinationAddress
@@ -100,6 +108,14 @@ enum SwapKitError: Error, LocalizedError, Equatable {
             return "swapKitErrorOutputAmountDeviationTooHigh".localized
         case .noRoutesFound:
             return "swapKitErrorNoRoutesFound".localized
+        case .amountBelowProviderMinimum:
+            // Reuse the existing THORChain "amount too small" copy rather than
+            // introducing a SwapKit-specific key — the user-facing meaning is
+            // identical and the view layer normalizes this case to
+            // `SwapCryptoLogic.Errors.swapAmountTooSmall` anyway. This
+            // `errorDescription` is only the fallback used if a caller logs
+            // the localized message without going through the tooltip view.
+            return "swapErrorAmountTooSmallDescription".localized
         case .blackListAsset:
             return "swapKitErrorBlackListAsset".localized
         case .invalidSourceAddress:
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitProviderCache.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitProviderCache.swift
@@ -92,6 +92,41 @@ actor SwapKitProviderCache {
         return Self.chainEnabled(chain, in: providers)
     }
 
+    /// Synchronous variant for tests + offline computation. Returns `true`
+    /// iff at least one non-filtered provider entry's `enabledChainIds`
+    /// contains BOTH the source and destination chain ids. The intersection
+    /// requirement on the same provider entry — rather than a union across
+    /// providers — narrows the false-positive surface when re-classifying
+    /// `noRoutesFound` errors as below-minimum amounts: a provider that
+    /// enables both chains is far more likely to actually route between
+    /// them than two providers that each only handle one side.
+    nonisolated static func pairEnabled(
+        fromChain: Chain,
+        toChain: Chain,
+        in providers: [SwapKitProvider]
+    ) -> Bool {
+        let fromId = SwapKitChainIDMapper.swapKitChainId(for: fromChain)
+        let toId = SwapKitChainIDMapper.swapKitChainId(for: toChain)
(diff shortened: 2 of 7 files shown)


johnnyluo (approved):
LGTM

PR #4421, opened 2026-05-26
fix(solana): co-sign dApp raw txs via signSolana, not swap reconstruction

### LINKED ISSUE #4419: Solana: dApp raw-tx co-signing stalls — swapPayload path diverges from signSolana pre-image hash
## Summary
Secure Vault co-signing of Solana dApp transactions (swap aggregators, multi-step routes) never completes. One device stalls polling the relay for messages that never arrive (`[]`), and the other eventually shows "fail to sign transaction". The two devices compute **different pre-image hashes** for the same transaction, so they never agree on the MPC `messageId` / setup message.

## Root cause
A Solana dApp keysign payload carries both:
- `swapPayload` (generic/oneinch) — used for the swap display
- `signData.signSolana.rawTransactions` — the fully-formed dApp transaction to sign

iOS prioritizes `swapPayload` over `signSolana`:
- `KeysignMessageFactory.getKeysignMessages()` — the `.generic` → `.solana` branch calls `SolanaSwaps.getPreSignedImageHash` before the chain-specific `SolanaHelper` fallback.
- `KeysignViewModel.getSignedTransaction()` — same branch ordering.

`SolanaSwaps.getPreSignedInputData` rebuilds the message from the swap quote and **overwrites the recent blockhash** (`v0Message.recentBlockhash = recentBlockHash`) before hashing. The extension/SDK instead prioritizes `signData.signSolana` and signs the raw bytes verbatim (original blockhash). Different blockhash → different message bytes → different pre-image hash. The co-signer's `SchnorrKeysign` setup-message equality check (`signingMsg != messageToSign`) then throws, so it never emits TSS messages and the initiator polls forever.

## Fix
For Solana, prefer the `signSolana` raw path over `swapPayload` whenever `signSolana` is present (the dApp case), in both `KeysignMessageFactory.getKeysignMessages()` and `KeysignViewModel.getSignedTransaction()`. iOS-initiated swaps (no `signSolana`) keep using `SolanaSwaps` unchanged.

Additionally, `SolanaHelper.getPreSignedImageHashForRaw` / `signRawTransaction` derive the message bytes directly from the raw transaction (stripping the `[shortvec(numSigs)][sigs]` envelope) instead of round-tripping through `SolanaSigningInput.rawMessage` + `TransactionCompiler`, which is sensitive to WalletCore version drift across platforms.

## Reproduction
1. Secure Vault (extension + iOS co-signer), Solana.
2. Initiate a dApp swap (raw v0 transaction with address lookup tables) from the browser extension.
3. iOS computes a mismatching pre-image hash → keysign stalls → "fail to sign transaction"; extension polls `/router/message/...` and only ever receives `[]`.

Ehsan-saradar:
## Problem
Secure Vault co-signing of Solana dApp transactions (swap aggregators, multi-step routes) never completes. One device stalls polling `/router/message/...` for messages that never arrive (`[]`); the other eventually shows "fail to sign transaction". The two devices compute **different pre-image hashes** for the same transaction, so they never agree on the MPC `messageId` / setup message.

## Root cause
A Solana dApp keysign payload carries both:
- `swapPayload` (generic) — used for the swap display
- `signData.signSolana.rawTransactions` — the fully-formed dApp transaction

iOS prioritised `swapPayload`: the `.generic`/`.solana` branch in `KeysignMessageFactory.getKeysignMessages()` and `KeysignViewModel.getSignedTransaction()` routed through `SolanaSwaps`, which rebuilds the message from the swap quote and **overwrites the recent blockhash** (`v0Message.recentBlockhash = recentBlockHash`) before hashing. The extension/SDK instead prioritises `signData.signSolana` and signs the raw bytes verbatim (original blockhash). Different blockhash → different message bytes → different pre-image hash → the co-signer's `SchnorrKeysign` setup-message equality check throws, so it never emits TSS messages.

## Fix
- For Solana, prefer the `signSolana` raw path over the swap payload whenever `signSolana` is present (the dApp case), in both `getKeysignMessages()` and `getSignedTransaction()`. iOS-initiated swaps (no `signSolana`) keep using `SolanaSwaps` unchanged.
- `SolanaHelper.getPreSignedImageHashForRaw` / `signRawTransaction` now derive the message bytes directly from the raw transaction (stripping the `[shortvec(numSigs)][sigs]` envelope) instead of round-tripping through `SolanaSigningInput.rawMessage` + `TransactionCompiler`, which is sensitive to WalletCore version drift across platforms.

## Testing

TX: https://solscan.io/tx/34JgNVhew1aicrRuDHceyCG8nCorEteHp4TGDAucNkicNzoJ5DEapvgv72zZ5DsME9Hceo4ToTWitXXZzDf39uTo

Verified end-to-end: a dApp Solana swap (raw v0 tx with address lookup tables) now co-signs through to broadcast on a 2-of-2 vault, where it previously hung.

Fixes #4419

Also resolves the co-signing hang tracked in vultisig/vultisig-windows#3941

## Summary by CodeRabbit

* **Refactor**
  * Solana raw transactions are now processed directly: the transaction message is extracted, the MPC signature is derived and verified, then spliced into the transaction; the finalized signed payload and its hash are produced.
  * Keysign flow for generic Solana swaps now uses the direct Solana signing path for pre-image/hash generation and final transaction assembly.

* **Bug Fix**
  * Raw signing now rejects requests containing anything other than exactly one raw transaction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4421?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessageFactory.swift
```diff
@@ -43,8 +43,21 @@ struct KeysignMessageFactory {
...
+                    // dApp Solana swaps (e.g. via the browser extension) carry
+                    // the fully-formed transaction in signData.signSolana. Sign
+                    // those raw bytes directly to stay byte-identical with the
+                    // extension/SDK, which prioritises signSolana over the swap
+                    // payload. Reconstructing from the swap quote here rewrites
+                    // the blockhash and produces a different pre-image hash,
+                    // breaking Secure Vault co-signing. The swap payload still
+                    // drives the swap display; it just must not drive signing
+                    // when raw bytes are present.
+                    if payload.signSolana != nil {
```
gastonm5: Can we avoid looking into `signSolana` if there is a `swapPayload` @Ehsan-saradar ?
I think this is a code smell and could lead to not know which path to take in the future if something gets changed. 
Also let's try to unify these `ifs` as they're duplicated. We can put it in `SolanaHeper` inside some `signRaw` method or smt like that

gastonm5 (dismissed):

johnnyluo (approved):
LGTM

PR #4435, opened 2026-05-27
fix(swap): scale SwapKit inbound fee by native decimals, not sell-token decimals

aminsato:
## Description

`SwapKitService.inboundFee` scaled the source-chain inbound fee via `fromCoin.raw(for:)` = `value × 10^fromCoin.decimals`. But the inbound fee is denominated in the source chain's **native gas coin** (e.g. `ETH.ETH`, `SOL.SOL`), never the sell token. For an **ERC-20 / SPL token source** the sell-token decimals differ from the native coin's (USDC 6 vs ETH 18), so the fee was **under-counted by 10^12** — it reads as dust on the Network Fee row and weakens the `balanceError` gas-coin sufficiency check. Native-source routes were unaffected because the decimals already matched.

The fix scales by the source chain's native decimals via a new `nativeDecimals(for:)` switch (EVM 18; SOL/SUI/TON 9; TRON/ADA/XRP/ATOM 6; BTC/BCH/LTC/DOGE/DASH/ZEC 8). On a native-source route this equals `fromCoin.decimals`, so the existing EVM / Solana / Tron pins are unchanged.

Surfaced during an iOS↔Android SwapKit parity review (Android PR vultisig/vultisig-android#4619 already scales by native decimals; iOS was the outlier).

## Which feature is affected?
- [x] Swap — SwapKit source-chain (inbound) fee shown as Network Fee + used in the gas-coin balance check

## Checklist

- [x] My changes generate no new warnings (SwiftLint clean on both changed files)
- [x] I have added tests that prove my fix is effective (USDC source / native-ETH inbound fee → scales by 18, not 6)

## Additional context

Behavior is display + balance-check only; the inbound fee does not modify the signed tx (SwapKit pre-builds it with the fee baked in). Only the scaling base changed.

I could not run the full Xcode test suite in this environment — relying on CI for the test run. SwiftLint passed on both files, and the scaling mirrors the existing `Coin.raw(for:)` idiom (`pow` + `NSDecimalRound(.up)` + `BigInt(_:)`).

- **Agent**: Claude Code
- **Issue**: none (found via cross-platform parity review)
- **Confidence**: high
- **Needs human review for**: the `nativeDecimals(for:)` per-chain values (verify ADA/XRP/ATOM = 6, TON/SUI = 9)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inbound fee calculation for swaps so fees are scaled using the source chain’s native gas-coin decimals, ensuring correct fee amounts when swapping non-native tokens.

* **Tests**
  * Added a regression test that verifies correct fee scaling for non-native token swaps to prevent regressions and ensure consistent fee behavior across chains.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4435?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitService.swift
@@ -199,7 +199,15 @@ extension SwapKitService {
         else {
             return nil
         }
-        let raw = fromCoin.raw(for: amount)
+        // The inbound fee is denominated in the source chain's NATIVE gas coin (e.g. "ETH.ETH",
+        // "SOL.SOL"), never the sell token. Scaling by `fromCoin.decimals` under-counts the fee
+        // for an ERC-20 / SPL token source (e.g. USDC's 6 vs ETH's 18 — off by 10^12), which reads
+        // as dust on the Network Fee row and weakens the gas-coin balance check. Scale by the
+        // chain's native decimals — identical to `fromCoin.decimals` on a native-source route.
+        var scaled = amount * pow(Decimal(10), nativeDecimals(for: fromCoin.chain))
+        var rounded = Decimal()
+        NSDecimalRound(&rounded, &scaled, 0, .up)
+        let raw = BigInt(rounded.description) ?? .zero
         return raw == .zero ? nil : raw
     }
 }
@@ -247,6 +255,20 @@ private extension SwapKitService {
         case .thorChain, .thorChainChainnet, .thorChainStagenet: return "THOR"
         case .mayaChain: return "MAYA"
         default: return fallback
+        }
+    }
+
+    /// Native gas-coin decimals for a SwapKit source chain. The inbound fee is always denominated
+    /// in the source chain's native coin, so it must be scaled by these — not the sell token's
+    /// decimals (which differ for an ERC-20 / SPL token source). Equal to the source coin's own
+    /// decimals on a native-source route.
+    func nativeDecimals(for chain: Chain) -> Int {
+        switch chain {
+        case .solana, .sui, .ton: return 9
+        case .tron, .cardano, .ripple, .gaiaChain: return 6
+        case .bitcoin, .bitcoinCash, .litecoin, .dogecoin, .dash, .zcash: return 8
+        // EVM chains (ETH / BASE / OP / ARB / AVAX / BSC / POL).
+        default: return 18
         }
     }
 }
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitService.swift
@@ -203,8 +203,8 @@ extension SwapKitService {
         // "SOL.SOL"), never the sell token. Scaling by `fromCoin.decimals` under-counts the fee
         // for an ERC-20 / SPL token source (e.g. USDC's 6 vs ETH's 18 — off by 10^12), which reads
         // as dust on the Network Fee row and weakens the gas-coin balance check. Scale by the
-        // chain's native decimals — identical to `fromCoin.decimals` on a native-source route.
-        var scaled = amount * pow(Decimal(10), nativeDecimals(for: fromCoin.chain))
+        // native coin's decimals — identical to `fromCoin.decimals` on a native-source route.
+        var scaled = amount * pow(Decimal(10), nativeDecimals(for: fromCoin))
         var rounded = Decimal()
         NSDecimalRound(&rounded, &scaled, 0, .up)
         let raw = BigInt(rounded.description) ?? .zero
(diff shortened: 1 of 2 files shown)


gomesalexandre (approved):
## lgtm 🫡

stamped — verified the under-counting is real and matters more than the PR body lets on. pre-fix, an ERC-20 / SPL source path scales the inbound fee by sell-token decimals (USDC=6), then [`SwapCryptoLogic.balanceError`](https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCryptoLogic.swift#L528) reads it back via `feeCoin.decimal(for:)` which divides by **native** decimals (ETH=18). so the gas-coin check effectively went from "need 0.00009884 ETH" to "need 9.9e-17 ETH" → the sufficiency check basically never fired for ERC-20 sources. user could submit a swap with no ETH at all and watch it dust on-chain. not just a cosmetic Network Fee row issue.

### what I checked

- swap chains that include `.swapkit` per [`Coin+Swaps.naturalSwapProviders`](https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Core/Services/Actions/Coin+Swaps.swift#L72-L202) → universe is closed and every chain in that list maps to the correct native decimals via the new switch:
  - EVM (ETH/BASE/OP/ARB/AVAX/BSC/POL/MANTLE/ZKSYNC/BLAST) → 18 ✓ (default branch)
  - SOL/SUI/TON → 9 ✓
  - TRON/ADA/XRP → 6 ✓
  - BTC/BCH/LTC/DOGE/DASH/ZEC → 8 ✓
- test pin math: `0.000098846611703085 × 10^18 = 98_846_611_703_085` ✓ exact
- consumer wiring still works end-to-end ([`SwapCryptoLogic.fee` L77](https://github.com/vultisig/vultisig-ios/blob/main/VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapCryptoLogic.swift#L77) → `balanceError` → Network Fee row)
- signing path is genuinely untouched — `SwapKitSwapResponse.tx.value`/`.gas` already carry the fee baked in, this BigInt is read-side only as you said

### nits (non-blocking)

- **q:** the new test comment says `(98_846_611 instead of the real wei)` but pre-fix `Coin.raw(for:)` would actually produce `BigInt(99)` (`0.000098846611703085 × 10^6 = 98.846… → NSDecimalRound .up → 99`), not 98_846_611. order-of-magnitude story still holds, just the specific number is off and the gap between 98_846_611 and 98_846_611_703_085 is 10^6, not the 10^12 the same comment cites. worth tightening so future-you doesn't chase that number. [SwapKitServiceInboundFeeTests.swift:64](https://github.com/vultisig/vultisig-ios/blob/aminsato/swapkit_inbound_fee_native_decimals/VultisigApp/VultisigAppTests/Swap/SwapKit/SwapKitServiceInboundFeeTests.swift#L64)
- **suggestion (follow-up bead, not for this PR):** there's already a `CosmosCoinFinder.feeDecimals(for chain:)` parallel table and now this `nativeDecimals(for chain:)` — both are basically "Chain → native gas decimals" lookups. worth hoisting onto `Chain` itself eventually so polkadot (10) doesn't silently fall into the `default: 18` branch if `.swapkit` ever lands there.

otherwise ship it 🫡

VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitService.swift
```diff
@@ -264,4 +257,18 @@ private extension SwapKitService {
         default: return fallback
         }
     }
+
+    /// Native gas-coin decimals for a SwapKit source chain. The inbound fee is always denominated
+    /// in the source chain's native coin, so it must be scaled by these — not the sell token's
+    /// decimals (which differ for an ERC-20 / SPL token source). Equal to the source coin's own
+    /// decimals on a native-source route.
+    func nativeDecimals(for chain: Chain) -> Int {
```
gastonm5: No need to hardcode decimals, we can get them from `Coin`. You can fetch native coin and pass it to the method if needed

gastonm5 (changes requested):
Thanks @aminsato , small comment, please check coderabbit comment also

aminsato:
Addressed in b4efaabf4:

- **@gastonm5** — dropped the hardcoded `nativeDecimals(for:)` switch. It now returns `coin.decimals` for a native source and otherwise looks up the chain's native coin from `TokensStore.TokenSelectionAssets` and uses *its* decimals. No more per-chain table to maintain.
- **CodeRabbit (Kujira)** — moot with the above: KUJI (6dp) now resolves from the store, so it scales by 10^6 correctly instead of falling into the old `default: 18`. Same for any future chain (e.g. Polkadot=10) whose native coin lives in the store.
- **@gomesalexandre nit** — corrected the test comment math: pre-fix value is `BigInt(99)` (round-up of `0.000098846611703085 × 10^6`), and the scaling difference is the 10^12 factor between USDC's 6 and ETH's 18. Assertion unchanged.

SwiftLint clean on both files.

gomesalexandre (approved):
## stamped again 🫡

re-reviewed at b4efaabf. new commit replaces the hardcoded switch with a `TokensStore.TokenSelectionAssets.first { $0.chain == coin.chain && $0.isNativeToken }` lookup, which threads both needles in one go — addresses [@gastonm5](https://github.com/gastonm5)'s "don't hardcode, fetch from the Coin" and CodeRabbit's `kujira` finding (KUJI=6, which the old switch missed and would have scaled by 10^18 instead of 10^6). also corrects my prior nit on the test comment.

### what I re-checked

verified every swapkit-eligible source chain has a native entry with the right decimals in TokensStore:

| chain         | native decimals |
|---------------|-----------------|
| solana/sui/ton | 9 ✓            |
| tron/cardano/ripple/kujira/gaiaChain | 6 ✓ |
| bitcoin/bitcoinCash/litecoin/dogecoin/dash/zcash | 8 ✓ |
| ethereum/avalanche/base/bscChain/arbitrum/optimism/polygon/blast/zksync/mantle | 18 ✓ |

so for every selectable swap source today, the lookup hits and returns the right value. fallback to `coin.decimals` is only reached when the native isn't in TokensStore, which can't happen for any selectable chain.

### latent gap (non-blocking, calling it out for the record)

- `polygonV2` has **no** native entry in TokensStore (it's the rebrand-target case per `Chain.removedChainMigrations["polygon": .polygonV2]`). on a non-native polygonV2 source, the lookup misses and falls back to `coin.decimals` → if a user could pick USDC on polygonV2 as a swap source, the fee would regress to the pre-fix under-counting by 10^12. **today this is dead code** because [`Chain.isSwapAvailable`](https://github.com/vultisig/vultisig-ios/blob/aminsato/swapkit_inbound_fee_native_decimals/VultisigApp/VultisigApp/Model/Chain.swift#L535) returns `false` for `.polygonV2`, and `SwapCoinPickerView.availableChains` filters by `isSwapAvailable` → user literally can't pick it. but the moment that flag flips for the matic→pol cutover, the bug silently reactivates. follow-up bead: add a POL native entry under `.polygonV2` in TokensStore (or alias polygonV2→polygon in this lookup) before flipping `isSwapAvailable`. would also catch this in CI if the regression test gained a polygonV2 case.

otherwise lgtm, ship it 🫡

(my prior approval got overridden by @gastonm5's CHANGES_REQUESTED, so it'll still need their stamp to clear — flagging.)

gastonm5 (approved):
lgtm

johnnyluo (approved):
LGTM

PR #4436, opened 2026-05-27
fix(keysign): dispatch customMessageSignature on key type for EdDSA parity

### LINKED ISSUE #4434: Custom message signing: EdDSA signature display differs between iOS and Android/Extension
Follow-up from the review of vultisig/vultisig-android#4646 (which closes vultisig/vultisig-android#4624). Tracking the cross-platform divergence the author acknowledged at https://github.com/vultisig/vultisig-android/pull/4646#discussion_r3304874667

## Problem

PR vultisig/vultisig-android#4646 made Android dispatch the displayed custom-message signature on key type: ECDSA → `r + s + recoveryId` (65 bytes), EdDSA → `reversed(r) + reversed(s)` (64 bytes). This matches the **Extension** for both formats and matches **iOS** for ECDSA.

iOS does **not** dispatch — `KeysignViewModel.customMessageSignature()` calls `getSignatureWithRecoveryID()` unconditionally, so iOS renders `r + s + recoveryId` even for EdDSA chains:

```swift
// vultisig-ios .../Features/Keysign/ViewModels/KeysignViewModel.swift:912-920
func customMessageSignature() -> String {
    // currently keysign for custom message is using ETH , and the signature should be get signature with recoveryid
    switch signatures.first?.value.getSignatureWithRecoveryID() {
    case .success(let sig):
        return sig.hexString
    ...
}
```

So after vultisig/vultisig-android#4646 the **Android ↔ iOS** signature for an **EdDSA-chain** custom message still mismatches — the EdDSA case of vultisig/vultisig-android#4624 is not fully resolved. Appending a recovery id to an ed25519 signature is meaningless, so iOS most likely needs to dispatch on key type the way Android and the Extension do (probably an iOS-repo fix).

## Action

- Confirm the canonical displayed encoding for EdDSA custom messages across platforms.
- If iOS is the outlier, file/fix on the iOS side so `customMessageSignature()` dispatches on key type.
- Narrow the KDoc in `CustomMessageSignatureFormatTest.kt`, which currently implies full iOS parity — it holds only for ECDSA.

Refs: vultisig/vultisig-android#4624, vultisig/vultisig-android#4646

johnnyluo:
## Summary

- `customMessageSignature()` was calling `getSignatureWithRecoveryID()` unconditionally, producing `r + s + recoveryId` (65 bytes) for all chains including EdDSA
- EdDSA signatures should be `reversed(r) + reversed(s)` (64 bytes) — `getSignature()` already implements this correctly
- Added a `switch keysignType` to dispatch: EdDSA → `getSignature()`, ECDSA/MLDSA → `getSignatureWithRecoveryID()`
- This matches Android and Extension behavior post vultisig/vultisig-android#4646

## Test plan

- [ ] Custom message sign on an ECDSA chain (e.g. ETH) — signature should still be 65 bytes `r + s + recoveryId`
- [ ] Custom message sign on an EdDSA chain (e.g. Solana/Polkadot) — signature should now be 64 bytes `reversed(r) + reversed(s)`, matching Android/Extension

Closes #4434

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom message signature generation now returns algorithm-appropriate signatures for ECDSA, EdDSA, and MLDSA, with an empty fallback when extraction fails.
  * Header selection for certain relay API endpoints now sources the message_id from the intended override field.

* **Tests**
  * Added tests validating signature outputs and edge cases for ECDSA, EdDSA, and MLDSA.
  * Added tests verifying relay API header behavior across endpoints and ensuring JSON content-type is always set.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4436?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gastonm5 (approved):
Let's add a test also

PR #4440, opened 2026-05-28
fix(keysign): bind `toToken` image properly in keysign animation

### LINKED ISSUE #4439: Keysign animation: `toToken` binding doesn't set destination image
## Summary
On the keysign screen, the Rive animation's `toToken` property is not correctly bound to the destination token image, so the wrong (or missing) image is shown for the destination token during swap keysign.

## Expected
The `toToken` image bound in the keysign Rive animation matches the destination token of the swap.

## Actual
The destination image bound via `toToken` does not render the correct token image.

## Scope
- File: `VultisigApp/VultisigApp/Resources/Animations/keysign.riv`
- Surface: keysign animation during swap signing.

gastonm5:
## Summary
- Updates the keysign Rive animation (`keysign.riv`) so the `toToken` property correctly drives the destination token image.
- Resolves the wrong/missing destination-token image shown during swap keysign.

Closes #4439

## Test plan
- [ ] Run a swap keysign on device and confirm the destination token image rendered by the keysign animation matches the swap's `toToken`.
- [ ] Verify the source (`fromToken`) image still renders correctly and no other slots regress.
- [ ] Smoke-test a non-swap keysign to ensure the animation still plays without issues.


johnnyluo (approved):
LGTM

PR #4446, opened 2026-05-29
fix(swap): responsive quote fetching — fast start, no phantom gas/stale flash, animated resolve

### LINKED ISSUE #4445: Swap quote fetching: slow to start, phantom "insufficient gas", and stale-value flash while typing
## Summary

Entering a swap amount feels broken: the quote takes a long time to *start* fetching, the UI can briefly flash a previous quote, and a spurious **"insufficient gas"** error appears while typing even with ample balance.

## Symptoms

1. **Slow to start.** After typing an amount, the quote doesn't begin fetching for a noticeable beat.
2. **Phantom "insufficient gas".** While updating the amount, an insufficient-gas error flashes even when the wallet has more than enough.
3. **Stale-value flash.** When the amount changes while a quote is in flight, the to-amount and fees briefly show the *previous* quote's values before updating.
4. **Abrupt updates / late skeleton.** Values pop in without transition, and the loading skeleton is slow to appear.

## Root causes (iOS, `Features/Swap`)

- **Stacked debounce.** `SwapCryptoAmountTextField` debounced keystrokes by **3 seconds** (`DebounceHelper.shared.debounce(delay: 3)`), and `SwapDetailsViewModel.fetchQuotes` added another 200 ms on top — ~3.2 s before a fetch starts. `DebounceHelper.shared` is also a single shared work-item across unrelated fields.
- **Cancellation treated as failure.** Each keystroke cancels the in-flight `updateQuoteTask`. `updateFees` mapped *every* thrown error (including the cancellation) to `Errors.insufficientGas` via its catch-all; `updateQuotes` only ignored `URLError.cancelled`, not Swift `CancellationError`.
- **Cancelled fetch still wrote results.** A superseded task resumed after its `await` and wrote the old quote/gas/fee (no post-`await` cancellation check), and its `defer` cleared the loading skeleton — unmasking stale values.
- **No transition.** To-amount and fees snapped between skeleton and value.

## Expected

Typing an amount should start fetching promptly (single short debounce), never show a phantom gas error or a stale quote, show the loading skeleton instantly, and fade the resolved values/fees in smoothly.

## Notes

Fetch *duration* (multi-provider fan-out latency) is out of scope here — this is about start latency and UI correctness while typing. Follow-up: `DebounceHelper.shared` per-instance cleanup for the Send/FunctionCall fields.

gastonm5:
Closes #4445.

Makes the swap amount → quote flow responsive and correct while typing. Five focused commits:

## 1. Collapse the stacked debounce (`3d1bcc1d8`)
`SwapCryptoAmountTextField` debounced keystrokes by **3 seconds** (`DebounceHelper.shared.debounce(delay: 3)`), and `fetchQuotes` added another 200 ms — ~3.2 s before a fetch started. Removed the text-field debounce (the field now reports keystrokes immediately) and made `fetchQuotes` the single source of truth at **300 ms** (named constant). Loading flags are set synchronously on the leading edge so feedback is immediate.

## 2. Stop cancelled fetches surfacing as "insufficient gas" (`8fd0d24fd`)
Each keystroke cancels the in-flight `updateQuoteTask`. `updateFees`'s catch-all mapped *any* error — including the cancellation — to `Errors.insufficientGas`. Both `updateQuotes` and `updateFees` now ignore cancellation (`CancellationError` / `URLError.cancelled`) and return without writing an error. Genuine shortfalls still surface via the real fee-vs-balance check (unchanged).

## 3. Stop a cancelled quote writing stale to-amount/fees (`39ed49bed`)
A superseded task resumed after its `await` and wrote the old quote/gas/fee, then its `defer` cleared the skeleton — flashing the previous quote. Now: guard `!Task.isCancelled` after each `await` before writing; only the winning task clears the loading flags; the previous quote's outputs are reset on the leading edge.

## 4. Animate to-amount and fees as they resolve (`15f8e8834`)
Crossfade the values in when a quote lands/refreshes instead of snapping.

## 5. Show the skeleton instantly, animate only the resolve (`55d5339ad`)
Directional animation (`isLoadingQuotes ? nil : .easeInOut`) so the skeleton appears instantly when a fetch starts and only the resolved value/fees crossfade in.

## Net behavior
Type an amount → **instant skeleton** → ~300 ms → values **fade in cleanly**, with no phantom gas error and no stale flash.

## Scope / verification
- Touches `Features/Swap` only — no TSS/crypto/signing.
- `build-for-testing` (iPhone 16 / iOS 18.4) green; SwiftLint clean on all touched files.
- No `SwapDetailsViewModel` test exists and the debounce/cancel timing isn't cleanly testable without an injectable scheduler, so no flaky wall-clock test was added.

## Out of scope / follow-ups
- Quote *fetch duration* (multi-provider fan-out latency).
- `DebounceHelper.shared` per-instance cleanup for the Send / FunctionCall fields (it remains a single shared work-item).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading-state animations for swaps so placeholders crossfade to values only after quotes finish loading.
  * Made swap amount inputs more responsive by applying user edits immediately and invoking change handlers per keystroke.
  * Fixed race conditions in quote and fee fetching: centralized debounce timing, clear stale state early, and ignore cancelled intermediate requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4446?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Components/Swap/SwapFromToField.swift
+++ b/VultisigApp/VultisigApp/Components/Swap/SwapFromToField.swift
@@ -123,9 +123,12 @@ struct SwapFromToField: View {
             .disabled(title=="to")
         }
         .redacted(reason: title == "to" && detailsViewModel.isLoadingQuotes ? .placeholder : [])
-        // Crossfade the to-amount between its loading skeleton and the resolved
-        // value instead of snapping when a quote lands or refreshes.
-        .animation(.easeInOut(duration: 0.25), value: detailsViewModel.isLoadingQuotes)
+        // Show the skeleton instantly when a fetch starts (nil animation on the
+        // entering edge); crossfade to the resolved value only when it lands.
+        .animation(
+            detailsViewModel.isLoadingQuotes ? nil : .easeInOut(duration: 0.25),
+            value: detailsViewModel.isLoadingQuotes
+        )
         .animation(.easeInOut(duration: 0.25), value: amount)
     }
 
@@ -136,7 +139,10 @@ struct SwapFromToField: View {
             .frame(maxWidth: .infinity, alignment: .trailing)
             .opacity(isFiatVisible() ? 1 : 0)
             .redacted(reason: title == "to" && detailsViewModel.isLoadingQuotes ? .placeholder : [])
-            .animation(.easeInOut(duration: 0.25), value: detailsViewModel.isLoadingQuotes)
+            .animation(
+                detailsViewModel.isLoadingQuotes ? nil : .easeInOut(duration: 0.25),
+                value: detailsViewModel.isLoadingQuotes
+            )
             .animation(.easeInOut(duration: 0.25), value: fiatAmount)
     }
 
--- a/VultisigApp/VultisigApp/Components/Swap/SwapFromToField.swift
+++ b/VultisigApp/VultisigApp/Components/Swap/SwapFromToField.swift
@@ -123,6 +123,10 @@ struct SwapFromToField: View {
             .disabled(title=="to")
         }
         .redacted(reason: title == "to" && detailsViewModel.isLoadingQuotes ? .placeholder : [])
+        // Crossfade the to-amount between its loading skeleton and the resolved
+        // value instead of snapping when a quote lands or refreshes.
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4448, opened 2026-05-29
fix(send): instant percentage / Max amount fill (optimistic-then-refine)

### LINKED ISSUE #4447: Send: percentage / Max amount buttons have a delay before filling (block on a fee fetch)
## Summary

On the Send details screen, tapping a percentage preset (25 / 50 / 75 / Max) has a visible delay before the amount fills. Typing the amount manually is instant. The delay is a blocking network round-trip.

## Root cause (iOS, `Features/Send`)

`SendDetailsViewModel.setMaxAmount(percentage:)` sets `isLoading = true` and **awaits `interactor.fetchGasAndFee(...)`** before filling the field, because the preset amount is computed as `balance − fee` (`Model/Coin.getMaxValue`). Manual entry just writes the typed value via `convertToFiat` — no fetch. The details screen doesn't otherwise compute a fee (the verify screen does), so this fetch exists solely to subtract the fee from the balance.

The fetch is unnecessary in most cases:
- **Non-native coins (tokens):** the fetched fee is discarded (`maxFee = coin.isNativeToken ? result.fee : .zero`) — max is the full balance. The await accomplishes nothing.
- **Native, partial % (25 / 50 / 75):** the wallet isn't being drained, so reserving the full fee before taking a fraction isn't needed for a responsive fill; the verify screen validates the real fee before signing.
- **Native, Max (100%):** a fee is genuinely needed for correctness, but a *blocking* fetch isn't — the amount can be filled optimistically and refined in the background.

## Expected

Tapping a preset fills the amount instantly (like manual entry), without a blocking spinner. Native-coin Max may settle by the fee a moment later (the fee is amount-dependent for UTXO/Cardano/TON; exact for EVM).

## Notes

The verify-screen fee calculation remains the source of truth for the fee at signing — unchanged.

gastonm5:
Closes #4447.

Tapping a percentage preset (25 / 50 / 75 / Max) on the Send details screen blocked on a fee fetch before filling the amount; manual entry was instant. This makes the fill instant.

## Root cause
`SendDetailsViewModel.setMaxAmount(percentage:)` set `isLoading = true` and awaited `interactor.fetchGasAndFee(...)` to compute `balance − fee`. The details screen doesn't otherwise compute a fee (the verify screen does), so the fetch existed only to subtract the fee — and it's wasted for non-native coins (fee discarded → max = full balance) and unnecessary for partials.

## Change (Option B — optimistic fill + async refine)
`setMaxAmount(percentage:)` is now **synchronous**:
- Cancels any in-flight refine, sets `sendMaxAmount = (percentage == 100)`, fills instantly from balance (`SendCryptoLogic.computeMaxAmount(coin:, fee: .zero)`, scaled by `applyPercentage` for partials), then `convertToFiat`. No interactor call.
- **Only native-coin Max** falls through to a background `startFeeRefine()`: sets the non-blocking `isCalculatingFee` (not `isLoading`), runs the same `fetchGasAndFee` request (`sendMaxAmount: true`, feeMode threaded), then settles `amount = balance − fee`.
- Race guard mirrors `SwapDetailsViewModel`: `cancelFeeRefine()` at the top of each call + `guard !Task.isCancelled` after the await before writing, so a newer preset tap / manual edit can't be clobbered by a stale refine. `CancellationError` returns silently; other errors keep the optimistic balance and log (no field-wiping error).
- Percentage buttons are no longer disabled during the refine (dropped the `isLoading || isCalculatingFee` gate); both call sites drop the `Task { await … }` wrapper.

## Behavior
Every token and every partial preset fills **instantly**. Native-coin Max fills instantly and refines in the background.

**Expected (Option B):** on native UTXO / Cardano / TON, Max visibly settles down by the fee a moment after the fill (the fee is amount-dependent there); on EVM/account-based it's exact.

## Scope / verification
- `Features/Send` only — no TSS/crypto/signing; verify-screen fee calc unchanged.
- `build-for-testing` (iPhone 16 / iOS 18.4) green; SwiftLint clean on touched files.
- `SendDetailsViewModelMaxAmountTests` rewritten (9/9) — optimistic→refine for EVM + UTXO, **token + partial assert the interactor is never called**, race-guard (Max → 50% → 50% wins), refine-error keeps optimistic balance, `sendMaxAmount == (percentage == 100)`. Other Send suites unchanged and passing.

## Wiki
`projects/vultisig/send-percentage-amount-speed/instant-fill-plan.md`

## Summary by CodeRabbit

* **Improvements**
  * Send screen’s Max action now instantly fills amounts and refines fee in the background for native coins, giving faster feedback and smoother UX.
* **Bug Fixes**
  * Refinement failures no longer clear the optimistic amount; cancellation avoids stale overwrite.
* **Tests**
  * Updated test coverage for optimistic fill, background refine, cancellation and error paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-ios/pull/4448?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/VultisigApp/VultisigApp/Components/Send/SendDetailsAmountTab.swift
+++ b/VultisigApp/VultisigApp/Components/Send/SendDetailsAmountTab.swift
@@ -98,13 +98,10 @@ struct SendDetailsAmountTab: View {
 
     @ViewBuilder
     var percentageButtons: some View {
-        let isDisabled = viewModel.isLoading || viewModel.isCalculatingFee
         PercentageButtonsStack(selectedPercentage: $percentage)
-            .opacity(isDisabled ? 0.5 : 1.0)
-            .disabled(isDisabled)
             .onChange(of: percentage) { _, newValue in
                 guard let newValue else { return }
-                Task { await viewModel.setMaxAmount(percentage: newValue) }
+                viewModel.setMaxAmount(percentage: newValue)
             }
     }
 
--- a/VultisigApp/VultisigApp/Components/Send/SendDetailsAmountTextField.swift
+++ b/VultisigApp/VultisigApp/Components/Send/SendDetailsAmountTextField.swift
@@ -95,7 +95,7 @@ struct SendDetailsAmountTextField: View {
         SendCryptoAmountTextField(
             amount: $viewModel.amount,
             onChange: { viewModel.convertToFiat(newValue: $0) },
-            onMaxPressed: { Task { await viewModel.setMaxAmount() } }
+            onMaxPressed: { viewModel.setMaxAmount() }
         )
         .focused($focusedField, equals: .amount)
         .onChange(of: viewModel.coin) { _, _ in
--- a/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
@@ -457,7 +457,11 @@ final class SendDetailsViewModel {
                     feeMode: self.feeMode,
                     fromAddress: self.fromAddress
                 )))
-                guard !Task.isCancelled else { return }
+                // Skip the refine write if the user moved off Max in the
+                // meantime — a manual amount edit flips `sendMaxAmount` via
+                // `convertToFiat` without cancelling this task, so guard on it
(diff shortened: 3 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4454, opened 2026-06-01
fix(send/macOS): camera button opens QR scanner, not a file picker

johnnyluo:
## Problem

On the **macOS Send screen**, tapping the **camera button** on the "Send to" address field opened a **file picker** (`fileImporter`) instead of the camera — so users couldn't scan a QR code with their camera.

## Root cause

A complete macOS QR scanner (`MacAddressScannerView` — camera mode, screen-capture mode, and an "Upload QR code image" fallback) already exists and is wired through `KeygenRoute.macAddressScanner`. But in `AddressFieldAccessoryStack`:

1. The camera button toggled `showImagePicker` (the file importer) on macOS instead of `showScanner`, so the scanner was never reached.
2. Even if it had been reached, result delivery was broken: the `onChange(of: showScanner)` else-branch only runs on a `showScanner` change, but `showScanner` is reset to `false` immediately on navigate — so the branch that reads the scan result never fired after returning from the pushed scanner. This dead path was introduced in the navigation refactor (#3581) and was never exercised because the button pointed at the file picker.

## Fix

- **Camera button now opens the scanner** on both platforms (`showScanner.toggle()`). The redundant macOS `.fileImporter` is removed — file import is still available via the scanner's built-in **"Upload QR code image"** button, and drag-and-drop (`onDrop`) onto the field is unchanged.
- **Result delivery is now reactive**: the view observes `ScannerResultManager` and delivers the parsed address via `onChange(of: scannedAddress)` once the pushed scanner stores its result.
- Removed a stale unused `OSLog` import and the now-dead `showImagePicker`/`selectedImage` state on macOS.

## Testing

- `swiftlint` — 0 violations on the changed file.
- macOS build (`xcodebuild -destination 'generic/platform=macOS'`) — **BUILD SUCCEEDED**.
- iOS path is unchanged (camera button already opened the QR scanner there).

## Summary by CodeRabbit

* **Refactor**
  * Unified address/QR scanning flow across platforms for consistent behavior.
  * Camera accessory now always opens the scanner UI and uses navigation-driven presentation.
  * Non‑iOS now favors drag‑and‑drop for QR input; legacy image‑import flow removed.
  * Redesigned screen‑capture preview, segmented mode switching, and clearer start/stop/back behavior.

* **Chores**
  * Improved capture lifecycle logging and diagnostics.

* **Removed**
  * Legacy macOS camera scanner implementation replaced with the new unified approach.


johnnyluo:
Follow-up: addressed two more issues with the macOS address-field QR scanner reported on this branch.

**1. "Desktop" (screen-capture) mode didn't work** and **2. the layout differed from the main "Scan" screen** (`MacScannerView`).

Root cause: the address scanner (`MacAddressScannerView`) had its own older layout and a duplicate camera view model (`MacAddressScannerViewModel`) with a different screen-capture lifecycle than the working Scan screen.

Fix: rewrote `MacAddressScannerView` to mirror `MacScannerView`:
- Reuses the shared `MacCameraServiceViewModel` + the same camera/screen-capture lifecycle (`onNavigationStackChange` + `handleModeVisible`/`handleModeHidden`), so **Desktop mode** now starts/stops capture identically to the Scan screen.
- **Matching layout**: full-bleed preview with the rounded viewport overlay, centered Camera/Desktop segmented control, Help tooltip, and bottom "Upload QR Code" button.
- Delivers the parsed address via the existing binding and pops with `router.navigateBack()` (the codebase idiom).
- Removed the now-unused `MacAddressScannerViewModel` (duplicated `MacCameraServiceViewModel`).

Verified: SwiftLint clean + macOS `xcodebuild` **BUILD SUCCEEDED**.

VultisigApp/VultisigApp/Components/Forms/AddressFieldAccessoryStack/AddressFieldAccessoryStack.swift
```diff
@@ -121,6 +95,32 @@ private extension AddressFieldAccessoryStack {
 }
 #else
 private extension AddressFieldAccessoryStack {
+    /// The address of the scan result currently pending delivery, if any. Observing
+    /// `scannerResultManager` re-renders this view when the pushed scanner stores a
+    /// result, letting `onChange(of: scannedAddress)` deliver it after we return.
+    var scannedAddress: String? {
+        guard let scannerResultId else { return nil }
+        return scannerResultManager.getResult(for: scannerResultId)?.address
```
gastonm5, line 103: Can we somehow refactor this `scannerResultId` with some binding or callback property?

gastonm5 (changes requested):
Could you check QR upload on send? I can see double navigation on it

Image

VultisigApp/VultisigApp/Components/Forms/AddressFieldAccessoryStack/AddressFieldAccessoryStack.swift
johnnyluo: Unfortunately a plain binding/callback isn't possible here given the navigation architecture. On macOS the scanner isn't presented as a sheet — it's *pushed* via `router.navigate(to: KeygenRoute.macAddressScanner(...))`, and `KeygenRoute` is `Hashable`. A `Hashable` route can't carry a closure or a `Binding`, so there's no way to hand the scanner an `onResult` directly at navigation time.

`scannerResultId` (a `UUID`) is exactly the bridge across that boundary: it's the only thing that *can* travel through the route, and it keys the result that the pushed scanner writes into `ScannerResultManager.shared`. The caller then observes the manager and picks the result back up. So even a binding-based variant (`ScannerResultManager` already exposes `getBinding(for:)`) still needs the `resultId` as the key — it can't be eliminated without changing how the scanner is presented (e.g. moving it to a sheet, which would defeat the purpose of this fix).

Happy to revisit if we move the scanner off the navigation stack, but for the push-based flow the id-keyed manager is the cleanest option.

johnnyluo:
@gastonm5 fixed the double navigation bar on QR upload (Send → address scanner) in 1b6376719.

**Root cause:** the address scanner showed the upload UI by rendering `GeneralQRImportMacView` *inline*. That view carries its own `.crossPlatformToolbar`, so on macOS a second toolbar bar stacked under the scanner's own toolbar — the double navigation in your screenshot. As a side effect, that inline view never invoked its `onParsedAddress` closure (it routes through `deeplinkViewModel.extractParameters`), so an uploaded QR's address wasn't actually delivered back to the field via the callback.

**Fix:** replaced the inline `GeneralQRImportMacView` with a self-contained import (`FileQRCodeImporterMac`) that delivers the parsed address through the existing `handleScan` → `onParsedResult` path, and kept a single scanner toolbar. While in import mode the toolbar shows a back button that returns to the camera scanner instead of a second nav bar.

Verified: SwiftLint clean + macOS `xcodebuild` **BUILD SUCCEEDED**. Could you re-check when you get a chance?

gastonm5 (approved):
LGTM

PR #4460, opened 2026-06-01
fix(swap): restore switch button after error dismiss + raise tooltip above % buttons (macOS)

### LINKED ISSUE #4443: Swap: error message renders unstyled + switch button stops swapping tokens (macOS)
## Description

Two problems on the Swap screen (ETH → BTC):

1. **Error box renders incorrectly.** The "Unexpected Error / Address screening failed — contact support" tooltip shows as a light/white callout that does not match the dark theme — it looks like an unstyled default popover rather than a proper themed error message.
2. **Switch button stops working.** After the error appears, pressing the swap/switch button again (the red circle between From and To) does **not** switch the tokens.

## Platform

- macOS

## Steps to Reproduce

1. Open Swap
2. Set **From** = Ethereum (ETH Native), **To** = Bitcoin (BTC Native)
3. Trigger the address screening error state
4. Observe the error callout styling
5. Press the switch button (circle between From/To) again

## Expected

- The error message renders as a properly themed (dark) error component.
- Pressing the switch button swaps the From/To tokens.

## Actual

- The error message appears as an unstyled light/white popover.
- Pressing the switch button does nothing — tokens are not swapped.

## Screenshot

swap-error-and-switch

gastonm5:
## Problem
Two functional bugs on the Swap screen (reported on macOS, ETH→BTC / RUNE→ETH):

1. **Switch button stops working after an error.** When a swap quote error appears, a red warning icon replaces the switch button and shows a dismissible tooltip. Closing the tooltip (X) only hid the tooltip while leaving the error state set, so the red icon stayed and the switch button never came back — the user could no longer switch From/To tokens.
2. **macOS: error tooltip renders behind the percentage buttons.** The tooltip painted *under* the 25%/50%/75% row, cutting off the error text.

## Fix
`Features/Swap/Views/Screens/SwapDetailsScreen.swift` (6 insertions, 1 file):

1. Clear `detailsViewModel.error = nil` in the tooltip's `onDismissTooltip` closure (alongside the existing `showErrorTooltip = false`). Dismissing the toast now clears the error → the red icon disappears → the switch button is restored. Mirrors how `handleSwapTap()` already clears the error at the screen level.
2. Add `.zIndex(1)` to `swapContent`, `#if os(macOS)`-guarded. On macOS `swapContent` and `percentageButtons` are siblings in the same VStack, so the later-painted buttons covered the tooltip overlay; raising the zIndex layers it above. iOS layout is untouched (there the percentage buttons live in the keyboard toolbar, not as a VStack sibling).

## Out of scope
The issue also mentioned the error box looking "unstyled / light". That's **not** addressed here — the app doesn't manage in-app themes and the tooltip already renders through the design system (`InfoTooltip` → `Theme.colors.bgTooltip`); the light look is a macOS rendering artifact. See the [comment on the issue](https://github.com/vultisig/vultisig-ios/issues/4443#issuecomment-4592314045).

## Testing
- `xcodebuild -scheme VultisigApp -destination 'platform=macOS,arch=arm64'` → **BUILD SUCCEEDED**.
- SwiftLint (repo config) on the changed file → 0 violations.
- ⚠️ **Runtime/manual UI verification still pending** — these are SwiftUI view-state changes confirmed by compile + code analysis, not yet by clicking through a running build. Suggested manual check before merge: trigger the swap error → close the toast → confirm the switch button returns and swaps tokens; on macOS confirm the tooltip renders above the 25/50/75% row.

closes #4443

## Summary by CodeRabbit

* **Bug Fixes**
  * Dismissing the swap error tooltip now clears the underlying error state.
  * Ensured the error tooltip remains visible above neighboring UI elements on macOS.

* **Style**
  * Improved tooltip visuals with rounded transitions between the body and arrow.
  * Standardized tooltip sizing and spacing for consistent behavior across platforms.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Tooltip/TooltipShape.swift
+++ b/VultisigApp/VultisigApp/Components/Tooltip/TooltipShape.swift
@@ -16,6 +16,10 @@ struct TooltipShape: Shape {
     let arrowWidth: CGFloat = 20
     let arrowHeight: CGFloat = 10
     let arrowCornerRadius: CGFloat = 2
+    /// Radius of the rounded junction where the arrow base blends into the
+    /// tooltip body, so the triangle joins the bubble with a soft fillet
+    /// instead of a hard kink (matches the design-system tooltip).
+    let arrowJunctionRadius: CGFloat = 4
     var arrowXFraction: CGFloat = 0.5
     var arrowDirection: TooltipArrowDirection = .up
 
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4462, opened 2026-06-01
perf(swap): cache VULT/Thorguard discount tier per wallet for the session

### LINKED ISSUE #4457: Swap: show loading on input + cache VULT/Thorguard discount tier per wallet
Two changes to cut the perceived and real latency before a swap quote appears.

## 1. Cache the VULT discount tier (incl. Thorguard) per wallet

`DefaultSwapInteractor.fetchQuote` resolves the discount tier serially **before** the quote network call:
- `VultisigApp/VultisigApp/Features/Swap/Interactor/DefaultSwapInteractor.swift:39`

The VULT balance is already 3-min cached, but the Thorguard NFT check is uncached and fires a fresh Ethereum `eth_call` on **every** quote fetch — every debounced keystroke **and** the 59s auto-refresh — for any wallet below Platinum tier (i.e. anyone holding 0 VULT):
- `VultisigApp/VultisigApp/Features/VultDiscountTiers/Service/VultTierService.swift:86` (`checkThorguardBalance`, no cache)

**Ask:** resolve the tier once when the swap screen loads, cache it per wallet for the session (VULT balance + Thorguard), and have the quote path read the cached value. The tier doesn't change mid-swap-session, so it shouldn't sit on the per-quote critical path.

## 2. Show the loading state on the first keystroke, not after the debounce

`SwapDetailsViewModel.fetchQuotes` only sets `isLoadingQuotes`/`isLoadingFees = true` **after** the 200ms debounce sleep, so there's 200ms of dead air before the spinner / redacted summary appears:
- `VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapDetailsViewModel.swift:364-368`

**Ask:** set the loading flags synchronously before the `Task.sleep` (right after the empty-amount guard). The sleep still debounces the network call and the per-keystroke `updateQuoteTask?.cancel()` still coalesces input — only the UI feedback moves earlier, so no extra requests fire.

gastonm5:
### Problem
`DefaultSwapInteractor.fetchQuote` resolved the VULT discount tier serially before every quote network call. The VULT *balance* was already 3-min cached, but the **Thorguard NFT check was uncached** — `checkThorguardBalance(for:)` fired a fresh Ethereum `eth_call` (`fetchERC20TokenBalance`) on **every** debounced keystroke and every 59s auto-refresh for any wallet below Platinum (anyone holding 0 VULT). A no-VULT wallet paid an eth_call per keystroke.

### Approach (design choice (a) + warm-on-load)
- Added a per-wallet **session cache for the fully-resolved tier** (VULT balance result + Thorguard boost) to `VultTierService`, backed by a thread-safe `SessionTierCache` actor keyed by `vault.pubKeyEdDSA`. A `Box` wrapper distinguishes "not yet resolved" from "resolved to nil" so a genuinely tier-less wallet isn't re-resolved.
- New `resolveTierForSession(for:)` does the full (uncached) resolve **once** and caches it; later calls return the cached value with **no network access**.
- The interactor now depends on a small `SwapDiscountTierResolving` seam (implemented by `VultTierService`) and reads the cached tier on the per-quote path instead of constructing `VultTierService()` and re-resolving each time.
- The swap screen warms the cache on load: `SwapDetailsScreen.onLoad` → `SwapDetailsViewModel.warmDiscountTier(vault:)` → `interactor.warmDiscountTier(for:)`. `fetchQuotes`/loading-flag logic was **not** touched.

### Session-validity policy
In-memory, **process-lifetime** cache keyed by vault id — the tier doesn't change mid-swap-session, matching the issue's "for the session" wording. It survives reopening the swap screen within the same app session; a `clearSessionTier(for:)` escape hatch is provided. The existing 3-min `@AppStorage` VULT-balance window still gates the *first* resolve's balance freshness.

### Correctness
The discount that reaches `fetchQuote`/the quote (`vultDiscountBps`) is **identical** to before — the same resolution logic runs, just once and cached. If a quote fires before the warm-up completes, `resolveTierForSession` resolves-and-awaits once (and caches), so the discount is never silently zeroed.

### Thorguard eth_call
No longer runs per-quote. It runs at most once per wallet per session (on warm-up, or on the first quote if it beats the warm-up).

### Logging
`print(...)` in the touched `VultTierService` code is replaced with OSLog `Logger` per repo rules.

### Scope
**Item 1 only.** Item 2 (loading flag before debounce) was already fixed in main by #4446 — `SwapDetailsViewModel.fetchQuotes` and its loading flags are untouched.

### Validation
- SwiftLint: clean on all touched files (`swiftlint lint --config VultisigApp/.swiftlint.yml`).
- Build: **BUILD SUCCEEDED** (iOS Simulator, iPhone 16 Pro, Debug, `CODE_SIGNING_ALLOWED=NO`).
- Tests: `DefaultSwapInteractorTests` — **7/7 passed**, including two new tests:
  - `testFetchQuoteResolvesTierOncePerSessionAcrossManyFetches` — warm-up + 5 quote fetches resolve the underlying tier (Thorguard eth_call stand-in) exactly **once**.
  - `testFetchQuotePassesCachedTierDiscountToQuoteService` — the cached tier's `bpsDiscount` reaches the quote service unchanged.
- Not verified at runtime on-device (no manual swap-flow run).

closes #4457

## Summary by CodeRabbit

* **Refactor**
  * Session-level caching of discount tiers to avoid redundant lookups and speed up swap quotes.
  * Pre-warms discount-tier data when opening the swap screen (view model + screen) to improve perceived load times.
  * Replaced ad-hoc printing with structured logging for better observability.

* **Tests**
  * Added tests validating discount-tier session caching and that resolved discounts propagate into quote results.


VultisigApp/VultisigApp/Features/VultDiscountTiers/Service/VultTierService.swift
```diff
@@ -27,13 +36,30 @@ struct VultTierService {
...
 
+    /// Resolves the discount tier once per wallet for the session and caches the
+    /// fully-resolved value (post-Thorguard). Subsequent calls return the cached
+    /// tier without any network access, keeping the Thorguard `eth_call` off the
+    /// per-quote critical path.
+    func resolveTierForSession(for vault: Vault) async -> VultDiscountTier? {
+        let vaultId = vault.pubKeyEdDSA
+        return await Self.sessionCache.resolve(for: vaultId) {
+            await fetchDiscountTier(for: vault)
+        }
```
johnnyluo, line 57: **`@Sendable` closure captures a `@Model Vault` — contradicts `resolve`'s own contract.**

The `resolve` doc comment (line 205) says *"`work` must not capture `@Model` types — pass value-type inputs only,"* but this closure captures `vault` (a `@Model`) and then `await fetchDiscountTier(for: vault)` touches `vault.coins` inside `Task {}` on the `SessionTierCache` actor's executor — i.e. off `MainActor`, the mandatory SwiftData boundary.

The pre-existing code already ran `fetchDiscountTier` off-MainActor, so this isn't strictly *new* unsafe access, but wrapping a non-`Sendable` `Vault` in a `@Sendable` closure is a real concurrency hazard that only compiles because strict concurrency checking isn't fully on here. Either resolve the value-typed inputs the closure needs *before* entering the actor and pass those, or fix the now-misleading comment so the next maintainer isn't misled.

VultisigApp/VultisigApp/Features/VultDiscountTiers/Service/VultTierService.swift
```diff
@@ -27,13 +36,30 @@ struct VultTierService {
...
+    }
+
+    /// Drops the session-cached resolved tier for a vault, forcing the next
+    /// `resolveTierForSession` to re-resolve from the network.
+    func clearSessionTier(for vault: Vault) async {
+        await Self.sessionCache.clear(for: vault.pubKeyEdDSA)
+    }
```
johnnyluo: **`clearSessionTier` (and `SessionTierCache.clear`) are dead code.** Nothing calls them. Per `.claude/rules/code-quality.md` ("No dead code: remove unused functions"), either remove it or wire it up. See the related note on cache invalidation — there probably *should* be a caller.

VultisigApp/VultisigApp/Features/VultDiscountTiers/Service/VultTierService.swift
```diff
@@ -7,15 +7,24 @@
...
     let thorguardContractAddress = "0xa98b29a8f5a247802149c268ecf860b8308b7291"
 
     @AppStorage("vult_balance_cache") private var cacheEntries: [CacheEntry] = []
     private static let cacheValidityDuration: TimeInterval = 3 * 60 // 3 minutes
 
+    /// Process-lifetime, per-wallet cache of the fully-resolved discount tier
+    /// (VULT balance result + Thorguard boost). The tier doesn't change during a
+    /// swap session, so once resolved it's read back without re-running the
+    /// Thorguard NFT `eth_call` on every quote fetch.
+    private static let sessionCache = SessionTierCache()
```
johnnyluo: **"Session" here = entire app process lifetime — worth a maintainer sign-off.** This cache never expires and is never invalidated, and the key is process-global across vault switches/swaps. Consequences:

- A user holding 0 VULT when the swap screen opens who then acquires VULT / a Thorguard NFT will **never** get the upgraded discount until app restart.
- The existing 3-min `@AppStorage` balance window becomes irrelevant after the first resolve.

The PR description acknowledges this and it matches the issue's "for the session" wording, but since it changes the discount a user actually pays on a HIGH-tier wallet path, consider invoking `clearSessionTier` on an explicit balance refresh or vault switch (which also resolves the dead-code note above).

johnnyluo (commented):
Three inline notes on the session-cache implementation (correctness path looks good and well-tested).

johnnyluo (approved):
LGTM

PR #4468, opened 2026-06-01
fix(keysign): don't surface CancellationError as broadcast failure

### LINKED ISSUE #4465: [BUG] iOS: Pairing device shows 'Failed to broadcast (Swift.CancellationError 1)' for a tx that actually broadcast
## Summary

On a co-signing (pairing) device, the keysign screen shows **"Error al Firmar / Failed to broadcast transaction,error:No se ha podido completar la operación. (Error de Swift.CancellationError 1.)"** — but the transaction was actually broadcast and landed on-chain. The user's funds moved correctly; only the pairing iPhone shows a false failure.

Observed on **iOS v1.37.57**, LTE, on a **BTC (UTXO) send**. (Full error text transcribed above; screenshot to be drag-attached.) The UTXO broadcast path is `KeysignViewModel.swift:803` (`UTXOTransactionsService.broadcastTransaction`), which routes its `catch` into `handleBroadcastError` at line 807.

This is the same *class* as #4170 / #2541 (co-sign device shows failure for an on-chain tx), but a **different error bucket**: the failure is a `Swift.CancellationError`, not a mempool/duplicate/sequence error.

## Where

`VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:956`

```swift
errMessage = "Failed to broadcast transaction,error:\(error.localizedDescription)"
```

`error.localizedDescription` here is the localized `CancellationError` (cooperative task cancellation), surfaced verbatim to the user.

## Probable cause (unverified)

The chain-agnostic safety net added in #4170 — `isAlreadyOnChain(transactionType:)` at `KeysignViewModel.swift:964` — is what *should* catch this: it looks the tx up by hash before declaring failure. It does not recover here, and the likely reason is that **cancellation propagates into the verification loop itself**:

- `isAlreadyOnChain` runs `try await TransactionStatusService.shared.checkTransactionStatus(...)` inside the already-cancelled task → each lookup throws `CancellationError` → caught and logged, loop continues.
- The backoff is `try? await Task.sleep(for: backoff)` — under cancellation `Task.sleep` throws immediately and `try?` swallows it, so there is **no real backoff**. All 3 attempts burn instantly in the cancelled context and `isAlreadyOnChain` returns `false`.
- `handleBroadcastError` then falls through to `setStatus(.KeysignFailed)` with the raw cancellation string.

So the broadcast-race net is defeated whenever the underlying failure is a cancellation rather than an RPC mempool error.

## Suggested direction (for the assignee to confirm)

1. Treat `CancellationError` (and `NSURLErrorCancelled`) as a distinct case in `handleBroadcastError` — it is **not** conclusive evidence the tx failed.
2. Run the `isAlreadyOnChain` hash check in a **detached / non-cancelled** context (e.g. `Task.detached` or isolated lookups) so verification can actually reach the RPC even when the parent keysign task was cancelled.
3. Don't surface a raw localized `CancellationError` string to the user as a broadcast failure.

## Impact

False-failure screen on the pairing device for a successful, on-chain transaction. Risk: a user re-taps **"Intentar de Nuevo"** and double-broadcasts from a flow that already succeeded.

---
_Heads up — AI-assisted triage; file:line references verified against current `main`._

gastonm5:
## Problem

On a co-signing (pairing) device, the keysign screen showed **"Failed to broadcast transaction, error: …Swift.CancellationError 1…"** for a BTC/UTXO tx that **actually broadcast and landed on-chain**. Funds moved; only the device displayed a false failure — and the "Try Again" CTA invited a double-broadcast of an already-landed tx.

## Root cause (cancellation poisons broadcast + the shared verify net)

The keysign work runs in a SwiftUI `.task`-backed `Task` that gets cancelled on view teardown / scene change. When cancellation lands:

1. The in-flight broadcast HTTP call throws `CancellationError`. `HTTPClient.request` calls `try Task.checkCancellation()` as its **first statement** and re-throws `URLError.cancelled` as a fresh `CancellationError()`. The tx may already have reached the mempool/proxy.
2. `handleBroadcastError` received it. `CancellationError` is not a `RetryableBroadcastError`, so it hit the `default` branch and built the raw `"Failed to broadcast transaction,error:\(error.localizedDescription)"` message.
3. The safety net (`isAlreadyOnChain`, from #4252) was **defeated by the same cancellation**: every `checkTransactionStatus` lookup went through `HTTPClient.request` → `Task.checkCancellation()` threw before any network I/O, and the backoff `try? await Task.sleep(...)` threw immediately under cancellation and was swallowed by `try?` (no real wait). All 3 attempts burned instantly → returned false → `.KeysignFailed` with the raw cancellation string.

This is **chain-agnostic**: the verify/messaging net is shared across every chain's broadcast site (inline `do/catch` for UTXO/Cardano and the outer `do/catch` for THORChain/Maya/EVM/Cosmos/Solana/Sui/Polkadot/Bittensor/Ton/Ripple/Tron), so the fix lives in the shared `handleBroadcastError` / `isAlreadyOnChain`.

## The fix

- **(A) Classify cancellation as non-conclusive.** New `KeysignViewModel.isCancellation(_:)` matches `CancellationError`, `URLError.cancelled`, and a cancellation wrapped in `HTTPError.networkError`. Detected early in `handleBroadcastError` and routed into verification/messaging instead of an immediate `.KeysignFailed`. **Deliberately NOT reclassified as `RetryableBroadcastError`** — auto re-broadcast of a possibly-landed tx is exactly the double-spend we're avoiding.
- **(B) Detached verification (primary).** `isAlreadyOnChain` now runs its RPC lookups inside `Task.detached { … }.value`, capturing value-type inputs (`Chain` + tx-hash `String`) before detaching so the closure is Sendable-clean. A detached task has a fresh, non-cancelled context, so `HTTPClient`'s `Task.checkCancellation()` no longer short-circuits and the backoff `Task.sleep` actually waits (the `try?`-swallow is replaced by an explicit `do/catch`). Only the network lookup is detached; all `@Published`/`@MainActor` state changes are applied back on the MainActor. On confirmed (`.confirmed`/`.pending`) the tx is treated as success exactly like a normal broadcast (txid set, normal `.KeysignFinished` path).
- **(C) Neutral state, never a raw `CancellationError`.** New terminal state `KeysignBroadcastUnconfirmed`. When a cancelled broadcast can't be positively confirmed (`.notFound` / unreachable), the UI shows a neutral "Couldn't confirm transaction — check the explorer" screen with the deterministic tx hash + explorer link (`viewOnExplorer` → `getTransactionExplorerURL`) and **no one-tap re-broadcast**. `movetoDoneView` is guarded so the unconfirmed state never navigates to the success done-screen. New strings `broadcastCouldNotConfirm` / `broadcastCouldNotConfirmTitle` added to all locale files + `sort_localizable.py` run.
- **Idempotency guard (hardening).** Before broadcasting, short-circuit to success when a prior attempt for the deterministic hash exists (`StoredPendingTransactionStorage.get`) **and** the tx is already on-chain (detached-safe `isAlreadyOnChain`). Gated on the stored-record check so first broadcasts pay no extra round-trip; protects current + future retry paths from double-spend.

## Test seam + coverage

Added a `TransactionStatusChecking` protocol (`TransactionStatusService` conforms, defaults to `.shared`), injectable on `KeysignViewModel` (`transactionStatusChecker`) — production behavior unchanged. 9 new tests in `KeysignBroadcastCancellationTests`:
- cancellation + hash on-chain (`.confirmed`/`.pending`) → success state with txid, **not** `.KeysignFailed`;
- cancellation + `.notFound` → neutral `KeysignBroadcastUnconfirmed`, user string contains **no** "cancellation" text;
- `URLError.cancelled` and `HTTPError.networkError(CancellationError())` are classified as cancellations;
- verification still calls the checker even when the **parent task is cancelled** (detached works);
- genuine non-cancellation RPC failure still ends in `.KeysignFailed` (no regression), and the existing peer-race rescue still resolves an on-chain hash;
- classifier unit coverage.

## Results

- **Build:** `xcodebuild … -destination 'platform=iOS Simulator,name=iPhone 17 Pro' build` → **BUILD SUCCEEDED**; `build-for-testing` → **TEST BUILD SUCCEEDED**.
- **Tests:** `KeysignBroadcastCancellationTests` → **9 passed, 0 failures**.
- **Lint:** `swiftlint` clean on all touched files/dirs (0 violations).

## Scope / caveats

- **Signing / TSS / keygen crypto is untouched** — only broadcast-result handling, verification, and messaging.
- No unrelated chains' broadcast **logic** changed — only the shared error/verify/messaging net.
- **Runtime verification is PENDING:** I cannot reproduce the live peer-broadcast race / cancellation timing in this environment. Logic, build, lint, and unit tests are green, but a live cosign repro on-device has not been performed.

Builds on the prior safety net in #4252.

**Maintainer review requested** — HIGH security tier, broadcast path.

closes #4465

## Summary by CodeRabbit

* **Bug Fixes**
  * Better handling when transaction broadcasts can’t be confirmed; avoids false failures, preserves tx IDs and explorer links, and continues verification even if the initiating task is cancelled.

* **New Features**
  * In-place “broadcast unconfirmed” error surface with blockchain explorer link and navigation guards.

* **Localization**
  * Added localized title and message for broadcast confirmation failures in multiple languages.

* **Tests**
  * New unit tests covering cancellation, on-chain verification outcomes, and terminal-state guards.


VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
```diff
@@ -984,6 +1033,50 @@ class KeysignViewModel: ObservableObject {
...
+            logger.info("cancelled broadcast confirmed on-chain — using hash \(transactionType.transactionHash, privacy: .public)")
+            await applyBroadcastSuccess(transactionType: transactionType)
+            return
+        }
+
+        logger.warning("cancelled broadcast could not be confirmed on-chain — neutral unconfirmed state for hash \(transactionType.transactionHash, privacy: .public)")
+        setTxid(transactionType.transactionHash)
+        self.approveTxid = transactionType.approveTransactionHash
+        self.keysignError = "broadcastCouldNotConfirm".localized
+        setStatus(.KeysignBroadcastUnconfirmed)
```
johnnyluo, line 1081: 🔴 **Blocking: this `.KeysignBroadcastUnconfirmed` state gets overwritten in the real flow.**

`handleCancelledBroadcast` is reached via `broadcastTransaction()`, which is called from `startKeysignDKLS` (line ~396) and `startKeysignGG20` (line ~427). Immediately after that call both do:
```swift
if status != .KeysignFailed && status != .KeysignRetryRequested {
    setStatus(.KeysignFinished)
}
```
Since `.KeysignBroadcastUnconfirmed` is excluded from neither check, it gets clobbered to `.KeysignFinished`. The view body switches on `viewModel.status` (`KeysignView.swift:74`), so in production the user lands on the **success done-screen**, not the neutral "couldn't confirm" screen — defeating feature (C), the central goal of this PR.

The added `movetoDoneView()` guard only blocks the `onChange(of: txid)` navigation path; it doesn't affect the status-driven body switch.

**Why CI is green:** all 9 tests call `handleBroadcastError(...)` directly, bypassing `startKeysignDKLS`/`startKeysignGG20`, so this integration seam is never exercised.

**Fix:** exclude `.KeysignBroadcastUnconfirmed` from both guards:
```swift
if status != .KeysignFailed,
   status != .KeysignRetryRequested,
   status != .KeysignBroadcastUnconfirmed {
    setStatus(.KeysignFinished)
}
```
Please also add a regression test that drives the actual `broadcastTransaction()` path (e.g. inject a broadcaster that throws `CancellationError`) so this seam is covered end-to-end.

johnnyluo (commented):
Code review: one blocking integration issue flagged inline. Otherwise a well-reasoned fix — detached verification and the no-auto-rebroadcast stance are right.

johnnyluo (approved):
LGTM

Issue #4470, opened 2026-06-01
[BUG][CRASH][macOS] NSException in AppKit constraint-update cycle — SwiftUI state mutation during render commit (v1.37.57)


realpaaao:
## Summary

macOS app crashes with `EXC_BREAKPOINT (SIGTRAP)` on the **main thread**. The terminating event is an **`NSException` thrown from AppKit's constraint-update cycle** (`-[NSWindow(NSDisplayCycle) _postWindowNeedsUpdateConstraints]`), triggered while SwiftUI is committing a render transaction inside `NSHostingView.layout`.

This persists **after** the prior macOS-crash work (#4124 / #4125 `asyncAfter` sweep, and #4327 / #4379 keysign race) — it is a **different signature**, not a timing/`asyncAfter` crash.

## Observed

- **Version:** 1.37 (57)
- **OS:** macOS 15.7.7 (24G720), Apple Silicon (ARM-64)
- **Time Awake Since Boot:** 380s — crash happens ~6 min into the session, not at launch.
- **Exception:** `EXC_BREAKPOINT (SIGTRAP)`, `Crashed Thread: 0 / com.apple.main-thread`.

User reports recurring crashes even after substantial stability work already landed.

## Crash signature (the telling frames)

```
3  AppKit       -[NSWindow(NSDisplayCycle) _postWindowNeedsUpdateConstraints]   ← NSException thrown here
10 SwiftUI      NSHostingView.setNeedsUpdate
11 SwiftUI      NSHostingView.requestUpdate(after:)
12 SwiftUICore  ViewRendererHost.graphDidChange
13 SwiftUICore  GraphHost.graphInvalidation(from:)
17 SwiftUICore  ObservationGraphMutation.apply()      ← observable state mutated…
18 SwiftUICore  withTransaction
19 SwiftUICore  GraphHost.runTransaction
20 SwiftUICore  ViewRendererHost.render(...)          ← …during the render/commit pass
23 SwiftUI      NSHostingView.layout { NSAnimationContext }
27 SwiftUI      NSHostingView.layout
49 QuartzCore   CA::Transaction::commit
```

[...]

PR #4125, opened 2026-04-04
fix: replace top 3 asyncAfter crash sources with cancellable Tasks [macOS]

realpaaao:
## Summary
- Replace `DispatchQueue.main.asyncAfter` with cancellable `Task` + `Task.sleep` in the 3 highest-risk crash sites to prevent `EXC_BREAKPOINT` in `-[NSWindow _postWindowNeedsUpdateConstraints]` on macOS Sequoia
- **CrossPlatformSheet.swift**: Both sheet dismiss delays (bool and item variants) now use `@State dismissTask` with cancellation
- **ContentView.swift**: 2s splash view delay now stored in `@State dismissSplashTask` with cancellation guard
- **VaultManagementSheet.swift**: Nested double `asyncAfter` for detent animation replaced with single sequential `Task` in `@State detentAnimationTask`

## Context
The macOS app frequently crashes with `EXC_BREAKPOINT (SIGTRAP)` in `_postWindowNeedsUpdateConstraints` during SwiftUI render cycles. Root cause: `DispatchQueue.main.asyncAfter` closures fire unconditionally — even after the window closes — triggering re-entrant constraint updates on an invalidated NSWindow.

These 3 sites were identified as the highest-probability crash sources from crash report analysis. Remaining ~65 sites are tracked in #4124.

## Test plan
- [ ] macOS: Open and close sheets repeatedly — no crash
- [ ] macOS: Close window during splash screen 2s delay — no crash
- [ ] macOS: Toggle vault management sheet edit mode rapidly — no crash
- [ ] iOS: Verify sheet dismiss, splash, and vault management behavior unchanged
- [ ] SwiftLint passes (verified: 0 violations)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of UI dismissal timing for splash screens and sheet components to enhance app responsiveness and reliability.

Issue #4124, opened 2026-04-04
[Fix] Replace DispatchQueue.main.asyncAfter with structured concurrency to prevent macOS crashes [macOS]


realpaaao:
## Problem

The macOS app frequently crashes with `EXC_BREAKPOINT (SIGTRAP)` in `-[NSWindow _postWindowNeedsUpdateConstraints]`. The root cause is **~65 remaining `DispatchQueue.main.asyncAfter` calls** throughout the codebase that mutate SwiftUI `@State`/`@Binding`/`@Published` properties after arbitrary delays. These closures are **not cancellable** and fire unconditionally — even after the view disappears or the window closes — triggering re-entrant constraint updates on an invalidated NSWindow.

The top 3 crash sources (CrossPlatformSheet, ContentView splash delay, VaultManagementSheet nested delays) have been fixed separately. This issue tracks migrating all remaining sites.

## Expected Behavior

No crashes from delayed state mutations. All delayed UI state changes should be cancellable and automatically cleaned up when their owning view disappears.

## Solution

Replace all `DispatchQueue.main.asyncAfter` calls with structured concurrency:
- For delays inside views: use `.task(id:)` modifier with `Task.sleep` (auto-cancelled on view disappear)
- For delays in view functions: use `@State private var task: Task<Void, Never>?` pattern with `task?.cancel()` before reassignment
- For delays in ViewModels: use stored `Task` properties cancelled in `deinit` or `onDisappear`

### Must Do (by priority)

**HIGH — Mutates navigation/global state:**
- [ ] `HomeScreen.swift` — 9 asyncAfter calls (lines 78, 305, 355, 367, 406, 452, 525, 602, 611)
- [ ] `VaultDeletionConfirmView.swift` — line 99 (1s delay before SwiftData deletion)
- [ ] `ChainDetailScreen.swift` — line 123 (macOS-only 0.05s sheet state change, line 87 copy callback)

**MEDIUM — Sheet/modal presentation after delay:**
- [ ] `BiweeklyPasswordVerificationViewModifier.swift` — lines 23, 43 (0.5s delays presenting sheets)
- [ ] `MonthlyBackupWarningViewModifier.swift` — line 32 (0.5s delay presenting sheet)
- [ ] `VaultMainScreen.swift` — lines 84, 255 (scroll + search clear delays)
- [ ] `DefiMainScreen.swift` — lines 62, 173 (scroll + search clear delays)
[...]

PR #4477, opened 2026-06-03
fix(macos): prevent reentrant state mutation during render commit (banners carousel + sheet blur)

gastonm5:
## The crash class

macOS-only hard crash: `EXC_BREAKPOINT (SIGTRAP)` from an `NSException` thrown inside AppKit's constraint cycle (`-[NSWindow _postWindowNeedsUpdateConstraints]`) **while SwiftUI is committing a render transaction inside `NSHostingView.layout`**.

Root cause: a view mutates observable `@State`/`@Published` state **synchronously during an animated layout/commit** that is driven by a timer/publisher (runloop) tick — not a user gesture. That mid-cycle write re-invalidates the SwiftUI graph, which reenters AppKit's constraint update and throws a fatal `NSException`. On iOS the same situation is only the purple "Modifying state during view update" warning; on macOS it is a hard crash. The crashed thread runs through a runloop observer / display-link commit (`CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER` -> `CATransaction NS_setFlushesWithDisplayLink`), confirming a timer/publisher-driven animated mutation.

This is the **same class as #4125** (which swept three `asyncAfter` sites and documented this exact signature), here at two sites that PR did not cover. The fix shape deliberately follows the durable pattern from that work — **idempotent guarded writes + breaking the animated feedback loop + cancellable `Task`**, not the `asyncAfter`/`Task {}` "deferral" anti-pattern that #4124/#4125 were removing.

### 1. `BannersCarousel` (`Features/Common/Views/Banners/BannersCarousel.swift`)
Sits on the idle Home tab (`VaultMainScreen`), matching the "several minutes in, not at launch" repro. It had a **mutually-recursive animated loop**: a 3s repeating `Timer` did `withAnimation { currentIndex += 1 }`; `onChange(currentIndex)` did `withAnimation { scrollPosition = … }`; and `onChange(scrollPosition)` did `withAnimation { currentIndex = …; startTimer() }`. `scrollPosition` drives `.scrollPosition(id:)` (layout). Timer (runloop) + nested `withAnimation` (NSAnimationContext) + reciprocal writes = the crash signature.

- **Broke the reciprocal loop with no-op guards on every write.** `currentIndex -> scrollPosition` only writes when the values differ; `scrollPosition -> currentIndex` only writes when they differ. A no-op write still emits a graph mutation, so a sync that lands mid-commit with an unchanged value previously re-invalidated the graph; the guards make it a true no-op and stop the handlers ping-ponging.
- **Pulled `startTimer()` out of `withAnimation`** (both in the `scrollPosition` handler and inside the timer tick) so a timer (re)start never runs an `NSAnimationContext` concurrently with a runloop tick committing layout.
- **Guarded the auto-advance write** (skip when the next index equals the current).
- **Replaced both `DispatchQueue.main.asyncAfter` calls** in `removeBanner` with a single cancellable `delayedTask` (the existing #4125-style helper backed by `Task { @MainActor }` + `Task.sleep`), stored in `@State` and **cancelled in `onDisappear`** so a stale closure can't mutate state on a torn-down view. The auto-advance timer is also already invalidated in `onDisappear`.
- **Extracted the wrap-around + post-removal index math** into a pure `BannersCarouselIndex` enum and unit-tested it.
- **Replaced `print()` with OSLog `Logger`.**

Behavior preserved: still auto-advances every 3s, still wraps around, still tracks manual swipe, indicators still follow the page. Only the reentrancy/loop is removed.

### 2. `SheetPresentedViewModifier` (`Components/Sheet/SheetPresentedCounterManager/SheetPresentedViewModifier.swift`)
`onReceive(counterManager.$counter)` (Combine, lands on the runloop) set `blurContent` (`@State`) under `.animation(.easeInOut, value: blurContent)`, animating a `.blur` over the entire NavigationStack (mounted once at `ContentView`). If the publisher fired during a sheet's own animated transition commit, it re-invalidated the laying-out root.

- **No-op guard:** compute `let shouldBlur = newValue > 0` and only assign when `blurContent != shouldBlur`. The counter changes on every increment/decrement (more often than the boolean flips), so the guard collapses most ticks to no-ops and keeps the write out of the commit.
- **Why guard, not a derived/computed blur:** the manager is read via a plain `@Environment(\.sheetPresentedCounterManager)` value (the env key's `static let` default instance), **not** `@EnvironmentObject`. Plain environment values do not re-render on `@Published` changes, so `.blur(radius: counter > 0 ? … : 0)` would not update reactively — only `onReceive` observes the publisher. Keeping `onReceive` + guarding the write is the option that both removes the in-commit mutation and preserves behavior. No `asyncAfter` introduced.

Behavior preserved: content still blurs when a sheet is presented and unblurs on dismiss, same easeInOut feel.

## Tests
Added `BannersCarouselIndexTests` (10 cases) covering `BannersCarouselIndex.next(after:count:)` (wrap-around, single/empty no-ops) and `afterRemoval(removedIndex:currentIndex:countBeforeRemoval:)` (remove before/after/at current, last-item step-back, single-banner reset). The reciprocal-loop and guard logic that lives in the view body is not directly unit-testable without a view host; no fake tests were added for it — it relies on the build + the reasoning above. All 10 new tests pass.

## Build results (both required, both green)
- **macOS** (`platform=macOS,arch=arm64`, the crashing platform): **BUILD SUCCEEDED**
- **iOS Simulator** (iPhone 17, regression check): **BUILD SUCCEEDED**
- SwiftLint clean on all touched files (0 violations).
- New unit tests: 10 passed, 0 failed (run on iPhone 17 sim; the unit-test target is iOS-only).

## Scope and honest caveats
- **Runtime macOS crash repro is NOT verified** — no macOS exception-breakpoint repro was available here. The two sites were root-caused from the crash signature and the #4125 precedent; the exact crashing view is not runtime-confirmed.
- The crash class can fire from multiple sites. The other ranked suspects from the investigation — **CrossPlatformSheet**, **KeysignDiscoveryView**, **StepsAnimationView** — are intentionally **out of scope** for this PR (lower confidence) and remain follow-ups that need a runtime exception-breakpoint session to confirm.
- Because of the above, this uses **`Addresses #4470` without an auto-close keyword** rather than `closes`.

Addresses #4470

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sheet transition performance by preventing unnecessary blur state updates.
  * Fixed banner carousel navigation to prevent redundant state changes and improved state synchronization.
  * Enhanced banner removal timing with better task cancellation handling.

* **Tests**
  * Added comprehensive unit tests for banner carousel navigation logic.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Sheet/SheetPresentedCounterManager/SheetPresentedViewModifier.swift
+++ b/VultisigApp/VultisigApp/Components/Sheet/SheetPresentedCounterManager/SheetPresentedViewModifier.swift
@@ -25,7 +25,17 @@ private struct SheetPresentedViewModifier: ViewModifier {
             .blur(radius: blurContent ? 6 : 0)
             .animation(.easeInOut(duration: 0.1), value: blurContent)
             .onReceive(sheetPresentedCounterManager.$counter) { newValue in
-                blurContent = newValue > 0
+                // Guard the write so an unchanged value never emits a graph
+                // mutation. The publisher lands on the runloop and can fire
+                // during a sheet's own animated transition commit; a no-op
+                // assignment mid-commit re-invalidates the laying-out root and
+                // triggers a reentrant AppKit constraint update that crashes on
+                // macOS. The counter changes more often than the boolean (every
+                // increment/decrement), so this guard collapses most ticks to
+                // no-ops.
+                let shouldBlur = newValue > 0
+                guard blurContent != shouldBlur else { return }
+                blurContent = shouldBlur
             }
             #if os(iOS)
             // Remove blur when enter background as the iOS dismisses sheet
--- a/VultisigApp/VultisigApp/Features/Common/Views/Banners/BannersCarousel.swift
+++ b/VultisigApp/VultisigApp/Features/Common/Views/Banners/BannersCarousel.swift
@@ -6,6 +6,9 @@
 //
 
 import SwiftUI
+import OSLog
+
+private let logger = Logger(subsystem: "com.vultisig.app", category: "banners-carousel")
 
 struct BannersCarousel<Banner: CarouselBannerType>: View {
     @Binding var banners: [Banner]
@@ -20,6 +23,7 @@ struct BannersCarousel<Banner: CarouselBannerType>: View {
     @State var bannersCount: Int = 0
     @State var bannersToRemove: Set<AnyHashable> = []
     @State private var timer: Timer?
+    @State private var removalTask: Task<Void, Never>?
     @State var showCarousel: Bool = false
 
     @State var internalBanners: [Banner] = []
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4506, opened 2026-06-08
fix(keygen): show nav bar on PeerDiscoveryScreen for secure vault

### LINKED ISSUE #4505: Secure vault: navigation bar hidden on PeerDiscoveryScreen (iPhone + iPad)
## Observed
Creating a **Secure** vault on iPhone (and iPad) shows **no navigation bar** — no title, back/close, or QR-share — on `PeerDiscoveryScreen` (the "waiting for devices" pairing step). Fast vaults are unaffected (they're intentionally chromeless and auto-proceed).

## Root cause
`PeerDiscoveryScreen` hides its nav bar **unconditionally** (`.screenNavigationBarHidden()`, `PeerDiscoveryScreen.swift:174` → `Screen` applies `.toolbar(.hidden, for: .navigationBar)` on iOS), then tries to re-introduce a bar for the secure path from **outside** the `Screen` via `.crossPlatformToolbar("", …)` (`:191-203`). That ancestor toolbar can't override the inner hidden-toolbar and passes an **empty title**, so the secure path renders no bar.

- The external toolbar is gated on `selectedTab.hasOtherDevices` (`.fast` → false, `.secure` → true). Fast never adds it (chromeless, expected); secure adds the broken one.
- iOS-only: `Screen`'s suppression is `#if os(iOS)`; macOS uses a custom toolbar. **iPad is affected too.** Reshare-secure hits the same screen.
- Not a recent regression — latent since the screen was created (2026-02-15).

## Fix direction
Adopt the app's working pattern (`SendPairScreen.swift`): make `.screenNavigationBarHidden(...)` conditional (hide only on the fast/auto path), add a real `.screenTitle(...)`, and route the QR-share button via `.screenToolbar { … }` (plus `.screenBackButtonHidden(hideBackButton)`) so `Screen` renders the title + items itself.

**Platform:** iOS (iPhone + iPad). macOS unaffected.

gastonm5:
Closes #4505

Make `PeerDiscoveryScreen` route its title, conditional nav-bar hide, back-button state, and QR-share button through `Screen` (the `SendPairScreen` pattern) so the secure-vault pairing step renders a proper navigation bar instead of none.

Also fixes **iPad** and **Reshare-secure**, which hit the same screen, and **preserves the fast-path chromeless behavior** (the bar stays hidden when `!selectedTab.hasOtherDevices`).

## Summary by CodeRabbit

* **Refactor**
  * Improved peer discovery screen UI organization by consolidating title, navigation button, and toolbar management into a unified configuration approach, enhancing maintainability and consistency.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keygen/Views/PeerDiscoveryScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/Views/PeerDiscoveryScreen.swift
@@ -171,7 +171,20 @@ struct PeerDiscoveryScreen: View {
             Screen {
                 states
             }
-            .screenNavigationBarHidden()
+            .screenTitle("pair".localized)
+            .screenNavigationBarHidden(!selectedTab.hasOtherDevices)
+            .screenBackButtonHidden(hideBackButton)
+            .screenToolbar {
+                CustomToolbarItem(placement: .trailing) {
+                    if isShareButtonVisible {
+                        NavigationQRShareButton(
+                            vault: vault,
+                            type: .Keygen,
+                            viewModel: shareSheetViewModel
+                        )
+                    }
+                }
+            }
             .screenEdgeInsets(ScreenEdgeInsets(top: 0, leading: 0, bottom: 0, trailing: 0))
             .screenBackground(.gradient)
             .onAppear {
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4508, opened 2026-06-08
fix(keygen): discard aborted secure vault so its name is reusable (#4493)

### LINKED ISSUE #4493: Keygen: aborting at "Looks Good" verification still persists the vault (cosigner) — name then blocked as "already in use"
## Observed

iOS as **cosigner** (Join Keygen), version 1.38.58:

1. Join a keygen, DKLS completes, reach the **"Review Your Vaults / Looks Good"** verification screen.
2. Something looks wrong → user **aborts** (back / "Something's wrong").
3. The vault is **already saved** to storage. Re-attempting keygen with the **same vault name** fails: *"Vault name already in use."*

Expected: an aborted keygen is completely discarded — no vault persisted, name reusable.

## Probable cause (light)

The vault is committed to SwiftData the instant DKLS keygen finishes, **before** the verification screen is shown, and there is no rollback on abort.

- `KeygenViewModel.finalizeDKLSKeygen` → `context.insert(self.vault)` + `try context.save()` (`KeygenViewModel.swift:790-793`). Runs inside `startKeygen` (kicked off in `KeygenView.onLoad`), i.e. before navigation to `ReviewYourVaultsScreen`.
- Abort path `ReviewYourVaultsScreen.handleSomethingsWrong()` for a cosigner just calls `router.navigateToRoot()` — no `context.delete` / rollback (`ReviewYourVaultsScreen.swift:122-136`). A repo-wide grep for `context.delete` / `deleteVault` / `rollback` in `Features/Keygen/` returns zero matches.
- Name uniqueness is then tripped by the leftover vault: `VaultNameValidator.swift:14-27` rejects any case-insensitive name already in storage.

## Fix direction

Either defer `context.insert` + `context.save()` until after the user confirms "Looks Good", or add `context.delete(vault)` + save on every abort/back path out of the post-keygen screens.

gastonm5:
## Problem

Closes #4493

As a cosigner (Join Keygen), aborting at the **"Review Your Vaults / Looks Good"** verification screen (back, "Something's wrong") still left the vault **persisted** in SwiftData, so re-running keygen with the same name failed *"Vault name already in use."*

## Root cause

For a secure keygen, `KeygenViewModel` inserted + saved the new vault the instant DKLS/GG20 finished — inside `finalizeDKLSKeygen` / `startKeygenGG20` / `startKeyImportKeygen` — **before** navigating to `ReviewYourVaultsScreen` and **before** any "Looks Good" confirmation. The review-screen abort paths (`handleSomethingsWrong` → `navigateToRoot` for cosigner, → `peerDiscovery` for initiator; back; back-swipe; app kill) performed no `delete`/`rollback`, so the vault stayed in the store. The unique `@Attribute(.unique) name` and `VaultNameValidator` (case-insensitive) then reject the name on retry.

## Approach: Option B — defer persistence to confirm

I chose **deferring the insert to the "Looks Good" confirmation** over delete-on-abort, because the persisted vault is **not read** anywhere between keygen-finish and confirm:

- `ReviewYourVaultsScreen`, `OnboardingOverviewScreen`, and the backup chain (`VaultBackupScreen` → `VaultBackupContainerView`) all receive the **in-memory** `vault` **by reference** — none of them issue a `FetchDescriptor`/`@Query` for it.
- Vault selection after backup (`AppViewModel.set(selectedVault:)`) and `markBackedUp()` mutate that **same in-memory object**; the flag persists via the main context's default autosave. `AppViewModel.loadSelectedVault` matches by persisted name + pubKey, which only matters after the vault is legitimately saved.
- The only pre-confirm SwiftData reads are the duplicate-vault collision guard (keyed on `pubKeyECDSA`, decides the upsert) and the name validator (which must *not* see the aborted vault).

So deferring is safe and discards the vault on **every** abort path automatically — including **app kill**, which delete-on-abort (Option A) structurally cannot cover.

### Why scoped to the secure path
- **Fast** vaults skip the review screen and already handle their own abort by deleting the vault at the email-verification step (`ServerBackupVerificationView.deleteVault()`), so they keep persisting at keygen-finish.
- **Reshare on an existing committee member** (`needsInsert == false`) updates an already-persisted, managed vault and still saves at keygen-finish.
- Only the **secure new-vault** path (the one gated by the review screen) is changed — for both **cosigner and initiator** (same screen; only `handleSomethingsWrong` differs).

## Changes
- `KeygenViewModel`: new `deferVaultPersistence` flag (set via `setData`); new `static commitVault(_:context:)` that does default-coins + insert + save. The three keygen-finish blocks skip insert/save (and default-coin creation) when deferring — default coins are kept out of the deferred finish path because they insert `Coin` rows the autosave could otherwise flush as orphans.
- The **duplicate-vault alert is unchanged**: its detection still runs at keygen-finish (a read-only fetch, no persistence side effect); only the insert is deferred. Inserting on the unique `pubKeyECDSA` still upserts, so re-keygen that reproduces an existing vault replaces it as before.
- `KeygenView`: computes `shouldDeferVaultPersistence` (secure, non-fast `.Keygen/.Reshare/.KeyImport`).
- `ReviewYourVaultsScreen`: "Looks Good" calls `commitVault` before navigating; surfaces a save error via an alert instead of silently proceeding.

## Abort paths covered (secure)
"Something's wrong" (cosigner → root; initiator → peer discovery), back button, system back-swipe, and app background/kill — all leave nothing persisted, so the name is reusable.

## Verification
- **SwiftLint**: 0 violations on changed files.
- **Build** (iPhone 16 Pro sim): **BUILD SUCCEEDED**.
- **Tests** — new `KeygenVaultCommitTests` (3/3) pins the contract: aborted keygen persists no vault and the name stays reusable; confirm persists exactly one vault and claims the name case-insensitively; abort-then-retry with the same name succeeds. Regression suites green: `KeygenPeerDiscoveryViewModelTests` (12/12), `KeyImportChainRestrictionTests` (11/11).

## Manual re-test suggested
- Cosigner: secure keygen → at review tap "Something's wrong" / back / kill app → vault gone, **name reusable**.
- Initiator: same, plus "Something's wrong" returns to peer discovery; restart + "Looks Good" creates the vault once.
- Happy path (cosigner + initiator): "Looks Good" → overview → backup → vault appears, is selected, `isBackedUp` persists.
- Fast vault: unchanged; email-verification "use different email" still discards.

## Release Notes

* **Bug Fixes**
  * Enhanced error handling for vault persistence; users now receive alerts if vault commitment fails during key generation workflows.

* **Improvements**
  * Optimized vault persistence timing for secure key generation and key import flows to ensure better reliability and control.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift
@@ -82,6 +82,14 @@ class KeygenViewModel: ObservableObject {
     var singleKeygenType: SingleKeygenType?
     var isTssBatch: Bool = false
 
+    /// When `true`, a brand-new vault is *not* inserted/saved when keygen finishes;
+    /// persistence is deferred to the "Review Your Vaults" confirmation
+    /// (`KeygenViewModel.commitVault`). This guarantees an aborted secure keygen
+    /// discards the vault on every exit path (back, "Something's wrong", app kill),
+    /// keeping the vault name reusable. The fast path persists immediately because
+    /// it manages its own abort cleanup at the email-verification step.
+    var deferVaultPersistence: Bool = false
+
     @Published var isLinkActive = false
     @Published var keygenError: String = ""
     @Published var status = KeygenStatus.CreatingInstance
@@ -122,7 +130,8 @@ class KeygenViewModel: ObservableObject {
                  initiateDevice: Bool,
                  keyImportInput: KeyImportInput? = nil,
                  singleKeygenType: SingleKeygenType? = nil,
-                 isTssBatch: Bool = false
+                 isTssBatch: Bool = false,
+                 deferVaultPersistence: Bool = false
     ) async {
         self.vault = vault
         self.tssType = tssType
@@ -136,6 +145,7 @@ class KeygenViewModel: ObservableObject {
         self.keyImportInput = keyImportInput
         self.singleKeygenType = singleKeygenType
         self.isTssBatch = isTssBatch
+        self.deferVaultPersistence = deferVaultPersistence
         let isEncryptGCM = await FeatureFlagService().isFeatureEnabled(feature: .EncryptGCM)
         messagePuller = MessagePuller(encryptionKeyHex: encryptionKeyHex, pubKey: vault.pubKeyECDSA,
                                       encryptGCM: isEncryptGCM)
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #4512, opened 2026-06-08
fix(qbtc-claim): hide UTXOs below the confirmation threshold

gastonm5:
## Problem
A BTC→QBTC claim offered Bitcoin UTXOs that had not yet reached the chain's required confirmation count (MinUtxoConfirmationBlocks = 144). Selecting one failed at broadcast: `gas simulation failed: ... no valid claimable UTXOs found`. The claimable filter (QBTCChainService.filterClaimable) gated on entitlement (claimed / not-indexed) but never on confirmations, and fails open on transient errors — so under-confirmed UTXOs leaked into the selectable list.

## Fix
- Add `QBTCChainService.minUtxoConfirmationBlocks()` to read the param.
- Surface the BTC chain tip from the Blockchair response (`context.state`) alongside the claimable UTXOs.
- Add a confirmation gate hiding any UTXO with fewer than the threshold confirmations (`tip - blockHeight + 1`; missing block height = mempool = 0, always hidden). Fail-open when the tip/param is unavailable.
- Show an 'awaiting confirmations' reason when the list is empty only because everything is under-confirmed.

User-reported (no issue). swiftlint clean, build green, 35 unit tests pass (confirmation math, threshold-from-param, nil-height, fail-open).

## Summary by CodeRabbit

* **New Features**
  * QBTC claim flow enforces Bitcoin confirmation thresholds, hiding under‑confirmed UTXOs and showing an “awaiting confirmations” blocked state (with updated blocked-view icon).

* **Behavior Changes**
  * Bitcoin chain-tip from Blockchair responses is surfaced to improve confirmation accuracy.
  * Fetch now fails when expected address data is missing, returning an explicit error.

* **Localization**
  * Added awaiting‑confirmations strings in en, de, es, hr, it, ko, pt, zh-Hans.

* **Tests**
  * Added tests for confirmation parsing, confirmation math, gating behavior, and Blockchair response handling.


VultisigApp/VultisigApp/Features/QBTCClaim/ViewModels/QBTCClaimViewModel.swift
```diff
@@ -338,19 +345,62 @@ final class QBTCClaimViewModel: ObservableObject {
...
+        // The list is empty *purely* because of confirmations when there were
+        // raw UTXOs but the confirmation gate removed all of them.
+        let emptiedByConfirmationGate = !blockchairUtxos.isEmpty && confirmed.isEmpty
+        return FetchedUtxos(utxos: filtered, emptiedByConfirmationGate: emptiedByConfirmationGate)
+    }
+
+    /// Applies the confirmation gate. Fetches `MinUtxoConfirmationBlocks` and
+    /// drops under-confirmed UTXOs; on a param-fetch failure returns the input
+    /// unchanged (fail-open, consistent with `filterClaimable`).
+    private func confirmationGated(
```
johnnyluo: This `confirmationGated(_:btcTipHeight:)` helper is duplicated verbatim in `QBTCClaimEligibilityChecker.swift` — same body, same doc comment, same fail-open warning log. Since both call sites depend only on the `QBTCChainServiceClaimable` protocol, this could be a single default implementation in an `extension QBTCChainServiceClaimable` (or a shared free helper), so the claim screen and the eligibility pre-check can't drift apart later. Non-blocking.

gastonm5: Good catch — and this is already the case on the current branch. `confirmationGated(_:btcTipHeight:)` now lives as a single default implementation on `extension QBTCChainServiceClaimable` (QBTCClaimEligibilityChecker.swift, ~lines 54-79), with the doc comment and the fail-open warning log in that one place. Both call sites depend only on the protocol and use that shared default impl — the claim screen at `QBTCClaimViewModel.swift:370` and the eligibility pre-check at `QBTCClaimEligibilityChecker.swift:225` — so `grep -rn 'func confirmationGated'` returns exactly one definition. The verbatim concrete copy you flagged was already removed in eed4784 (hence this comment showing as outdated), so the two surfaces can no longer drift. No further change needed; resolving.

VultisigApp/VultisigAppTests/Chains/BlockchairServiceQBTCClaimTests.swift
```diff
@@ -0,0 +1,120 @@
...
+        )
+
+        XCTAssertEqual(result.btcTipHeight, 952_877)
+        XCTAssertEqual(result.utxos.count, 1)
+        let utxo = try XCTUnwrap(result.utxos.first)
+        XCTAssertEqual(utxo.txid, Self.txid)
+        XCTAssertEqual(utxo.vout, 0)
+        XCTAssertEqual(utxo.amount, 50_000)
+        XCTAssertEqual(utxo.blockHeight, 952_734)
+    }
```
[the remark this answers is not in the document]
gastonm5: Added in 8aa5dcc57. `testFetchReturnsNilTipWhenContextMissing` constructs a Blockchair response that omits `context` entirely, then asserts `result.btcTipHeight == nil` while the UTXO still comes through (`result.utxos.count == 1`, txid/amount checked) — alongside the existing `testFetchAdaptsUtxosAndReadsTip`. Verified: passed (3/3 in BlockchairServiceQBTCClaimTests, 47/47 across the QBTC suites).

johnnyluo (approved):
LGTM

PR #4522, opened 2026-06-10
fix(solana): decode WalletCore signed tx as base58 and pin base64 on sendTransaction

### LINKED ISSUE #4521: Solana sends/swaps fail after keysign: signed tx is base58 but decoded as base64 (regression from #4421)
## Observed

Solana sends fail after the keysign ceremony completes (verified manually on a current build). The app throws `Failed to decode signed Solana transaction`, broadcast is never attempted, and the full TSS ceremony is wasted. Affected paths: native SOL sends, SPL token sends, ATA-create sends, and aggregator swaps — all route through `SolanaHelper.getSignedTransaction`.

## Cause

`VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift:248`:

```swift
guard let txData = Data(base64Encoded: output.encoded) else {
    throw HelperError.runtimeError("Failed to decode signed Solana transaction")
}
```

`SolanaSigningInput.txEncoding` is never set anywhere in the app (repo-wide grep: zero hits), so WalletCore uses the proto default `Encoding.Base58 = 0` and `output.encoded` is **base58**, not base64. Confirmed empirically against the pinned walletcore-spm 4.3.19 binary: compiling a signed Solana tx returns a base58 string (~367 chars); strict base64 decode fails.

Consequences:
- ~3 of 4 base58 strings have length % 4 != 0 → the guard throws → send/swap fails post-ceremony.
- The remaining ~1 in 4 decode "successfully" (the base58 alphabet is a subset of base64's) into garbage bytes → `transactionHash` from `getHashFromRawTransaction` is junk → the idempotency guard and peer-broadcast dedup in `KeysignViewModel` are silently broken. Broadcast itself still works in that minority case because the base58 `rawTransaction` is passed through unchanged.

Introduced by da464f3ad — "fix(solana): co-sign dApp raw txs via signSolana, not swap reconstruction" (#4421, merged 2026-05-31). The pre-#4421 code passed `output.encoded` straight through without decoding. The regression is present in the `release-v1.39.60` branch.

## Related (same root cause: wire encoding never pinned)

The dApp raw-tx path has the mirror-image bug: `Solana.swift:336` broadcasts `signedTx.base64EncodedString()`, but `SolanaAPI.swift:52` sends `sendTransaction` with no `{"encoding":"base64"}` config object, and the RPC default is base58. Probed live against the Solana RPC proxy: `-32602 invalid base58 encoding`. Any `signSolana` payload with `skipBroadcast == false` fails at broadcast.

## Fix shape

1. Set `SolanaSigningInput.txEncoding` explicitly on the signing input and decode `output.encoded` to match (or simply treat it as base58).
2. Pass an explicit `encoding` option on `sendTransaction` so the wire format is pinned end-to-end for both the send path and the `signSolana` path.

---
Analysis was AI-assisted; the code citations and the failure were verified by hand.

gastonm5:
## Problem

Every Solana send and swap fails after the TSS keysign ceremony with **"Failed to decode signed Solana transaction"** — the signing ceremony completes (and is wasted), then the post-keysign assembly throws before broadcast. The dApp raw path is broken on the other end: `signRawTransaction` returns base64 but `sendTransaction` sent no encoding config, so the RPC decoded it as base58 and rejected the broadcast with `-32602`.

## Root cause

The co-sign raw-tx change switched `SolanaHelper.getSignedTransaction` to decode `SolanaSigningOutput.encoded` with `Data(base64Encoded:)`. But `SolanaSigningInput.txEncoding` is never set anywhere (transfer, token-transfer, create-ATA, and swap `rawMessage` inputs alike), so WalletCore (walletcore-spm 4.3.19) emits **base58** — the proto default. The strict base64 guard therefore throws for every regular send and every swap (Jupiter and THORChain both funnel through the same `getSignedTransaction(coinHexPubKey:inputData:signatures:)` overload).

## Fix (surgical — zero signing-input/MPC changes)

- `Solana.swift` `getSignedTransaction`: decode `output.encoded` with `Base58.decodeNoCheck` and normalize `rawTransaction` to `txData.base64EncodedString()`. `transactionHash` stays base58 of the first 64-byte signature (the correct Solana txid).
- `SolanaAPI.swift` `.sendTransaction`: params are now `[tx, {"encoding": "base64"}]` — pins the wire encoding instead of relying on the RPC's base58 default. Both signing paths (regular + dApp raw) now uniformly hand base64 to a base64-pinned RPC call.
- `getZeroSignedTransaction`: **unchanged** (comment added) — base58 is the Blockaid security-scanner contract (`SecurityScannerTransactionFactory` feeds it straight to Blockaid).
- `signRawTransaction`: no functional change (already base64).

Setting `txEncoding = .base64` on the signing inputs was rejected instead: it would diverge from the sibling platforms' signing inputs (cross-platform pre-image parity is load-bearing for secure-vault co-signing) and would silently flip `getZeroSignedTransaction`'s output, breaking the Blockaid base58 contract.

## Tests

New `VultisigAppTests/Chains/SolanaHelperTests.swift` (6 tests, all passing; `ChainHelperTests` fixture suite still green):

- `testZeroSignedTransactionEncodingIsBase58NotBase64` — pins the WalletCore base58 output contract; fails on any future walletcore bump that changes the default before users hit a broken decode.
- `testGetSignedTransactionReturnsBase64RawTransactionAndBase58Hash` — full path with a synthetic ed25519 keypair (signs the real pre-image so the in-function `publicKey.verify` passes); asserts base64 `rawTransaction`, hash == base58 of the first 64 signature bytes, and that the spliced signature verifies.
- `testGetHashFromRawTransactionReturnsBase58OfFirstSignature` / `testGetHashFromRawTransactionThrowsOnGarbageBytes` — hand-built `[0x01][64-byte sig][message]` fixture + garbage fixtures.
- `testSignRawTransactionSplicesSignatureAndKeepsBase64Encoding` — dApp raw-path round trip: spliced bytes, base64 round trip, base58 hash.
- `testSendTransactionRequestPinsBase64Encoding` — asserts the JSON-RPC params are `[tx, {"encoding": "base64"}]`.

## Verification

- `make generate` (new test file picked up)
- `swiftlint lint` — 0 violations on all 3 changed files
- `xcodebuild test` (iPhone 17 Pro Max sim) — **TEST SUCCEEDED**, 7/7 tests, 0 failures (app target built green as part of the test run)

## Manual validation checklist

- [ ] Native SOL send (default Vultisig proxy RPC)
- [ ] SPL token send — recipient with existing ATA
- [ ] SPL token send — recipient without ATA (create-and-transfer path)
- [ ] Jupiter swap
- [ ] THORChain swap from SOL
- [ ] dApp `signSolana` with `skipBroadcast = false` (app broadcasts)
- [ ] dApp `signSolana` with `skipBroadcast = true` (dApp broadcasts)
- [ ] Two-device secure-vault send (peer dedup / co-sign parity)
- [ ] Custom Solana RPC override (no proxy path)

Closes #4521

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved Solana transaction encoding inconsistency by normalizing format during signing to ensure proper compatibility with RPC submission, improving downstream reliability and transaction handling.

* **Tests**
  * Added comprehensive test suite for Solana transaction signing and encoding, validating transaction formatting, signature handling, and RPC request parameters.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
@@ -49,7 +49,12 @@ struct SolanaAPI: TargetType {
         // HTTPClient owns the JSON serialization.
         switch rpcMethod {
         case .sendTransaction(let encodedTransaction):
-            return .requestParameters(rpcEnvelope(method: "sendTransaction", params: [encodedTransaction]), .jsonEncoding)
+            // Pin the encoding: signed transactions are normalized to base64
+            // before broadcast, while the RPC default is base58.
+            return .requestParameters(
+                rpcEnvelope(method: "sendTransaction", params: [encodedTransaction, ["encoding": "base64"]]),
+                .jsonEncoding
+            )
         case .getBalance(let address):
             return .requestParameters(rpcEnvelope(method: "getBalance", params: [address]), .jsonEncoding)
         case .getRecentPrioritizationFees:
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
@@ -245,11 +245,15 @@ enum SolanaHelper {
                                                                              publicKeys: publicKeys)
         let output = try SolanaSigningOutput(serializedBytes: compileWithSignature)
 
-        guard let txData = Data(base64Encoded: output.encoded) else {
+        // WalletCore emits `output.encoded` as base58: the signing input never
+        // sets `txEncoding`, so the proto default (base58) applies. Normalize
+        // `rawTransaction` to base64 to match the encoding pinned on the
+        // sendTransaction RPC call.
+        guard let txData = Base58.decodeNoCheck(string: output.encoded) else {
             throw HelperError.runtimeError("Failed to decode signed Solana transaction")
         }
         let result = SignedTransactionResult(
-            rawTransaction: output.encoded,
+            rawTransaction: txData.base64EncodedString(),
             transactionHash: try getHashFromRawTransaction(txData: txData)
         )
 
@@ -380,6 +384,8 @@ enum SolanaHelper {
         return Base58.encodeNoCheck(data: sigBytes)
     }
 
+    /// Returns WalletCore's base58 output unchanged — base58 is the Blockaid
+    /// security-scanner contract; do not normalize to base64 here.
     static func getZeroSignedTransaction(keysignPayload: KeysignPayload) throws -> String {
         let coinHexPublicKey = keysignPayload.coin.hexPublicKey
         guard let publicKey = PublicKey(data: Data(hex: coinHexPublicKey), type: PublicKeyType.ed25519) else {
--- a/VultisigApp/VultisigAppTests/Chains/SolanaHelperTests.swift
+++ b/VultisigApp/VultisigAppTests/Chains/SolanaHelperTests.swift
@@ -1,0 +1,227 @@
+//
+//  SolanaHelperTests.swift
+//  VultisigApp
+//
+//  Pins the Solana signed-transaction encoding contract: WalletCore compiles
+//  with the proto-default base58 output (txEncoding is never set on the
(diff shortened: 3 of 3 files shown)


johnnyluo (approved):
LGTM

Issue #4529, opened 2026-06-10
UTXO fee ~6.25x suggested by default; custom byte-fee override never reaches the signed tx


realpaaao:
## Observed

Two related UTXO fee problems on BTC/LTC/BCH/DASH:
1. The **default** fee is ~6.25× Blockchair's suggested next-block rate — systematic overpayment on every send.
2. The user's **custom byte-fee** override (advanced settings) is cosmetic — it's displayed but never reaches the signed transaction.

## Cause

**6.25× default** — `Core/Services/BlockChainService.swift:19-21`:
```swift
static func normalizeUTXOFee(_ value: BigInt) -> BigInt {
    return value * 2 + value / 2 // x2.5 fee
}
```
applied to Blockchair's already-next-block `suggested_transaction_fee_per_byte_sat`, then multiplied again by `feeMode.utxoMultiplier` (`BlockChainService.swift:252-269`). The default mode is `.fast` (`Fee/FeeMode.swift:15`) = 2.5× (`:37`). Net: **2.5 × 2.5 = 6.25×**. Even "Low" (0.75) = 1.875× — users can never pay below ~1.9× suggested. Swaps hardcode `.fast` (`BlockChainService.swift:215,245`).

**Custom byte-fee ignored** — `SendGasSettingsViewModel.swift:24` / `SendDetailsScreen.swift:320` set `customByteFee` and it's shown in the UI (`SendDetailsViewModel.swift:215`), but `SendChainSpecificRequest` (`Features/Send/Interactor/SendInteractor.swift:20-30`) has **no byteFee field**, and `BlockChainSpecific.UTXO(byteFee:)` is only ever populated from `fetchUTXOFee`. A repo-wide grep for `tx.customByteFee` returns zero consumers. The plan/keysign always uses the API-derived 6.25× rate. (EVM's `customGasLimit` IS threaded through — UTXO is the asymmetry.)

## Fix shape

1. Recalibrate the default multiplier (2.5 normalize × 2.5 fast is double-applied; pick one, target ~1.2–1.5× next-block for `.fast`).
2. Thread `customByteFee` into `SendChainSpecificRequest` → `BlockChainSpecific.UTXO(byteFee:)` so the user's override actually signs. Today the advanced fee control is a no-op.

---
Heads up — AI-assisted analysis; verified by hand against current `main`.

johnnyluo:
Let's leave it as it is

PR #4543, opened 2026-06-10
fix(utxo): drop accidental 2.5x double-multiply on UTXO fee (6.25x → 2.5x, match Android)

realpaaao:
## What

Fixes the UTXO fee being **6.25× the suggested next-block rate** by default. Closes the iOS half of #4529.

`normalizeUTXOFee` applied a hardcoded ×2.5, and `fetchUTXOFee` then multiplied the result **again** by `feeMode.utxoMultiplier` (`.fast` = 2.5). Default sends and all swaps (which hardcode `.fast`) therefore paid 2.5 × 2.5 = **6.25×**.

## Change

Remove the redundant `normalizeUTXOFee` step; apply only the fee-mode tier to the API rate (which is already a next-block suggestion):

| Mode | Before | After |
|------|--------|-------|
| Low | 1.875× | 0.75× |
| Normal | 2.5× | 1× |
| **Fast (default)** | **6.25×** | **2.5×** |

2.5× for the default matches Android's `UtxoFeeService` (`gas × 5 / 2`). DOGE keeps its separate `/10` path (unchanged).

Cross-device safe: the computed `byteFee` is baked into `BlockChainSpecific.UTXO` and shared via the keysign payload, so both devices sign the same plan — this only changes how the initiator derives the rate.

## receipts

- **Scope:** one function removed (single caller, no other refs repo-wide — `grep -rn normalizeUTXOFee` confirms), one expression simplified.
- **Math verified by hand:** `.fast` default 2.5 × 2.5 = 6.25 → 2.5; matches Android `UtxoFeeService.kt` non-DOGE `gas.multiply(5).divide(2)`.
- **Not compiled locally** (no iOS build env on this machine) — relying on CI to build. Change is a 3-line arithmetic simplification with no new symbols or API changes.

Heads up — AI-assisted change; reviewer please confirm the CI build + a real UTXO send fee looks right before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTXO fee calculation logic for more accurate transaction prioritization across supported networks.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/BlockChainService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/BlockChainService.swift
@@ -16,10 +16,6 @@ struct BlockSpecificCacheItem {
 }
 final class BlockChainService {
 
-    static func normalizeUTXOFee(_ value: BigInt) -> BigInt {
-        return value * 2 + value / 2 // x2.5 fee
-    }
-
     static func normalizeEVMFee(_ value: BigInt) -> BigInt {
         let normalized = value + value / 2 // x1.5 fee
         return max(normalized, 1) // To avoid 0 miner tips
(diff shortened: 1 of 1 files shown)


johnnyluo:
Let's leave as it is for now , it is a bit high , but we can guarantee the tx get commit to block faster

johnnyluo:
Also , as suggested by coderabbitai , it might potentially make use pay 0 sats as well , if they choose slow

PR #4553, opened 2026-06-11
fix(doge): apply fee tier + Android-aligned base instead of flat /10 divisor

### LINKED ISSUE #4545: DOGE fee ignores fee tier and underprices vs Android (flat /10 magic divisor, no feeMode)
## Observed

Dogecoin sends use a bespoke fee path that (a) ignores the user's fee tier (Low/Normal/Fast all produce the same fee) and (b) ends up materially cheaper than the equivalent Android rate. Surfaced while fixing the non-DOGE UTXO double-multiply (#4529 / PR #4543) — the else-branch is now a clean tier-only model, which makes the DOGE special-case the odd one out.

## Cause

`Core/Services/BlockChainService.swift:257-261` (`fetchUTXOFee`):
```swift
if coin.chain == .dogecoin {
    // For DOGE, the API returns 500k sats/byte which is too high for WalletCore
    // Use a much lower value that WalletCore can work with: divide by 10
    result = sats / 10 // 500k / 10 = 50k sats/byte (still high but workable)
} else {
    let prioritized = Float(sats) * feeMode.utxoMultiplier   // honors fast/normal/low
    result = BigInt(prioritized)
}
```

The DOGE branch divides the API rate by 10 (≈0.1×) and **never applies `feeMode.utxoMultiplier`**, so the fee tier selector is a no-op for DOGE. Android's equivalent (`UtxoFeeService.kt`, `gas.multiply(5).divide(20)`) is ≈0.25× — so iOS DOGE is also ~2.5× cheaper than Android, on a chain where underpaying risks slow/stuck confirmations. The `/10` is an undocumented magic divisor, not a tier.

## Fix shape

Bring DOGE into the same tier model as the other UTXO chains: pick a base multiplier that matches Android's ~0.25× intent and apply `feeMode.utxoMultiplier` on top (so Low/Normal/Fast actually do something), rather than a flat `/10`. Confirm the resulting sat/byte against a live DOGE next-block rate before shipping.

Related: PR #4543 (non-DOGE 6.25×→2.5×), Android DOGE fee issue vultisig-android#4855.

---
Heads up — AI-assisted analysis; verified against current vultisig-ios main.

gastonm5:
## Problem

Dogecoin sends use a bespoke fee path that (a) ignores the user's fee tier — Low/Normal/Fast all produce the **same** fee — and (b) underprices vs Android by ~2.5×. On DOGE, underpaying risks slow/stuck confirmations.

## Root cause

`BlockChainService.fetchUTXOFee` special-cased DOGE with `result = sats / 10` (≈0.1×) and never applied `feeMode.utxoMultiplier`, so the tier selector was a no-op. Android (`UtxoFeeService.kt`) uses `gas.multiply(5).divide(20)` = **0.25×** for DOGE, so iOS at 0.1× was ~2.5× cheaper.

> Note: the issue's premise that the else-branch is "now a clean tier-only model (post-#4543)" does **not** hold against current `main` — PR #4543 was **closed unmerged**, so the else-branch still has the 6.25× double-multiply bug (#4529). #4545 is scoped to DOGE only, so this PR makes the DOGE branch self-contained rather than mirroring the still-buggy else-branch. The non-DOGE double-multiply is explicitly out of scope.

## Fix

Replace the magic `/10` with a pure, tier-aware helper:

```swift
static let dogeBaseFeeMultiplier: Float = 0.25
static func dogeByteFee(suggestedSatsPerByte sats: BigInt, feeMode: FeeMode) -> BigInt {
    let prioritized = Float(sats) * dogeBaseFeeMultiplier * feeMode.utxoMultiplier
    return BigInt(prioritized)
}
```

Base = **0.25** matches Android's DOGE rate; iOS `.normal` has `utxoMultiplier = 1.0`, so `.normal` lands exactly on Android's number and Low/Fast fan out around it.

## sat/byte cross-check vs live next-block rate

Live `suggested_transaction_fee_per_byte_sat` for dogecoin (fetched 2026-06-11): **500,000**.

| Path | Multiplier | sat/byte @ 500k |
|------|-----------|------------------|
| Raw API suggestion | 1.0× | 500,000 |
| iOS **old** (`/10`) | 0.1× | 50,000 |
| Android DOGE (`gas*5/20`) | 0.25× | 125,000 |
| **New** `.safeLow` (0.25 × 0.75) | 0.1875× | 93,750 |
| **New** `.normal` (0.25 × 1.0) | 0.25× | **125,000** (= Android) |
| **New** `.fast`, default (0.25 × 2.5) | 0.625× | 312,500 |

`.normal` reproduces Android's 0.25× exactly; tiers now differ; default `.fast` is no longer underpriced. Max tier (312,500) stays below the raw 500k.

**WalletCore constraint:** the old "/10 because raw is too high for WalletCore" comment does not hold — `UTXOChainsHelper` casts `byteFee` to `Int64` (no cap, 500k fits trivially); the divisor was just an undocumented underprice.

## Verification

- `swiftlint` — 0 violations on changed files.
- `make generate` re-ran (new test file).
- `xcodebuild test -only-testing:VultisigAppTests/DogeFeeTierTests` (worktree-local `-derivedDataPath`): **4/4 pass**, full app target compiled (`** TEST SUCCEEDED **`).
- New `DogeFeeTierTests`: `.normal` == Android 125k; tiers distinct; fix pricier than old `/10`; `.fast` below raw rate.

## Risks

- 0.25× base validated against a **single** live snapshot (500k sats/byte, 2026-06-11). If Blockchair ever reports DOGE at a different (already-normalized) scale, 0.25× could mis-price — but Android applies the identical 0.25× against the same endpoint, so iOS/Android stay in lockstep regardless of absolute scale.
- Does not fix the non-DOGE 6.25× double-multiply (out of scope; #4529 / closed PR #4543).

Closes #4545

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Dogecoin transaction fee calculation with refined tier-based fee scaling for more accurate fee estimates.

* **Tests**
  * Added comprehensive test suite validating Dogecoin fee tier calculations across all fee modes.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/BlockChainService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/BlockChainService.swift
@@ -20,6 +20,20 @@ final class BlockChainService {
         return value * 2 + value / 2 // x2.5 fee
     }
 
+    /// Dogecoin's `suggested_transaction_fee_per_byte_sat` from Blockchair is
+    /// reported at the chain's relay-floor scale (~500k sats/byte), an order of
+    /// magnitude above the rate other UTXO chains report. Android scales it by
+    /// 0.25 (`gas * 5 / 20`) to land on a sane next-block rate; this base
+    /// multiplier reproduces that intent and the fee mode tier is applied on
+    /// top, so Low/Normal/Fast actually differ. At a live 500k sats/byte:
+    /// Low ≈ 93,750, Normal = 125,000 (== Android), Fast = 312,500 sats/byte.
+    static let dogeBaseFeeMultiplier: Float = 0.25
+
+    static func dogeByteFee(suggestedSatsPerByte sats: BigInt, feeMode: FeeMode) -> BigInt {
+        let prioritized = Float(sats) * dogeBaseFeeMultiplier * feeMode.utxoMultiplier
+        return BigInt(prioritized)
+    }
+
     static func normalizeEVMFee(_ value: BigInt) -> BigInt {
         let normalized = value + value / 2 // x1.5 fee
         return max(normalized, 1) // To avoid 0 miner tips
@@ -252,12 +266,11 @@ final class BlockChainService {
     func fetchUTXOFee(coin: Coin, feeMode: FeeMode) async throws -> BigInt {
         let sats = try await utxo.fetchSatsPrice(coin: coin)
 
-        // DOGE has extremely high base fees from API, need to reduce significantly
         let result: BigInt
         if coin.chain == .dogecoin {
-            // For DOGE, the API returns 500k sats/byte which is too high for WalletCore
-            // Use a much lower value that WalletCore can work with: divide by 10
-            result = sats / 10 // 500k / 10 = 50k sats/byte (still high but workable)
+            // DOGE reports its rate at the relay-floor scale; rescale to a sane
+            // next-block rate and apply the fee mode tier (see dogeByteFee).
+            result = Self.dogeByteFee(suggestedSatsPerByte: sats, feeMode: feeMode)
         } else {
             // For other chains, use normal normalization and multipliers
             let normalized = Self.normalizeUTXOFee(sats)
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4554, opened 2026-06-11
fix(sui): select coin objects by exact coinType + paginate getAllCoins

### LINKED ISSUE #4538: SUI coin-object selection by ticker substring -> wrong objects in native sends, some tokens unsendable
## Observed

Two SUI send failures driven by coin-object selection by **ticker substring**:
1. A native SUI send from a wallet that *also* holds any coin whose type string contains "SUI" (sSUI, haSUI and other LSTs, SUI-named memecoins) can feed **non-SUI object references** into the transaction → invalid tx → dry-run fails (silently) → broadcast fails post-ceremony (and per the error-as-txid bug, may be reported as success).
2. A token whose on-chain type doesn't contain its display ticker (classic case: Wormhole-bridged assets typed `...::coin::COIN`) matches nothing → `"Non-native token transaction requires the token to be present"` → **unsendable**, and its balance also mis-displays.

## Cause

Filtering is `coinType.contains(ticker)` in three places:
- `Blockchain/Sui/Service/SuiService.swift:160`: `coins.filter { $0.coinType.uppercased().contains("SUI") || ...contains(coin.ticker.uppercased()) }`
- `Blockchain/Sui/Signing/Sui.swift:38` (native) and `:73` (token): same `.contains(...ticker...)` filter.

Substring matching can't distinguish `0x2::sui::SUI` from `0x…::xSUI::XSUI`, and fails for tokens whose type symbol ≠ display ticker.

Also `SuiService.getAllCoins` (`SuiService.swift:157`) does not paginate `suix_getAllCoins` (default page ~50 objects), so heavy wallets get a truncated object set.

## Fix shape

Match coin objects by **exact `coinType`** (the fully-qualified `address::module::struct`), carried on the `Coin` record, not by ticker substring. Paginate `suix_getAllCoins`. This is the same exact-equality-over-substring lesson as past review-matching bugs.

---
Heads up — AI-assisted analysis; the `.contains(ticker)` filters verified by hand against current `main` (trigger requires a matching token in the vault).

gastonm5:
## Problem

SUI sends can fail or be impossible because coin-object selection filters by `coinType.contains(ticker)`:

1. **Native SUI sends pull in wrong objects.** A wallet that also holds any coin whose type string contains "SUI" (sSUI / haSUI and other LSTs, SUI-named memecoins) feeds non-SUI object references into the native `paySui` input → invalid tx → dry-run fails silently → broadcast fails *after* the TSS ceremony.
2. **Some tokens are unsendable.** A token whose on-chain type doesn't contain its display ticker (classic case: Wormhole-bridged assets typed `…::coin::COIN`) matches nothing → `"Non-native token transaction requires the token to be present"`, and its balance mis-displays.

`suix_getAllCoins` was also unpaginated (default page ~50 objects), truncating the object set on heavy wallets.

## Root cause

Substring matching cannot distinguish `0x2::sui::SUI` from `0x…::xsui::XSUI`, and breaks when a token's on-chain struct symbol differs from its display ticker. The three filters lived at `SuiService.getAllCoins` and the native + token branches of `SuiHelper.getPreSignedInputData`.

The `Coin` `@Model` has no `coinType` field — the fully-qualified `address::module::struct` type is carried in `Coin.contractAddress` (empty for native SUI, whose canonical type is `0x2::sui::SUI`). The coin-object dicts already carry the node's `coinType`, so exact matching is possible on both sides.

## Fix

- **`SuiCoinType`** (new, in `SuiConstants.swift`): normalization-aware exact matching. `normalize(_:)` collapses the package-address segment so the short `0x2` and long `0x0000…0002` forms compare equal; `matches`, `isNative`, and `expectedType(isNativeToken:contractAddress:)` build on it.
- Replaced all three substring filters in `Sui.swift` with exact matching (native → `isNative`; token → `matches(expectedType)`; gas → `isNative`).
- `getAllCoins` now returns **all** owned coin objects (so a token send still sees its SUI gas objects) and **paginates** via `nextCursor` / `hasNextPage`. Touched `print` calls converted to `OSLog`.
- `getAllBalances` substring filter (balance display, not object selection) left unchanged to keep the fix focused.

### `0x2` normalization

`normalize` lowercases the type, strips `0x`, drops leading zeros from the address segment, and re-prefixes `0x`, leaving module/struct intact. So both `0x2::sui::SUI` and the node's `0x0000…0002::sui::SUI` normalize identically, while `…::xsui::XSUI` stays distinct.

## Verification

- New unit test `VultisigAppTests/Chains/SuiCoinTypeTests.swift` (9 cases): native excludes xSUI / `::coin::COIN` / SUI-named look-alikes, address-form normalization (short ↔ long), bridged `::coin::COIN` token selection + native gas split. **9/9 pass** (iPhone 16 Pro sim, worktree-local derived data).
- `swiftlint` clean on all changed files; `make generate` run for the new test file.

Closes #4538

## Summary by CodeRabbit

* **New Features**
  * Canonical native SUI coin type and robust coin-type matching/normalization for precise coin selection

* **Bug Fixes**
  * More accurate SUI coin selection and gas picking using normalization-aware matching
  * Improved coin fetching to follow paginated results reliably and surface decoding/RPC errors
  * SUI-specific payload now includes the correct set of coin objects for sends

* **Tests**
  * Added tests for coin-type selection and paginated coin fetching (including failure cases)

* **Chores**
  * Enhanced logging for coin retrieval failures

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
@@ -176,8 +176,8 @@ class SuiService {
                 )
 
                 guard let coins: [SuiCoin] = Utils.extractResultFromJson(fromData: data, path: "result.data") else {
-                    logger.error("Failed to decode coins")
-                    break
+                    logger.error("Failed to decode coin page at cursor \(cursor ?? "<start>", privacy: .public)")
+                    throw Errors.coinPageDecodeFailed(cursor: cursor)
                 }
 
                 allCoins.append(contentsOf: coins.map { suiCoin in
@@ -347,6 +347,7 @@ private extension SuiService {
         case simulationFailed(String)
         case failedToParseGasEstimate
         case dryRunFailed(String)
+        case coinPageDecodeFailed(cursor: String?)
 
         var errorDescription: String? {
             switch self {
@@ -358,6 +359,8 @@ private extension SuiService {
                 return "Failed to parse gas estimate from dry run"
             case .dryRunFailed(let error):
                 return "Dry run failed: \(error)"
+            case .coinPageDecodeFailed(let cursor):
+                return "Failed to decode coin page from suix_getAllCoins at cursor \(cursor ?? "<start>"). Aborting to avoid a truncated coin set."
             }
         }
     }
--- a/VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
@@ -8,10 +8,13 @@
 import Foundation
 import SwiftUI
 import BigInt
+import OSLog
 import WalletCore
 
 class SuiService {
     static let shared = SuiService()
+
+    private let logger = Logger(subsystem: "com.vultisig.app", category: "sui-service")
 
     /// Default Sui JSON-RPC host.
     static let defaultRPCURL: URL = {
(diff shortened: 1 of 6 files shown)


VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
```diff
@@ -151,31 +154,51 @@ class SuiService {
...
+                let data = try await Utils.PostRequestRpc(
+                    rpcURL: rpcURL,
+                    method: "suix_getAllCoins",
+                    params: [coin.address, cursor]
+                )
+
+                guard let coins: [SuiCoin] = Utils.extractResultFromJson(fromData: data, path: "result.data") else {
+                    logger.error("Failed to decode coins")
+                    break
+                }
```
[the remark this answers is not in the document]
gastonm5: Good catch — fixed in e3d83b9e0.

`getAllCoins` now throws `Errors.coinPageDecodeFailed(cursor:)` when a page can't be decoded mid-pagination instead of `break`ing and returning the pages gathered so far. A truncated coin set is worse than failing: downstream selection in `SuiHelper` would silently miss the SUI gas object or the token's objects and build an invalid tx — exactly the silent-failure class this PR closes (#4538). The error carries the failing cursor for diagnostics and propagates to the caller so tx building aborts.

Happy-path pagination is unchanged: it follows `nextCursor`/`hasNextPage` and terminates on `hasNextPage == false`.

Added `SuiServiceGetAllCoinsTests` (URLProtocol-stubbed) covering both full pagination to completion and a mid-pagination decode failure that throws rather than returning the first page.

VultisigApp/VultisigApp/Blockchain/Sui/Service/SuiService.swift
```diff
@@ -151,31 +154,51 @@ class SuiService {
...
 
-                return allCoins
-            } else {
-                print("Failed to decode coins")
-            }
+                let hasNextPage = Utils.extractResultFromJson(fromData: data, path: "result.hasNextPage") as? Bool ?? false
+                cursor = hasNextPage ? Utils.extractResultFromJson(fromData: data, path: "result.nextCursor") as? String : nil
+            } while cursor != nil
+
+            return allCoins
```
johnnyluo, line 197: **Keysign payload / QR bloat (medium).**

`getAllCoins` now returns *every* owned coin object (all `coinType`s), and that full set flows unfiltered into the keysign payload: `BlockChainService.swift:486` builds `.Sui(coins: allCoins, …)` into the `KeysignPayload`, which is serialized into the **pairing QR code and the TSS relay payload**. The exact-type selection only happens later in `SuiHelper.getPreSignedInputData`.

Consequences:
- For a **native SUI send**, the payload now carries every memecoin/LST/token object the wallet holds — none of which the native tx needs.
- For a **token send**, it carries every *other* unrelated token too, not just that token + SUI gas.
- Combined with removing the ~50-object page cap (pagination), a heavy wallet's payload and QR can grow substantially — and QR density directly affects pairing-scan reliability.

Suggestion: filter to `native SUI ∪ {the specific token type}` before embedding into the payload (the `coin` context is available in `BlockChainService`), rather than carrying the whole wallet across the wire. That preserves the fix's intent (token sends still get their SUI gas objects) without the bloat. As written, pagination and the unfiltered return work against each other — you fetch *more* completely, then ship *all* of it.

gastonm5: Fixed in `fec16f1ab`. The payload now carries only what the signer consumes, not the whole wallet.

At the Sui payload-build site (`BlockChainService` `.sui` branch) I filter `getGasInfo`'s coin set down to **native SUI ∪ {the target token's exact coinType}** before it goes into `.Sui(coins:)` / the `KeysignPayload`. The `coin` context is right there, so the filter uses the same matcher `SuiHelper.getPreSignedInputData` does:

- **Native SUI send** → only the native objects (which are both the inputs and the gas). No memecoin/LST objects.
- **Token send** → the target token's objects ∪ the native SUI objects (needed for gas). No *other* held tokens.

This is exactly the set `getPreSignedInputData` selects downstream, so nothing it needs is dropped (native SUI for gas is preserved on token sends). `getAllCoins` itself is unchanged — it still fetches/paginates *all* owned objects (gas-object discovery + the fail-loud pagination fix stand); the trim happens at assembly, not by re-capping the RPC. So pagination and the payload no longer work against each other.

New helper `SuiCoinType.payloadCoins(_:isNativeToken:contractAddress:)` does the filtering; tests pin that a native send embeds only native objects, a token send embeds native ∪ that token only (no unrelated tokens), and multiple matching objects aren't capped. `SuiCoinTypeTests` (12) + `SuiServiceGetAllCoinsTests` (2) green.

johnnyluo (approved):
LGTM

PR #4555, opened 2026-06-11
fix(keysign): cancel orphaned signing body on stage timeout & gate broadcast (#4541)

### LINKED ISSUE #4541: Keysign 90s timeout doesn't cancel signing body; orphaned ceremony broadcasts unconditionally -> double-broadcast vector
## Observed

The keysign 90-second stage timeout does not actually stop the in-flight signing ceremony. The abandoned ceremony body runs to completion and **broadcasts unconditionally**, while the user — shown a "try again" retry view — may have already started a second ceremony. On nonce/gas-varying chains (EVM) the two ceremonies produce different tx hashes, so the idempotency guard misses and **both transactions can land**.

## Cause

`KeysignViewModel.swift:246-264` races the keysign body against a 90s timeout (`keysignStageTimeout`, `:70`); on timeout it sets `.KeysignRetryRequested` and calls `group.cancelAll()`. But the DKLS/Schnorr/Dilithium bodies **never check `Task.isCancelled`** (grep of all three + `DKLSMessenger` → zero cancellation checks) — their loops gate only on a 60s elapsed-time check (`DKLSKeysign.swift:252-253`). So `cancelAll()` is silently ignored and the body runs to its own completion.

When the orphaned body finishes, `startKeysignDKLS` calls `await broadcastTransaction()` **unconditionally** (`:404`) — the `isTerminalStatus` guard at `:411` protects only the `.KeysignFinished` status write, not the broadcast.

Meanwhile the user sees the retry view (`KeysignView.swift:140-153`) and can re-sign. `retryBroadcast` rebuilds the payload from scratch (re-fetched gas/nonce, possibly reselected UTXOs) → **different tx hash** → the idempotency guard (`:778-784`, keyed on `transactionType.transactionHash`) doesn't match → both can land where chain-level protection is weak. (On deterministic-payload chains the rebuilt hash matches and the guard fires — so the double-broadcast risk is chain-dependent, worst on EVM.)

Compounding: a single message can legitimately exceed 90s (4 attempts × 60s per message, `DKLSKeysign.swift:391,417-421`, messages signed sequentially), so on a slow network the timeout fires *while signing is still healthy*, manufacturing the race rather than catching a hang.

## Fix shape

- Make the signing bodies cooperatively cancellable: check `Task.isCancelled` in the poll loops so `cancelAll()` actually stops them.
- Gate `broadcastTransaction()` on `!isTerminalStatus(status)` so an orphaned/timed-out body can't broadcast.
- Reconsider the 90s stage timeout vs. the 4×60s-per-message worst case — they're inconsistent and the timeout fires on healthy slow ceremonies.

---
Heads up — AI-assisted analysis; the missing cancellation checks and unconditional broadcast verified by hand. The double-broadcast outcome is chain-dependent (worst on EVM) and timing-dependent — worth a repro before estimating impact.

gastonm5:
## Problem

The keysign 90s stage timeout does not actually stop the in-flight signing ceremony. The abandoned body runs to completion and **broadcasts unconditionally**, while the user — shown the retry view — may have started a second ceremony. On nonce/gas-varying chains (EVM) the two ceremonies produce different tx hashes, so the idempotency guard (keyed on `transactionHash`) misses and **both transactions can land**.

Closes #4541.

## Root cause (verified against current `main` source)

1. **Cancellation silently ignored.** `KeysignViewModel.startKeysign()` races the body against a stage-timeout sleeper in a task group; on timeout it sets `.KeysignRetryRequested` and calls `cancelAll()`. But the poll loops — `pullInboundMessages` and the unbounded `processXxxOutboundMessage` loop in `DKLSKeysign` / `SchnorrKeysign` / `DilithiumKeysign` — had **zero** `Task.isCancelled` / `checkCancellation()` checks. `cancelAll()` was a no-op.
2. **Broadcast was unconditional.** When the orphaned body finished, `startKeysignDKLS` (and `startKeysignGG20`) called `await broadcastTransaction()` unconditionally; the `isTerminalStatus` guard only protected the *post*-broadcast `.KeysignFinished` write.
3. **Idempotency guard is chain-dependent.** The user's retry rebuilds the payload (refetched gas/nonce, possibly reselected UTXOs). On EVM the rebuilt hash differs → guard misses → double-broadcast. On deterministic-payload chains the hash matches and the guard already fires. **Worst on EVM.**
4. **Timeout manufactured the race.** Messages sign sequentially and each inbound poll caps at 60s/attempt (×3 retries), so a flat 90s budget could expire while a healthy multi-message ceremony was still progressing.

## Fix (control flow only — cryptographic protocol untouched)

- **Cooperative cancellation:** `try Task.checkCancellation()` at the head of each poll-loop iteration (`pullInboundMessages`, `processXxxOutboundMessage`) in all three signing libs; the per-message retry loops now **rethrow** `CancellationError` instead of treating it as a signing failure and retrying. Same boundary handling for the GG20 path in `KeysignViewModel`.
- **Broadcast gate:** `startKeysignDKLS` and `startKeysignGG20` now gate `broadcastTransaction()` on `!isTerminalStatus(status) && !Task.isCancelled`; `startKeysignDKLS` adds a `catch is CancellationError` arm so a propagated cancel doesn't overwrite the timeout's terminal status with `.KeysignFailed`.
- **Timeout rescale:** the flat `keysignStageTimeout` becomes a message-count-scaled budget (`90s/message * max(count,1) + 30s` broadcast headroom), snapshotted before entering the task group.

## Verification

- `make generate` (new test file), SwiftLint clean on all changed files (0 violations).
- Built the full app and ran the keysign test suites on a worktree-local derived-data path: **`** TEST SUCCEEDED **` — 16/16** (`KeysignTimeoutCancellationTests` 5 new + `KeysignBroadcastCancellationTests` 11 existing).

## Test notes

New `KeysignTimeoutCancellationTests`:
- `testDKLSPullInboundMessagesThrowsCancellationWhenTaskCancelled` — injects an empty-polling `HTTPClientProtocol` fake to keep the loop stalled, cancels the task, asserts the loop surfaces `CancellationError` (the core cooperative-cancellation fix).
- `testStageTimeoutScalesWithMessageCount` / `…FloorForEmptyMessageList` — the timeout budget grows with message count and stays positive for an empty list.
- Broadcast-gate predicate tests — `.KeysignRetryRequested` is terminal so the gate fires and the finish guard can't overwrite it.

The full orphan-vs-retry double-broadcast race needs a live relay and can't be exercised in a unit test; it's covered by inspection + a repro note in the wiki plan.

## Review flag

⚠️ This is **signing-path control flow** (TSS keysign ceremony + broadcast gating), HIGH security tier. The crypto protocol, message ordering, and signature handling are untouched, but cancellation timing against real relays warrants careful **human + on-device review** before merge.

## Release Notes

* **Bug Fixes**
  * Improved cancellation handling in keysign ceremonies for faster, more reliable operation unwinding
  * Enhanced timeout logic now scales dynamically with message complexity
  * Added safeguards to prevent broadcast during timeout or cancellation scenarios

* **Tests**
  * Added comprehensive test coverage for keysign timeout and cancellation behaviors

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/States/Keysign/DKLSKeysign.swift
+++ b/VultisigApp/VultisigApp/Blockchain/States/Keysign/DKLSKeysign.swift
@@ -184,6 +184,11 @@ final class DKLSKeysign {
 
     func processDKLSOutboundMessage(handle: godkls.Handle) async throws {
         repeat {
+            // Cooperative cancellation: the stage-level timeout in
+            // KeysignViewModel cancels this task group on a stall. Without an
+            // explicit check the C-backed loop runs to completion and the
+            // orphaned ceremony broadcasts behind the user's back.
+            try Task.checkCancellation()
             let (result, outboundMessage) = GetDKLSOutboundMessage(handle: handle)
             if result != DKLS_LIB_OK {
                 print("fail to get outbound message,\(result)")
@@ -220,6 +225,10 @@ final class DKLSKeysign {
         var isFinished = false
         let start = DispatchTime.now()
         repeat {
+            // Cooperative cancellation: honour cancelAll() from the stage-level
+            // timeout so a stalled poll stops instead of running to completion
+            // and broadcasting an abandoned ceremony.
+            try Task.checkCancellation()
             let response: HTTPResponse<Data>
             do {
                 response = try await httpClient.request(TssRelayAPI(
@@ -387,6 +396,9 @@ final class DKLSKeysign {
                 try await Task.sleep(for: .milliseconds(500))
             }
         } catch {
+            // A cancellation is not a signing failure — never retry it,
+            // propagate so the abandoned ceremony unwinds immediately.
+            if error is CancellationError { throw error }
             logger.error("Failed to sign message (\(messageToSign, privacy: .public)) on attempt \(attempt, privacy: .public): \(error.localizedDescription, privacy: .public)")
             if attempt < 3 {
                 try await DKLSKeysignOneMessageWithRetry(attempt: attempt+1, messageToSign: messageToSign)
--- a/VultisigApp/VultisigApp/Blockchain/States/Keysign/DilithiumKeysign.swift
+++ b/VultisigApp/VultisigApp/Blockchain/States/Keysign/DilithiumKeysign.swift
@@ -168,6 +168,11 @@ final class DilithiumKeysign {
 
     func processDilithiumOutboundMessage(handle: vscore.Handle) async throws {
         repeat {
+            // Cooperative cancellation: the stage-level timeout in
+            // KeysignViewModel cancels this task group on a stall. Without an
+            // explicit check the C-backed loop runs to completion and the
+            // orphaned ceremony broadcasts behind the user's back.
+            try Task.checkCancellation()
             let (result, outboundMessage) = GetDilithiumOutboundMessage(handle: handle)
             if result != MLDSA_LIB_OK {
                 print("fail to get outbound message,\(result)")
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4556, opened 2026-06-11
fix(swap): send tolerance_bps so THORChain/Maya swaps carry a minimum-output limit

### LINKED ISSUE #4539: THORChain/Maya swaps signed with no minimum-output limit (toAmountLimit:0, no tolerance_bps) -> unbounded slippage
## Observed

THORChain and MayaChain swaps are signed with **no minimum-output limit** — the swap fills at whatever the pool gives, with no slippage floor. Combined with the 30s+ ceremony plus unbounded pairing time, price can move arbitrarily between quote and execution. This is the largest fund-protection gap in the swap stack.

## Cause

The quote request never sends `tolerance_bps` — `Blockchain/THORChain/Service/ThorchainMainnetAPI.swift:156-167` builds only asset/amount/destination/streaming/affiliate params (repo-wide grep for `tolerance` finds only Kyber's `slippageTolerance` and a SwapKit comment). Without it, THORChain returns the memo with an empty/zero LIM field.

The payload then hardcodes the limit to zero — `Features/Swap/Logic/SwapPayloadBuilder.swift:102-104`:
```swift
toAmountLimit: "0",
streamingInterval: String(provider.streamingInterval),
streamingQuantity: "0",
```
and the client signs that memo verbatim:
- EVM router: `evm.swift:192-203` (`depositWithExpiry` memo param)
- UTXO: `UTXOChainsHelper.swift:83-85` (`outputOpReturn = memoData`)
- RUNE source: `thorchain.swift` (`memo = keysignPayload.memo ?? ""`)

`streamingQuantity: "0"` means streaming swaps also run with no abort-and-refund floor on price collapse.

## Fix shape

Pass a `tolerance_bps` on the quote request and set a real `toAmountLimit` in the THORChain/Maya memo (derived from the quoted output minus the user's slippage tolerance). Cross-check Android/SDK for parity — the swap-routing/limit logic has diverged per platform before, so confirm whether they already set a limit before assuming this is iOS-only.

---
Heads up — AI-assisted analysis; `toAmountLimit: "0"` / `streamingQuantity: "0"` verified verbatim against current `main`.

gastonm5:
## Problem

THORChain and MayaChain swaps are signed with **no minimum-output limit** — the swap fills at whatever the pool gives, with no slippage floor. Combined with the 30s+ ceremony and unbounded pairing time, price can move arbitrarily between quote and execution. This is the largest fund-protection gap in the swap stack.

## Root cause

The signed swap memo is the THORChain/Maya-returned `memo` string, verbatim, on every source chain (RUNE `thorchain.swift`, UTXO `UTXOChainsHelper.swift`, EVM router `evm.swift`). The on-chain minimum-output floor lives in the memo's `LIM` field.

The quote request never sent `tolerance_bps` (`ThorchainMainnetAPI`/`MayaChainAPI`/`ThorchainStagenetAPI` `swapQuote` built only asset/amount/destination/streaming/affiliate), so the node returned a memo with `LIM=0` (unbounded). The proto payload's `toAmountLimit`/`streamingQuantity` were hardcoded to `"0"`, but those fields do **not** rebuild the signed memo — iOS does not use WalletCore's native `THORChainSwap.build` path — so patching only those would be cosmetic.

## Fix

- Add `tolerance_bps` to the `swapQuote` endpoint of all three THORChain/Maya API targets and thread a `toleranceBps` parameter through `ThorchainSwapProvider.fetchSwapQuotes` and every service impl (THORChain, Maya, both stagenet services).
- `SwapService` passes `defaultThorchainToleranceBps` (100 bps / 1%) on both the rapid and streaming quote fetches, so the node bakes a real `LIM` into the returned memo. This also gives streaming swaps an abort-and-refund floor.
- `buildThorchainSwapPayload` now derives a matching `toAmountLimit` and a real `streamingQuantity` for the proto payload (previously `"0"`) so cross-device display stays consistent with the signed memo.

## Derivation

Node-authoritative: `LIM = expected_amount_out × (1 − tolerance_bps/10_000)`, baked into the memo by THORChain when `tolerance_bps` is present. Preferred over computing `LIM` client-side and rewriting the memo string (which would risk drift from the node's trimming/streaming-suffix format and reintroduce a sign-vs-display mismatch). The payload's `toAmountLimit` mirrors that formula (`floor`).

No per-swap user slippage UI exists for swaps; 100 bps matches the existing anti-rekt streaming threshold and is a conservative default.

## Cross-platform parity (verified, not assumed)

All three platforms share the **same** gap and the **same** anti-rekt streaming-upgrade mechanism, and **none** sent `tolerance_bps`:
- **SDK/Windows** (`packages/core` `getNativeSwapQuote.ts`): no `tolerance_bps`; `nativeSwapQuoteToSwapPayload.ts` hardcodes `toAmountLimit: '0'`; keysign signs `native: ({ memo }) => memo`.
- **Android** (`ThorChainQuoteSource` + swap `build.ts`): same 1% streaming cutoff; memo signed verbatim.
- **iOS**: as above.

So this is not iOS-specific divergence — there was no "correct" platform to copy a derivation from. iOS leads; SDK and Android need the same change in follow-ups.

## Verification

- `swiftlint` clean on all changed files.
- `xcodebuild test` (worktree-local derivedDataPath, iPhone 16 / iOS 18.4): `SwapPayloadBuilderTests` + `ThorchainAntiRektTests` — **33 tests, 0 failures**, including 5 new limit-math/tolerance assertions.

## Risk

The default tolerance value (100 bps) is a product decision — it matches the streaming threshold and is a conservative floor, but Product should confirm the exact number and whether it should eventually be user-configurable.

Closes #4539

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed swap quote requests to enforce configurable slippage tolerance, preventing unbounded minimum output floors on THORChain and MayaChain swaps.

* **New Features**
  * Swap payloads now calculate protective minimum output amounts based on tolerance thresholds, improving protection against unfavorable price execution.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -22,6 +22,14 @@ struct SwapService {
     /// and `vultisig-android` (`STREAMING_SLIPPAGE_THRESHOLD_BPS`).
     static let streamingSlippageThresholdBps = 100
 
+    /// Minimum-output tolerance (basis points) sent on every THORChain/Maya
+    /// quote request as `tolerance_bps`. The node bakes a real `LIM` into the
+    /// returned swap memo — `expected_amount_out × (1 − tolerance_bps/10_000)` —
+    /// so the signed memo carries a slippage floor instead of `LIM=0`
+    /// (unbounded). 100 bps (1%) is a conservative default aligned with the
+    /// streaming-upgrade threshold; there is no per-swap user slippage control.
+    static let defaultThorchainToleranceBps = 100
+
     func fetchQuote(
         amount: Decimal,
         fromCoin: Coin,
@@ -318,6 +326,7 @@ private extension SwapService {
                 toAsset: toCoin.swapAsset,
                 amount: truncatedAmount.description,
                 interval: provider.streamingInterval,
+                toleranceBps: Self.defaultThorchainToleranceBps,
                 referredCode: referredCode,
                 vultTierDiscount: vultTierDiscount
             )
@@ -610,6 +619,7 @@ extension SwapService {
                 amount: amount,
                 interval: 1,
                 streamingQuantity: streamingQuantity,
+                toleranceBps: Self.defaultThorchainToleranceBps,
                 referredCode: referredCode,
                 vultTierDiscount: vultTierDiscount
             )
--- a/VultisigApp/VultisigApp/Blockchain/THORChain/Service/MayaChainAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/THORChain/Service/MayaChainAPI.swift
@@ -28,7 +28,8 @@ struct MayaChainAPI: TargetType {
             streamingInterval: String,
             streamingQuantity: String?,
             affiliate: String?,
-            affiliateBps: String?
+            affiliateBps: String?,
+            toleranceBps: String?
         )
         case broadcast(body: Data)
         case pools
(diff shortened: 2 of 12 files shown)


johnnyluo (approved):
LGTM

Issue #730, opened 2024-06-20
[BUG] Password prompt when importing other files


realpaaao:
**Describe the bug**
When trying to import other files than viable .dat files, you always get the password input prompted.

**Expected behavior**
Check if viable .dat file is imported and if not give an appropriate error `Please select a viable file type`

**Screenshots**
<img width="1005" alt="image" src="https://github.com/vultisig/vultisig-ios/assets/167348323/39c036fd-8501-456b-b297-21efaccfc806">

**Additional context**
I would also amend the UI Screen to display that only .dat files are viable to upload.

realpaaao:
@Rockindash do we have `upload files, text or image` prompted on purpose?
As I think we only can use .dat files as input.

johnnyluo:
This should have been fixed in https://github.com/vultisig/vultisig-ios/pull/744

PR #4561, opened 2026-06-11
fix(cardano): shared initiator-computed dynamic fee for MPC sighash parity (#4531)

### LINKED ISSUE #693: Add TON coin to Wallet Core
https://github.com/enriquesouza/wallet-core/commits/master/

### LINKED ISSUE #4531: Cardano flat 180k-lovelace fee insufficient above ~560 bytes (CNT/multi-input txs rejected)
## Observed

Cardano transactions are signed with a **flat 180,000-lovelace fee**, independent of transaction size. Simple 1-in/1-out ADA sends overpay slightly; anything larger (native-token / CNT transfers, multi-input consolidations, metadata) is **rejected at broadcast with `FeeTooSmallUTxO`** after the keysign ceremony.

## Cause

`Blockchain/Cardano/Service/CardanoService.swift:88`:
```swift
return 180000 // 0.18 ADA
```
fed into `Cardano.swift:217` as `forceFee = UInt64(byteFee)` before `AnySigner.plan`.

Cardano's protocol fee is `minFeeA × size + minFeeB`. **Live protocol params (koios `cli_protocol_params`, 2026-06-10):**
- `minFeeA` (txFeePerByte) = **44**
- `minFeeB` (txFeeFixed) = **155,381**

The fixed component alone (155,381) already consumes 86% of the 180,000 budget, leaving only 24,619 lovelace for per-byte. Break-even: `180000 = 44 × size + 155381` → **size ≈ 560 bytes**. A simple transfer is ~200–300 B (fine), but multi-asset/CNT transfers and multi-input txs routinely exceed 560 B and fail.

The comment at `Cardano.swift:53-54` admits dynamic min-ADA is "a future enhancement"; the same applies to the fee.

## Fix shape

Fetch live protocol params (minFeeA/minFeeB) and compute the fee from the planned tx size — or run the WalletCore planner without `forceFee` and use `plan.fee` directly. Note: `CardanoService.swift` carries a comment claiming SDK/Windows MPC-sighash parity, so changing the fee derivation needs cross-platform coordination to keep the co-signed bytes identical.

---
Heads up — AI-assisted analysis; protocol params confirmed via live koios query; code verified by hand against current `main`.

gastonm5:
## Problem

Cardano transactions are signed with a **flat 180,000-lovelace fee**, independent of tx size. Simple 1-in/1-out ADA sends pass, but anything larger — native-token / CNT transfers, multi-input consolidations, CIP-20 metadata — is **rejected at broadcast with `FeeTooSmallUTxO`**, *after* the full keysign ceremony (the worst failure point). Closes #4531.

Cardano's protocol fee is `minFeeA × size + minFeeB` (live mainnet: `minFeeA=44`, `minFeeB=155,381`). The fixed `minFeeB` already eats 86% of the 180k budget, so the break-even is `(180000 − 155381) / 44 ≈ 560 bytes`.

## The constraint that shaped the fix: MPC sighash parity

In TSS keysign every co-signing device must build the **byte-identical** unsigned tx body, or the Blake2b sighash differs and the signature won't verify — keysign fails with no signature. The Cardano fee is the variable that determines those bytes, so it can't be changed on one platform unilaterally.

Investigation across the fleet found the fee handling is **split three ways** — and that two WalletCore facts make naïve approaches unsafe:

| Platform | WalletCore | Body fee today |
|---|---|---|
| iOS (`main`) | `4.3.19` | flat 180k (forces `byteFee`) |
| Android | `4.3.22` | **dynamic already** (derives, *discards* `byteFee`) |
| SDK / Windows | `4.6.7` | flat 180k (forces `byteFee`) |

- Android already derives a dynamic fee, so **Android ↔ iOS/Windows/SDK Cardano co-signing is already latently broken** for any tx whose real fee ≠ 180k.
- **Three different WalletCore versions** are pinned. Since the fee math, coin selection and CBOR encoding are all WalletCore-internal, having each device *derive independently* (the obvious fix) risks **divergent fees across versions → mismatched sighash**.

## Approach — Option 1: shared, initiator-computed, force-everywhere

Rather than have every device derive the fee (version-coupled, unsafe across `4.3.19/4.3.22/4.6.7`), the **initiator computes the size-derived fee once** into the shared payload (`chainSpecific.byteFee`), and **every co-signer forces that exact value** into the body. The fee becomes a transmitted constant — **independent of each peer's WalletCore version** — which preserves the invariant that already lets flat-fee iOS `4.3.19` and SDK `4.6.7` co-sign today, while fixing the underpayment.

### iOS changes (this PR)
- `Cardano.swift` — restore `transferMessage.forceFee = UInt64(byteFee)` at sign time so the body carries the transmitted fee (and `plan.fee == byteFee`).
- `KeysignPayloadFactory` — new `estimateDynamicByteFee(...)`: after UTXO selection, build the input, run `AnySigner.plan` with `forceFee = 0`, and seed `plan.fee` into `chainSpecific.byteFee`. Falls back to flat `180_000` (logged) on plan error — never crashes.
- `calculateDynamicFee` now returns the real (dynamic) fee for the Verify screen.
- `CardanoDynamicFeeTests` rewritten: body fee equals the transmitted `byteFee`; the initiator's estimate is size-derived (small < 180k, large > 180k, monotonic). `cardano.json` fixtures reverted to forced-fee hashes.

## ⚠️ Draft — cross-platform rollout required before merge

This PR makes **iOS-initiated** Cardano sends correct and parity-safe. It does **not** fully fix the fleet on its own:

- **SDK / Windows / vultiagent** — their signing path already forces `byteFee` ✅, so they'll honor a dynamic value an iOS initiator sends. But their **initiator** seeding still uses the flat `cardanoDefaultFee` — must switch to the computed `plan.fee`, or sends *they* initiate still underpay. → **[vultisig-sdk#730](https://github.com/vultisig/vultisig-sdk/pull/730)** (Closes #693), open.
- **Android** — must switch from discard-and-derive to **forcing the transmitted `byteFee`**, and seed a dynamic fee when it initiates. → **[vultisig-android#4891](https://github.com/vultisig/vultisig-android/pull/4891)**, open.
- Verify WalletCore Cardano fee parity across `4.3.19 / 4.3.22 / 4.6.7`.

Until those land, mixed-version / cross-platform co-signs of >560 B txs can still mismatch. Full analysis + four-repo task list: `wiki/pages/projects/vultisig/cardano-dynamic-fee/plan.md`.

**Sequencing:** this PR + [vultisig-sdk#730](https://github.com/vultisig/vultisig-sdk/pull/730) + [vultisig-android#4891](https://github.com/vultisig/vultisig-android/pull/4891) all adopt the same Option 1 (initiator derives, every device forces `byteFee`) — land them close together so no platform underpays as initiator.

### ⚠️ Known limitation — send-max
WalletCore **ignores `forceFee` for max-amount sends** (the change output is `total − fee`, so it always derives). So a Cardano **send-max** falls back to per-device fee derivation and is **not** pinned by the transmitted `byteFee` — cross-version (`4.3.19` ↔ `4.3.22` ↔ `4.6.7`) derivation must agree or the sighash diverges. Pre-existing and equal across platforms (not introduced here), but the one Cardano case this model can't cover. Verify send-max co-signs across the version spread, or gate it.

## Testing
- `xcodebuild` build (iPhone 16 Pro / OS 18.4): **BUILD SUCCEEDED**.
- `CardanoDynamicFeeTests` 5/5; `ChainHelperTests` green (confirms reverted fixtures + iOS/SDK body parity).
- SwiftLint clean on touched files.

> HIGH security tier — changes signed tx bytes. Maintainer review required.

## Summary by CodeRabbit

* **New Features**
  * Cardano transfers now calculate fees dynamically from the transaction’s size (using the selected UTXOs), improving accuracy versus earlier placeholder behavior.
  * Ensures the same computed fee is preserved across all signers, keeping displayed and signed fees aligned for multi-signature flows.
* **Tests**
  * Added coverage for dynamic fee honoring, cross-seed fee consistency, and initiator fee behavior across small/large transfers and increasing transaction sizes (including expected monotonic scaling).


johnnyluo:
Currently IOS also use transaction plan to calculate the fee on initiate device and join device separately 

Technically , that should be fine , because with the same input , walletcore should spit out the same result. All UTXO chain does the same

Maybe check how the estimate fee show on screen is calculated? make sure it it is getting from the transaction plan?

gastonm5:
## Cosigning parity review — vs SDK #693/#730 and Android #4891

**Verdict: ✅ for ordinary sends.** iOS ↔ SDK (and Windows, same `@vultisig/core` path) Cardano co-signing holds: both sides **force the transmitted `byteFee`** and read `ttl`/`utxos` from the shared payload, so a single initiator-built `KeysignPayload` yields byte-identical bodies → matching Blake2b sighash.

- iOS (this PR): `forceFee = byteFee` → `plan.fee == byteFee`, pinned by `testBodyFeeEqualsTransmittedByteFee`.
- SDK `mpc/keysign/signingInputs/resolvers/cardano.ts:47` — `forceFee: Long.fromString(byteFee.toString())`, with `byteFee`/`ttl` read from the **transmitted** `blockchainSpecific` (`:19`), not re-resolved.

The load-bearing property of Option 1: the cosigner **forces, never re-derives**, so the fee *algorithm* need not match across WalletCore `4.3.19`/`4.3.22`/`4.6.7` — only the field + the force mechanism, which all three platforms share. Correct.

### Fleet is converging — sequence these together
- **SDK #730** (Closes #693): initiator now computes `byteFee` from WalletCore tx-body bytes (dynamic); signing path already forces it.
- **Android #4891**: drops its derive-and-discard and forces the transmitted `byteFee` — closes the previously-latent iOS↔Android break.

Each is safe for *cosigning* on its own (forcing never breaks sighash). But until a platform's **initiator** computes a dynamic fee, that platform still underpays large (>560 B) txs *as initiator* → `FeeTooSmallUTxO` at broadcast. So this PR fixes iOS-initiated sends; SDK/Windows-initiated need #730, Android-initiated needs #4891. Ideally land the three close together so no platform underpays as initiator.

### ⚠️ One case Option 1 does NOT cover — send-max
Per Android #4891's own test note: **WalletCore ignores `forceFee` for max-amount sends** (output = total − fee, so it always derives). So for a Cardano **send-max**, every device derives the fee independently again → cross-version (`4.3.19`↔`4.6.7`) derivation must agree or the sighash diverges. Pre-existing and equal across platforms (not introduced here), but it's the one Cardano case the transmitted-fee model can't pin. Worth confirming send-max Cardano co-signs across the version spread (or gating it).

### Minor (iOS-internal)
`estimateDynamicByteFee` derives from a `forceFee=0` preliminary plan then forces that value — exact only if WalletCore selects the same UTXOs in both passes. Tests cover fee *scaling*, not that selection determinism. Low risk; worth a note.

Code is correct and the test suite pins the right invariant. 👍

johnnyluo (approved):
LGTM

PR #4566, opened 2026-06-12
fix: chain badge renders the plain chain logo on a dark disc

### LINKED ISSUE #4565: [Fix] chain badge icon doesn't match the chain logo used everywhere else [Swap coin picker]
bsc mismatch

The little chain badge on token cells (bottom-right of the asset icon) is not the same artwork as the chain logo in the "Select chain" carousel right below it. BSC: dark mono cube on the badge vs gold BNB diamond in the carousel. Ethereum: navy diamond vs the standard grey one. Same screen, two different icons for the same chain.

## Problem

`AsyncImageView` renders the token badge from a parallel asset family: `ChainIconView(icon: "chain-" + chainIcon)` (`Components/ImageView/AsyncImageView.swift:58`) pulling from `Assets.xcassets/Crypto/chains/chain-*.imageset` - while the chain carousel (`SwapCoinPickerView.swift:209`), portfolio rows, and everything else render the plain `chain.logo` assets (`bsc`, `eth`, ...). Two icon families for the same concept, maintained by hand, already drifted in artwork.

Bonus rot: 4 chains have no `chain-*` asset at all (`sei`, `hyperliquid`, `qbtc`, `bittensor`), so their token badges render an empty white circle.

## Expected Behavior

One source of truth: the badge shows the same chain icon as the carousel / the rest of the app, for every chain, no hand-maintained parallel set.

## Steps to Reproduce

1. Swap screen, tap the from/to asset pill to open the coin picker
2. Pick BSC (or Ethereum) in the chain carousel
3. Compare the badge on any token cell (AAVE, USDC, ...) with the chain icon in the carousel at the bottom

## Solution

Render the badge from `chain.logo` directly (drop the `"chain-" +` prefix in `AsyncImageView`) and delete the now-dead `Crypto/chains/` family - it has exactly one consumer. That also auto-fixes the 4 blank-badge chains, and any future chain gets a badge for free. If design actually wants a distinct mono badge style, the alternative is keeping the family but deriving it from a single accessor on `Chain` and backfilling the 4 missing assets - but the simple unification matches what every other surface in the app does.

### Must Do
- [ ] Badge and carousel render identical chain artwork for every chain in the picker
- [ ] No chain renders an empty badge
- [ ] Remove whatever icon family/lookup ends up dead

### Must NOT Do
- Don't touch token (coin.logo) resolution
- Don't refactor AsyncImageView beyond the badge source

### Out of Scope
- aave-wrapped tokens (aBnbUSDC, stataBnbUSDC, aEthUSDC...) render no token icon at all - that's missing token logo content, not the chain badge: https://litter.catbox.moe/qnh73z.png - separate follow-up

## Acceptance Criteria
- [ ] SwiftLint passes
- [ ] Build succeeds
- [ ] Badge == carousel artwork for BSC + Ethereum in the sim (screenshot proof)
- [ ] sei / hyperliquid / qbtc / bittensor tokens show a non-empty badge

Screenshots are litterbox, 72h.

gomesalexandre:
closes #4565

## What

The little chain badge on token cells was rendered from a parallel mono asset family (`Crypto/chains/chain-*`, 34 imagesets) that visibly drifted from the plain `chain.logo` assets every other surface uses - same picker showed a dark mono cube badge next to a gold BNB carousel logo. The family was also never extended for sei / hyperliquid / qbtc / bittensor, so their badges rendered as empty white circles. Users now see one brand mark per chain everywhere, and every future chain gets a badge for free.

## How

- `AsyncImageView` renders the badge from `chain.logo` directly (dropped the `"chain-"` prefix) - single source of truth with the carousel / portfolio / everything
- the white badge disc from the design revamp stays EXACTLY as it was - only the glyph artwork changes. Inset tightened to `size/3` so full-color logos occupy the visual space the tight-cropped mono glyphs did
- deleted the dead `Crypto/chains/` family (sole consumer was the prefix)
- ride-along: `EditReferralDetailsView` passed `?? .empty` for the badge logo, painting a stray empty circle when no preferred asset is set - `nil` now suppresses the badge

Backed by a multi-agent research pass (codebase blast radius, cross-platform parity vs Android/Windows/extension, per-asset legibility audit): https://gist.github.com/gomesalexandre/254a4fa4c08cf730e306e92b29fc0922

## Known gap (design input wanted - @realpaaao)

Six plain chain logos are white-fill SVGs tuned for dark backgrounds: arbitrum, mantle, zsync_era, sui, cro, xrp. On the white disc those badges render invisible (cro/xrp never badge in practice - no tokens; arbitrum/mantle/zksync/sui are real token surfaces). The mono family had the mirror-image problem (4 chains with NO badge at all + drifted artwork). Proper fix is giving those 6 assets brand colors, which also makes them legible everywhere else they render on light surfaces - happy to do that as a follow-up if design blesses it.

## receipts

Before / after (BSC picker - the issue repro; white disc unchanged, glyph now matches the carousel):

before-after

Badge vs carousel, zoomed - badge artwork now IS the carousel artwork on the same white disc as before:

bsc zoom

Full picker screen: https://litter.catbox.moe/pwthjj.png

Build green on iPhone 16 Pro sim, swiftlint clean on touched files, no test references to the badge path, codex adversarial review on the unification commit: approve with zero real findings. App is dark-only (fixed hex palette - verified pixel-identical content under light/dark system appearance), so no appearance matrix applies. Screenshots are litterbox, 72h.

## Risk

Low - one rendering chokepoint, asset deletions are dead weight, no data migration (history records persist plain `chain.logo` already). Only judgment call is the white-fill-logo gap above.


gastonm5:
@gomesalexandre black chain logo with white circle was part of the design revamp we did last year, I believe we need to confirm this with @realpaaao 

Some issues I see with this:
1. The new blue bg blends with the general bg, so we should use another color if we want to move on this direction.
2. It looks there's a double background or some stroke in the mid of the bg

bsc zoom

gomesalexandre:
@gastonm5 ye you're right on both counts - and you commented mid-iteration, the dark disc is already dead. Pushed `4dd189cd3`:

- white disc from the design revamp stays EXACTLY as you shipped it, untouched. The only change left is the glyph: badge now renders the same `chain.logo` artwork as the carousel/portfolio instead of the parallel `chain-*` mono family (which had drifted + was missing sei/hyperliquid/qbtc/bittensor entirely)
- the blending + "double background" you saw was the dark variant's `bgSurface1` disc sitting on `bgSurface1` rows with a same-color hairline stroke on top - antialiasing made it read as a phantom ring. Gone with the white disc
- PR body re-receipted with the white-disc build: https://litter.catbox.moe/xsvfu0.png (before/after) + https://litter.catbox.moe/to11l4.png (zoom)

One genuine design question remains for @realpaaao: arbitrum / mantle / zsync_era / sui / cro / xrp plain logos are white-fill SVGs (dark-bg-tuned), so their badges are invisible on the white disc (cro/xrp never badge - no tokens; the other four are real). The mono family had the mirror problem (4 chains with no badge asset at all). Cleanest fix imo is brand-coloring those 6 SVGs in a follow-up - they'd also stop vanishing on any future light surface. Happy to do that once design blesses.

realpaaao:
@gomesalexandre @gastonm5 I think this PR is crossing two separate icon contracts.

The carousel should keep rendering the normal `Chain.logo` assets. The small token-chain badge should keep rendering the mono `Crypto/chains/chain-*` assets on the white disc. So changing `AsyncImageView` from `"chain-" + chainIcon` to `chainIcon` makes the badge use the normal carousel/icon family, including colored logos and some white-on-transparent logos, which is not the intended badge style.

The actual gap looks smaller: current `main` is missing mono badge assets for `hyperliquid`, `sei`, `qbtc`, and `bittensor`. I opened #4567 as the low-priority follow-up for those missing `chain-*` assets. For this PR, I would keep the badge lookup separate, fix the `EditReferralDetailsView` native-token suppression comment, and avoid deleting the existing mono badge asset family unless design explicitly wants badges to stop being mono.

— ⚒️ _Mjölnir, Paaao's agent — posting on his behalf, not Paaao directly._

gomesalexandre:
@realpaaao appreciate the clear contract framing, and #4567 is worth doing regardless. But before we settle on "two contracts", honest challenge on two points:

**1. The two contracts share a viewport.** This isn't badge-in-one-screen, carousel-in-another - the swap coin picker renders BOTH families for the SAME chain about 40pt apart:

same viewport

Whatever the system intent is, what the user sees is two different BSC marks side by side. That doesn't read as a design system, it reads as drift. And BSC is the FIRST chain in the picker's default order, so this exact pairing is everyone's first impression of the screen - and the mono cube at ~7pt on the white disc frankly reads broken next to the gold mark.

**2. The mono family's track record says it won't stay healthy.** From the research pass on this PR: 3 of 34 badges were dead on arrival at launch (chain-bitcoin / chain-cardano / akash naming bugs, fixed later in drive-bys), and then 4 consecutive chain integrations (sei, hyperliquid, qbtc, bittensor) shipped without mono assets - which is the blank-badge bug that started all this. `Image("chain-x")` fails silently (no compile error, empty render), so chain #39 will miss again unless the mono asset becomes a hard checklist item per integration.

If design's verdict after seeing the same-viewport shot is still mono-on-white, zero drama - the revert is one parked commit, #4567 covers the gap, and I'd add a runtime fallback (mono asset missing -> plain `chain.logo`) so the silent-blank class dies forever. But genuinely asking for a look at that screenshot first: is two marks for one chain in one viewport the intended experience, or should the badge derive from the same artwork?

cc @gastonm5

gomesalexandre:
Follow-up to my own challenge above, to be fair to the other side: a deliberate badge variant family IS a legit design pattern (plenty of systems run a simplified mono mark at small sizes), so if that's the intended system here - badge = mono on white disc, full-size = colored brand mark - just confirm it explicitly and we comply: revert is parked and ready, #4567 covers the missing four, and I'll add the missing-asset -> plain-logo runtime fallback plus a per-chain-integration checklist item so the family stops rotting.

One concrete ask if mono is confirmed though: the current `chain-bsc` mono glyph needs a better cut. At ~7pt on the white disc it reads muddy (see the zoom in my previous comment) - and since BSC is the first chain in the picker order, it's the worst possible place for the weakest glyph. A cleaner mono cube with heavier strokes tuned for badge size would fix the "looks broken" first impression without touching the system.

So @realpaaao, two yes/nos: (1) two icon families intended - yes/no? (2) if yes, ok to redo the bsc mono glyph (+ any others that read muddy at badge size) as part of #4567?

NeOMakinG (approved):
## testing approach

Triple-lane static review (Claude + Codex gpt-5.4 + GLM 4.6). Code-only — no runtime QA per re-review protocol for this class of pure rendering fix. All three lanes reviewed independently and converged on the same verdict.

## blocking

None.

## preferably-blocking

None.

## should-fix

None.

## pre-existing bugs i bumped into

`fallbackText` in `AsyncImageView.swift:93` uses `Color.white` (hardcoded) instead of a theme token. Not introduced by this PR — pre-existing, out of scope.

## risk

**Low.** Key invariants verified:

- **Asset coverage complete.** All 38 `Chain.logo` return values (`ada`, `akash`, `arbitrum`, `atom`, `avax`, `base`, `bch`, `bittensor`, `blast`, `bsc`, `btc`, `cro`, `dash`, `doge`, `dot`, `dydx`, `eth`, `hyperliquid`, `kuji`, `ltc`, `luna`, `lunc`, `mantle`, `matic`, `maya`, `noble`, `optimism`, `osmo`, `qbtc`, `rune`, `sei`, `solana`, `sui`, `ton`, `tron`, `xrp`, `zec`, `zsync_era`) have corresponding `.imageset` entries in `Assets.xcassets/Crypto/`. No silent blank-disc regression.
- **`chan-sui` typo** was in the *deleted* `chain-sui.imageset/Contents.json`, not in any live codepath.
- **Badge guard correct.** `logo != tokenChainLogo` in `AsyncImageView.swift:56` still correctly suppresses the badge for native tokens (e.g. ETH coin on Ethereum: `logo="eth"`, `tokenChainLogo="eth"` → guard fires, no badge). `CoinMeta.tokenChainLogo` and `Coin.tokenChainLogo` both already returned `chain.logo` directly since commit `bf79252` (2026-03-04) — the only mismatch was `AsyncImageView` prepending `"chain-"` on top, which this PR removes.
- **No persisted stale data.** `TransactionHistoryRecorder` sources `coinChainLogo` from `coin.tokenChainLogo` which always produced the un-prefixed asset name. No data migration needed.
- **Theme conformance clean.** `ChainIconView` uses `Theme.colors.textPrimary` (disc fill) and `Theme.colors.bgSurface1` (stroke) — both are `ColorSystem` design tokens, not hardcoded hex. No new inline colour constants introduced.
- **`EditReferralDetailsView` nil fix correct.** Old `?? .empty` passed `""` as non-nil, which triggered `ChainIconView(icon: "chain-" + "")` = `Image("chain-")` = blank white circle when no preferred asset was set. New `nil` suppresses the badge via the `if let` guard.
- **No remaining `"chain-"` prefix callsites** in Swift source (verified by grep across all `tokenChainLogo` callsites).
- **White-fill logo gap** (arbitrum / mantle / zksync / sui invisible on white disc) acknowledged by author and deferred to design. This is a pre-existing legibility gap that this PR makes no worse — those four chains had *no badge at all* in the old mono family.

## verdict

`APPROVE` — clean unification, complete asset coverage, no dead prefix callsites, correct nil guard, design tokens throughout. The known white-fill gap is correctly deferred.

---
*reviewed by [vultisig-ops reviewer](https://github.com/NeOMakinG) — claude-sonnet-4-6 + codex gpt-5.4 + glm-4.6*

PR #4571, opened 2026-06-13
fix(polkadot): check tx status via node RPC instead of Subscan

johnnyluo:
## Summary
- Subscan's `assethub-polkadot.api.subscan.io` endpoint now requires an API key, breaking Polkadot Asset Hub transaction-status checks.
- Switches status checks to the Vultisig node-RPC proxy `https://api.vultisig.com/dot/` — the same node `PolkadotService` already uses for balances and broadcasting (verified live as "Polkadot Asset Hub").
- A bare Substrate node doesn't index extrinsics by hash, so status is derived from the transaction pool via `author_pendingExtrinsics`: each pending extrinsic is blake2b-256 hashed (the value `author_submitExtrinsic` returns) and matched against the tx hash.
  - **In the pool** → `pending`
  - **Gone from the pool** (included in a block) → `confirmed`
- Removed the dead Subscan `polkadotTransactionStatusRpc` endpoint constant.

## Behavior note / tradeoff
Node RPC can't replay an already-included extrinsic's events, so on-chain dispatch *failures* are no longer distinguished (the previous Subscan path could surface a `failed` reason). The node still rejects malformed/invalid extrinsics at broadcast time via `author_submitExtrinsic`. Detecting post-inclusion failures would require block/event scanning — a heavier follow-up.

## Test Plan
- [x] Added `PolkadotTransactionStatusProviderTests` (6 cases: in-pool→pending, prefix/case-insensitive match, not-in-pool/empty/nil→confirmed, RPC error→throws) — all passing
- [x] SwiftLint passes (0 violations on changed files)
- [ ] Manual: send a Polkadot/DOT tx and confirm the status screen transitions pending → confirmed

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable Polkadot transaction status: now verifies against node RPC block data to reduce false "not found"/"failed" results.

* **Tests**
  * Added tests covering confirmed/pending detection, case- and prefix-insensitive hash matching, empty-hash handling, and RPC error propagation.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/TransactionStatus/Models/PolkadotTransactionStatusResponse.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TransactionStatus/Models/PolkadotTransactionStatusResponse.swift
@@ -7,13 +7,29 @@
 
 import Foundation
 
-/// JSON-RPC response for `author_pendingExtrinsics`.
+/// JSON-RPC response for `chain_getBlock`.
 ///
-/// `result` is the list of hex-encoded extrinsics currently in the node's
-/// transaction pool. `error` is populated when the node rejects the call.
+/// `result.block` carries the signed block: `extrinsics` is the list of
+/// hex-encoded extrinsics in the block (each blake2b-256 hashed to match a tx
+/// hash), and `header.parentHash` links to the previous block so the status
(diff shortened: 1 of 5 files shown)


gastonm5 (approved):
lgtm

PR #4585, opened 2026-06-16
fix(cardano): enforce 1.4 ADA minimum send amount

johnnyluo:
## Summary
- Native-ADA sends below Cardano's protocol minimum UTXO value (~1.4 ADA) were accepted by the iOS/macOS send form but silently **dropped by the node**. The validation logic existed in `CardanoHelper` but was never wired into the send flow. Android already enforces this.
- Added `SendCryptoLogic.isBelowCardanoMinimum(coin:amount:)` and called it from both validation layers before the keysign ceremony:
  - `SendDetailsViewModel.validateBalance` (Details screen)
  - `SendCryptoVerifyLogic.validateBalanceWithFee` (Verify screen)
- The check validates the **recipient output amount**, so it also blocks **MAX sends** when the whole vault holds less than the floor (`computeMaxAmount` already nets out the fee).
- CNT (Cardano native token) sends carry their own ADA floor on the bundled output and are exempt; non-Cardano chains are unaffected.
- Added a localized `cardanoMinimumSendAmountError` string to all 7 locales (sorted via `sort_localizable.py`).

## Behavior
| Scenario | Result |
|---|---|
| Send 1.0 ADA | Blocked |
| Send ≥ 1.4 ADA | Allowed |
| MAX send, vault < 1.4 ADA | Blocked |
| MAX send, vault ≥ 1.4 ADA (+fee) | Allowed |
| CNT token / non-Cardano | Unaffected |

## Test Plan
- [x] SwiftLint passes (changed files)
- [x] `xcodebuild` build succeeds
- [x] `SendValidationTests` — 42/42 pass, incl. 7 new `isBelowCardanoMinimum` cases (below/at/above floor, MAX below/above floor, CNT token, non-Cardano)
- [x] Manual: attempt a sub-1.4 ADA ADA send → blocked with error; attempt MAX from a vault holding < 1.4 ADA → blocked

## Summary by CodeRabbit

* **New Features**
  * Added Cardano-native transaction validation to prevent sending amounts below the network minimum send amount.

* **Documentation**
  * Added localized error messaging for the “minimum send amount” rule (English, German, Spanish, Croatian, Italian, Portuguese, and Simplified Chinese).

* **Tests**
  * Added unit tests covering minimum-send behavior across Cardano ADA, MAX scenarios, and non-native/exempt and non-Cardano cases.

CleanShot 2026-06-16 at 18 38 31@2x

### DIFF
--- a/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
@@ -653,11 +653,12 @@ final class SendDetailsViewModel {
             }
         }
 
-        // Cardano enforces a protocol minimum UTXO value on every output; a
-        // native-ADA send below it is silently dropped by the node. Match
-        // Android by blocking it here before the keysign ceremony.
-        if SendCryptoLogic.isBelowCardanoMinimum(coin: coin, amount: amount) {
-            let minAmount = CardanoHelper.defaultMinUTXOValue.toADAString
+        // Some chains enforce a protocol minimum value on every output; a
+        // native send below it is silently dropped by the node. Match Android
+        // by blocking it here before the keysign ceremony.
+        if SendCryptoLogic.isBelowMinimumSendAmount(coin: coin, amount: amount),
+           let minimum = coin.chain.minimumSendAmount {
+            let minAmount = coin.decimal(for: minimum).description
             setAmountError(message: String(format: "cardanoMinimumSendAmountError".localized, minAmount))
             return false
         }
--- a/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
@@ -653,6 +653,15 @@ final class SendDetailsViewModel {
             }
         }
 
+        // Cardano enforces a protocol minimum UTXO value on every output; a
+        // native-ADA send below it is silently dropped by the node. Match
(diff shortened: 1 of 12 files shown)


VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
```diff
@@ -653,6 +653,15 @@ final class SendDetailsViewModel {
             }
         }
 
+        // Cardano enforces a protocol minimum UTXO value on every output; a
+        // native-ADA send below it is silently dropped by the node. Match
+        // Android by blocking it here before the keysign ceremony.
+        if SendCryptoLogic.isBelowCardanoMinimum(coin: coin, amount: amount) {
```
gastonm5: Can we make this generic instead of cardano specific?

gastonm5 (changes requested):

VultisigApp/VultisigApp/Features/Send/ViewModels/SendDetailsViewModel.swift
johnnyluo: Done in fe44f5b. Introduced a `Chain.minimumSendAmount: BigInt?` extension returning the per-chain protocol floor (`nil` when a chain has none), and renamed `isBelowCardanoMinimum` → `isBelowMinimumSendAmount` to drive off it. Cardano is the only case today, but adding another chain is now a one-line switch case. The displayed amount is derived generically via `coin.decimal(for:)`.

gastonm5 (approved):
lgtm

Issue #4586, opened 2026-06-16
feat(qbtc): in-app governance proposals — list / vote / weighted vote on the DeFi tab (ML-DSA)


gastonm5:
## Context
Add a **Governance / Proposals tab** to the **QBTC DeFi chain page**: list active + past governance proposals, view details + tally, and vote — **single-option and weighted**. QBTC is a standard Cosmos-SDK chain running **vanilla `x/gov` v1**, and iOS already signs QBTC governance votes with **ML-DSA** (`QBTCHelper.buildMsgVote`). Builds on the merged QBTC Cosmos-staking DeFi tab (#4481) and the QBTC vote-cosign work (#4046 / #4518).

iOS is the **first native Vultisig client** with an in-app proposals UI (Windows / SDK / Android are signers only; proposal browsing currently lives in the external `qbtc-explorer` dApp).

## Scope
- **In:** read (list active + past, detail, tally, my-vote) + **`MsgVote`** (Yes / No / NoWithVeto / Abstain) + **`MsgVoteWeighted`** (split voting power across options).
- **Out:** submit proposals (`MsgSubmitProposal`), deposits (`MsgDeposit`).

## Chain facts (verified 2026-06-16)
- **Module:** stock `x/gov` **v1** (depinject; the forked `btcq-org/cosmos-sdk` only diverges on IBC/Wasm — no gov customization).
- **Params:** denom `qbtc`; `voting_period` & `max_deposit_period` 172800s (2d); `quorum` 0.334; `threshold` 0.5; `veto_threshold` 0.334; `min_deposit` 10000000; expedited 86400s / 0.667.
- **LCD (live, HTTP 200):** base `https://api.vultisig.com/qbtc-rpc`. **Query with v1**; the **vote message stays `/cosmos.gov.v1beta1.MsgVote`** (byte-equality with cosigning peers + the #4518 gate — do NOT change the vote type URL).
- **Disambiguation:** QBTC's custom `x/qbtc` `MsgGovClaimUTXO` is gov-*gated*, not a vote; but proposals can *wrap* arbitrary messages (the live testnet proposal #1 wraps a `MsgGovClaimUTXO`) → render `messages[].@type` **generically**.

## Reused vs net-new
[...]

PR #4602, opened 2026-06-18
[PoC] feat(qbtc): in-app governance — proposals list / vote / weighted vote (ML-DSA)

gastonm5:
> **Proof of concept.** This implements the QBTC in-app governance flow end-to-end, but the **UI is preliminary and will likely change** — opening as a draft for early feedback on the approach, not as final design.

Implements #4586: an in-app governance tab on the QBTC DeFi page — read-only proposals list, single-option vote, and weighted vote, signed via ML-DSA.

## What's included
- **Query layer** — gov DTOs + service + decode tests (`CosmosGov*` models, generic over wrapped message `@type`, incl. `/qbtc.qbtc.v1.*`).
- **Proposals tab** — active + past proposals as cards with status badge, voting-end countdown, compact tally bar, and the user's recorded-vote badge (`QBTCGovernanceView`, embedded in the DeFi chain page).
- **Proposal detail sheet** — title/summary/status, generic message-type rendering, full tally breakdown, voting window, quorum, and the vote controls.
- **Voting** — single-option (`QBTC_VOTE:`) and weighted (`QBTC_VOTEW:`, per-option weights summing to 100%) → verify → ML-DSA keysign. Weighted votes gated on byte-equality.
- **Design-system pass** — sheets now render on `Theme.colors.bgPrimary` (were falling back to the system gray); colors/fonts via `Theme.*` throughout.

## Status / caveats
- UI is a PoC — layout, copy, and styling are expected to evolve.
- Unit tests cover the gov query/decode layer and the governance view model.
- Built (iPhone 16 / iOS 18.4) and SwiftLint-clean.

Refs #4586

## Summary by CodeRabbit

* **New Features**
  * Added a QBTC governance section with active/past proposals, proposal details, live tallies/quorum, countdowns, and “you voted” badges.
  * Enabled weighted voting via a dedicated weighted-vote sheet and MsgVoteWeighted transaction signing.
* **Bug Fixes**
  * Improved governance LCD decoding with safer fallbacks, robust proposal ID parsing, and treating “not voted” (404) as empty.
* **Localization**
  * Added governance UI strings across multiple languages.
* **Tests**
  * Expanded governance decoding, view-model, memo, and weighted-vote signing/vector coverage.


VultisigApp/VultisigApp/Blockchain/Cosmos/Models/Gov/CosmosGovDTOs.swift
```diff
@@ -0,0 +1,385 @@
...
+    let depositEndTime: Date?
+    let expedited: Bool
+    let failedReason: String
+}
+
+/// Wire shape for a single proposal in the v1 list / detail response.
+struct CosmosGovProposalWire: Decodable {
+    let id: String
+    let messages: [WireMessage]?
+    let status: String
```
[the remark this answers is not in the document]
gastonm5: Fixed in a1f2a908e. `status` is now decoded as `String?` and mapped via `CosmosGovProposalStatus(wire: status ?? "")`, so a missing/null status falls through to `.unspecified` instead of failing decode — only an unparseable `id` drops a proposal. Added a regression test (`testMissingStatusDecodesAndMapsToUnspecified`).

VultisigApp/VultisigApp/Blockchain/Cosmos/Signing/QBTCHelper.swift
```diff
@@ -308,15 +315,110 @@ struct QBTCHelper {
...
+    /// proto enum integer. Shared by the single-option and weighted paths.
+    static func voteOptionInt(from description: String) -> UInt64 {
         switch description.uppercased() {
         case "YES": return 1
         case "ABSTAIN": return 2
         case "NO": return 3
         case "NO_WITH_VETO", "NOWITHVETO": return 4
-        default: return 0 // UNSPECIFIED
+        default: return 0
         }
```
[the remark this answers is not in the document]
gastonm5: Fixed in a1f2a908e. `voteOptionInt(from:)` now returns `UInt64?` (nil for unknown tokens), and both signing paths fail fast: the single-vote path guards and throws `HelperError.runtimeError("QBTC: invalid vote option ...")`, and `parseWeightedOptions` throws `QBTC: invalid weighted-vote option ...`. The redundant `voteOptionValue` wrapper was removed. Malformed memos can no longer be signed as `VOTE_OPTION_UNSPECIFIED`.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/QBTCGovernanceViewModel.swift
```diff
@@ -0,0 +1,174 @@
...
+            return try await service.fetchGovParams()
+        } catch {
+            logger.warning("Failed to fetch gov params: \(error.localizedDescription, privacy: .public)")
+            return nil
+        }
+    }
+
+    private func resolveTallies(for proposals: [CosmosGovProposal]) async {
+        guard !proposals.isEmpty else { return }
+        let resolved = await withTaskGroup(of: (UInt64, CosmosGovTallyResult?).self) { group in
```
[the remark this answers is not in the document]
gastonm5: Fixed in a1f2a908e. The empty-proposals guard in `resolveTallies` now sets `liveTallies = [:]` before returning, so stale live tallies no longer override `finalTally` for past proposals once active proposals drop to zero.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/DefiChainMainViewModel.swift
```diff
@@ -87,12 +87,17 @@ final class DefiChainMainViewModel: ObservableObject {
...
+            // Stake-only segment for Cosmos-SDK staking on LUNA / LUNC.
             // These chains route to `CosmosStakeDefiView` (see
             // `DefiChainMainScreen.cosmosStakeView`); the polished
             // position-card UI lands in a follow-up.
             [.stake]
+        case .qbtc:
+            // QBTC adds an in-app governance proposals segment alongside
+            // staking. Governance is not a selectable asset, so it is
+            // always present here and skips `setupSelectablePositions()`.
+            [.stake, .governance]
```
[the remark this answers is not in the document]
gastonm5: Fixed in a1f2a908e. `moveToNextPosition`/`moveToPreviousPosition` now cycle over `positions.map(\.value)` (the chain-available segments built from `getDefiPositionTypes()`) with an empty-guard, instead of `DefiChainPositionType.allCases`. Navigation can no longer land on a segment that isn't available for the current chain.

VultisigApp/VultisigApp/Features/Defi/DefiChain/View/Governance/QBTCGovernancePresentation.swift
```diff
@@ -0,0 +1,92 @@
...
+            return "\(days)d \(hours)h"
+        }
+        if hours > 0 {
+            return "\(hours)h \(minutes)m"
+        }
+        if minutes > 0 {
+            return "\(minutes)m"
+        }
+        return "<1m"
+    }
```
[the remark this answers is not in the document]
gastonm5: Fixed in a1f2a908e. The unit tokens are now localized: `shortDuration` uses format keys `governanceDurationDaysHours`/`governanceDurationHoursMinutes`/`governanceDurationMinutes` and `governanceLessThanOneMinute`, added to all 8 locale files (en, de, es, hr, it, ko, pt, zh-Hans) and sorted via `sort_localizable.py`. Negative intervals are clamped to 0. Kept the existing two-units-max structure with `%d`-formatted unit tokens rather than `DateComponentsFormatter` so locales can reorder/translate the tokens.

VultisigApp/VultisigApp/Features/Defi/DefiChain/View/Governance/GovernanceWeightedVoteSheet.swift
```diff
@@ -0,0 +1,127 @@
...
+        .yes: 25, .no: 25, .noWithVeto: 25, .abstain: 25
+    ]
+
+    private var total: Int {
+        CosmosGovVoteChoice.allCases.reduce(0) { $0 + (weights[$1] ?? 0) }
+    }
+
+    private var isValid: Bool {
+        total == 100
+    }
```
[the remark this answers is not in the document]
gastonm5: Addressed in a1f2a908e. The weighted-vote rules now live in `GovernanceWeightedVoteViewModel` (a `@MainActor ObservableObject`): `weights` state, `total`, `isValid`, and `buildOptions()`. `GovernanceWeightedVoteSheet` holds it via `@StateObject` and is now presentation-only. The refactor was contained enough to do properly even though this is a PoC, since it directly enforces the MVVM boundary.

VultisigApp/VultisigApp/Blockchain/Cosmos/Service/QBTC/QBTCGovService.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+//  Hits `/cosmos/gov/v1/...` on the qbtc-rpc proxy (verified live: the
+//  proxy passes the standard cosmos gov paths). The vote *message* the user
+//  signs stays at gov v1beta1 (`QBTCHelper`) — only the read path is v1.
+//
+
+import Foundation
+import OSLog
+
+protocol QBTCGovServiceProtocol {
```
johnnyluo: Swift 6 Sendable risk: `QBTCGovernanceViewModel.resolveTallies`/`resolveMyVotes` capture `service` (`any QBTCGovServiceProtocol`) into `group.addTask { [service, logger] ... }`, which is a `@Sendable` closure. Under strict concurrency that's an error since the protocol isn't `Sendable` (builds today presumably because the target is still Swift 5 mode). Suggest `protocol QBTCGovServiceProtocol: Sendable` to future-proof — the concrete `QBTCGovService` struct already satisfies it.

johnnyluo (commented):
Reviewed the QBTC governance PoC end-to-end — clean layering, strong decode resilience, and good test coverage (the cross-platform byte-equality gate is the right call for a hand-rolled proto encoder). A few refinements inline; given this is an explicit draft/PoC, the architecture/UX notes can reasonably defer to follow-up. The crypto-boundary items worth a maintainer's eyes are the Sendable note and the fixture provenance.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/QBTCGovernanceViewModel.swift
```diff
@@ -0,0 +1,177 @@
...
+        loadFailed = false
+
+        // Active first (newest id first), then past (newest id first).
+        let active = proposals.filter { $0.status.isActive }.sorted { $0.id > $1.id }
+        let past = proposals.filter { !$0.status.isActive }.sorted { $0.id > $1.id }
+        activeProposals = active
+        pastProposals = past
+
+        await resolveTallies(for: active)
+        await resolveMyVotes(for: proposals, voterAddress: voterAddress)
```
johnnyluo: This fans out a my-vote query for *every* proposal (active + past), whereas tallies are correctly scoped to active only. With `defaultProposalLimit = 100`, a populated chain fires up to ~100 concurrent requests on each `refresh()` and every pull-to-refresh. Fine for testnet (1 proposal), but worth bounding — e.g. limit to active + recently-ended — or noting it as a known PoC scaling limit.

VultisigApp/VultisigApp/Blockchain/Cosmos/Signing/QBTCHelper.swift
```diff
@@ -307,16 +316,111 @@ struct QBTCHelper {
...
+    private func buildVoteWeightedAny(keysignPayload: KeysignPayload) throws -> Data {
+        let msg = try buildMsgVoteWeighted(keysignPayload: keysignPayload)
+        var anyMsg = Data()
+        anyMsg.appendProtoString(fieldNumber: 1, value: Self.msgVoteWeightedTypeURL)
+        anyMsg.appendProtoBytes(fieldNumber: 2, data: msg)
+        return anyMsg
+    }
+
+    private func buildMsgVoteWeighted(keysignPayload: KeysignPayload) throws -> Data {
+        // Memo format: "QBTC_VOTEW:PROPOSAL_ID:OPTION=WEIGHT,OPTION=WEIGHT,..."
```
johnnyluo, line 336: Argument order diverges between the two vote memos: single is `QBTC_VOTE:<OPTION>:<ID>` but weighted is `QBTC_VOTEW:<ID>:<OPTIONS>`. Each parser is internally consistent, but the flipped proposal-id position is a footgun for future edits. Consider aligning both to id-first, or at least documenting the divergence here.

VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
```diff
@@ -293,6 +305,66 @@ struct DefiChainMainScreen: View {
         return coin
     }
 
+    /// Builds a single-option QBTC governance vote tx straight from the
+    /// proposal + chosen option and pushes it to the existing verify → ML-DSA
+    /// keysign flow. The memo (`QBTC_VOTE:<OPTION>:<ID>`) is what
+    /// `QBTCHelper.buildMsgVote` consumes; the dictionary is display-only so
+    /// verify reads "Vote <OPTION> on Proposal #N" rather than the raw memo.
+    func onGovernanceVote(proposal: CosmosGovProposal, choice: CosmosGovVoteChoice) {
```
johnnyluo, line 333: Memo construction (`QBTC_VOTE:`/`QBTC_VOTEW:` strings + the `weightString`/`weightPercentString` helpers) is business logic living in the View, which `architecture.md` says belongs in a ViewModel/Service. It mirrors the screen's existing navigation-wiring style, so this is a soft flag — but pulling the memo assembly into a small pure helper next to `QBTCHelper` would make it unit-testable independent of the view.

VultisigApp/VultisigApp/Features/Defi/DefiChain/View/Governance/QBTCGovernanceView.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    @State private var selectedProposal: CosmosGovProposal?
+
+    var body: some View {
+        Group {
+            if viewModel.isLoading && viewModel.activeProposals.isEmpty && viewModel.pastProposals.isEmpty {
+                ProgressView()
+                    .frame(maxWidth: .infinity)
+                    .padding(.top, 32)
+            } else if viewModel.loadFailed && viewModel.isEmpty {
```
johnnyluo, line 34: Minor: if a refresh fails *after* data is already loaded, `loadFailed` is true but `isEmpty` is false, so this branch is skipped and the user sees stale proposals with no indication the refresh failed. Acceptable, but a transient toast/banner on refresh failure would be friendlier.

VultisigApp/VultisigAppTests/Blockchain/Cosmos/Signing/QBTCVoteWeightedVectorFixture.swift
```diff
@@ -0,0 +1,41 @@
...
+
+enum QBTCVoteWeightedVector {
+    static let voter = "qbtc1voter00000000000000000000000000000000"
+    static let mldsaPubKeyHex = String(repeating: "ab", count: 1312)
+    static let chainID = "qbtc-testnet"
+
+    /// The memo the weighted-vote UI emits for proposal 42, YES=0.7 ABSTAIN=0.3.
+    static let memo = "QBTC_VOTEW:42:YES=0.7,ABSTAIN=0.3"
+
+    static let bodyBytesB64 = "CokBCiMvY29zbW9zLmdvdi52MWJldGExLk1zZ1ZvdGVXZWlnaHRlZBJiCCoSKnFidGMxdm90ZXIwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMBoYCAESFDAuNzAwMDAwMDAwMDAwMDAwMDAwGhgIAhIUMC4zMDAwMDAwMDAwMDAwMDAwMDA="
```
johnnyluo, line 46: Please confirm how these literals were generated. The header says they're derived from the proto/cosmjs contract (not iOS output) — if they were actually captured from iOS's own encoder, the byte-equality test is circular and proves nothing. (The fake `signatureHex = "deadbeef…"` correctly keeps the TxRaw assertion about encoding rather than crypto — good.)

johnnyluo (commented):
One more follow-up on the vote-initiation path.

VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
```diff
@@ -293,6 +305,66 @@ struct DefiChainMainScreen: View {
...
+        guard let nativeCoin else { return }
```
johnnyluo: Both vote entry points only guard on `nativeCoin` existing, not on it having enough QBTC to cover the tx fee. Casting a vote is still an on-chain tx that costs gas, so a user with 0 (or dust) balance can walk all the way into verify → ML-DSA keysign only to have the broadcast fail. Please add a balance check before starting the vote flow — ideally disable/grey the vote controls (with a hint) when the native QBTC balance can't cover the fee, and/or guard here so we fail fast with a clear message rather than at broadcast. Same applies to `onGovernanceWeightedVote`.

VultisigApp/VultisigApp/Blockchain/Cosmos/Service/QBTC/QBTCGovService.swift
gastonm5: Done. Made `QBTCGovServiceProtocol: Sendable` (commit bf7f820). The concrete `QBTCGovService` struct already satisfies it, and this future-proofs the `group.addTask { [service, logger] ... }` captures in `resolveTallies`/`resolveMyVotes` against strict concurrency.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/QBTCGovernanceViewModel.swift
gastonm5: Done. Bounded the my-vote fan-out to active + a recent-past slice instead of every proposal. It now resolves my-vote for `active + past.prefix(myVoteRecentPastLimit=10)`, matching the scope tallies use (active) plus a small recently-ended window for the "you voted" badge. On a populated chain this caps the per-refresh concurrency at the active window + 10 rather than scaling with `defaultProposalLimit` (~100).

VultisigApp/VultisigApp/Blockchain/Cosmos/Signing/QBTCHelper.swift
gastonm5: Order is chain-defined, documented, not changed. The flipped proposal-id position is dictated by the QBTC chain on-chain memo parser (each parser is internally consistent), so realigning would break parsing/consensus. Left both orders as-is and added doc comments at both `QBTCHelper.buildMsgVote` and `buildMsgVoteWeighted` (and in the new `QBTCGovVoteMemo` builder) noting the divergence is intentional and chain-contract-defined so a future edit does not "fix" it.

VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
gastonm5: Extracted to `QBTCGovVoteMemo`, unit-tested. The `QBTC_VOTE:`/`QBTC_VOTEW:` memo construction plus the `weightString`/`weightPercentString` (and the weighted display-value join) helpers now live in a pure `QBTCGovVoteMemo` enum next to `QBTCHelper` (`Blockchain/Cosmos/Signing/QBTCGovVoteMemo.swift`). The view calls `QBTCGovVoteMemo.singleVote(...)` / `.weightedVote(...)` instead of assembling strings inline. Added `QBTCGovVoteMemoTests` (7 tests) covering both memo shapes, the arg-order divergence, and the weight formatting — all green.

VultisigApp/VultisigApp/Features/Defi/DefiChain/View/Governance/QBTCGovernanceView.swift
gastonm5: Acknowledged — acceptable for the PoC, not building the toast now. The refresh-fail-after-load case (stale proposals shown with no indication) is left as-is and tracked as a follow-up (a transient refresh-failure banner). Added a short code comment at that branch noting the accepted behavior + follow-up so it is not mistaken for an oversight.

VultisigApp/VultisigAppTests/Blockchain/Cosmos/Signing/QBTCVoteWeightedVectorFixture.swift
gastonm5: Honest finding: I could not confirm these literals were generated from an external (proto/cosmjs) source, and there is no committed generator script or recording in the repo to prove it. The fixture, the `buildMsgVoteWeighted` encoder, and the test all landed together in one commit, with no external derivation artifact. So as it stands the byte-equality assertions are effectively a **regression guard** (they lock the current iOS encoder output); they do not *independently* prove cross-platform agreement — i.e. this is closer to circular than the single-vote cosign vector, whose AuthInfo/SignDoc bytes were genuinely recorded from the Windows/SDK cosmjs encoder.

What I verified objectively: the `bodyBytes` decode to canonical proto3 for `/cosmos.gov.v1beta1.MsgVoteWeighted` (field 1 proposal_id, 2 voter, repeated 3 WeightedVoteOption{1 option, 2 weight}, canonical 18-dec LegacyDec weights), and vultisig-windows uses the cosmjs `MsgVoteWeighted` registry codec (`core/ui/qbtc/dapp/messageRegistry.ts`) for the same type URL — so the *shape* matches the contract. But shape-matching is not the same as bytes reproduced from an external run.

I did not fake a fix. I added an explicit PROVENANCE CAVEAT to the fixture header documenting this and recommending the bytes be regenerated from the cosmjs encoder to make the gate non-circular. Recommendation: regenerate `bodyBytesB64`/`authInfoBytesB64`/`signDocSHA256Hex`/`txRawSHA256HexUpper` from a committed cosmjs/proto script before relying on this as a cross-platform gate.

VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
gastonm5: Added a pre-flight balance guard. New `canCoverVoteFee` computed property on the screen compares the native QBTC `rawBalance` against the chain flat fee from `CosmosStakingConfig.feeAmount(for:)` (the single source of truth for QBTC fee; 800 base units `min_tx_fee`). Because `min_gas_price` is 0 on qbtc-testnet, gas is free and the fee does not vary by message, so this is a **precise** pre-flight (not an approximation): the vote is affordable iff `rawBalance >= feeAmount`. When it is false: (1) both vote entry points fast-fail (`guard let nativeCoin, canCoverVoteFee ...`), and (2) the detail-sheet vote controls (Yes/No/Abstain/NoWithVeto + "vote with weights") are disabled and dimmed with a hint — new localized key `governanceInsufficientBalanceForFee` ("Not enough QBTC to cover the voting transaction fee.", added to all locales). For non-QBTC chains with no fee config the guard is permissive (does not block).

VultisigApp/VultisigAppTests/Blockchain/Cosmos/Signing/QBTCVoteWeightedVectorFixture.swift
gastonm5: Good catch — addressed in 7806b8995. The literals are now **generated by a committed, independent cosmjs script**, so the gate is genuinely cross-platform rather than circular.

- Generator: `VultisigApp/VultisigAppTests/Blockchain/Cosmos/Signing/fixtures/qbtc-voteweighted/generate.mjs` (+ pinned `package.json` / `package-lock.json` / `README.md`). It encodes the full MsgVoteWeighted signing recipe with the cosmjs codec family — `cosmjs-types` `MsgVoteWeighted` / `TxBody` / `AuthInfo` / `SignDoc` / `TxRaw` + `google.protobuf.Any` (the custom `/cosmos.crypto.mldsa.PubKey` value hand-encoded as proto field 1, since cosmjs has no codec for the ML-DSA key), `@cosmjs/crypto` for SHA-256 — the same codecs a Windows/SDK peer uses, computed **independently of iOS's hand-rolled encoder**. Run: `npm install && node generate.mjs`.
- **iOS matches the cosmjs output byte-for-byte** — all four values (`bodyBytesB64`, `authInfoBytesB64`, `signDocSHA256Hex`, `txRawSHA256HexUpper`) are identical to the cosmjs run, and `QBTCVoteWeightedByteEqualityTests` passes (6/6) against the regenerated vectors. So iOS's `buildMsgVoteWeighted` agrees with what a cosmjs peer signs. The provenance header now points at the generator and the regression-only caveat is dropped.
- Two findings while decoding the signed tx: `TxBody.memo` is **empty** (the `QBTC_VOTEW:` memo carries the vote into the message, not into `TxBody.memo`), and the fee denom is **`qbtc`** (not `uqbtc` as the old comment said), amount `800` — both now documented in the header/README.

Leaving this thread open for you to close.

gastonm5: Confirmed non-circular: the literals are generated by the committed cosmjs script `fixtures/qbtc-voteweighted/generate.mjs` (cosmjs-types `MsgVoteWeighted`/`SignDoc`/`TxRaw` + `@cosmjs/crypto` SHA-256), independently of iOS's hand-rolled encoder — so the test genuinely pins iOS against what a cosmjs peer signs, not against iOS's own output.

This actually just surfaced via `3d984a175`: #4613/#4614 switched `QBTCChain.chainID` to the mainnet constant `"qbtc"`, but the vector was still generated for `"qbtc-testnet"`, so the initiator-path SignDoc hash diverged and CI failed. I regenerated the vector via `generate.mjs` with `CHAIN_ID = "qbtc"` — `bodyBytes`/`authInfoBytes`/`txRawSHA256` are byte-identical (chain-id only lives in SignDoc field 3), and `signDocSHA256Hex` moved `ccfd8e7c… → 09d6d990…`, which equals iOS's own output. That confirms iOS's `buildMsgVoteWeighted` is byte-for-byte correct vs the cosmjs contract; only the fixture's chain-id was stale.

johnnyluo (approved):
LGTM

PR #4628, opened 2026-06-22
fix(custom-rpc): show Polygon once and alias polygonV2 overrides (#4606)

### LINKED ISSUE #4606: Custom RPC Endpoints lists Polygon twice (.polygon + .polygonV2, same chainId 137)
## Problem
The **Custom RPC Endpoints** list (Settings → Advanced) shows **"Polygon" twice** — two identical-looking rows (same name, same icon, both "Default endpoint"). Should match Figma ([node 77059-75288](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=77059-75288)), where Polygon appears once.

## Cause
`Features/CustomRPC/Model/CustomRPCSupportedChains.swift` lists both `.polygon` and `.polygonV2`:

```swift
.polygon,
.polygonV2,
```

These are two enum cases for the **same network** (chainId 137) — a MATIC→POL migration artifact. Both resolve to the same display name and icon (`Chain.swift:106-107` both return "Polygon"; differ only by ticker MATIC vs POL), so the list renders two indistinguishable Polygon rows.

## Fix
Expose Polygon **once** in `CustomRPCSupportedChains.all` (drop the redundant case). Since both enum cases share chainId 137, make sure a custom RPC set on the single Polygon row applies to both `.polygon` and `.polygonV2` through the resolution funnel, so no vault loses its override.

gastonm5:
## Summary

Custom RPC Endpoints (Settings → Advanced) listed **Polygon twice**. This shows it once and makes a single override apply to Polygon RPC traffic regardless of which internal Polygon enum case a coin runs as.

## Root cause

`.polygon` and `.polygonV2` are the same network (both `Chain.name` = "Polygon", both icon "matic", both chainId 137 — they differ only by ticker MATIC vs POL). `CustomRPCSupportedChains.all` listed both back-to-back and the picker did not de-dup, so two identical "Polygon" rows rendered (Figma shows one).

The subtler bug: the override store is keyed by `Chain.rawValue` (the enum **case name**), not by chainId. `CustomRPCOverride.chainRaw` is `@Attribute(.unique)`, and `CustomRPCStore.url(for:)` / `set` / `reset` all key on `chain.rawValue`. Because `"polygon" != "polygonV2"`, the two cases held **independent override slots**. Real EVM traffic can resolve through either case (`EvmService.forChain(coin.chain)` → `EvmServiceConfig.getConfig(forChain:resolver:)` → `resolver.url(for: chain)`), so naively deleting `.polygonV2` from the list would fix the visuals but silently drop overrides for a Polygon coin running as `.polygonV2`.

## Fix

- Removed `.polygonV2` from `CustomRPCSupportedChains.all` — one "Polygon" row, matching Figma.
- Added a `storageKey(for:)` normalizer in `CustomRPCStore` that maps `.polygonV2 → Chain.polygon.rawValue`, applied at all three key sites (`url(for:)`, `set`, `reset`). Aliasing lives in the **store** — the single choke point used by both RPC resolution and the UI badge/editor state — so one override covers both cases and the badge logic isn't blind to the V2 case.
- One-time migration in `reloadFromStore()`: a legacy `polygonV2`-keyed row is folded onto the canonical `polygon` key and the orphan row is deleted (required — `chainRaw` is `.unique`, so a stale dup would be unreachable and risk the constraint). When both rows exist, `polygon` wins.
- `CustomRPCDefaultEndpoint.string(for:)` left as-is (both polygon cases already map to `Endpoint.polygonServiceRpcService`).

Closes #4606

## Gate receipts

- `make generate` — Created project ✅
- `make build-check` — **BUILD SUCCEEDED** ✅
- `make test` — **1048 tests, 1 failure**: only `QBTCClaimSecureVaultRoundDriverTests.testRunBtcRoundDoesNotReKickoff` (on the known-flaky list; passed **3/3 in isolation**, fails only under full-suite timing contention; unrelated to this change). All CustomRPC suites (52 tests) pass ✅
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 0 violations in touched files (71 pre-existing warnings elsewhere, e.g. `Chain.swift` trailing comma — not introduced here) ✅

## Tests added

- **Store aliasing** (`CustomRPCStoreTests`): override on `.polygon` is visible via `url(for: .polygonV2)` and vice-versa; `reset(.polygon)` clears both; persisted row always carries the canonical `polygon` key.
- **Migration** (`CustomRPCStoreTests`): a persisted `polygonV2` row surfaces under the `.polygon` lookup after `reloadFromStore()` and the orphan row is removed; both-exist case prefers the existing `polygon` row.
- **EVM funnel** (`CustomRPCPolygonFunnelTests`, real store): override stored under polygon applies to `getConfig(forChain: .polygonV2)` and vice-versa; no override → default for both.
- **Dedup invariant** (`CustomRPCSelectChainFilterTests`): `Set(all.map(\.name)).count == all.count`; list contains `.polygon` exactly once, not `.polygonV2`.

## Manual verify

- Custom RPC list shows a single Polygon row.
- Setting an override on it applies to Polygon RPC calls.
- An override previously set on the old V2 row is preserved after upgrade (migration).

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified Polygon and PolygonV2 custom RPC overrides to share a single storage slot, ensuring consistent settings across both identifiers.
  * Removed PolygonV2 from supported chains list to establish a single source of truth.

* **Data Migration**
  * Automatic migration of legacy PolygonV2 custom RPC configurations to the consolidated Polygon storage on startup.

* **Tests**
  * Enhanced test coverage for Polygon/PolygonV2 aliasing and migration behavior.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Stores/CustomRPCStore.swift
+++ b/VultisigApp/VultisigApp/Core/Stores/CustomRPCStore.swift
@@ -42,7 +42,7 @@ final class CustomRPCStore: @unchecked Sendable {
     func url(for chain: Chain) -> String? {
         lock.lock()
         defer { lock.unlock() }
-        return mirror[chain.rawValue]
+        return mirror[storageKey(for: chain)]
     }
 
     // MARK: - Write path (MainActor + SwiftData)
@@ -56,7 +56,7 @@ final class CustomRPCStore: @unchecked Sendable {
         }
 
         let trimmed = url.trimmingCharacters(in: .whitespacesAndNewlines)
-        let chainRaw = chain.rawValue
+        let chainRaw = storageKey(for: chain)
 
         do {
             let descriptor = FetchDescriptor<CustomRPCOverride>(
@@ -83,7 +83,7 @@ final class CustomRPCStore: @unchecked Sendable {
             return
         }
 
-        let chainRaw = chain.rawValue
+        let chainRaw = storageKey(for: chain)
 
         do {
             let descriptor = FetchDescriptor<CustomRPCOverride>(
@@ -109,7 +109,8 @@ final class CustomRPCStore: @unchecked Sendable {
         }
 
         do {
-            let overrides = try context.fetch(FetchDescriptor<CustomRPCOverride>())
+            var overrides = try context.fetch(FetchDescriptor<CustomRPCOverride>())
+            migrateLegacyPolygonV2(in: &overrides, context: context)
             let snapshot = Dictionary(
                 overrides.map { ($0.chainRaw, $0.url) },
                 uniquingKeysWith: { _, last in last }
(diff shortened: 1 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4633, opened 2026-06-22
feat(swap): cache select-asset token list per chain

### LINKED ISSUE #4610: Swap 'Select asset' re-fetches token list on every chain switch (no per-chain cache)
## Problem
The swap **Select asset** sheet has no per-chain cache. Load Solana's assets, switch to another chain, then select Solana again → it shows **"Loading…"** and re-fetches the whole token list from the network. Every chain (re)selection is a fresh fetch.

## Cause
- `Features/Swap/Views/SwapCoinPickerView.swift` — `.onChange(of: selectedChain)` calls `viewModel.fetchCoins(chain:)` on every chain switch.
- `SwapCoinSelectionViewModel.fetchCoins` always sets `isLoading = true` and calls `logic.fetchCoins` → `TokenSearchService.loadTokens(for:)`.
- `Core/Services/TokenSearchService.swift:17` `loadTokens` has **no cache** — `fetchExternalTokens` hits 1inch (EVM) / Jupiter (Solana) on every call. Nothing keyed by chain is retained.

## Fix
Cache the fetched token list **per chain** (e.g. `[Chain: [CoinMeta]]` in `TokenSearchService` or the view model). On re-select, serve the cached list instantly — skip `isLoading`/the network fetch — and optionally refresh in the background. Add a TTL if staleness matters.

## Cross-platform note
Windows already caches this (React Query `['whitelistedCoins', chain]`), and Android sources from the cached vault-address flow. **This bug is iOS-only.**

gastonm5:
## Summary

The swap **Select asset** sheet had no per-chain token cache. Switching chains and re-selecting one showed **"Loading…"** and re-fetched the entire 1inch/Jupiter token list every time. This adds a per-chain cache so the list serves **instantly on re-select** (no spinner, no network within the TTL window), while the vault's held coins are re-merged on every open.

Closes #4610

## Cache design

- **New `SwapTokenListCache`** (`Core/Services/SwapTokenListCache.swift`) — a `@MainActor final class` (`static let shared`) modeled on `SwapKitTokensCache`, holding `[Chain: (tokens, fetchedAt)]`:
  - sync peek `cached(for:)` + `isStale(_:)` so the MainActor view model can serve the picker synchronously,
  - async `tokens(for:fetch:)` returning a fresh entry within TTL, else running `fetch` and storing it, with **per-chain in-flight `Task` coalescing** (concurrent opens share one fetch) and **fail-open to last-good** on fetch error (returns stale rather than throwing),
  - `setCached(_:for:fetchedAt:)` / `clearCache()` test+debug seams and an injectable `now: Date` on the TTL path.
- **Cached the vault-independent list only.** `TokenSearchService.loadTokens(for:)` now delegates: `SwapTokenListCache.shared.tokens(for: chain) { try await fetchUncached(chain) }` (the prior 1inch/Jupiter + preset + filter body). Both callers — the swap picker and the wallet add-token flow (`TokenSelectionViewModel`) — get caching consistently.
- **TTL = 6h** (`SwapKitConfig.swapTokenListCacheTTL = 6 * 60 * 60`). Token catalogs change rarely; a long window keeps re-selects fully offline. Tunable.
- **Loading UX (the user-visible fix):** `SwapCoinSelectionViewModel.fetchCoins(chain:)` does a sync cache peek on the MainActor. If the chain is cached, it runs the cheap local merge (native + cached external/preset + destination registry + `vault.coins(for:)` + sort) and publishes **without setting `isLoading`** → instant, no spinner. A **stale** entry triggers a silent background re-fetch (still no spinner). **Only a cold load** (no cached entry for that chain) shows the loader. `SwapCoinSelectionLogic.fetchCoins` was split into the network fetch step and a reusable `merge(externalTokens:chain:)` the VM calls with the cached list — dedup/sort/`merge`/`mergeExternal` behavior is unchanged.

## Why the merged result is NOT cached

The merged `SwapCoinSelectionResult` embeds `vault.coins(for:)`, which changes at runtime (e.g. `onSelect → CoinService.addToChain` adds a coin). Caching it would hide newly-added tokens. We cache only the vault-independent external/preset list and re-run the cheap merge (live balance read + sort, on MainActor) on every open.

## Cancellation

`SwapCoinPickerView.reloadCoins()` now cancels the prior in-flight load before starting a new one, and the VM checks `Task.checkCancellation()` before publishing, so a fast chain-switch superseding an in-flight cold fetch cannot publish results for the wrong chain.

## Cross-platform parity

- **Windows** caches the whitelist per chain via React Query keyed `['whitelistedCoins', chain]` (vault-agnostic, `gcTime 24h`) and serves cached data while refetching — instant re-select. This PR matches that model and is slightly better (non-zero TTL avoids even the background refetch within the window).
- **Android** has no token-list cache (its win is instant local paint + progressive streaming).
- All three merge a per-chain remote whitelist with the vault's held coins at the display layer; none cache the merged result.

## Gate receipts

- `make generate` — OK (new files auto-picked; project.pbxproj untouched)
- `make build-check` — **BUILD SUCCEEDED**
- `make test` — **1880 tests, 0 failures** (incl. 7 new `SwapTokenListCacheTests`)
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 0 violations in touched files (71 pre-existing elsewhere, untouched)

New tests (`SwapTokenListCacheTests`): hit-within-TTL serves stored tokens without invoking fetch; past-TTL invokes fetch and updates; per-chain isolation; fail-open returns last-good on fetch error (and error propagates with no prior entry); in-flight coalescing (two concurrent calls → one fetch); clearCache.

## Manual verify checklist

- [ ] Load Solana assets → switch to another chain → re-select Solana = list appears **instantly, no "Loading…"**, no network call.
- [ ] After the TTL elapses, re-select refreshes silently (no spinner).
- [ ] Add a custom token, reopen the picker = the new token appears next open.

## Release Notes

* **New Features**
  * Swap “Select asset” token lists are now cached per chain for **6 hours** to speed up the picker.
  * The app **prefetches** tokens for the active “from” and “to” chains to reduce initial loading.

* **Bug Fixes**
  * Cached tokens are shown instantly; loading indicators appear only for cold fetches.
  * Refresh failures **fall back to the last cached tokens**, while respecting cancellation behavior.
  * Concurrent refreshes for the same chain are **coalesced**, and overlapping picker reloads are prevented.

* **Tests**
  * Added coverage for TTL/staleness, concurrency coalescing, fail-open, cancellation propagation, and cache clearing.


gastonm5:
Thanks @coderabbitai — addressed both nitpicks:

**1. Cancellation test / contract (SwapTokenListCacheTests + SwapTokenListCache):** fixed in 6620bde94. Good catch on the gap — but the cache was actually failing **open on `CancellationError` too**, so the diff's "propagate" expectation and the summary's "return last-good" contradicted each other. The correct contract is: fail-open serves last-good only on real fetch *failures*; cooperative **cancellation propagates** (the caller is tearing down, so masking it with stale would be wrong). Re-throw `CancellationError` in the catch and added `testCancellationErrorPropagatesWithStaleEntry` to lock it.

**2. Inject token-loading into SwapDetailsViewModel:** skipping this one. `prefetchSwapTokens()` is a fire-and-forget cache warmer, and the injectable seam already exists one layer down — `SwapTokenListCache` takes `init(ttl:)` + `setCached` + an injectable `now:`, and its behavior is fully unit-tested independent of the singleton (`SwapTokenListCacheTests`). Adding a VM-level `loadTokens` dependency for an untested fire-and-forget warmer is churn without a consuming test; the singleton call mirrors how the picker's `SwapCoinSelectionLogic` already defaults `service: TokenSearchService = .shared`.

johnnyluo (approved):
LGTM

PR #4642, opened 2026-06-23
fix(swap): in-band lower-gas tie-break + tighten preference band to 50bps

### LINKED ISSUE #4634: Swap routing: include source gas in net-output ranking; tighten preference band 100→50bps
## Summary

iOS already ranks swap quotes by **net destination output** (`expectedNetToAmount`) — it's the reference implementation and the best of the three platforms. Two gaps remain to fully deliver "best net user output": **source gas is not in the ranking metric**, and the **preference band is 1%** (can cost more than the fee we charge). This issue closes those so iOS matches the unified cross-platform metric.

## Current behavior

`VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift`

- `selectBestQuote()` ranks on `quote.expectedNetToAmount(toCoin:)` — net of destination-side provider fees (affiliate/integrator/outbound). ✅ Good.
- **Source gas is excluded** from the metric (`SwapQuote+Ranking.swift`). On Ethereum mainnet a route with higher gross/net-of-fee output but much higher gas can win while netting the user less.
- `providerPreferenceBand = 0.01` (1%) — a near-tie can hand a "preferred" provider up to 1% of user value, 2× the 50 bps affiliate fee.

## Desired behavior
- Include source gas for same-chain EVM routes in the comparable metric (or as an in-band tie-break: among quotes within the band, prefer the lower-gas one before provider-preference order). Document the choice.
- Tighten `providerPreferenceBand` to **0.005 (50 bps)**.
- Keep `expectedNetToAmount` as the destination-side net metric — it's the canonical reference the SDK and Android are converging to.

## Acceptance
- [ ] Source gas included for same-chain EVM (or in-band tie-break, documented)
- [ ] `providerPreferenceBand` = 0.005
- [ ] Existing best-net regression tests still pass
- [ ] Shared cross-platform test vector matches SDK/Android selection

## Context
Follow-up to closed convergence issue #4450. Companion issues on SDK (`vultisig-sdk`) and Android (`vultisig-android`). iOS `expectedNetToAmount` is the reference the other two platforms are matching.

gastonm5:
## What

Improves swap-quote ranking in `SwapService.selectBestQuote` along two axes (closes #4634):

1. **In-band lower-gas tie-break.** Among quotes already within the provider-preference band — economically equivalent on net destination output — prefer the one with lower **source-chain gas** *before* falling back to provider preference. The lower-gas check fires only when **both** compared quotes expose gas in the same native-wei unit (same-chain EVM aggregators: 1inch / KyberSwap / LI.FI). A gas-unknown quote (THORChain / Maya / SwapKit, which expose no router gas at quote time) can never falsely win it; it falls through to the existing `priority(of:)` ordering, then higher net output.

2. **Tighten the preference band 100bps → 50bps.** A "preferred" provider could previously be handed up to 1% (2× the 50bps affiliate fee) over a marginally-better-rate aggregator. 50bps closes that gap.

## How

- New `SwapQuote.sourceGasWei: BigInt?` derives `gas × gasPrice` from the EVM quote tx for the `.oneinch` / `.kyberswap` / `.lifi` cases; `nil` for THOR / Maya / SwapKit.
- Gas is **not** folded into `expectedNetToAmount` — that would require a cross-asset price (source-native wei → destination token) the ranker doesn't have and would disadvantage THORChain/Maya. The destination-side net metric stays the cross-platform anchor; gas is applied only as the in-band tie-break, where the two compared EVM quotes share a unit.

## Scope

Same-chain EVM aggregators only (where source gas is comparable in the same native-wei unit). Cross-chain routes are unaffected.

## Tests

`VultisigAppTests/Services/SwapQuoteRankingTests.swift`:
- Recalibrated the band-boundary fixture for 0.5% (`best * 0.995`).
- Added `test_selectBestQuote_inBandLowerGasWins` (lower-gas in-band EVM quote beats higher net + higher priority) and `test_selectBestQuote_outsideBandBetterRateWins_regardlessOfGas` (a real rate advantage outside the band always wins, regardless of gas).
- All existing best-net / priority tests stay green unchanged.

## Gate

- `make build-check`: **BUILD SUCCEEDED**
- Ranking suite: **21/21 passing, 0 failures**
- SwiftLint: **0 violations** on the changed files (pre-existing repo-wide warnings untouched)

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Improved swap quote selection with gas-aware in-band tie-breaking, preferring lower source-chain gas for closely matched EVM quotes.
  * Refined in-band “effectively tied on rate” threshold and updated selection order for more consistent results.

* **Tests**
  * Updated swap quote ranking tests for the revised in-band calculation.
  * Added regression coverage to confirm lower-gas wins only within the in-band range, while better rates win outside it.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
@@ -16,10 +16,12 @@ extension SwapQuote {
     /// - 1inch/KyberSwap: `dstAmount` is net of swap fees.
     /// - LiFi: `dstAmount` minus integrator fee (charged on the output side).
     ///
-    /// Source-chain gas is intentionally excluded: aggregators on a given chain consume similar
-    /// gas (~200k units), so the destination output dominates the comparison. THORChain Router
-    /// gas on EVM is large but isn't exposed at quote time. A future refinement can subtract gas
-    /// once a native-token price lookup is wired in.
+    /// Source-chain gas is intentionally excluded from this destination-side metric: folding it
+    /// in would require a cross-asset price (source-native wei → destination token) the ranker
+    /// doesn't have, and it only applies to same-chain EVM aggregators — an asymmetric term that
+    /// would disadvantage THORChain/Maya (no router gas at quote time). Source gas is instead
+    /// applied as an in-band lower-gas tie-break in `selectBestQuote` via `sourceGasWei`, where
+    /// the two compared EVM quotes share the same native-wei unit and no price lookup is needed.
     func expectedNetToAmount(toCoin: Coin) -> Decimal? {
         switch self {
         case .thorchain(let quote),
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -195,16 +195,21 @@ struct SwapService {
 
     /// Width of the priority band, as a fraction of the best net output. Quotes whose net
     /// output lands within this band of the best are treated as effectively tied on rate, so
-    /// the higher-priority provider is preferred over a marginally larger raw output. 1%.
-    static let providerPreferenceBand: Decimal = 0.01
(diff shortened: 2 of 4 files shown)


johnnyluo (approved):
LGTM

Issue #4635, opened 2026-06-22
SwapKit: token catalog 5-min TTL hides new tokens; provider gate fails open


realpaaao:
## Summary

Two minor SwapKit cache behaviors that can hurt swap coverage/reliability.

## Details

`VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitConfig.swift`

- **Token catalog TTL (5 min) vs provider TTL (24h).** `tokensCacheTTL = 5 * 60` (`:50`) while `providerCacheTTL = 24 * 60 * 60` (`:42`). New tokens added to Chainflip / NEAR Intents / other SwapKit sub-providers can be invisible in the picker for up to 5 minutes on first open. Acceptable by design, but worth a refresh-on-picker-open so newly-listed pairs surface immediately.
- **Provider gate can fail open.** `isFeatureEnabled` (`:69`) and the provider-cache path: if the `/providers` fetch fails, SwapKit is treated as enabled anyway. The app may then offer SwapKit routes that fail downstream. Consider fail-closed, or an explicit degraded state, when the providers list cannot be loaded.

## Fix
- Refresh the token catalog when the asset picker opens (or shorten/condition the TTL).
- Decide fail-closed vs explicit-degraded for the provider gate on `/providers` fetch failure.

## Notes
Low priority (reliability/coverage, not fund-at-risk). Android may share the pattern via `SwapKitQuoteSource` — check for parity.

PR #4643, opened 2026-06-23
fix(swap): SwapKit provider gate fail-closed on cold cache miss + token catalog refresh-on-open

gastonm5:
## Summary
Two SwapKit cache reliability fixes for #4635.

### 1. Provider gate fails CLOSED on the no-snapshot edge
`SwapKitProviderCache.isEnabled(chain:)` previously returned `true` when the providers list couldn't be loaded at all. On a cold launch where the first `/providers` fetch fails and there is no prior snapshot, SwapKit was advertised for the chain and offered routes that fail downstream.

Now it fails **closed** (`false`) on that edge: SwapKit simply isn't offered for the chain until a refresh succeeds. Other providers still populate the picker, and `SwapService.fetchSwapKitQuote` throws `providerNotEnabled` cleanly. This matches Android's already-fail-closed behavior.

- `isPairSupported(...)` deliberately stays **fail-OPEN** on the same edge — it only governs a `noRoutesFound` 404 -> `amountBelowProviderMinimum` error re-label, so failing it closed would merely degrade a message, not block a swap. The asymmetry is documented in code.
- The fetch-fail-**with**-prior-snapshot last-good path is untouched (benign, correct degraded behavior).

### 2. Token catalog refresh-on-picker-open
Threaded a `forceRefresh` flag through `SwapKitTokensCache.tokens(for:forceRefresh:)` -> `ensureSnapshot`, the `DestinationTokenProvider` protocol, `DestinationTokenRegistry`, and the picker view model. When `true` it skips the TTL-fresh early-return but **keeps** `inFlight` coalescing + last-good fallback.

The destination picker's **first open per presentation** passes `forceRefresh: true` (so a stale catalog, or one that missed the cold-launch fetch, is re-fetched). In-session re-merges (search debounce, chain re-select) stay `false` (instant/offline). `tokensCacheTTL` stays 5m as the background floor.

## Tests
- `SwapKitProviderCacheTests`: no-snapshot fetch-fail now asserts **fail-closed**; added a prior-snapshot **last-good** guard against over-correcting; renamed the pair-support no-snapshot test (still fail-open).
- New `SwapKitTokensCacheTests`: `forceRefresh` bypasses fresh TTL, coalesces concurrent calls, and serves last-good on failure.

## Gate
- `make build-check`: BUILD SUCCEEDED
- `make test`: TEST SUCCEEDED, 0 failures
- `swiftlint`: no new warnings in touched files

## Summary by CodeRabbit

* **Improvements**
  * Coin picker now forces a token refresh on initial sheet load for fresher swap options, while keeping cached data for later in-session picks.
  * Token and destination-token fetching support a `forceRefresh` option that bypasses freshness checks, coalesces concurrent requests, and continues serving the last known good snapshot if refresh fails.
  * Provider availability gating is now more conservative when no prior data exists (fail-closed), improving reliability of SwapKit enablement.
* **Tests**
  * Updated and added coverage for `forceRefresh`, last-good fallback, provider gating edge cases, and concurrent request coalescing.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -543,8 +543,10 @@ private extension SwapService {
         recipientAddress: String?
     ) async throws -> SwapQuote {
         // Provider-cache gate — refuse to call `/v3/quote` for a chain SwapKit
-        // doesn't enable. Fails open if the cache can't be loaded so we don't
-        // silently disable the aggregator on a bad network day.
+        // doesn't enable. Fails CLOSED on the no-snapshot edge (throws
+        // `providerNotEnabled`) rather than offering routes that fail
+        // downstream; once a snapshot exists, `SwapKitProviderCache` serves it
+        // as last-good so a transient network blip doesn't disable SwapKit.
         let fromEnabled = await service.isChainEnabled(fromCoin.chain)
         let toEnabled = await service.isChainEnabled(toCoin.chain)
         guard fromEnabled, toEnabled else {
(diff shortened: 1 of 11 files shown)


johnnyluo (approved):
LGTM

PR #4695, opened 2026-06-29
feat(qbtc): derive Bitcoin/QBTC claim accounts on the fly (#4679)

### LINKED ISSUE #4679: QBTC claim: derive Bitcoin/QBTC accounts on the fly instead of blocking when the chain isn't enabled
## Summary

The QBTC claim flow on iOS requires the vault's Bitcoin and QBTC chains to already be enabled. `QBTCClaimViewModel` reads `vault.nativeCoin(for: .bitcoin)` / `vault.nativeCoin(for: .qbtc)` and blocks with `.missingCoin` when either coin is absent, so a vault that hasn't added those chains can't claim.

Android (PR vultisig/vultisig-android#5065) and Windows derive the Bitcoin/QBTC accounts on the fly from the vault's own keys when the chain isn't enabled, so the claim works regardless of which chains the user has added. iOS should do the same for parity.

## Suggested change

Instead of blocking on `.missingCoin`, derive the native coin in-memory from the native-token template + the vault's keys when it isn't enabled (the claim hash is still recomputed locally from the vault's own keys, never trusted from the QR). Only a genuine derivation failure (vault lacks the key material) should fall back to the blocked/missing-account state.

References:
- `VultisigApp/VultisigApp/Features/QBTCClaim/ViewModels/QBTCClaimViewModel.swift` — `btcCoin`/`qbtcCoin` via `vault.nativeCoin(for:)`, `.blocked(reason: .missingCoin(...))`
- Android equivalent: `ResolveQbtcClaimCoinsUseCase` in vultisig/vultisig-android#5065

Raised from review feedback on the Android PR (cc @rkokhatskyi).

gastonm5:
## QBTC claim: derive Bitcoin/QBTC accounts on the fly

Closes #4679

Today the QBTC claim flow hard-blocks with `.missingCoin` unless the vault already has **both** the Bitcoin and QBTC chains enabled. This brings iOS to parity with Android ([vultisig-android#5065](https://github.com/vultisig/vultisig-android/pull/5065)'s `ResolveQbtcClaimCoinsUseCase`) and Windows: the two native coins are derived **in-memory** from the vault's own keys when the chain isn't enabled, and the claim blocks only on a **genuine** derivation failure (the vault lacks the key material).

Design + grounding: internal wiki plan `wiki/pages/projects/vultisig/qbtc-claim/on-the-fly-coin-derivation-4679-plan.md` (plus the QBTC-claim `overview.md` for context).

### What changed

- **New resolver** `Features/QBTCClaim/Services/QBTCClaimCoinResolver.swift`. Per chain: returns `vault.nativeCoin(for:)` if enabled; otherwise derives via the pure, **non-persisting** `CoinFactory.create(...)` from the native-token template + vault keys (the same precedent `SwapCryptoLogic.getDefaultCoin` uses); otherwise throws `Error.derivationFailed(chainName:)`. Templates: QBTC → `TokensStore.qbtc`, Bitcoin → the inline native entry in `TokensStore.TokenSelectionAssets`. Injected via initializers (default-constructed) for testability.
- **Three call sites** rerouted through the resolver (mirrors Android's three wirings):
  - Initiator `QBTCClaimViewModel` — resolves once at the top of `load()` and stores `resolvedCoins`; the address-type guard, `fetchUtxos`, `startClaim()`, `prepareSecureVaultPair()` and the keysign payload all read the stored coins. On `derivationFailed` it lands on the existing `.blocked(reason: .missingCoin(chainName:))` (reusing the existing UI + `qbtcClaimMissingCoinDetail` string — **no new locale keys**).
  - Peer `QBTCClaimJoinDriver.runInternal()` + `QBTCClaimJoinView.doneScreen` — derive from the peer's own vault; the now-unused `missingBitcoinCoin`/`missingQbtcCoin` error cases were removed.
  - `KeysignDiscoveryViewModel` — the `vault.nativeCoin(for: .qbtc)` read now resolves/derives the QBTC claimer address.
- **Entry-point parity** — `ChainDetailViewModel.qbtcClaimBitcoinCoin()` derives the Bitcoin coin in-memory for the claim-eligibility check, so the Claim button on the QBTC chain-detail screen appears even when the Bitcoin chain isn't enabled (matching the claim flow). Logic lives in the ViewModel, not the `View`.
- **Tests**: new `QBTCClaimCoinResolverTests` (enabled-as-is / derives-when-missing / blocked-on-failure / single-chain), new `QBTCClaimViewModelTests` (non-quantum vault → `.blocked(.missingCoin)`), `QBTCClaimJoinDriverTests` gains a derive-on-the-fly peer regression test, and `ChainDetailViewModelTests` covers the entry-point BTC resolution (BTC screen / QBTC-with-BTC-enabled / derive-when-not-enabled).

### Eligibility nuance

Bitcoin (ECDSA) derives for any standard vault, so it effectively never fails. QBTC (MLDSA) requires a non-empty `vault.publicKeyMLDSA44` — `CoinFactory.create` routes to `createMLDSACoin` which throws without it. So the **only** genuine derivation failure is a non-quantum vault, and the real eligibility gate is now "MLDSA-capable vault" (matching Android). The derived coins are **never** appended to `vault.coins` or inserted into Storage — they stay local to the claim flow.

### Security note

Unchanged. The claim hash is always recomputed locally from the vault's own (derived) coins, never trusted from the QR/relay. Deriving the coin in-memory yields the *identical* deterministic hash on every device (initiator and peer derive the same BTC ECDSA + QBTC MLDSA coins from the same vault keys), and the proof service's echoed hashes are still re-verified against the local values.

### KeyImport allow-list divergence (deliberate)

`CoinFactory.create` does not call `CoinService.assertChainAllowed` (that guard lives only in the persisting `addToChain` path). Deriving in-memory therefore bypasses the key-import chain allow-list. This is acceptable for the claim — it's read-only derivation, nothing is persisted — but flagging it as a deliberate divergence for review.

### Surfaced reason for a non-quantum vault

For a non-quantum vault we intentionally reuse `.missingCoin(chainName: "QBTC")` ("Your vault is missing the QBTC coin needed to claim.") — no new strings, per the plan and product sign-off. A dedicated "vault is not quantum-capable" reason can be a follow-up if desired.

### Entry-point parity (now included)

The QBTC chain-detail Claim button drives off a real BTC-UTXO eligibility check, which previously needed the Bitcoin chain to already be enabled. It now derives the Bitcoin coin in-memory (via the same resolver) when the chain isn't enabled, so a quantum vault with QBTC enabled but Bitcoin not added still surfaces the Claim entry point. Since being on the QBTC chain-detail screen implies QBTC is enabled (which requires an MLDSA vault), this path only ever derives for quantum-capable vaults. The BTC promo banner on the Bitcoin screen is unchanged (it already implies Bitcoin is enabled).

## Summary by CodeRabbit

* **New Features**
  * QBTC claim flows now derive and reuse BTC/QBTC coin accounts in-memory when chains aren’t enabled (when key material is available).
  * Claim screens share the same resolved coin accounts across steps for more consistent UI and completion handling.

* **Bug Fixes**
  * Improved claim join flow so it no longer fails early due to temporarily unavailable enabled coins.
  * Enhanced QBTC eligibility checks to use the correct BTC coin source across Bitcoin vs. non-Bitcoin screens.

* **Tests**
  * Added coverage for coin resolution, join driver behavior, and blocked-state handling.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignDiscoveryViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignDiscoveryViewModel.swift
@@ -124,7 +124,10 @@ class KeysignDiscoveryViewModel: ObservableObject {
                 // both devices share the same hash. The claimer's QBTC
                 // address comes from this device's own vault.
                 let btcCoin = keysignPayload.coin
-                if let qbtcCoin = vault.nativeCoin(for: .qbtc),
+                // Resolve QBTC from this device's own vault, deriving it
+                // in-memory when the chain isn't enabled so both devices
+                // share the same claimer address (and thus the same hash).
+                if let qbtcCoin = try? QBTCClaimCoinResolver().resolve(vault: vault, chain: .qbtc),
                    let compressedPubkey = Data(hexString: btcCoin.hexPublicKey) {
                     do {
                         let hashes = try QBTCClaimHashes.computeAll(
--- a/VultisigApp/VultisigApp/Features/QBTCClaim/Services/QBTCClaimCoinResolver.swift
+++ b/VultisigApp/VultisigApp/Features/QBTCClaim/Services/QBTCClaimCoinResolver.swift
@@ -1,0 +1,97 @@
+//
+//  QBTCClaimCoinResolver.swift
+//  VultisigApp
+//
+//  Resolves the Bitcoin and QBTC accounts a QBTC claim is derived from.
+//  Neither chain has to be enabled in the vault: an already-enabled
+//  native coin is used as-is, otherwise the account is derived in-memory
+//  from the native-token template + the vault's own keys via the pure,
+//  non-persisting `CoinFactory.create` (the same precedent
+//  `SwapCryptoLogic.getDefaultCoin` uses). Nothing is appended to
+//  `vault.coins` or written to storage — the resolved coins stay local
+//  to the claim flow, so the deterministic claim hash is identical to
+//  the one computed when the chains are enabled.
+//
+//  Bitcoin (ECDSA) derives for any standard vault; QBTC (MLDSA) requires
+//  a non-empty `publicKeyMLDSA44`, so the only genuine failure is a
+//  non-quantum vault — surfaced as `Error.derivationFailed`.
+//
+
+import Foundation
+
+@MainActor
+struct QBTCClaimCoinResolver {
+
+    /// `nonisolated` so the resolver can be default-constructed as an
+    /// initializer argument (default arguments are evaluated in a
+    /// nonisolated context). The synthesized initializer can't be used —
+    /// it would inherit `@MainActor` and fail as a default argument — so
+    /// this explicit one is required despite looking redundant. The
+    /// stateful work (the `@Model`-reading `resolve` methods) stays
+    /// main-actor isolated.
+    nonisolated init() {} // swiftlint:disable:this unneeded_synthesized_initializer
+
+    enum Error: LocalizedError, Equatable {
+        /// The native coin for `chainName` is neither enabled nor
+        /// derivable from the vault's keys (e.g. a non-quantum vault has
+        /// no MLDSA-44 key to derive the QBTC account).
+        case derivationFailed(chainName: String)
+
+        var errorDescription: String? {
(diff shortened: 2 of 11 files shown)


johnnyluo (approved):
LGTM

PR #4705, opened 2026-07-01
feat(solana): fetch staking min-delegation dynamically from a public node

### LINKED ISSUE #4694: Solana staking: fetch min delegation dynamically (getStakeMinimumDelegation) instead of hardcoding 1 SOL
## Problem
The Solana native-staking delegate form hardcodes the minimum delegation at **1 SOL** (`SolanaStakingConfig.minDelegationFloorLamports = 1_000_000_000`, enforced via `SolanaDelegateTransactionViewModel.minimumDelegationDecimal`) because the Vultisig RPC proxy (`api.vultisig.com/solana/`) **blocks** the `getStakeMinimumDelegation` method.

This is correct **today**, but the cluster minimum delegation is **not a constant** — it changed recently:
- Historically it was effectively ~1 lamport.
- The "1 SOL minimum" feature gate has since activated on mainnet (validators on Agave 4.0+), so `getStakeMinimumDelegation` now returns `1_000_000_000`.
- Verified live (June 2026): both `api.mainnet-beta.solana.com` and `solana-rpc.publicnode.com` return `1000000000`. Tracking: `solana-program/stake#250` (was still pre-activation as of Feb 2026).

A hardcoded constant goes stale on the next change — either rejecting valid sub-minimum stakes, or (worse) letting through amounts the chain reverts with `StakeError::InsufficientDelegation` (custom error 12).

## Proposed fix
1. **Backend:** unblock `getStakeMinimumDelegation` on the `api.vultisig.com` Solana RPC proxy (currently on the method denylist).
2. **iOS:** add `getStakeMinimumDelegation` to `SolanaAPI`/`SolanaService`; query it (cache ~24h, like the rent-exempt reserve); use it in `SolanaDelegateTransactionViewModel.minimumDelegationDecimal` instead of the hardcoded constant. Keep `minDelegationFloorLamports` (1 SOL) as the offline/last-known **fallback** when the RPC is unavailable.

The effective wallet minimum is already mostly dynamic: `minimumUserInput = minDelegation + rentExemptReserve + txFee`, where `rentExemptReserve` is already fetched live via `getMinimumBalanceForRentExemption(200)`. This issue makes the `minDelegation` term dynamic too.

## Acceptance
- [ ] `getStakeMinimumDelegation` allowed by the Vultisig Solana proxy.
- [ ] iOS reads the min delegation live (cached), not a hardcoded constant.
- [ ] Falls back to the 1 SOL constant when the RPC is unavailable.
- [ ] The delegate form's enforced minimum reflects the live value.

## Context
Follow-up to the Solana native staking work (PR #4690). Refs: [getStakeMinimumDelegation docs](https://solana.com/docs/rpc/http/getstakeminimumdelegation) · [solana-program/stake#250](https://github.com/solana-program/stake/issues/250).

gastonm5:
## What & why

Closes #4694.

The Solana delegate form enforced a **hardcoded 1 SOL** minimum active delegation. That value is a **feature-gate-activated network constant**, not a permanent one — and the Vultisig RPC proxy (`api.vultisig.com/solana/`) **blocks** `getStakeMinimumDelegation`:

```
→ -32600 "Unsupported method: getStakeMinimumDelegation on SOLANA_MAINNET"
```

so it can't be read through the normal path. If the network minimum changes, the stale hardcode either rejects valid stakes or lets through amounts that revert on-chain (`StakeError.InsufficientDelegation`).

## Approach

The method is standard (Solana ≥1.14) and **public nodes serve it keyless** (verified live):

| Endpoint | `getStakeMinimumDelegation` |
|---|---|
| `api.vultisig.com/solana/` (proxy) | ❌ unsupported |
| `solana-rpc.publicnode.com` | ✅ `1000000000` (primary) |
| `api.mainnet-beta.solana.com` | ✅ `1000000000` (fallback) |

So this reads the value **directly from a public node**, cached 24h, with the 1 SOL constant kept as the offline fallback.

The min-delegation is **network-global** and used **only as a form-validation floor** — it never touches the keysign payload or MPC byte-parity. It is deliberately **not** routed through the custom-RPC override (`resolver.resolvedURL(for: .solana)`), which would repoint *all* Solana RPC including `sendTransaction`; instead a dedicated read targets a hardcoded public host with `usesProxyPath: false`.

## Changes

- **`SolanaAPI`** — `getStakeMinimumDelegation` method (no params) + envelope; `minDelegationPublicHosts` (PublicNode → mainnet-beta); `SolanaGetStakeMinimumDelegationResponse`.
- **`SolanaService.fetchSolanaMinDelegation()`** — 24h cache, iterates public hosts, throws only if all fail.
- **`SolanaStakingService`** — `fetchMinDelegation()` pass-through on both protocols (caching lives on `SolanaService`).
- **`SolanaDelegateTransactionViewModel`** — `minimumDelegationLamports` seeded to the 1 SOL floor, adopted live in `onLoad`, **falls back to the floor on failure**.
- **Tests** — envelope + DTO decode; VM adopt/fallback/seed; updated 3 staking-service fakes.

## Acceptance criteria (#4694)

- [x] App fetches the minimum dynamically (public node, since the proxy blocks it)
- [x] Cached (~24h)
- [x] Graceful fallback when RPC unavailable (documented 1 SOL floor)
- [x] Delegate form reflects the live value

> Note: the issue's option 1 (unblock the method on the Vultisig proxy) is a **backend** change and remains a nice-to-have. This PR unblocks the iOS side without it; if the proxy later allows the method, the host list just gets the proxy prepended and the public nodes stay as fallback.

## Testing

- SwiftLint: **0 violations** (9 changed files).
- `make test` (full iOS suite): all 6 new tests pass + the whole Solana staking suite passes. One **pre-existing, unrelated** failure remains — `StakingValidatorProjectionTests.testCosmosEmptyMonikerFallsBackToTruncatedAddress` (host-locale decimal separator: asserts `"12.3%"`, gets `"12,3%"`; a Cosmos APY formatting flake outside this diff).
- Pending: on-device delegate-form check.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading Solana’s live minimum stake delegation value.
  * Staking screens now use the network value to validate delegation amounts, with a fallback if the network lookup fails.

* **Bug Fixes**
  * Improved validation so delegation amounts stay accurate even when the live value changes.
  * Added safer fallback behavior to keep staking actions usable offline or during RPC failures.

* **Tests**
  * Expanded automated coverage for staking minimum delegation fetching, decoding, caching, and validation behavior.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
@@ -16,6 +16,18 @@ struct SolanaAPI: TargetType {
     /// The proxy path appended to the default host.
     static let proxyPath = "/solana/"
 
+    /// Keyless public JSON-RPC endpoints used ONLY for
+    /// `getStakeMinimumDelegation`, which the Vultisig proxy blocks. The value
+    /// is a network-global constant (identical on every mainnet node), used just
+    /// as a form-validation floor, and never touches signing — so reading it off
+    /// a public node is safe. Tried in order: PublicNode is app-friendly and
+    /// keyless; mainnet-beta is the official keyless fallback. Each is a complete
+    /// endpoint, so `usesProxyPath` is `false` when targeting them.
+    static let minDelegationPublicHosts: [URL] = [
+        URL(staticString: "https://solana-rpc.publicnode.com"),
+        URL(staticString: "https://api.mainnet-beta.solana.com")
+    ]
+
     let baseURL: URL
     /// `true` for the default Vultisig proxy (RPC under `/solana/`); `false`
     /// when a custom override supplies a complete JSON-RPC endpoint.
@@ -53,6 +65,11 @@ struct SolanaAPI: TargetType {
         case getMinimumBalanceForRentExemption(size: Int)
         /// Network inflation rate for the current epoch.
         case getInflationRate
+        /// Network minimum active delegation (lamports) enforced by the Stake
+        /// program. Takes no params. The Vultisig proxy blocks this method, so
+        /// it is read from a public node; the value is a network-global constant
+        /// used only as a form-validation floor and never touches signing.
+        case getStakeMinimumDelegation
     }
 
     enum TokenAccountFilter {
@@ -146,6 +163,8 @@ struct SolanaAPI: TargetType {
             )
         case .getInflationRate:
             return .requestParameters(rpcEnvelope(method: "getInflationRate", params: [] as [Any]), .jsonEncoding)
+        case .getStakeMinimumDelegation:
+            return .requestParameters(rpcEnvelope(method: "getStakeMinimumDelegation", params: [] as [Any]), .jsonEncoding)
         }
     }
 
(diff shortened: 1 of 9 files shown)


johnnyluo (approved):
LGTM

PR #4718, opened 2026-07-01
fix(ui): round the consent-checkbox circle on Send/Swap verify (#4715)

### LINKED ISSUE #4715: iOS: consent checkbox checkmark circles clipped on left edge (Send/Swap verify)
### Observed
On the verify screen (Send/Swap), the green circular checkmark icons next to the consent checkboxes ("The amount is correct", "I'm sending to the correct address") are **clipped on their left edge** — the circle has a flat left side instead of being fully round.

### Probable cause (leads — clip container to confirm)
- The consent checkboxes are rendered as the footer content with **zero horizontal padding**: `SendCryptoVerifySummaryView.swift:44-48` puts `contentFooter()` in a `VStack(...).padding(.horizontal, contentPadding)`, and `SendVerifyScreen.swift:134-138` passes no `contentPadding` (defaults to `0`). So the checkbox row sits flush at the container's left edge.
- The check circle's 1pt border sits directly on the frame edge — `Checkbox.swift:57-60` uses `.frame(width: 24, height: 24).overlay(Circle().stroke(color, lineWidth: 1))`, so half the stroke is on the very edge of the 24×24 box.
- Combined with a clipping ancestor (the verify card / scroll container), the leftmost sliver of the circle gets cut. Everything else on the screen is inset (the summary card uses `.padding(24)`), so the checkboxes are the one row with no left inset.

### Fix directions
- Give the checkbox footer horizontal padding (pass a non-zero `contentPadding`, or pad the `checkboxes` VStack in `SendVerifyScreen`/`SwapVerifyScreen`), and/or
- Inset the circle so its stroke isn't on the frame edge (e.g. `.padding(1)` around the check, or draw the stroke inside).

Please confirm the exact clipping ancestor when fixing — I traced the padding asymmetry but didn't run it on-device.

### Repro
Send (or Swap) → verify screen → observe the two consent checkmark circles clipped on the left.

Files:
- `VultisigApp/VultisigApp/Components/Checkbox.swift:49-61`
- `VultisigApp/VultisigApp/Features/Send/Views/SendCryptoVerifySummary/SendCryptoVerifySummaryView.swift:44-48`
- `VultisigApp/VultisigApp/Features/Send/Views/Screens/SendVerifyScreen.swift:134-138` (same pattern in `Features/Swap/Views/Screens/SwapVerifyScreen.swift`)

gastonm5:
## Problem

The green circular checkmark ring next to the consent checkboxes on the Send/Swap **verify** screens was clipped on its left edge — a flat left side instead of a fully round circle.

## Root cause (two compounding factors)

1. **Stroke overflow** — `Components/Checkbox.swift` `check` draws the ring as a 24×24 frame with `.overlay(Circle().stroke(color, lineWidth: 1))`. `.stroke` centers the 1pt line on the circle path, so ~0.5pt of the stroke spills **outside** the 24×24 frame on every side.
2. **Flush row clipped by the ScrollView** — the consent checkboxes are the bottom content of the verify scroll area with no left inset, so the checkbox frame sits flush at x≈0 of the scroll container and the ~0.5pt sliver at x<0 lands outside the bounds, which the ScrollView clips → flat left edge.

### Confirmed clipping ancestor
- **Send:** the `ScrollView` in `SendCryptoVerifySummaryView.fields` (`Features/Send/Views/SendCryptoVerifySummary/SendCryptoVerifySummaryView.swift:42-51`). Its inner `VStack { summary; contentFooter() }` uses `.padding(.horizontal, contentPadding)` where `contentPadding` defaults to **0** and the Send call site passes none → the checkbox footer is flush and the ScrollView clips the overflow.
- **Swap:** the Swap verify screen does **not** use `SendCryptoVerifySummaryView`; it has its own inline `summary`/`checkboxes` inside a `ScrollView` in `SwapVerifyScreen.fields` (`Features/Swap/Views/Screens/SwapVerifyScreen.swift:94-102`). Same mechanism.

## Fix

`Checkbox.swift`: `Circle().stroke(color, lineWidth: 1)` → `Circle().strokeBorder(color, lineWidth: 1)`. `strokeBorder` insets the stroke so it draws **inside** the 24×24 bounds and never overflows to be clipped, fixing the clip regardless of ancestor padding, for all `Checkbox` usages (Send verify, Swap verify, VaultDeletionConfirm, VaultBackupNow). The ring's outer diameter becomes exactly 24 (was 24+1) — imperceptible elsewhere, and the intended behavior.

One-line change, no layout touched.

## Visual verification (render-time)

Reproduced the clipping ancestor with a SwiftUI `ImageRenderer` harness (no app changes committed): with `.stroke` the ring clips flat against the flush-left boundary (a "D" shape); with `.strokeBorder` it renders fully round and contained.

> Before/after render PNG is available locally (`checkbox_before_after.png`) and can be drag-dropped into this PR for inline display. Summary: `.stroke` → ring clipped flat on the left; `.strokeBorder` → ring fully round and contained.

## Tests

No snapshot regression test added: a device-layout perceptual snapshot (repo convention, 0.99999 precision) would not reliably catch a sub-pixel edge clip on a 24pt ring, and a tight strict-pixel snapshot of an anti-aliased circle risks CI flakiness across machines/renderers. The `strokeBorder` change is self-evidently correct and covered by the visual before/after above.

## Gate

`xcodebuild test` (VultisigApp scheme, iPhone 17 Pro Max, worktree-local derived data) + SwiftLint on the touched file. Build SUCCEEDED; the only failing test is the known `"12,3%"` locale non-failure (`StakingValidatorProjectionTests.testCosmosEmptyMonikerFallsBackToTruncatedAddress`), unrelated to this change.

Closes #4715

### DIFF
--- a/VultisigApp/VultisigApp/Components/Checkbox.swift
+++ b/VultisigApp/VultisigApp/Components/Checkbox.swift
@@ -56,7 +56,7 @@ struct Checkbox: View {
             .opacity(isChecked ? 1 : 0)
             .overlay(
                 Circle()
-                    .stroke(color, lineWidth: 1)
+                    .strokeBorder(color, lineWidth: 1)
             )
     }
 


johnnyluo (approved):
LGTM

PR #4728, opened 2026-07-03
fix(zcash): re-plan memo txs to the ZIP-317 conventional fee

### LINKED ISSUE #4726: [BUG] ZEC memo txs (Maya-route swaps) plan an underpaying ZIP-317 fee — extension co-sign fails, mobile-only broadcast rejected
## Summary

ZEC transactions **with a memo** (i.e. MayaChain-routed ZEC swaps — the swap instruction rides in an OP_RETURN) are planned with an underpaying ZIP-317 fee on iOS. Two user-visible failures:

1. **Extension-initiated swap, iOS co-signs → keysign fails every time.** The extension's SDK signer plans the correct (higher) conventional fee; iOS plans WalletCore's underpaying `zip0317` fee. Different fee → different change output → different preimage hashes → MPC keysign cannot converge.
2. **Mobile-only signing → tx signs but the network rejects it at broadcast** (fee below the ZIP-317 conventional minimum).

Live user report (via Android co-sign, same class): ZEC → USDC initiated on extension, co-sign failed, trailing relay HTTP 404 shown after the failed sign — the 404 is the torn-down session, not the cause.

## Root cause

WalletCore's `zip_0317` planner sizes an OP_RETURN output at a flat ~34 bytes — one logical action short of what ZIP-317 charges — and ignores `byteFee` in that mode (verified against real WalletCore in vultisig/vultisig-windows#4145 / vultisig/vultisig-windows#4153). So any memo tx plans exactly one action short (e.g. 15000 zats planned vs 20000 required). WalletCore 4.7.0 does **not** fix this (no Zcash changes in 4.6.15→4.7.0).

The TS SDK fixed this in vultisig/vultisig-sdk#773: when the `zip_0317` plan underpays the byte-accurate conventional fee, re-plan with `zip_0317` **off** (where `byteFee` is honoured) and bump `byteFee` until the fee clears (`packages/core/mpc/keysign/signingInputs/resolvers/utxo.ts`, `planZcashConventionalFee`).

**iOS never got the equivalent fix.** `VultisigApp/VultisigApp/Blockchain/UTXO/Signing/UTXOChainsHelper.swift` sets `zip0317 = coin == .zcash` on both the swap path (memo → `outputOpReturn`) and the send path, with no conventional-fee check or re-plan.

Note: SwapKit ZEC swaps are unaffected — `SwapKitZcashSigner` uses a frozen plan from the SwapKit PSBT, and NEAR-route swaps are plain transfers with no memo. Only the memo path (Maya-native swaps, sends with memo) is broken.

## Fix

Port sdk#773's `planZcashConventionalFee` algorithm into `UTXOChainsHelper`:
1. Plan with `zip0317 = true` as today.
2. Compute the byte-accurate ZIP-317 conventional fee for the planned tx (recipient + change + OP_RETURN with PUSHDATA/CompactSize thresholds — see `packages/core/chain/chains/utxo/fee/zip317.ts` in the SDK for the reference math).
3. If the plan's fee is below it, re-plan with `zip0317 = false` and bump `byteFee` until the fee meets the conventional fee (SDK caps at 5 bumps).
4. Plain (no-memo) sends keep the `zip0317` plan unchanged.

⚠️ The re-planned fee must be **byte-identical to the SDK's result** — MPC co-signing requires all devices to derive the same digests. Divergence here just moves the keysign failure. This is exactly the class of bug the cross-platform signing-vector net (vultisig/commondata#94) should pin; a ZEC-memo vector should be added there as part of this fix.

Long-term, the clean fix is upstreaming the OP_RETURN action-count correction to trustwallet/wallet-core so all platforms converge without workarounds.

## Verification (all reproduced live on 2026-07-03)

- SwapKit rail for ZEC→USDC is healthy on all three proxies (quote + swap 200s) — separate from this bug.
- Maya quote for ZEC→ETH.USDC returns a valid quote, so Maya-route swaps are offered and users do hit this path.
- Extension prod (0.2.22) + desktop v1.0.64+ carry sdk#773; iOS and Android carry no equivalent.

Sibling issue filed on vultisig-android (same gap in `UtxoHelper.kt`).

gastonm5:
## Summary

Closes #4726

ZEC transactions **with a memo** — MayaChain-routed ZEC swaps (the swap instruction rides in an OP_RETURN) and plain sends with memo — were planned with an underpaying ZIP-317 fee on iOS:

- **Extension-initiated swap, iOS co-signs → keysign fails every time** (the SDK plans the correct higher fee; iOS plans WalletCore's underpaying `zip0317` fee → different fee → different change output → preimage-hash mismatch → MPC cannot converge).
- **Mobile-only signing → network rejects at broadcast** (`tx unpaid action limit exceeded` — fee below the ZIP-317 conventional minimum).

## Root cause

WalletCore's `zip_0317` planner sizes an OP_RETURN output at a flat ~34 bytes — one logical action short of what ZIP-317 charges — and **ignores `byteFee` in that mode** (verified against real WalletCore in vultisig/vultisig-windows#4145 / #4153; WalletCore 4.7.0 does not fix it). So any memo tx plans exactly one action short (e.g. 15,000 zats planned vs 20,000 required). The TS SDK fixed this in vultisig/vultisig-sdk#773; iOS never got the equivalent.

Traced iOS path: Maya-native ZEC swaps ride `SwapPayload.mayachain` + native token → `KeysignMessageFactory` falls through to the **plain UTXO send path** (`UTXOChainsHelper.getBitcoinPreSigningInputData`) with the swap memo in `outputOpReturn` — that path (and the fee-preview/UTXO-selection path) planned once with `zip0317 = true` and no conventional-fee check.

## Fix — byte-parity port of the SDK guard

1. `ZcashConventionalFee` (new): ZIP-317 conventional-fee math ported from the SDK's canonical `packages/core/chain/chains/utxo/fee/zip317.ts` — 5,000 zats per logical action, two-action grace, logical actions = max(ceil(tx_in bytes/150), ceil(tx_out bytes/34)), OP_RETURN sized byte-accurately with PUSHDATA + CompactSize thresholds.
2. `UTXOChainsHelper.planZcashConventionalFee` (ported from `planZcashConventionalFee` in the SDK's `packages/core/mpc/keysign/signingInputs/resolvers/utxo.ts`): plan with `zip0317` as today; if the planned fee underpays the conventional fee for the planned shape, re-plan with `zip0317 = false` (where WalletCore honours `byteFee`), bumping `byteFee` from 1 using the planner's byteFee-independent vsize, max 5 bumps (same cap as the SDK), then throw. Plain no-memo sends keep the `zip0317` plan; empty (insufficient-funds) plans pass through untouched.
3. All four `AnySigner.plan` call sites route through one choke point, so the presign path, the THORChain-style swap path, fee preview / UTXO selection (`getBitcoinTransactionPlan` — used by `KeysignPayloadFactory`, `DefaultSendInteractor`, `JoinKeysignGasViewModel`, `SwapPayloadBuilder`), and the Blockaid placeholder all agree. Non-Zcash coins take the early exit — zero behaviour change.
4. `SwapKitZcashSigner` deliberately untouched (frozen plan from the SwapKit PSBT).

## Byte-parity argument (why co-signs converge)

The MPC preimage digests derive from the WalletCore *plan* (selected UTXOs, amount, fee, change). Both platforms pin **WalletCore 4.7.0** (iOS `walletcore-spm` exactVersion 4.7.0; SDK `@trustwallet/wallet-core` 4.7.0) and identical dust floors (1,000 zats), and the loop is an exact port (initial byteFee 1, same bump formula, same cap, same empty-plan pass-through). Pinned with golden vectors **extracted from a live run of the SDK resolver** (`getUtxoSigningInputs` on real wallet-core WASM 4.7.0, from `utxo.zcashFee.test.ts` scenarios):

| scenario (byteFee 100, 1 UTXO) | fee | change | final byteFee | zip0317 |
|---|---|---|---|---|
| plain send 5,000,000 / 8,300,000 | 10,000 | 3,290,000 | 100 (untouched) | true |
| memo `m`×40 | 20,020 | 3,279,980 | 77 | false |
| memo `m`×40, sendMax 2,200,000 | 15,142 | 0 (amount 2,184,858) | 67 | false |
| memo `m`×200 | 45,240 | 3,254,760 | 174 | false |
| Maya-shaped 109-char memo | 30,160 | 3,269,840 | 116 | false |
| non-ASCII memo (34 UTF-8 bytes) | 20,020 | 3,279,980 | 77 | false |
| insufficient 90,000 / 100,000 + memo | 0 (empty plan pass-through) | 0 | 100 (untouched) | true |

iOS reproduces every value exactly (`ZcashZip317PlanTests`, real WalletCore). Math vectors mirror the SDK's `zip317.test.ts` one-for-one plus upper PUSHDATA2→4 / CompactSize 3→5 boundaries (`ZcashConventionalFeeTests`).

## Gate

- `make test`: **2282 tests, 1 failure** — the failure is the known pre-existing decimal-comma locale flake (`StakingValidatorProjectionTests.testCosmosEmptyMonikerFallsBackToTruncatedAddress`, `("12,3%") != ("12.3%")`), unrelated to this change. All 26 new Zcash tests pass.
- SwiftLint (`VultisigApp/.swiftlint.yml` over `VultisigApp/`): zero warnings in touched files (43 pre-existing repo-wide, none in this diff).
- Cross-model review: per-commit + whole-branch Codex passes — no findings on the final branch (boundary vectors and UTF-8 byte counts independently recalculated).

## ⚠️ HIGH-tier signing path — review requirements

This changes fee planning inside the MPC signing path. Divergence between co-signing devices moves the keysign failure instead of fixing it. Before merge this needs:
- **Human review** of the parity argument above.
- **Cross-device validation on a real Maya ZEC swap**: initiate on the extension (0.2.22+, carries sdk#773), co-sign on an iOS build with this patch — keysign must converge and the broadcast must be accepted; plus a mobile-only ZEC send with memo.

## Follow-ups (out of scope here)

- Add a ZEC-memo vector to the cross-platform signing-vector net (vultisig/commondata#94).
- Upstream the OP_RETURN action-count correction to trustwallet/wallet-core so all platforms converge without workarounds.
- Android sibling (`UtxoHelper.kt`) tracked on vultisig-android.

## Summary by CodeRabbit

* **New Features**
  * Added Zcash-specific conventional-fee replanning to ensure memo/payload-based transactions are planned correctly.
* **Bug Fixes**
  * Prevented Zcash from producing fees below the required conventional minimum.
  * Made pause/halt detection more resilient when pause flags are missing, treating absent values as unpaused.
* **Tests**
  * Added coverage for Zcash conventional-fee calculation, memo sizing, ZIP-317 planning, and swap-style payload behavior.
  * Expanded halt/pause gate tests for nil pause-flag decoding.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapHaltGate.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapHaltGate.swift
@@ -20,6 +20,6 @@ enum SwapHaltGate {
         guard let entry = inbound.first(where: { $0.chain.caseInsensitiveCompare(chainName) == .orderedSame }) else {
             return false
         }
-        return entry.halted || entry.global_trading_paused || entry.chain_trading_paused
+        return entry.halted || (entry.global_trading_paused ?? false) || (entry.chain_trading_paused ?? false)
     }
 }
--- a/VultisigApp/VultisigApp/Blockchain/THORChain/Service/InboundAddress.swift
+++ b/VultisigApp/VultisigApp/Blockchain/THORChain/Service/InboundAddress.swift
@@ -28,9 +28,9 @@ struct InboundAddress: Codable {
     let address: String
     let router: String? // Router address for ERC20 tokens
     let halted: Bool
-    let global_trading_paused: Bool
-    let chain_trading_paused: Bool
-    let chain_lp_actions_paused: Bool
+    let global_trading_paused: Bool?
+    let chain_trading_paused: Bool?
+    let chain_lp_actions_paused: Bool?
     let gas_rate: String
     let gas_rate_units: String
     let dust_threshold: String? // Dust threshold for the chain
--- a/VultisigApp/VultisigApp/Blockchain/UTXO/Signing/UTXOChainsHelper.swift
+++ b/VultisigApp/VultisigApp/Blockchain/UTXO/Signing/UTXOChainsHelper.swift
@@ -47,6 +47,77 @@ class UTXOChainsHelper {
         plan.branchID = try zcashBranchID(keysignPayload: keysignPayload)
     }
 
+    /// Max non-ZIP-317 re-plans before giving up raising the Zcash fee.
+    /// Matches the SDK's cap so every co-signing device fails identically
+    /// instead of diverging on the plan.
+    private static let maxZcashFeeBumps = 5
+
+    /// Plan the transaction, guarding Zcash plans against WalletCore's
+    /// underpaying ZIP-317 mode. Every other coin plans directly.
+    private func planTransaction(_ input: inout BitcoinSigningInput) throws -> BitcoinTransactionPlan {
+        guard coin == .zcash else {
+            return AnySigner.plan(input: input, coin: coin)
+        }
+        return try planZcashConventionalFee(&input)
+    }
+
+    /// Produce a Zcash plan whose fee meets the ZIP-317 conventional fee.
+    ///
+    /// WalletCore's `zip0317` planner sizes an OP_RETURN output as a flat ~34
+    /// bytes and ignores `byteFee`, so memo transactions (MayaChain-routed
+    /// swaps carry the swap instruction in an OP_RETURN; sends with memo too)
+    /// plan one logical action short — e.g. 15,000 zats where the network
+    /// requires 20,000 — with no way to raise the fee in that mode. When the
+    /// `zip0317` plan underpays the byte-accurate conventional fee, re-plan
+    /// with `zip0317` off — where WalletCore honours `byteFee` — and bump
+    /// `byteFee` until the fee clears. Plain (no-memo) sends already meet the
+    /// fee and keep the `zip0317` plan.
+    ///
+    /// An empty plan (no selected UTXOs) is returned untouched: coin selection
+    /// produced nothing (insufficient funds), and that flow owns the outcome.
+    ///
+    /// Byte-parity port of the SDK's `planZcashConventionalFee` — every
+    /// co-signing device must derive the same plan or the MPC preimage
+    /// digests diverge and keysign fails. Keep in lockstep with the SDK.
+    private func planZcashConventionalFee(_ input: inout BitcoinSigningInput) throws -> BitcoinTransactionPlan {
+        let memoSize = input.outputOpReturn.count
+
+        func conventionalFee(for plan: BitcoinTransactionPlan) -> Int64 {
+            ZcashConventionalFee.conventionalFee(
+                inputCount: plan.utxos.count,
+                outputSizes: ZcashConventionalFee.transparentOutputSizes(change: plan.change, memoSize: memoSize)
+            )
+        }
+
+        // An empty plan has no shape to charge for; leave the fee flow to the caller.
+        func meetsConventionalFee(_ plan: BitcoinTransactionPlan) -> Bool {
+            plan.utxos.isEmpty || plan.fee >= conventionalFee(for: plan)
+        }
+
+        let zipPlan: BitcoinTransactionPlan = AnySigner.plan(input: input, coin: coin)
+        if meetsConventionalFee(zipPlan) {
+            return zipPlan
+        }
+
+        input.zip0317 = false
+        var byteFee: Int64 = 1
+        var plan = zipPlan
+        for _ in 0..<Self.maxZcashFeeBumps {
+            input.byteFee = byteFee
+            plan = AnySigner.plan(input: input, coin: coin)
+            if meetsConventionalFee(plan) {
+                return plan
+            }
+
+            // byteFee mode scales fee linearly with vsize; derive the byteFee that
+            // clears the conventional fee for this plan's (byteFee-independent) vsize.
+            let plannerVsize = plan.fee > 0 ? plan.fee / byteFee : 1
+            byteFee = max(ZcashConventionalFee.ceilDiv(conventionalFee(for: plan), plannerVsize), byteFee + 1)
+        }
+
+        throw HelperError.runtimeError("Failed to meet the Zcash minimum network fee (ZIP-317): planned \(plan.fee) zats, required \(conventionalFee(for: plan))")
+    }
+
     static func getHelper(coin: Coin) -> UTXOChainsHelper? {
         switch coin.chainType {
         case .UTXO:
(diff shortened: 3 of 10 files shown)


gastonm5:
## On-device mainnet validation ✅ (2026-07-03)

Three transactions run from builds carrying this fix, all accepted on mainnet — covering the broken path (memo/OP_RETURN), the regression path (no memo), and the memo-with-change shape:

| tx | shape | fee | proves |
|---|---|---|---|
| [`ea1b22e8…`](https://blockchair.com/zcash/transaction/ea1b22e811eb25a2a06f473531d69e65168f06b9c9e97be730d13f9edfc5b8fc) | 1-in / recipient + **OP_RETURN Maya memo** (`=:c:maya1…`), no change | **20,114** zats | the exact previously-broken path: memo tx re-planned above the ZIP-317 conventional minimum (raw `zip0317` planning underpaid this class); accepted by the network |
| [`d0a611b7…`](https://blockchair.com/zcash/transaction/d0a611b7585ccec6d8e0fc2ad4c20804a52dfc970ccd812f42136377d0168f79) | 1-in / recipient + change + **OP_RETURN Maya swap memo** (`=:c:maya1…:139956209441/3/0:v0:50`) | **25,220** zats | memo + change output shape — fee scales with the larger output side (long memo), still clears conventional and confirms |
| [`87b28802…`](https://blockchair.com/zcash/transaction/87b2880283cef858eceefbfd8d3eab1855623fda8b0a1c63542437857ec2c492) | 4-in / 1-out, **no memo** | **20,000** zats | regression check: the untouched plain-`zip0317` path still plans exactly the conventional fee (5,000 × 4 actions) and lands |

The two memo fees sit just above their conventional minimums, which is the expected signature of the `planZcashConventionalFee` byteFee-bump loop (it raises `byteFee` until the plan's fee clears the conventional fee, so the result can overshoot by a granularity step) — matching the SDK's behavior on the same algorithm.

gastonm5:
## Scope addition — Maya inbound decode fix (`7792c5a53`)

Testing the Maya-route ZEC swap surfaced a second, independent blocker: **every Maya-route swap on iOS was failing before quoting**, because `InboundAddress` required THORChain's three pause-flag keys (`global_trading_paused`, `chain_trading_paused`, `chain_lp_actions_paused`) and MayaChain's `/mayachain/inbound_addresses` does not serve them (verified against live mayanode — it returns only `halted` + address/gas/fee fields). `JSONDecoder` threw `keyNotFound`, the inbound fetch failed, and no Maya swap could be built. The mainnet proof txs above were signed with this fix in place — without it the swap flow never reaches planning.

Fix (user-authored, tests added here): the three pause flags become optional and read as not-paused when absent, at all four use sites (`SwapHaltGate`, `FunctionCallAddThorLP`, `FunctionCallCosmosSwitch`, `FunctionCallSecuredAsset`). `halted` — the only halt signal Maya exposes — is still honored, and THORChain behavior is unchanged (present flags still block). Regression tests pin the exact Maya-shaped payload decode and the nil-flag gate semantics.

Codex read-only review: no findings; it confirmed no other use sites exist. One accepted trade-off it noted: if THORChain ever *omitted* these fields, the gate would read them as not-paused rather than fail closed — `halted` remains authoritative in that case.

Gate re-run on the full suite: 2,280 passed, sole failure = the known decimal-comma locale flake (pre-existing baseline).

johnnyluo (approved):
LGTM

PR #4734, opened 2026-07-03
fix(sui): select only the coin objects a send needs (fixes tx-size-limit + gas failures on scattered balances)

DanialV:
## Problem

On a wallet whose SUI balance is spread across many coin objects, sends fail — because the code referenced/embedded **every** owned object instead of only the ones the send needs. Three symptoms, one root cause:

1. **Keysign can't even start:** the co-signer shows `HTTP 404` (it polls the relay but the oversized payload never landed) and the initiator shows `Transaction data expired`. The keysign payload embedded every object, making the pairing QR / TSS relay message too large.
2. **Broadcast rejected — transaction too large:**
   > Sui broadcast failed: … **Size limit exceeded: serialized transaction size exceeded maximum of 131072 is 303638**

   A native `PaySui` send passes its whole input set as the gas payment; thousands of objects overflow Sui's **128 KiB transaction-size** limit and **256 gas-payment-object** cap ([Sui docs](https://docs.sui.io/concepts/transactions), [Move Book — limits](https://move-book.com/guides/building-against-limits.html)).
3. **`InsufficientCoinBalance in command 0` / token gas failures:** same over-inclusion, plus token sends picking an arbitrary SUI object for gas ([Sui issue #14696](https://github.com/MystenLabs/sui/issues/14696)).

## Fix — select only the coin objects a send needs

Reference/embed only the fewest **largest** objects that cover the send, at both layers:

- **Transaction** (`SuiHelper.getPreSignedInputData`):
  - native (`PaySui`): objects covering `amount + gasBudget` — Sui gas-smashes them into one `GasCoin`, so a scattered balance is still **merged**;
  - token (`Pay`): token objects covering `amount`, plus a **single SUI gas object chosen to cover `gasBudget`** (smallest covering object, matching Android) instead of `suiObjectForGas.first`.
- **Keysign payload** (`BlockChainService` Sui branch, via `SuiCoinType.selectPayloadCoins`): embed only the covering objects (native) / covering token objects + the largest few SUI gas candidates (token). The dry-run simulates over the same covering subset, so it also stays small on dusty wallets.

Selection is **deterministic** (balance desc, then `objectID`) so every co-signing device — which signs the received payload verbatim — builds the identical transaction. Bounded by `SuiConstants.maxInputCoinObjects` (255 — one under Sui's 256 gas-payment cap). The repeated `SuiObjectRef` mapping is factored into `objectRef(from:)`.

### Commits
1. token-send gas object = smallest SUI object covering the budget (Android parity);
2. transaction-level coin selection (fixes the 128 KiB broadcast failure);
3. payload-level coin selection (fixes the relay 404 / "data expired" so keysign can start).

## Tests

39 Sui unit tests pass locally; SwiftLint clean.
- `selectInputCoins` / `selectGasObject` / `selectPayloadCoins` unit-covered (fewest-covering, single-covers, cap, deterministic tie-break, gas-candidate bound, dusty-wallet bound).
- `SuiHelperInputDataTests` decodes the built `Sui.SigningInput`: native small send → one object, near-max → all, **800-object dusty wallet → only the one large object it needs**; token send → covering token objects + smallest covering gas object.

## Known limitation

A balance split across **more than 255** objects that a single send would need to *fully* cover can't be sent in one Sui transaction (protocol limit) and would require a coin-consolidation step (not yet in the app). Worth a follow-up "merge coins" action.

on chain tx link: https://suiscan.xyz/mainnet/tx/Wm4etgsWPRRAWaUp4uuTD2ctVBcS4pwaSst92FNjDyH

## Summary by CodeRabbit

* **New Features**
  * Improved Sui transaction building to include only the coins needed for a send, helping keep payloads smaller and more deterministic.
  * Added smarter selection of gas coins for token transfers, choosing an appropriate native SUI coin when available.

* **Bug Fixes**
  * Fixed cases where coin selection could include too many objects or pick an unsuitable gas coin.
  * Improved handling for wallets with many small coins and for missing or malformed balance values.

## Cross-platform parity

- Android: [vultisig/vultisig-android#5385](https://github.com/vultisig/vultisig-android/pull/5385) — merged
- SDK / Windows: [vultisig/vultisig-sdk#1275](https://github.com/vultisig/vultisig-sdk/pull/1275) — merged

The deterministic coin-selection, gas-object, fail-fast, and coin-type normalization rules now match across iOS, Android, and SDK/Windows.


VultisigApp/VultisigApp/Blockchain/Sui/Signing/Sui.swift
```diff
@@ -46,22 +46,19 @@ enum SuiHelper {
...
                 throw HelperError.runtimeError("Native token transaction requires at least one SUI coin")
             }
 
+            // Reference only the largest objects needed to fund amount + gas.
+            // PaySui uses its whole input set as the gas payment, which Sui
+            // gas-smashes into one coin — so a scattered balance is still merged,
+            // but the transaction stays within Sui's 128 KiB size / 256-gas-object
+            // limits instead of referencing every object and failing at broadcast.
+            let target = keysignPayload.toAmount + gasBudget
+            let suiCoins = SuiCoinType.selectInputCoins(nativeCoins, covering: target).map(objectRef(from:))
```
realpaaao: **🔴 Blocker — cross-device / cross-version keysign divergence.**

Each co-signer independently recomputes this transaction (and its sighash) from the shared payload, and all devices must produce byte-identical bytes — see `JoinKeysignViewModel.swift:252-253`, and the Zcash committee-matching note at `:244-249` that exists for exactly this reason.

This PR now emits a **balance-sorted covering subset**. Every other implementation references **all** payload objects in RPC order: Android `SuiHelper.kt:50-95` + `SuiFeeService.kt:106-108`, and the SDK/Windows `signingInputs/resolvers/sui.ts:60,80`. So when a non-updated / non-iOS device initiates a send from a multi-object wallet (the common case), an updated iOS co-signer trims to a different set/order → different bytes → **keysign fails** (liveness, not fund-loss). Even a near-max send diverges on ordering.

Before this PR iOS and the others agreed. Fix requires the identical algorithm on all platforms shipped together (or a version gate) + a cross-platform byte-equality interop test. Parity: https://github.com/vultisig/vultisig-android/issues/5252 , https://github.com/vultisig/vultisig-sdk/issues/1132

VultisigApp/VultisigApp/Blockchain/Sui/Signing/SuiConstants.swift
```diff
@@ -92,6 +108,118 @@ enum SuiCoinType {
...
+    /// the selected objects into one spendable coin. `target` is the send amount
+    /// plus, for a native send, the gas budget (the native input set also pays
+    /// gas); for a token send it is just the token amount.
+    ///
+    /// Selection is deterministic (balance descending, then `objectID`) so every
+    /// co-signing device selects the identical set and signs the identical
+    /// transaction. If even `maxObjects` largest objects don't reach `target`,
+    /// they are still returned (best effort) — the caller decides how to handle
+    /// an under-funded selection.
+    static func selectInputCoins(
```
realpaaao, line 190: **Under-funded selection builds a doomed tx.** When even 255 largest objects don't reach `target` (the >255-object known limitation), this returns an under-covering set — and no caller checks. `getPreSignedInputData` builds and signs it, so it fails later at broadcast with a cryptic node error (or a native send that can't cover amount+gas). Prefer throwing a clear, localized error when `accumulated < target`, instead of shipping a tx that can't succeed.

VultisigApp/VultisigApp/Blockchain/Sui/Signing/SuiConstants.swift
```diff
@@ -92,6 +108,118 @@ enum SuiCoinType {
...
+    /// whose balance is spread across thousands of objects that produces a
+    /// keysign payload too large to relay — the co-signer's poll 404s and the
+    /// initiator's transaction data expires before signing can start. Bounding
+    /// the embedded set to the objects the send actually needs keeps the pairing
+    /// QR / TSS relay message small.
+    ///
+    /// Native send: the largest objects covering `amount + gasBudget` (the input
+    /// set also pays gas). Token send: the largest token objects covering
+    /// `amount`, plus the largest few native SUI objects as gas candidates.
+    static func selectPayloadCoins(
```
realpaaao, line 109: `selectPayloadCoins` replaces `payloadCoins`, which is now unused in production — only 3 tests still reference it (`SuiCoinTypeTests.swift:113,124,144`). Delete `payloadCoins` and its now-obsolete tests, or wire it back if the split was intentional. (Dead-code removal is the repo standard.)

VultisigApp/VultisigApp/Blockchain/Sui/Signing/SuiConstants.swift
```diff
@@ -92,6 +108,118 @@ enum SuiCoinType {
...
+    /// object at all.
+    static func selectGasObject(_ coins: [[String: String]], gasBudget: BigInt) -> [String: String]? {
+        let suiObjects = coins.filter { isNative($0["coinType"] ?? .empty) }
+        guard !suiObjects.isEmpty else { return nil }
+
+        let covering = suiObjects.filter { balance(of: $0) >= gasBudget }
+        if let smallestCovering = covering.min(by: { balance(of: $0) < balance(of: $1) }) {
+            return smallestCovering
+        }
+        return suiObjects.max(by: { balance(of: $0) < balance(of: $1) })
```
realpaaao: **Parity nit:** this largest-object fallback diverges from Android's `selectSuiGasCoin`, which returns `nil` / fails fast when no object covers the budget (`SuiHelper.kt:151-153`). Building a token send that can't pay gas just defers the failure to broadcast, and it's another cross-platform digest divergence for token sends. Prefer matching Android — return `nil`/throw when nothing covers — and pin the chosen behavior across all platforms.

VultisigApp/VultisigApp/Core/Services/BlockChainService.swift
```diff
@@ -673,27 +684,21 @@ private extension BlockChainService {
...
                     let totalCost = computationCost + storageCost
-                    gasBudget = (totalCost * 115) / 100
-
-                    // Ensure minimum gas budget of 2000 (network requirement)
-                    let finalGasBudget = max(gasBudget, BigInt(2000))
-
-                    return .Sui(referenceGasPrice: referenceGasPrice, coins: allCoins, gasBudget: finalGasBudget)
+                    gasBudget = max((totalCost * 115) / 100, BigInt(2000))
                 } catch {
                     print("⚠️ Sui dry run failed, using default gas budget: \(error.localizedDescription)")
```
realpaaao: Use the existing `logger` (`BlockChainService.swift:20`) instead of `print()`, matching the file convention (also flagged by CodeRabbit). Pre-existing, but cheap to fix while adjacent.

realpaaao (dismissed):
Thorough review of the Sui coin-selection change (reviewed against PR head `6834d99`, cross-checked with Android and the SDK/`@core`). The selection math, the `objectID` tie-break, and the test coverage are genuinely solid work.

**But this must not merge as an iOS-only change.** It alters how iOS deterministically builds a Sui transaction from a shared keysign payload, and that construction has to be byte-identical on **every** co-signing device (Android / Windows / SDK / older iOS) or the MPC signatures don't combine and the send fails. It currently isn't — see the blocker inline on `Sui.swift`.

**Parity issues filed (please link from this PR):**
- Android — https://github.com/vultisig/vultisig-android/issues/5252
- SDK, covers Windows via `@core` — https://github.com/vultisig/vultisig-sdk/issues/1132

**⚠️ Please keep this PR open — do not merge to a public iOS release — until the *identical* selection algorithm ships on Android + SDK/Windows (or all platforms gate the new behavior behind a version/capability flag).** Until then, a Sui send initiated on a non-updated or non-iOS device with an updated iOS co-signer will diverge on the sighash and fail to keysign. A cross-platform interop test (same coin fixture → identical `Sui.SigningInput` bytes on iOS + Android) should gate the coordinated rollout.

Non-blocking findings (under-funded-selection guard, dead `payloadCoins`, gas-fallback parity, `print()`→`Logger`) are inline. Note: CodeRabbit's gas-candidate tie-break comment is a false positive for consensus — `selectPayloadCoins` runs once on the initiator and its output is shared, so it can't diverge across the committee.

johnnyluo (approved):
LGTM

PR #4735, opened 2026-07-03
fix(keysign): pad EdDSA r/s to 32 bytes before assembling the signature

DanialV:
## Problem

`TssKeysignResponse.getSignature()` assembles an Ed25519 signature by reversing `r` and `s` (big-endian scalars from TSS) into little-endian and concatenating them into `R || S` — but it never left-padded each half to 32 bytes.

tss-lib emits `r`/`s` via Go's `hex.EncodeToString(bigInt.Bytes())`, and `big.Int.Bytes()` **strips leading zero bytes**. So a scalar with a high-order zero byte — an `S` value < 2²⁴⁸, which happens for a meaningful fraction of signatures — arrives as **fewer than 64 hex chars**. Reversing that short value produces a truncated half and an assembled signature **under 64 bytes**, which fails `PublicKey.verify`.

This surfaces as an **intermittent "signature verification failed"** during the signing phase on EdDSA chains (Sui, Solana, Polkadot, TON…). It's value- and vault-dependent:
- The **DKLS/Schnorr** path is unaffected — it hex-encodes fixed 32-byte arrays (leading `00`s preserved).
- The **GG20** path strips zeros, so it hits the bug — consistent with the intermittent, vault-dependent reports.

## Fix

Left-pad each big-endian half to a full 32 bytes **before** reversing. This is a **no-op for well-formed 64-char components** (the common case), so already-correct signatures are byte-for-byte unchanged; only the previously-broken short case is fixed. Odd-length hex is left-padded a nibble so it parses instead of being rejected, and an over-long (> 32-byte) component is still rejected as corruption (fail-closed).

## Tests

`TssSignatureTests`:
- Byte-level padding: short half padded before reversal, full-width no-op, odd-length parsed, over-long rejected.
- **Real-signature round-trip**: reformats 40 real Ed25519 signatures into the exact tss-lib wire form (big-endian halves, leading zeros stripped) and asserts each rebuilds bit-for-bit through `getSignature()` **and verifies** against the key — exercising the padding path end-to-end.

`CustomMessageSignatureTests`: updated the two EdDSA cases that had pinned the old *un-padded* output using toy 2-byte values (a real EdDSA signature is always 64 bytes), plus a new short-component padding assertion.

All affected suites pass locally (`TssSignatureTests`, `CustomMessageSignatureTests`, `SolanaHelperTests`, `SuiSignSuiTests`); SwiftLint clean.

> ⚠️ Touches shared TSS signature assembly used by **all** EdDSA chains — flagged for maintainer/crypto review per the repo's HIGH security tier.


johnnyluo (approved):
LGTM

PR #4745, opened 2026-07-06
fix(solana): staking DeFi fixes — validator 520 fallback, position-selection gating, cache-first seed heal

gastonm5:
Three user-reported Solana staking issues on the DeFi tab, each root-caused with endpoint-level and simulator evidence. Issue 1 turned out to be server-side and was fixed at the proxy origin; issues 2 and 3 are fixed here. None of the changes touches signing/keysign byte construction — everything is presentation or display-persistence.

## 1. Validators screen: "Validator fetch failed: HTTP status code: 520" — resolved server-side, no client change

**Root cause.** The picker reads `getVoteAccounts` from the Vultisig proxy (`api.vultisig.com/solana/`), whose origin persistently returned a Cloudflare 520 for this one method — verified with curl: 3/3 failures with body `error code: 520`, unaffected by browser-like headers, while `getEpochInfo` / `getInflationRate` / `getProgramAccounts` on the same proxy returned 200. With no last-known list, the picker rendered a dead error screen, and the stake rows lost their validator names (metadata enrichment is keyed onto the empty validator list).

**Resolution.** The proxy origin was fixed server-side by @johnnyluo on 2026-07-06 — re-verified 3/3 × HTTP 200 (~310 KB, ~1.6 s) through `api.vultisig.com/solana/`. This PR originally carried a client-side public-node fallback + stale-cache serve for this path; per the discussion below, that commit was dropped to keep the client proxy-only and lean, so the validator fetch is exactly as on `main`.

## 2. Solana staking card ignored the DeFi position selection

**Root cause.** `DefiChainMainScreen.solanaStakeView` rendered the card unconditionally, and the Solana balance/count roll-ups in `DefiBalanceService` were deliberately ungated ("like Tron"). A vault that never selected the SOL staking position under `DefiChainSelectPositionsScreen` still saw the full card (reproduced in the simulator against a vault with staked SOL and no `DefiPositions` row for Solana).

**Fix** — mirror the LUNA/LUNC (Cosmos) opt-in pattern end to end:
- `solanaStakeView` computes `isPositionEnabled` from `vault.defiPositions[chain].staking` (identical to `cosmosStakeView`) and passes the shared manage-positions empty state through.
- `SolanaStakeDefiView` is generic over the empty state and renders it until the position is opted in (exactly `CosmosStakeDefiView`'s shape).
- `solanaStakingTotalBalanceFiatDecimal` / `solanaStakingPositionCount` gate on the same opt-in, so an un-opted-in vault's delegated SOL no longer leaks into the DeFi total or the chain cell badge.
- Enabling SOL no longer inserts the coin-keyed zero-amount placeholder row (Solana rows are per-stake-account and on-chain-discovered; the placeholder would alias every account onto one coin-keyed id).

**Runtime verification:** un-opted-in vault now sees the "No positions selected" banner; selecting the position via Manage Positions brings the card up with the stake accounts.

## 3. Stake segment showed a loading state on every entry

**Root cause.** The cache-first seed (`seedFromPersistedSnapshot`) never painted: stores written by an earlier build carry the stake-account pubkey in the persisted row's `id` suffix but `NULL` in the `stakeAccountPubkey` column, and the seed drops rows without a pubkey. Because `upsert(solanaStake:)` matches rows by `id` and `StakePosition.apply()` deliberately skipped the field, successful refreshes updated the row in place forever without healing it — verified live: after three successful refreshes on main, the column was still NULL (sqlite inspection), and every screen entry spun for 1.5–5.5 s.

**Fix:**
- `StakePosition.apply()` backfills a nil/empty `stakeAccountPubkey` from the DTO. The upsert matches by `id` and the `id` embeds the pubkey, so the DTO value is by construction the same suffix — an existing non-nil field is never rewritten. One successful refresh heals the store; every subsequent entry paints instantly from the seed and refreshes in the background.
- A genuine first-ever load (nothing to seed) now shows the `DefiChainStakedPositionSkeletonView` placeholder used by the persisted THOR/Maya/TON stake path, not a bare spinner.

**Runtime verification:** after one refresh on the fixed build, the store's pubkey column is backfilled and re-entering the screen paints the stake accounts instantly with no spinner.

## Tests
- Balance/count zero without the opt-in even with delegated SOL; existing Solana balance tests opt in first.
- Storage-level pubkey backfill on a legacy row; VM-level heal-then-seed (first entry can't seed → refresh heals → fresh VM paints).

## Notes
- Pre-existing, unrelated: the `VultisigAppUITests` target does not compile on `main` (`AccessibilityID` isn't in its sources and CI never runs `ui_test`). Not addressed here.

## Summary by CodeRabbit

* **New Features**
  * Solana staking views now respect vault opt-in settings, showing staking details only when enabled.
  * Validator and stake data can now recover more reliably from network issues by trying fallback RPC endpoints.

* **Bug Fixes**
  * Improved Solana staking resilience when the primary RPC endpoint fails.
  * Restored missing stake-account identifiers for older staking records, helping cached data load correctly.
  * Stale cached validator data can now still be shown if refresh attempts fail.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/Common/Service/DefiBalanceService.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/Common/Service/DefiBalanceService.swift
@@ -155,12 +155,19 @@ private extension DefiBalanceService {
     /// stake accounts. `BalanceService.fetchStakedBalance(.solana)` sums the
     /// delegated lamports of every stake account and writes the total (base
     /// units) to `Coin.stakedBalance`; `stakedBalanceDecimal` divides by 10^9
-    /// to get SOL. Solana has no per-coin opt-in (a real stake is always the
-    /// vault's SOL DeFi position), so — like Tron — we don't gate on
-    /// `defiPositions`. Returning the persisted value keeps the balance stable
-    /// through a transient stake-account read failure (which writes nothing).
+    /// to get SOL. Gated on the per-vault opt-in
+    /// (`defiPositions[.solana].staking`) so a vault that hasn't selected the
+    /// staking position doesn't have its delegated balance silently rolled into
+    /// the DeFi total — matches the empty-state semantic in
+    /// `SolanaStakeDefiView` and mirrors the Cosmos staking roll-up. Returning
+    /// the persisted value keeps the balance stable through a transient
+    /// stake-account read failure (which writes nothing).
     func solanaStakingTotalBalanceFiatDecimal(for vault: Vault) -> Decimal {
-        guard let solCoin = vault.nativeCoin(for: .solana) else { return .zero }
+        guard
+            let solCoin = vault.nativeCoin(for: .solana),
+            let enabledPositions = vault.defiPositions.first(where: { $0.chain == .solana }),
+            enabledPositions.staking.contains(solCoin.toCoinMeta())
+        else { return .zero }
         return solCoin.fiat(decimal: solCoin.stakedBalanceDecimal)
     }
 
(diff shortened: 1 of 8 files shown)


johnnyluo:
@gastonm5 , I fixed the getVoteAccount issue on server side , it should work now, double check

gastonm5:
Verified — `getVoteAccounts` through the proxy now returns 200 consistently (3/3, ~310 KB, ~1.6 s). Thanks for the quick server-side fix @johnnyluo!

**Update:** since the origin is fixed, we're dropping the client-side fallback commit entirely to keep the client strictly proxy-only. The PR is being rebased down to the gating + caching fixes and will be re-gated and updated shortly.

gastonm5:
Dropped the client-side fallback commit as discussed above — with the proxy origin fixed server-side (re-verified 3/3 × HTTP 200, ~310 KB through `api.vultisig.com/solana/`), the client stays proxy-only and the validator fetch path is exactly as on `main` again (including `SolanaAPI.minDelegationPublicHosts`, which is back to serving only the min-delegation read from #4705).

The branch is now 2 commits: the position-selection gating fix and the stake-row backfill + first-load skeleton fix, both unchanged. Full gate re-run after the rebase: 2368 tests, 0 failures; SwiftLint 0 violations on the 8 touched files.

johnnyluo (approved):
LGTM

PR #4746, opened 2026-07-06
fix(ripple): gate broadcast on tesSUCCESS + verify-by-hash dedup for already-submitted results

### LINKED ISSUE #4743: fix(ripple): gate broadcast on tesSUCCESS + tefALREADY dedup (port SDK result-code handling)
## What
XRP broadcast result handling on iOS is weaker than the SDK/Windows resolver: it reads `tx_json.hash` but **does not gate on `tesSUCCESS` at submit**, and has **no `tefALREADY`/already-submitted dedup** (`RippleService.swift:83`). The SDK correctly classifies tec/tef/tem codes and verifies-by-hash.

## Fix
Port the SDK's ripple broadcast classification: gate on engine result (`tesSUCCESS`), on `tefALREADY`/`tefPAST_SEQ` verify the tx by hash instead of treating as failure, surface the real rejection reason otherwise, and only ever set a real tx hash.

## Acceptance
- [ ] Submit gated on tesSUCCESS; non-success maps to a real error, not silent success
- [ ] tefALREADY / already-applied → verify-by-hash, no double-broadcast
- [ ] Real result code surfaced to the friendly error screen
- [ ] Unit test on the code-classification paths

Ref: XRPL reliable submission https://xrpl.org/docs/concepts/transactions/reliable-transaction-submission · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gastonm5:
## Root cause

`RippleService.broadcastTransaction` gated success on **hash presence only**: any submit response carrying `result.tx_json.hash` was returned as a successful broadcast. But the XRPL `submit` method echoes `tx_json.hash` — the deterministic hash of the exact submitted blob — for **every** engine result, including `temREDUNDANT`, `tefPAST_SEQ`, `telINSUF_FEE_P`, and every other rejection. The code's own comment documented the echo behavior while relying on it as a success signal.

Consequences of the old gate:

- Any `tef*`/`tem*`/`tel*`/`ter*` rejection was reported to the user as a **successful broadcast** with a success screen and explorer link.
- Since a rejected transaction is never applied to any ledger, the persisted hash resolved to `txnNotFound` forever — an untrackable, dead txid ("silent success").
- `engine_result` was decoded but never consulted for the success/failure decision.

## Fix design

Ports the semantics of the SDK broadcast resolver (`vultisig-sdk` `packages/core/chain/tx/broadcast/resolvers/ripple.ts`) into the iOS broadcast layer. Everything lives in `Blockchain/Ripple/Service/` — no signing-path or `KeysignViewModel` changes.

**1. Pure classifier — `RippleSubmitDisposition.classify`** (new file):

| Engine result | Disposition |
|---|---|
| `tesSUCCESS` (or defensively-missing `engine_result`, SDK parity) + echoed hash | `.accepted(hash:)` — return the hash |
| `tesSUCCESS`/missing, but no hash | `.rejected` — never persist an empty txid as a fake success (preserves the existing guard) |
| `tefALREADY`, `tefPAST_SEQ`, `terQUEUED` | `.verifyByHash` — resolve by `tx` lookup |
| everything else (`tem*`/`tec*`/`tel*`/remaining `tef*`/`ter*`) | `.rejected(code:message:)` — throw |

Rejections throw the existing `RippleBroadcastError.broadcastFailed(code:message:)`, which flows through `handleBroadcastError` to the keysign friendly-error screen with the **real engine code and message** rendered — no ViewModel changes needed.

**2. Verify-by-hash resolution — `RippleService.resolveSubmitByHash`**: `tefALREADY`/`tefPAST_SEQ` mean this exact transaction (or its sequence) was already applied or queued — for an MPC wallet that almost always means a **faster co-signing peer landed the same signed blob first**, so it must not surface as a failure. The echoed deterministic hash is looked up with the `tx` method via a new `RippleAPI.tx` endpoint — against the same **custom-RPC-override-aware** node the submit went to (the existing status provider hardcodes `xrplcluster.com`, which would risk cross-node propagation false negatives):

- validated `tesSUCCESS` → return the hash (idempotent success; asserted single submit, no double-broadcast)
- known but unvalidated → return the hash; the existing status poller resolves the final outcome
- validated non-`tes` result → throw with that final on-chain code
- `txnNotFound` / lookup errors → 3 attempts × 2 s backoff (cluster nodes can lag a peer's submit; injectable backoff for tests), then throw the **original** engine code — an unverified duplicate is never converted into success

**terQUEUED deviation from the SDK (deliberate):** the SDK throws on all `ter*`. `terQUEUED`, however, means the server queued the transaction to apply in a **future ledger** — it is in flight and likely to land. Hard-failing would show a false failure screen for a transaction that lands seconds later, and the error screen's "Try Again" restarts the flow: the user would re-sign with a fresh sequence and could **double-send** once the queued original also applies. Routing `terQUEUED` through the same verify-by-hash path returns the trackable hash when the network knows the transaction (poller resolves the outcome) and still falls back to the real engine code if it genuinely can't be found. All other `ter*` codes keep SDK parity and hard-fail.

**Interpretation parity:** the new pure `RippleTxLookupOutcome.interpret` mirrors `RippleTransactionStatusProvider`'s reading of the `tx` response (validated-only finality, known-but-unvalidated = pending, unusable responses carry no evidence).

## Test coverage

35 new tests in `VultisigAppTests/Blockchain/Ripple/`:

- `RippleSubmitDispositionTests` (15) — exhaustive classifier matrix: `tesSUCCESS` ± hash, missing engine result ± hash (defensive default), `tefALREADY`/`tefPAST_SEQ`/`terQUEUED`, `tem`/`tec`/`ter`/`tel`/other `tef` rejections, code+message propagation, empty-hash normalization.
- `RippleTxLookupOutcomeTests` (8) — pure `tx`-lookup interpretation: validated success/failure, validated-without-meta, known-pending, result-level and top-level `txnNotFound`, non-evidence lookup errors, empty response.
- `RippleBroadcastGatingTests` (12) — service-level via stub `HTTPClientProtocol`: hash returned on `tesSUCCESS`, rejection throws with the engine code in `localizedDescription` (the string the error screen renders), dedup resolution for validated-success / known-pending (incl. `terQUEUED`), validated-failure surfacing, `txnNotFound` ×3 retry then original code, missing echoed hash, lookup errors falling through to the original error.

## Gate

- `make test` (full suite): **2400 tests, 0 failures** (iPhone 17 Pro Max, iOS 26.5).
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/`: **0 new warnings** (0 in touched files).
- Cross-model review loop: 2 per-commit Codex reviews + whole-branch pass — all clean, no findings.

Closes #4743

Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

## Summary by CodeRabbit

* **New Features**
  * Improved XRP transaction handling with better follow-up checks after broadcast submissions.
  * Added support for looking up transaction details by hash.

* **Bug Fixes**
  * More reliably returns the correct transaction hash after submission.
  * Better handles duplicate or queued transactions by checking their final on-chain status.
  * Clearer errors are shown when a transaction is rejected or cannot be verified.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleAPI.swift
@@ -13,6 +13,7 @@ import Foundation
 struct RippleAPI: TargetType {
     enum Endpoint {
         case submit(txBlob: String)
+        case tx(hash: String)
         case serverState
         case accountInfo(account: String)
     }
@@ -40,6 +41,14 @@ struct RippleAPI: TargetType {
         case .submit(let txBlob):
             return .requestCodable(
                 RippleRpcRequest(method: "submit", params: [RippleSubmitParams(txBlob: txBlob)]),
+                .jsonEncoding
+            )
+        case .tx(let hash):
+            return .requestCodable(
+                RippleRpcRequest(
+                    method: "tx",
+                    params: [RippleTxParams(transaction: hash, binary: false, apiVersion: 2)]
+                ),
                 .jsonEncoding
             )
         case .serverState:
@@ -73,6 +82,18 @@ struct RippleSubmitParams: Encodable {
 
     enum CodingKeys: String, CodingKey {
         case txBlob = "tx_blob"
+    }
+}
+
+struct RippleTxParams: Encodable {
+    let transaction: String
+    let binary: Bool
+    let apiVersion: Int
+
+    enum CodingKeys: String, CodingKey {
+        case transaction
+        case binary
+        case apiVersion = "api_version"
     }
 }
 
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
@@ -47,12 +47,18 @@ class RippleService {
     /// live (the shared mirror updates without a relaunch).
     private let resolver: RPCEndpointResolving
 
+    /// Backoff between `tx` lookups while resolving a verify-by-hash submit;
+    /// injectable so tests run without delay.
+    private let verifyByHashBackoff: Duration
+
     init(
         resolver: RPCEndpointResolving = CustomRPCStore.shared,
-        httpClient: HTTPClientProtocol = HTTPClient()
+        httpClient: HTTPClientProtocol = HTTPClient(),
+        verifyByHashBackoff: Duration = .seconds(2)
     ) {
         self.resolver = resolver
         self.httpClient = httpClient
+        self.verifyByHashBackoff = verifyByHashBackoff
     }
 
     /// The override-aware XRPL host. Falls back to the default host when no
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4748, opened 2026-07-06
fix(ripple): owner-aware reserve for MAX send — remove the double-counted static 1 XRP

### LINKED ISSUE #4744: fix(ripple): owner-aware reserve for MAX send (match balance calc, not static 1 XRP)
## What
iOS XRP MAX-send uses a **static 1 XRP** existential deposit (`RippleService.swift:104`, `SendCryptoLogic.swift:166`) even though the iOS balance calc is **owner-aware** (base + OwnerCount*reserve_inc). For accounts holding objects (trust lines, offers, tickets) this over/under-reserves and the MAX send can fail on-chain.

## Fix
Use the **same owner-aware reserve** (`reserve_base + OwnerCount * reserve_inc`, live from server_state) for MAX/available-send as the balance calc already uses, then subtract the fee.

## Acceptance
- [ ] MAX-send reserve == balance-calc reserve (owner-aware), not a static 1 XRP
- [ ] MAX send from an account holding N objects succeeds on-chain
- [ ] Unit test with OwnerCount > 0

Companion to the SDK fix vultisig-sdk#959 and android#5148. Ref: XRPL reserves https://xrpl.org/docs/concepts/accounts/reserves · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gastonm5:
## Root cause — the reserve was counted twice

XRP's `coin.rawBalance` is **already the reserve-net available balance**: `RippleService.getBalance` (main @ `VultisigApp/Blockchain/Ripple/Service/RippleService.swift:103-108`) fetches `account_info` + `server_state` and returns `total − (reserve_base + OwnerCount × reserve_inc)` — owner-aware, live values. `BalanceService` routes `.ripple` straight to it, so every consumer (wallet display, send, swap) sees the spendable balance.

The send path then subtracted a **second, static 1 XRP** on top of that reserve-net balance:

- `RippleHelper.defaultExistentialDeposit` (main @ `VultisigApp/Blockchain/Ripple/Signing/Ripple.swift:19`) — hardcoded `1 XRP`.
- `SendCryptoLogic.existentialDeposit(for:)` (main @ `VultisigApp/Features/Send/Logic/SendCryptoLogic.swift:73-74`) returned it for `.ripple`, feeding:
  - MAX-send math `computeMaxAmount` (`SendCryptoLogic.swift:166`): `MAX = rawBalance − fee − 1 XRP` → **exactly 1 XRP of spendable funds stranded** — the UI showed an "available" balance the user could never send.
  - the Verify-screen recompute (`SendCryptoVerifyViewModel.swift:117`): same double-count on refine.
  - the reap guard `canBeReaped` (`SendCryptoLogic.swift:85-94`), hard-wired in Details (`SendDetailsViewModel.swift:650`) and Verify (`SendCryptoVerifyLogic.swift:107`): **valid partial sends** leaving < 1 XRP *spendable* were blocked with `belowExistentialDepositError`, even though the account keeps its full reserve on-chain and XRPL never deletes accounts (the reserve is a floor, not a deletion threshold).

Per the XRPL reserve rules (https://xrpl.org/docs/concepts/accounts/reserves), the reserve floor applies to the Payment **amount** only — the **fee is exempt** and may even take the account below the reserve. So the true max is exactly `available − fee`, and landing exactly at the reserve is valid.

> Note: the issue cites `RippleService.swift:104` as a static-1-XRP site; on current `main` that line is the `?? 1000000` **seed fallback inside the already-correct owner-aware calc** — the line refs drifted. The real static sites are the ones above.

## Fix design (one commit per step, codex-review-loop gated)

1. **`refactor(ripple): extract owner-aware reserve math into RippleReserve`** — pure enum (sibling of `RippleFee`): `reservedDrops(ownerCount:reserveBase:reserveInc:)` / `availableDrops(...)`, with the mainnet values (1 XRP base / 0.2 XRP per owned object, Dec 2024 validator vote) as named last-resort seeds. `getBalance` delegates; zero behavior change. New `RippleReserveTests` pin the formula (incl. the OwnerCount > 0 acceptance case), seed fallbacks, and the zero clamp.
2. **`feat(ripple): live-cache-seed fallback for server_state reserve values`** — reserve values must come live from `server_state`, but a transient outage of that one RPC used to throw away the whole balance refresh. Now: live → cached last-good (24h `TTLCache`, host-scoped key so a custom-RPC change never reuses another endpoint's snapshot) → `RippleReserve` seeds; only an `account_info` failure stays fatal; cancellation propagates. `httpClient` injected for testability (mirrors the Solana min-delegation pattern). `RippleReserveFallbackTests` cover live/warm-cache/cold-cache/fatal/host-scoping/TTL-reuse.
3. **`fix(ripple): stop double-reserving the XRP MAX send`** — **the actual fix.** `existentialDeposit(for: .ripple)` → `.zero` (with a load-bearing rationale comment); `RippleHelper.defaultExistentialDeposit` deleted. MAX settles at `available − fee` in both the Details fill and the Verify recompute (they share `existentialDeposit(for:)`). `canBeReaped` becomes structurally inert for XRP — correct by design: with a reserve-net balance, "would drop below the reserve" is exactly `amount + fee > rawBalance`, which `isAmountExceeded` / `validateBalanceWithFee` already block. DOT (total-balance model, `transfer_keep_alive`) untouched.
4. **`feat(ripple): block sub-reserve sends to unfunded XRP destinations`** — Sign-time guard: destination `account_info` lookup; unfunded + `amount < live base reserve` fails with localized `xrpDestinationNotActivatedError` *before* the ceremony (on-chain this is `tecNO_DST_INSUF_XRP`, after the fee burns). Funded destinations pass untouched; exactly the base reserve activates; lookup failures fail closed. Errors are rewrapped as `HelperError` so the Verify screen's alert plumbing actually presents them. Copy added to all 8 locales, sorted.

## Intentional test flip

`SendValidationTests.testCanBeReapedTrueForRippleWhenRemainderBelowExistentialDeposit` asserted the buggy double-count behavior (partial send leaving < 1 XRP spendable must be blocked). It flips to `testCanBeReapedFalseForRippleReserveNetBalance`, plus new pins: `testExistentialDepositZeroForRipple` and `testComputeMaxAmountForRippleSubtractsOnlyFee` (MAX == R − f exactly, OwnerCount > 0 fixture). DOT and TAO expectations unchanged and green.

## Cross-platform parity

- **vultisig-sdk#959**: iOS never had the fee-side bug (`BlockChainService` builds `.Ripple(gas:)` from the network fee only); what iOS lacked was the unfunded-destination guard — folded in here (commit 4).
- **vultisig-android#5148 / PR #5180**: same destination guard shipped there (static base reserve). Android's *sender-side* XRP reserve still has the same double-count pattern this PR fixes (`checkIsReapable` static 1 XRP vs. its own reserve-net `RippleApi.getBalance`) — a parity note for tracker vultisig-sdk#979 follows once this lands.
- Converged semantics: **MAX = (total − reserve_base − OwnerCount × reserve_inc) − fee, values live from `server_state`**.

## Gate

- Full unit suite (`make test`, iPhone 17 Pro Max / iOS 26.5, en_US): **2389 tests, 0 failures — TEST SUCCEEDED**
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/`: 43 warnings, all pre-existing in untouched files — 0 new.
- Codex cross-review: 4 per-step gates (2 rounds max) + whole-branch pass; all findings fixed or ledgered.

**Runtime note:** the on-chain acceptance item ("MAX send from an account holding N objects succeeds") needs a real device broadcast — the owner-aware math is pinned by unit fixtures modelling OwnerCount > 0; a mainnet probe of `RippleReserve.reservedDrops` matches explorer reserve figures for the current 1 / 0.2 XRP parameters.

Closes #4744

## Summary by CodeRabbit

* **New Features**
  * Added reserve-aware XRP spendable balance calculation using live network reserve values.
  * Added pre-send XRP destination activation verification with clearer, reserve-based thresholds.
  * Added localized XRP destination error messages (activation required / lookup failed) across multiple languages.
* **Bug Fixes**
  * Improved XRP max-send and balance validation to correctly reflect reserve-net behavior.
  * Enhanced fail-open vs fail-closed handling for destination checks when network data is missing or ambiguous.
* **Tests**
  * Added/expanded unit tests for XRP reserve math, caching/fallback behavior, and destination activation guard flows.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
@@ -423,7 +423,7 @@ enum RippleSendError: Error, LocalizedError {
         case .destinationNotActivated(let minimumXRP):
             return String(format: "xrpDestinationNotActivatedError".localized, minimumXRP)
         case .destinationLookupFailed(let code):
-            return "XRP destination lookup failed (\(code))"
+            return String(format: "xrpDestinationLookupFailedError".localized, code)
         }
     }
 }
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
@@ -274,6 +274,17 @@ class RippleService {
                     throw HelperError.runtimeError("RippleService deallocated")
                 }
                 let ledger = try await self.fetchServerState(host: host)?.result?.state?.validatedLedger
+                // A server that is up but still syncing answers HTTP 200 with
+                // `validated_ledger == null`, so both reserve fields come back
+                // nil. Caching that fully-empty response would pin the seeds for
+                // the whole 24h TTL and never refetch, even after the node
+                // recovers. Throw instead so TTLCache keeps serving the last-good
+                // snapshot (or the caller seeds) and retries next time — a thrown
+                // error is not cached. A partial response with at least one real
+                // field is still worth caching, seeds filling the gap.
+                guard ledger?.reserveBase != nil || ledger?.reserveInc != nil else {
+                    throw HelperError.runtimeError("server_state returned no reserve values (validated_ledger unavailable)")
+                }
                 return RippleReserveValues(reserveBase: ledger?.reserveBase, reserveInc: ledger?.reserveInc)
             }
         } catch is CancellationError {
@@ -288,20 +299,38 @@ class RippleService {
     /// would create the destination with less than the base reserve is
     /// rejected on-chain (`tecNO_DST_INSUF_XRP`) — after the ceremony, with
     /// the fee burned. Throws `RippleSendError.destinationNotActivated` when
-    /// the destination is unfunded and `amountDrops` is below the live base
-    /// reserve. A lookup failure also throws (fail closed, matching the
-    /// SDK/Android companions) so the ceremony never starts on an
-    /// unverifiable destination.
+    /// the destination is unfunded (a definitive `actNotFound`) and
+    /// `amountDrops` is below the live base reserve.
+    ///
+    /// The block fires ONLY on proof the destination is unfunded. A transport
+    /// or RPC error (offline, timeout, 429/5xx, an exhausted node-error retry)
+    /// is NOT such proof: before this guard existed the same send to a funded
+    /// address proceeded, so a transient lookup failure must not start blocking
+    /// it — fail open and let the on-chain guard remain the backstop. Only a
+    /// successful-but-uninterpretable response (HTTP 200, no account data, no
+    /// `actNotFound`) fails closed, with a localized message.
     func validateDestinationActivation(address: String, amountDrops: BigInt) async throws {
-        let result = try await fetchAccountsInfo(for: address)?.result
+        let result: RippleAccountResponse.Result?
+        do {
+            result = try await fetchAccountsInfo(for: address)?.result
+        } catch is CancellationError {
+            throw CancellationError()
+        } catch {
+            // Transport/RPC error — not evidence the destination is unfunded.
+            // Don't gate an otherwise-valid send behind a fresh, uncached
+            // account_info call whose transient failure would block it.
+            logger.warning("validateDestinationActivation: destination lookup failed, allowing send: \(error.localizedDescription)")
+            return
+        }
+
         guard result?.accountData == nil else {
             return // Funded destination — any amount is valid.
         }
 
-        // Only a definitive `actNotFound` means "unfunded". Any other
-        // account_data-less shape — rate limiting, a malformed address, a
-        // proxy error page decoding to all-nils — is an unverifiable
-        // destination and fails closed, exactly like a transport error.
+        // No account data: only a definitive `actNotFound` proves the account
+        // is unfunded. Any other account_data-less-but-successful shape (a
+        // proxy error page decoding to all-nils, an unexpected token) is
+        // unverifiable, so fail closed with a localized message.
         guard result?.error == "actNotFound" else {
             throw RippleSendError.destinationLookupFailed(code: result?.error ?? "unknown")
         }
@@ -316,7 +345,10 @@ class RippleService {
             reserveInc: reserveValues?.reserveInc
         )
         if amountDrops < baseReserve {
-            let minimumXRP = (Decimal(string: baseReserve.description) ?? 1) / pow(10, 6)
+            // drops → XRP. `toDecimal(decimals:)` only truncates, so divide to
+            // scale (the getMaxValue / SwapDetailsScreen idiom); reserves are
+            // always positive, so no fallback default is needed.
+            let minimumXRP = baseReserve.toDecimal(decimals: 6) / pow(10, 6)
             throw RippleSendError.destinationNotActivated(minimumXRP: minimumXRP.description)
         }
     }
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
@@ -34,6 +34,36 @@ enum RippleFee {
     }
 }
 
+/// Owner-aware XRPL account-reserve math. The reserve floor is
+/// `reserve_base + OwnerCount × reserve_inc` — every ledger object the account
+/// owns (trustline, offer, ticket, escrow, …) adds one increment. The floor
+/// applies to the Payment amount only; the transaction fee is exempt and may
+/// take the account below the reserve.
+/// - https://xrpl.org/docs/concepts/accounts/reserves
+enum RippleReserve {
+    /// Mainnet base reserve — 1 XRP (validator vote, Dec 2024). Last-resort
+    /// seed only; live values come from `server_state`.
+    static let seedReserveBaseDrops = BigInt(1_000_000)
+    /// Mainnet per-object owner reserve — 0.2 XRP (validator vote, Dec 2024).
+    /// Seed only, as above.
+    static let seedReserveIncDrops = BigInt(200_000)
+
+    /// Total reserved balance in drops: `reserve_base + OwnerCount × reserve_inc`.
+    /// Missing `server_state` fields fall back to the mainnet seeds; a missing
+    /// owner count counts as zero owned objects.
+    static func reservedDrops(ownerCount: Int?, reserveBase: Int?, reserveInc: Int?) -> BigInt {
+        let base = reserveBase.map { BigInt($0) } ?? seedReserveBaseDrops
+        let inc = reserveInc.map { BigInt($0) } ?? seedReserveIncDrops
+        return base + BigInt(ownerCount ?? 0) * inc
+    }
+
+    /// Spendable balance in drops: `max(total − reservedDrops, 0)`.
+    static func availableDrops(totalDrops: BigInt, ownerCount: Int?, reserveBase: Int?, reserveInc: Int?) -> BigInt {
+        let reserved = reservedDrops(ownerCount: ownerCount, reserveBase: reserveBase, reserveInc: reserveInc)
+        return max(totalDrops - reserved, BigInt(0))
+    }
+}
+
 class RippleService {
 
     static let shared = RippleService()
@@ -100,12 +130,13 @@ class RippleService {
             return "0"
         }
 
-        let ownerCount = BigInt(accountInfo?.result?.accountData?.ownerCount ?? 0)
-        let reservedBase = BigInt(serverState?.result?.state?.validatedLedger?.reserveBase ?? 1000000)
-        let reserveInc = BigInt(serverState?.result?.state?.validatedLedger?.reserveInc ?? 200000)
-
-        let reservedBalance = reservedBase + (ownerCount * reserveInc)
-        let availableBalance = max(totalBalance - reservedBalance, BigInt(0))
+        let validatedLedger = serverState?.result?.state?.validatedLedger
+        let availableBalance = RippleReserve.availableDrops(
+            totalDrops: totalBalance,
+            ownerCount: accountInfo?.result?.accountData?.ownerCount,
+            reserveBase: validatedLedger?.reserveBase,
+            reserveInc: validatedLedger?.reserveInc
+        )
 
         return availableBalance.description
     }
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
@@ -64,12 +64,19 @@ enum RippleReserve {
     }
 }
 
+/// Reserve values reported by `server_state`. Fields stay optional so a
+/// partial response still caches, with `RippleReserve` seeding the gaps.
+struct RippleReserveValues {
+    let reserveBase: Int?
+    let reserveInc: Int?
+}
+
 class RippleService {
 
     static let shared = RippleService()
 
     private let logger = Logger(subsystem: "com.vultisig.app", category: "ripple-service")
-    private let httpClient: HTTPClientProtocol = HTTPClient()
+    private let httpClient: HTTPClientProtocol
 
     /// Resolves the Ripple custom RPC override. Injected so the API values are
     /// built from a dependency rather than a global reach-in; resolution happens
@@ -77,8 +84,28 @@ class RippleService {
     /// live (the shared mirror updates without a relaunch).
     private let resolver: RPCEndpointResolving
 
-    init(resolver: RPCEndpointResolving = CustomRPCStore.shared) {
+    /// Last-good `server_state` reserve values. Reserves change only by rare
+    /// validator vote, so a generous TTL is safe; `TTLCache` coalesces
+    /// concurrent refreshes and fails open to the last-good snapshot when a
+    /// refresh throws. Internal (not `private`) together with the key so tests
+    /// can seed stale entries via `setCached`.
+    let reserveValuesCache = TTLCache<String, RippleReserveValues>()
+    private static let reserveValuesTTL: TimeInterval = 60 * 60 * 24
+
+    /// Cache key for the reserve values, scoped to a resolved host: a custom
+    /// RPC override can point at a network with different reserves (e.g. a
+    /// testnet), so a snapshot cached for one endpoint must never be served
+    /// for another after a runtime override change.
+    static func reserveValuesCacheKey(for host: URL) -> String {
+        "xrpl-reserve-values|\(host.absoluteString)"
+    }
+
+    init(
+        resolver: RPCEndpointResolving = CustomRPCStore.shared,
+        httpClient: HTTPClientProtocol = HTTPClient()
+    ) {
         self.resolver = resolver
+        self.httpClient = httpClient
     }
 
     /// The override-aware XRPL host. Falls back to the default host when no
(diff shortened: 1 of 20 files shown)


gastonm5:
## QA proof — verified on-chain (XRP mainnet)

A real MAX send from a build with this branch, showing the reserve is no longer double-counted. On-chain metadata pulled from a public XRPL node (`tx` method), not the explorer UI.

[`06D0E9398FBBF6A530B21DEC4100249AF40A9362AC77E0C6130DB5EDA44385C9`](https://xrpscan.com/tx/06D0E9398FBBF6A530B21DEC4100249AF40A9362AC77E0C6130DB5EDA44385C9) — `Payment`, `validated: true`, `tesSUCCESS`

| | drops | XRP |
|---|---|---|
| Sender balance before | 9,944,476 | 9.944476 |
| Amount sent | 8,944,456 | 8.944456 |
| Fee | 20 | — |
| **Sender balance after** | **1,000,000** | **1.000000** |
| Sender `OwnerCount` | 0 | — |

- The account landed at **exactly the base reserve** (1.000000 XRP): `remaining − reserve = 0`.
- MAX amount = `balance − reserve − fee` → `9,944,476 − 1,000,000 − 20 = 8,944,456` ✓.
- Under the previous double-count, MAX would have subtracted the reserve **plus** a second static 1 XRP, sending ~7.94 XRP and stranding 2 XRP. Instead it sent the full spendable amount and left the single reserve.
- `tesSUCCESS` confirms the ledger accepts a send that lands exactly at the reserve (the fee is reserve-exempt) — the assumption the fix relies on.

**Coverage note:** this account's `OwnerCount` is 0, so this proves the double-count is removed; it does not exercise the owner-aware *scaling* (for `OwnerCount 0` the owner-aware reserve equals the base reserve). A MAX send from an account holding an object (e.g. a trust line → `OwnerCount ≥ 1`) would confirm the reserve correctly grows to `base + N × 0.2 XRP` (leaving 1.2 XRP instead of 1.0). Still pending as the remaining manual variant.

VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyLogic.swift
```diff
@@ -170,6 +177,28 @@ struct SendCryptoVerifyLogic {
...
+
+    /// XRPL rejects a Payment that would create the destination account with
+    /// less than the base reserve (`tecNO_DST_INSUF_XRP`) — on-chain, after
+    /// the keysign ceremony, with the fee burned. Gate it here so the failure
+    /// surfaces before signing starts; no-op for every other chain. Matches
+    /// the guard the SDK and Android run at submit time.
+    func validateDestinationIfNeeded(tx: SendTransaction) async throws {
+        guard tx.coin.chain == .ripple, tx.coin.isNativeToken else { return }
+        do {
+            try await rippleService.validateDestinationActivation(
```
johnnyluo, line 190: **Fails closed on any destination-lookup transport error, blocking previously-working sends to *funded* XRP destinations.**

`validateDestinationActivation` → `fetchAccountsInfo` rethrows on a transient network blip or a 429/rate-limit from the RPC. That error propagates out and is rewrapped here as `HelperError.runtimeError`, blocking signing with a message like "The Internet connection appears to be offline." Before this PR the send path did no destination lookup, so the same send to a funded address proceeded and succeeded on-chain.

Fail-closed is right for verifying an *unfunded* destination, but it now gates *every* native XRP send behind a fresh, uncached `account_info` call whose transient failure blocks an otherwise-valid transaction. Consider failing closed only when the response is genuinely account-data-less (`actNotFound` vs. any other shape), and letting a transport/RPC error through — or caching the funded-destination result.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
```diff
@@ -91,25 +148,99 @@ class RippleService {
...
+            return try await reserveValuesCache.value(
+                for: Self.reserveValuesCacheKey(for: host),
+                now: Date(),
+                ttl: Self.reserveValuesTTL
+            ) { [weak self] in
+                guard let self else {
+                    throw HelperError.runtimeError("RippleService deallocated")
+                }
+                let ledger = try await self.fetchServerState(host: host)?.result?.state?.validatedLedger
+                return RippleReserveValues(reserveBase: ledger?.reserveBase, reserveInc: ledger?.reserveInc)
```
johnnyluo, line 288: **An all-nil `server_state` response gets cached as authoritative for the full 24h TTL, pinning the seed fallback.**

A rippled endpoint that is up but not fully synced can return HTTP 200 with `validated_ledger == null` (common right after a restart/resync). This closure does **not** throw — it returns `RippleReserveValues(nil, nil)`, which `TTLCache` stores as a valid entry for 24h. For that window `getBalance` and `validateDestinationActivation` fall back to the 1 / 0.2 XRP seeds and never refetch, even after the server recovers.

Latent today (seeds == mainnet), but if a validator vote later raises the reserve, the app overstates spendable balance, computes a too-large MAX send (fails on-chain after the ceremony, fee burned), and under-blocks sub-reserve sends to unfunded destinations. A thrown error is correctly *not* cached — only the empty-but-successful response poisons the cache. Consider not caching a fully-empty response, or caching it with a short TTL.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
```diff
@@ -159,6 +294,27 @@ class RippleService {
...
+    /// blocked (fail closed). Technical copy, same style as
+    /// `RippleBroadcastError`.
+    case destinationLookupFailed(code: String)
+
+    var errorDescription: String? {
+        switch self {
+        case .destinationNotActivated(let minimumXRP):
+            return String(format: "xrpDestinationNotActivatedError".localized, minimumXRP)
+        case .destinationLookupFailed(let code):
+            return "XRP destination lookup failed (\(code))"
```
johnnyluo: **Hardcoded English string reaches the UI unlocalized.**

On any non-`actNotFound` lookup shape, this `destinationLookupFailed` literal is rewrapped into `HelperError` and shown verbatim in the Verify-screen alert (`SendVerifyScreen.swift:168`) in every locale. Violates `.claude/rules/localization.md` ("Never hardcode user-facing strings — use `"key".localized`"). The sibling `destinationNotActivated` case is correctly localized via `xrpDestinationNotActivatedError`. Low severity — consistent with the pre-existing `RippleBroadcastError` style — but still a rule violation.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
```diff
@@ -91,25 +148,99 @@ class RippleService {
...
+        // least the base reserve. OwnerCount is 0 by definition for an
+        // account being created.
+        let reserveValues = try await fetchReserveValues()
+        let baseReserve = RippleReserve.reservedDrops(
+            ownerCount: 0,
+            reserveBase: reserveValues?.reserveBase,
+            reserveInc: reserveValues?.reserveInc
+        )
+        if amountDrops < baseReserve {
+            let minimumXRP = (Decimal(string: baseReserve.description) ?? 1) / pow(10, 6)
```
johnnyluo: **Re-implements an existing helper.**

`(Decimal(string: baseReserve.description) ?? 1) / pow(10, 6)` round-trips through a string and hand-divides. The repo already has `BigInt.toDecimal(decimals:)` (`Core/Extensions/BigInt.swift:30`), so this is just `baseReserve.toDecimal(decimals: 6)`. That also drops the misleading `?? 1` fallback, which would silently report a "1 XRP" minimum if the string ever failed to parse.

johnnyluo (commented):
Automated review of the XRP owner-aware reserve fix. The core change is sound — reserve math, `canBeReaped` inertness for XRP, `computeMaxAmount`, and the TTLCache fail-open chain all check out. Four inline findings below; #1 and #2 are worth a decision before merge, #3 and #4 are cleanups.

johnnyluo:
### 🐛 Sub-reserve send to an unfunded XRP destination is not blocked on the Verify screen

Sending **0.1 XRP** to a destination account that **does not exist yet** reaches the Verify screen with no error and an enabled **Sign transaction** button. The destination-activation guard added in this PR (`RippleService.validateDestinationActivation`) should reject this: 0.1 XRP = 100,000 drops < the 1 XRP base reserve, so on-chain it would fail with `tecNO_DST_INSUF_XRP` after the fee is burned.
CleanShot 2026-07-07 at 08 46 02@2x

> _Screenshot: on the Verify screen for a 0.1 XRP send to the unfunded address below, no error banner is shown and **Sign transaction** is enabled._

**Repro**
- From: `rMwdVSrJte3z8zJsdDySGSgBq27xWqt9VW`
- To (unfunded): `rEh4M58svDByyqkXGGcUFjxV7iXN5XuE7K`
- Amount: 0.1 XRP → Verify screen shows no error; Sign is enabled.

**Where the guard runs**
`SendCryptoVerifyViewModel.validateForm()` calls `logic.validateDestinationIfNeeded(tx:)` — but `validateForm()` only runs from `SendVerifyScreen.signAndMoveToNextView()` **when Sign is pressed**, not when the Verify screen loads. So the screen renders clean regardless. Worth confirming whether the intent was to surface this pre-Sign (like the balance/amount checks) rather than only on the Sign tap.

**Then, on Sign**, please double-check the guard actually fires end-to-end for this case. Candidate reasons it may still not surface:
1. The real `account_info` response for a non-existent account decodes to `result.error == "actNotFound"` and `accountData == nil` — verify the newly-added `error` CodingKey actually populates against the production endpoint (the guard's unfunded branch hinges on it; if `error` comes back nil, it throws `destinationLookupFailed` instead — still an error, but a different one).
2. Confirm this send does not take the `prebuiltKeysignPayload` early-return in `validateForm()` (line 189), which skips `validateDestinationIfNeeded` entirely.

If the guard *is* meant to be tap-time only, at minimum the Verify screen should reflect the invalid state; if it's meant to be load-time, `validateDestinationIfNeeded` needs to be wired into the same validation pass as the amount/balance checks.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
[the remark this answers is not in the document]
gastonm5: Done in `51bd24dfc` — added `xrpDestinationLookupFailedError` to all 8 `Localizable.strings` and changed the case to `String(format: "xrpDestinationLookupFailedError".localized, code)`. Resolving.

VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyLogic.swift
gastonm5: Fixed in `f2390404e`: the guard now fails **open** on transport/RPC/retry-exhausted errors — only a definitive `actNotFound` fails closed and runs the reserve threshold. A send to a funded destination is no longer blocked by a transient `account_info` blip.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
gastonm5: Fixed in `f2390404e`: the reserve-values fetch closure now throws on an all-nil `server_state` response, so `TTLCache` doesn't store it and refetches next time — the empty-but-successful response can no longer pin the seed for the 24h window.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
gastonm5: Fixed in `51bd24dfc`: `destinationLookupFailed` now uses `String(format: "xrpDestinationLookupFailedError".localized, code)`, with the key added to all 8 locales.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleService.swift
gastonm5: Fixed in `f2390404e`: switched to `baseReserve.toDecimal(decimals: 6)` and dropped the misleading `?? 1` fallback. (Kept the explicit scaling since `toDecimal` truncates rather than scales — please sanity-check the final expression reads as whole XRP.)

gastonm5:
## Update: merged `main` + addressed review

**Merged `origin/main`** (merge commit 4783dd344) to clear the conflict. The only content conflict was `RippleService.swift`, reconciling this branch's reserve cache / destination guard with main's Ripple retry + verify-by-hash landing (#4746 / #4750 / #4754): kept both the injected `httpClient` and the new `RippleRequestRetrier`, both the reserve-values TTL cache and the `verifyByHashBackoff` init params, routed `fetchServerState` through the retrier while keeping the explicit `host` the reserve cache key depends on, and merged the `RippleAccountResponse.error` doc. `Localizable.strings` did not conflict.

### Review dispositions

| # | Finding | Disposition |
|---|---|---|
| 1 | Sub-reserve send to an unfunded destination not blocked on the Verify screen (johnnyluo) | **Fixed** `20479b452` — guard now runs in `loadGasInfoForSending` (load pass); Sign disabled on load. See the reply on that comment for the `actNotFound` / `prebuiltKeysignPayload` verification. |
| 2 | Fails closed on any transport lookup error, blocking funded sends (johnnyluo) | **Fixed** `f2390404e` — fail **open** on transport/RPC/retry-exhausted errors; only a definitive `actNotFound` fails closed. |
| 3 | All-nil `server_state` cached for 24h (johnnyluo) | **Fixed** `f2390404e` — the fetch closure throws on an all-nil response, so `TTLCache` doesn't cache it and refetches. |
| 4 | `destinationLookupFailed` hardcoded English (johnnyluo + CodeRabbit) | **Fixed** `51bd24dfc` — `xrpDestinationLookupFailedError` added to all 8 locales; case uses `String(format:…)`. |
| 5 | Reinvents `BigInt.toDecimal` (johnnyluo) | **Fixed** `f2390404e` — `baseReserve.toDecimal(decimals: 6) / pow(10, 6)`, dropping the `?? 1`. (`toDecimal` only truncates, so the explicit `/ pow(10,6)` stays — see the reply.) |
| — | Cancellation contract (whole-branch review) | **Fixed** `87a609132` — `CancellationError` propagates through the guard at both layers instead of becoming a spurious balance error. |
| — | Duplicated test doubles (CodeRabbit nitpick, Trivial) | Deferred — pre-existing across the Ripple test files; extracting a shared scripted `HTTPClientProtocol` is a separate cleanup. |

### Gate + runtime verification

Full `make test` on a dedicated en_US iOS 26.5 simulator (iPhone 17 Pro Max clone, region verified `AppleLocale = en_US`): **2464 tests, 0 failures, `** TEST SUCCEEDED **`** + swiftlint 0 violations on the changed files.

Finding #1 runtime verification: the suite runs `testLoadGasInfoBlocksUnfundedXrpDestinationOnLoad` in the simulator, driving the real VM load pass through `RippleService.validateDestinationActivation` against the **captured production `actNotFound` response** for `rEh4M58svDByyqkXGGcUFjxV7iXN5XuE7K` — it asserts `hasBalanceError` + `signButtonDisabled` on load for the exact 0.1 XRP repro. The production `account_info` response shape was confirmed live against `xrplcluster.com`. A full manual UI drive from a funded vault wasn't performed (needs a funded on-device vault, same limitation as the DOT ED plan); the on-chain MAX-send acceptance receipt is already attached above.

Reviewed with the cross-model codex loop (read-only `gpt-5.5-high` gate per commit + whole-branch pass).

johnnyluo (approved):
LGTM

PR #4749, opened 2026-07-06
feat(ripple): XRP Destination Tag as a first-class field + restored memo & tag+memo support

### LINKED ISSUE #4742: feat(ripple): dedicated Destination Tag field under Memo in XRP send (+ X-addr autofill, RequireDest warning)
## What
The XRP send view has **no dedicated destination-tag field** — the tag is overloaded onto the memo field, which is error-prone and hides a fund-loss risk. Add a proper destination-tag UX for XRP sends.

## Fix
1. **Add a dedicated "Destination Tag" field as a second field directly under the Memo field** in the XRP send view (numeric, optional).
2. **X-addresses:** accept `X...` address input; auto-populate the destination-tag field from the decoded tag and lock/show it (consumes vultisig-sdk#981).
3. **RequireDestTag warning:** when the destination requires a tag and none is entered, block the send with a clear warning (consumes vultisig-sdk#980).

iOS: Ripple.swift builds the Payment; tag currently overloaded onto memo.

## Acceptance
- [ ] Dedicated Destination Tag field rendered as a second field under Memo in the XRP send flow
- [ ] Pasting an X-address auto-fills the tag from the decoded value
- [ ] Sending to a RequireDest account with no tag is blocked with a clear warning
- [ ] Tag is carried into the signed Payment (not via memo overload)

Ref: Figma send view · XRPL dest tags https://xrpl.org/docs/concepts/transactions/source-and-destination-tags · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979). Depends on vultisig-sdk#980, vultisig-sdk#981.

### LINKED ISSUE #4755: feat(ripple): restore memo on plain XRP sends + destination tag & memo together (rawJSON)
## What
#4749 made plain XRP sends **tag-only** (non-numeric memo rejected) because the memo slot was the tag's only carrier. Now that the destination tag has its own first-class field (`RippleSpecific.destination_tag`, commondata#95), the memo slot is free to carry an actual on-chain memo again. Restore memo support for plain XRP sends, and support a **destination tag AND a memo together** (XRPL allows both as independent fields).

## Behavior (plain payments; swap path unchanged)
Signer resolution in `Ripple.swift`:
- **Field-tag present, memo empty** → tag-only Payment (native `destinationTag`).
- **Field-tag present, memo == canonical(field-tag)** → the dual-write **echo** → tag-only (NOT a Memos blob). Preserves current byte output.
- **Field-tag present, memo is real text** → **combo**: `rawJson` Payment with BOTH `DestinationTag` (from field) and `Memos` (from memo).
- **Field-tag present, memo is a different number** → tag/memo conflict → reject (existing conflict rule).
- **Field-tag absent, memo numeric** → tag (legacy parse).
- **Field-tag absent, memo text** → `Memos` blob (legacy rawJson) — restores pre-#4749 memo-only capability.

Initiator population (dual-write preserved):
- tag-only → field=tag, memo=canonical(tag) [echo].
- memo-only → field absent, memo=text.
- tag+memo → field=tag, memo=text (tag rides the field only).

## Mixed-version safety (decided: allow the combo now)
- **tag-only** and **memo-only** are byte-identical on any device version (old co-signer reads the memo slot and agrees) — safe now.
- **tag+memo** requires the memo slot to hold the text, so an **old co-signer** (no field) builds a `Memos`-only tx without the tag → hashes diverge → the keysign **fails** (fail-safe: no fund loss, no wrong tx) until that device updates. Accepted.
- Add a clear advisory on the send form when both a tag and a memo are entered: "Sending a tag and a memo together requires all your signing devices to be up to date." (localize, 7 locales).

## Acceptance
- [ ] Plain XRP send accepts a text memo again (form no longer rejects it)
- [ ] Memo-only send lands an on-chain `Memos` blob; byte-identical old vs new device
- [ ] Tag-only send unchanged (echo recognized → no Memos); byte-identical old vs new
- [ ] Tag+memo send builds `rawJson` with both `DestinationTag` and `Memos` on an updated device
- [ ] Advisory shown when both fields are populated
- [ ] Unit + byte-parity tests for all cases; swap path untouched; no signing change beyond the resolution rules

Stacks on #4749 (needs the `RippleSpecific.destination_tag` field). Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gastonm5:
Closes #4742

## The fund-loss shape this closes

Today XRP has no destination-tag concept anywhere in the iOS UI, transaction model, or keysign payload — the tag is overloaded onto the generic memo field. `RippleHelper.getPreSignedInputData` branches on the memo string: a value that parses as an integer becomes a real `DestinationTag` (correct, but undiscoverable), while a non-numeric memo (`tag:12345`, a trailing space, `> UInt64.max`) falls through to a hand-built rawJSON transaction with an on-chain `Memos` blob and **no DestinationTag**. An exchange never credits such a deposit — funds are effectively lost. A user who omits a required tag entirely hits the same loss, previously unchecked.

This PR gives XRP a dedicated, validated Destination Tag field, autofills it from X-addresses, gates tagless sends to accounts that require a tag, and removes the non-numeric-memo → on-chain-Memos path outright.

## Wire format: memo is the permanent carrier, validated on both sides — NO commondata change

The tag rides the existing `KeysignPayload.memo` field on the wire as a **canonical uint32 decimal** (1..4294967295, no leading zeros; `"0"`/empty = no tag). This is byte-identical to what every shipped platform already produces — iOS/Android/SDK all implement the same numeric-memo → `DestinationTag` rule in their Ripple signing-input builders, so in MPC keysign every co-signer rebuilds an identical `RippleSigningInput` and the pre-image hash matches. A first-class proto field would break that: any co-signer that didn't read it would build a tagless payment → different pre-image → ceremony failure for every mixed-version/other-platform peer. So **no `RippleSpecific.destination_tag` proto field, ever** — memo is the permanent carrier.

Safety comes from **two-sided validation** instead: an XRP keysign payload's memo must be **empty or a canonical uint32 decimal**, else the payload is rejected — enforced on both the initiator (form) path and the co-signer (pre-sign) path (which also covers dApp/deeplink-originated payloads). This removes the old non-numeric-memo → on-chain-`Memos` path (the fund-loss shape). Consequence, accepted by design: **in-app arbitrary text memos for XRP are no longer supported.** SwapKit's numeric-memo → tag contract is preserved — swap payloads keep the legacy memo contract; plain sends are strict.

## Per-commit design

1. **`c3cde36de` — hand-rolled XLS-5d X-address codec.** WalletCore's Swift surface exposes no X-address codec, so this adds pure-Swift base58 (XRPL alphabet, not Bitcoin's) + sha256d-checksum decode/encode, pinned to canonical `ripple-address-codec` vectors (with-tag, no-tag, tag-0, max-u32, testnet reject, corrupt checksum).
2. **`df5d23237` — tag threading + canonical-u32 memo contract + two-sided validation.** `SendTransaction.destinationTag: UInt32?` carried separately from `memo` (honest Verify display), encoded into the memo slot at payload build; `Ripple.swift` hardened to a u32-bounded parse; `validatePayloadMemo` enforces the empty-or-canonical-uint32 contract on both initiator and co-signer paths. `parseTag` accepts only canonical, non-zero decimals (proto3 scalar can't express present-with-0, so `"0"` = no tag).
3. **`a0690cb89` — Destination Tag field (XRP-only, under Memo) + Verify row.** `Chain.supportsDestinationTag`, a `DestinationTagTextField` modeled on the memo field (numeric pad, paste, digit-filtered), a `Field.destinationTag` focus case, and a Destination Tag row on the Verify screen. Copy localized across 8 locales.
4. **`31f96c5b6` — X-address paste → normalize + autofill/lock.** Pasting an `X…` address normalizes to the classic `r…` address for the Payment and autofills + locks the tag from the embedded value; clearing the address releases the lock and drops the derived tag.
5. **`9a025c653` — RequireDest pre-send gate.** Reads `lsfRequireDestTag`; hard-blocks a tagless send to an account that requires a tag; fails **open** with an explicit acknowledgment on lookup failure.

## Fail-open + acknowledgment rationale (RequireDest)

The gate reads `lsfRequireDestTag` from the destination's AccountRoot flags. On `actNotFound` (unfunded account) or a lookup/network failure it **fails open but requires an explicit user acknowledgment**, keyed to the exact destination address. Rationale: fail-closed would make XRPL public-RPC availability a hard dependency for *all* XRP sends (a new outage class) — and the codebase already fails open on the same RPC for fees. The fund-loss cohort (exchange deposit addresses) are funded, high-availability accounts, so the happy-path check protects them; the ack keeps the user informed without bricking sends. The ack is address-scoped so it can't leak onto a different destination.

## X-address autofill + lock lifecycle

Normalization happens at the ViewModel's single address seam (covers paste, QR, address book, deeplinks). An `X…` with an embedded tag → classic address in the Payment + tag autofilled and locked; an `X…` without a tag → address normalized, tag stays editable. Clearing the address releases the lock and drops the *derived* tag (a manually typed tag is user intent and mirrors the memo field's order-independent lifecycle, so it persists — that seam was hardened during review so a stale derived tag can't ride onto a re-entered address). Validation runs address-resolution → tag/memo → RequireDest in that order, so a raw tag-0 X-address on a prefill path can't slip a displayed-but-dropped tag through.

## Test coverage

New XCTest suites (camelCase):
- `RippleXAddressTests` — codec vectors (with/without tag, tag-0 vs no-tag, max-u32, testnet reject, corrupt checksum, classic round-trip).
- `RippleDestinationTagThreadingTests` — tag encodes as canonical decimal memo; `validatePayloadMemo` rejects non-canonical / `"0"`; string-memo path removed; above-u32 typed error.
- `RippleXAddressAutofillTests` — paste normalizes + locks; no-tag leaves editable; clearing releases lock and drops derived tag; zero-tag X-address on prefill path rejected.
- `RippleDestinationTagRequirementTests` — `lsfRequireDestTag` classification, `actNotFound`, network error → unknown.
- `RippleRequireDestGateTests` + `SendFormFixture` — tagless send blocked; passes with tag; lookup-failure fail-open ack; non-XRP skip; chain-switch clears tag.

## Gate

- `make test` (VultisigApp scheme): green — `** TEST SUCCEEDED **`, full suite. One gate-forced fixup (`7df58c573`) removed a now-stale XRP test fixture (`xrp.json` "Send XRP with memo" = "helloworld") that exercised the deleted text-memo → on-chain-`Memos` path; the rejection it used to assert is now covered by `RippleDestinationTagThreadingTests`.
- SwiftLint (changed files): clean — 0 new violations (one pre-existing `trailing_comma` warning in `Chain.swift:58` predates this branch).
- Simulator: dedicated iPhone 17 Pro Max clone on iOS 26.5, region en_US (avoids the known decimal-comma locale flake).

Deferred to a manual on-device pass (per the tracker's real-broadcast requirement): a real XRP send carrying a tag to an exchange deposit address (confirm credit), X-address paste → autofill → send, and a tagless send to a RequireDest destination (confirm block).

## Summary by CodeRabbit

* **New Features**
  * Added XRP Destination Tag support in send and verify flows, including X-address auto-fill, tag locking, and destination-tag display.
  * Added a destination-tag input field with paste support and expanded localized text in multiple languages.
  * Added support for showing destination-tag requirements and unverified destination warnings.

* **Bug Fixes**
  * Improved handling of XRP memos and tags to prevent conflicts and invalid values.
  * Added support for decoding XRP X-addresses into classic addresses with embedded tags.

---

## QA / Testing guide

The X-address checks below need **no funds** — you're verifying decode / autofill / lock by watching the fields react. Paste each into the recipient field on an **XRP** send.

### X-address paste → expected autofill

| Paste this X-address | Recipient normalizes to | Tag field | Lock |
|---|---|---|---|
| `X7AcgcsBL6XDcUb289X4mJ8djcdyKaGZMhc9YTE92ehJ2Fu` | `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` | `1` | 🔒 locked |
| `X7AcgcsBL6XDcUb289X4mJ8djcdyKaB5hJDWMArnXr61cqZ` | `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` | *(stays empty / editable)* | unlocked |
| `XVLhHMPHU98es4dbozjVtdWzVrDjtV18pX8yuPT7y4xaEHi` | `rGWrZyQqhTp9Xu7G5Pkayo7bXjH4k4QYpf` | `4294967295` (uint32 max) | 🔒 locked |

**Expected:** the field **accepts** the X-address (no "invalid address" error). Under the hood the recipient is normalized to the classic `r...` form used for signing / the Blockaid scan / tx history, while the original X-address is preserved and shown as the **label** on the Verify and Done screens. The tag autofills from the embedded value and locks (first two are the same account with / without a tag).

**Negative X-address cases:**
- **Tag-0 X-address** `XVLhHMPHU98es4dbozjVtdWzVrDjtV8AqEL4xcZj5whKbmc` (`rGWr…` + tag 0): autofills `0` and locks, **but then blocks the send** — proto3 can't distinguish tag 0 from "no tag", so signing it would send untagged behind a displayed tag.
- **Testnet X-address** `T7oKJ3q7s94kDH6tpkBowhetT1JKfcfdSCmAXbS75iATyLD` (starts with `T`): **rejected as invalid** on mainnet, no autofill.

### Autofill/lock interactions
1. Type `999` in the tag field, then paste tagged `X7Acgcs…ehJ2Fu` → embedded tag `1` wins and locks.
2. After an X-paste, **clear** the recipient field → lock releases and the derived tag drops (a stale tag can't ride onto the next address).
3. After an X-paste, **hand-edit** the recipient into a different classic address → unlocks and clears the derived tag.

### Manual destination tag (classic `r...` address + typed tag)
- **Accepted:** `1`, `12345`, `4294967295`, or empty (no tag).
- **Rejected:** `0`, `4294967296` (> uint32 max), `007` (leading zero), `12.5`, `abc`, `-1`.
- **Memo field:** any text on a plain XRP send is **rejected** with a steer to the tag field (the old text-memo → on-chain `Memos` blob path is removed).

### Real end-to-end (needs funds)
Send a small amount to **your own second XRP account** with a tag (e.g. `12345`), sign, then confirm the `DestinationTag` shows on the Payment at [livenet.xrpl.org](https://livenet.xrpl.org). Don't send to the demo addresses above. An exchange XRP deposit that requires a tag is the most realistic check.

> X-address vectors are the canonical XLS-5d fixtures the codec is pinned to, so they decode identically in the build.

---

## Merged in: memo restoration + tag+memo combo (was #4759 — `Closes #4755`)

Folded the stacked memo PR into this one. Now that the destination tag has its own first-class commondata field, plain XRP sends can carry a **text memo** again, and a **destination tag AND a memo together** (via rawJSON `DestinationTag` + `Memos`).

Signer resolution (plain payments): field + empty memo → tag-only; field + memo == canonical(tag) → dual-write **echo** → tag-only (no Memos); field + real text → combo rawJSON (DestinationTag + Memos); no field + numeric → tag (legacy); no field + text → Memos blob (restored). Swap path unchanged.

**Byte-parity:** tag-only (field ≡ echo) and memo-only (field-unset ≡ old device) proven byte-identical across device versions. The **combo** requires the memo slot to hold the text, so an **old co-signer** builds a Memos-only tx without the tag → keysign fails (fail-safe: no fund loss, no wrong tx). Accepted with a send-form advisory prompting all devices to update. Combined tree gate: **2487 tests** (sole failure = the known non-en_US comma-decimal locale flake, unrelated).

`Closes #4755`


gastonm5:
## QA proof — verified on-chain (XRP mainnet)

Two real sends from a build with this branch, each covering one of the two ways to set a destination tag. On-chain data pulled from a public XRPL node (`tx` method), not the explorer UI.

### 1. X-address paste → autofill path
[`7DB89F3F96A57C818C1D9982131386CAFBB20547195A0A69832E221BBD887064`](https://xrpscan.com/tx/7DB89F3F96A57C818C1D9982131386CAFBB20547195A0A69832E221BBD887064) — ledger 105416812, `validated: true`

Sent to the tagged X-address `X7AcgcsBL6XDcUb289X4mJ8djcdyKaGZMhc9YTE92ehJ2Fu`:

| Field | On-chain value |
|---|---|
| `TransactionType` | `Payment` |
| `Destination` | `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` — X-address normalized to its classic form |
| `DestinationTag` | `1` — the tag embedded in the X-address |
| `Memos` | *(none)* |
| result | `tesSUCCESS` |

Proves: X-address accepted → normalized to the classic address for signing → embedded tag carried into the first-class `DestinationTag` field, **not** a `Memos` blob.

### 2. Manual tag entry path
[`8B4700AA3AED776A5D301C2AF101416D31D38AB19F17E9AC9A976D59EB1CA769`](https://xrpscan.com/tx/8B4700AA3AED776A5D301C2AF101416D31D38AB19F17E9AC9A976D59EB1CA769) — `validated: true`

Sent to a plain classic address with a hand-typed tag:

| Field | On-chain value |
|---|---|
| `TransactionType` | `Payment` |
| `Destination` | `rUsiHjDXFn3c2m33RASooD3e9BJ1qYLTZG` |
| `DestinationTag` | `1234` — typed into the Destination Tag field |
| `Memos` | *(none)* |
| result | `tesSUCCESS` |

Proves: a manually-entered tag threads into `DestinationTag` with no `Memos` overload.

### Hash cross-check
For both txs, independently recomputing `SHA-512Half("TXN\0" + signed blob)` from the on-chain binary reproduces the exact tx id — so the deterministic hash the app now computes is correct on real signed transactions (also validates sibling PR #4750).

**Coverage:** both tag-entry routes confirmed on-chain, no fund-loss `Memos` path. Still to exercise manually: a MAX send from an `OwnerCount > 0` account (#4744) and an induced broadcast rejection (#4743).

gastonm5:
### Refactor: extract `RippleDestinationTagViewModel` out of `SendDetailsViewModel`

Behavior-preserving VM-layer refactor (2 commits on top of the existing branch — no rewrite of the prior 6). The XRP-only destination-tag concern (field state, X-address autofill/lock lifecycle, RequireDest gate) is lifted out of the ~987-line `SendDetailsViewModel` into a dedicated `@MainActor @Observable` `RippleDestinationTagViewModel`, so the parent stays a general send-form VM.

**What moved vs stayed**
- **Moved to the sub-VM:** `destinationTag` / `isDestinationTagLocked` / `showDestinationTagUnverifiedAlert` / `destinationTagFieldNudge`, the RequireDest cache + ack-address + injected provider, and the decision logic as return-typed pure methods — `applyDecodedTag` / `handleAddressChangedAway` / `releaseLockedTag` / `clearForUnsupportedChain`, `validateTagAndMemo(memo:) -> TagMemoValidation`, `resolvedTag(memo:)`, `validateRequireDest(toAddress:memo:) async -> RequireDestOutcome`, `acknowledge(toAddress:)`, `reset()`.
- **Stayed on the parent (shared):** `toAddress` / `toAddressLabel` / `lastResolvedAddress` (also the ENS/name-resolution path), `memo`, and the `setGeneralError` banner. `normalizeRippleXAddressIfNeeded` stays but is now thin: it decodes + writes the address fields it owns, then delegates the tag decision. The parent keeps thin delegators (`validateRippleTagAndMemo` / `resolvedDestinationTag` / `validateRippleRequireDest` / `onToAddressCleared` / `acknowledgeUnverifiedDestinationTag`) and forwards the injected RequireDest provider through its own init so DI/tests are unchanged.
- **Views:** `SendDetailsAdditionalSection` + `SendDetailsScreen` read/bind through `viewModel.rippleTag` (local `@Bindable` for the nested-observable bindings, since a binding can't chain through the parent's `let rippleTag`).

**Behavior preservation** — the two verified mainnet sends are pinned by tests:
- X-address paste `X7Acgcs…9YTE92ehJ2Fu` → classic `r9cZA1…gnfk59` + autofilled/locked tag `1` — `RippleXAddressAutofillTests` (asserting through `parent.rippleTag`).
- Manual tag `1234` → signed `DestinationTag=1234`, no Memos — new `testManualTagThreadsEndToEndToSignedDestinationTag` pins VM field → `SendTransaction` → payload memo → `RippleSigningInput.opPayment.destinationTag == 1234`.

**Tests** — integration tests kept on the parent (now via `parent.rippleTag.…`); added `RippleDestinationTagViewModelTests` pinning the new pure seams. No coverage reduction. The signing path (`Ripple.swift`/keysign/payload) is untouched.

**Cross-model review:** Codex (gpt-5.5 high) — per-commit source pass clean; whole-branch pass = 1 MINOR (add the exact `1234` end-to-end pin), fixed.

**Gate:** `make test` → `** TEST SUCCEEDED **`, 2459 tests / 0 failures; SwiftLint clean on the touched files.

gastonm5:
## Update — upgraded to a first-class `RippleSpecific.destination_tag` proto field (dual-write)

This supersedes the **"Wire format: memo is the permanent carrier — NO commondata change"** section of the description above. Following a later decision, XRP's DestinationTag now gets a first-class home in the shared protobuf, rolled out with a **dual-write** so the cosigner sees a labeled tag while mixed-version keysigns stay byte-identical.

### commondata (its own PR)
- **vultisig/commondata#95** — adds `optional uint32 destination_tag = 4;` to `RippleSpecific` (+ regenerated Swift/Go bindings). Additive, backward-compatible. Branch head `b940d663d78f13f26a175048f3e7a47ea86811e2`.

### Dual-write rollout (fund-safety critical)
- **Initiator** writes the resolved tag into **BOTH** `RippleSpecific.destination_tag` **AND** the existing `KeysignPayload.memo` carrier — same canonical uint32, both derived from one value so they can't disagree (`SendCryptoVerifyLogic.dualWritingRippleTag`). A tagless send leaves the field **unset**.
- **Signer** (`RippleHelper`) **prefers** the field and **falls back** to parsing the memo. A tag delivered via the field produces a **byte-identical** pre-signed `RippleSigningInput` (and pre-image hash) to the same tag via the memo — so a new-version signer and an old memo-only co-signer hash the exact same transaction in an MPC ceremony. When both are present the field wins (a divergent memo can at worst fail a legacy peer's ceremony, never produce a signature); a present-0 field is rejected exactly as memo `"0"` is. Swap path untouched.
- **Cosigner display**: the keysign confirm screen now renders a labeled **Destination Tag** row (field-preferred, else canonical memo) and suppresses the numeric-memo row — the UX win. Scoped to plain payments; swaps keep their existing memo display.

### Safety tests
The pin is `testFieldSourcedTagProducesByteIdenticalInputToMemoFallback`: field-sourced tag vs memo-sourced tag (same value) → **identical serialized signing-input bytes AND identical pre-image hash**. Plus dual-write equality, field-preference/conflict rules, present-0 rejection, proto round-trip (`hasDestinationTag` set only when present), and the two on-chain-verified behaviors (X-address tag=1, manual tag=1234) still pass end-to-end. Full `make test`: **2477 tests, 0 failures**.

### Draft-blocking dependency
This PR stays **draft** until commondata#95 merges. The `VultisigCommonData` SPM pin currently points at the commondata **branch head** (`project.yml` + `Package.resolved` re-pinned to `b940d663d…`); it must be re-pinned to the merged `main` commit before this leaves draft.

### Cross-platform ripple
Android / Windows / SDK / backend regenerate their bindings when they adopt the read side. Until then they keep reading the memo carrier — which is exactly what dual-write preserves. The memo carrier is retired only once every platform reads the field.

VultisigApp/VultisigAppTests/TestData/xrp.json
```diff
@@ -28,36 +28,6 @@
         },
         "expected_image_hash": ["345d97cf0407b433caef1f764c88341f073ba62fa96ee3eacb04c59de26220c2"]
     },
-    {
```
johnnyluo, line 31: can we please keep this test case?

VultisigApp/project.yml
```diff
@@ -57,13 +57,15 @@ packages:
...
-    revision: 03e61a7454be7cebababb6b33708b85077962088
+    # TEMPORARY: pinned to the head of the commondata `feat/ripple-destination-tag`
+    # branch, which adds `RippleSpecific.destination_tag` (optional uint32) — the
+    # first-class XRPL DestinationTag field. Consumed here via dual-write: the
+    # initiator writes the tag into BOTH this field AND the legacy memo carrier,
+    # and signers prefer the field, falling back to the memo, so mixed-version
+    # device pairs stay byte-identical in MPC keysign. Re-pin to the merged `main`
+    # commit once that commondata PR lands (this branch stays draft until then).
+    # Supersedes the earlier pin at commondata #93's `CosmosSpecific.gas_limit`.
+    revision: b940d663d78f13f26a175048f3e7a47ea86811e2
```
johnnyluo: PR in commondata has been merged , please update it

VultisigApp/VultisigAppTests/TestData/xrp.json
gastonm5: Restored — re-added the "Send XRP with memo" case (`helloworld` text memo, expected hash `a6e55bd5…`). It passes now that the memo-restore work is folded into this branch: a plain text memo is written to the Memos field again, reproducing the original signing input. Verified via `ChainHelperTests`.

VultisigApp/project.yml
gastonm5: Done — re-pinned `VultisigCommonData` to the merged `main` commit `2b92938e269ab44d735dabb30b1779a254f42143` (project.yml + Package.resolved) and dropped the temporary branch-head wording. Also merged latest `main` (resolved the Send-flow conflicts), so the PR is mergeable again.

johnnyluo (approved):
LGTM

PR #4750, opened 2026-07-06
fix(ripple): compute the deterministic transaction hash for signed XRP txs

### LINKED ISSUE #4747: fix(ripple): populate deterministic transactionHash so keysign safety nets arm for XRP
## Problem

XRP's signing helper returns the signed transaction with an **empty** hash: `Ripple.getSignedTransaction` builds `SignedTransactionResult(rawTransaction: output.encoded.hexString, transactionHash: "")` — `VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift:212`.

`KeysignViewModel` has three chain-agnostic safety nets that all key off `transactionType.transactionHash`, and the empty string leaves **all three inert for XRP**:

1. **Verify-by-hash net** — `isAlreadyOnChain` guards `!hash.isEmpty` and immediately returns `false`, so a failed/ambiguous broadcast can never be rescued by confirming the tx is already on-chain.
2. **Pre-broadcast idempotency guard** — `hasPriorBroadcastAttempt` checks the empty string, so a re-entered keysign flow can't detect that this exact transaction was already broadcast.
3. **Broadcast-cancellation net** — the cancelled-broadcast path falls through to the "couldn't confirm" outcome with an **empty** txid instead of verifying whether the signed tx landed.

## Fix direction

Populate `transactionHash` with the deterministic XRPL transaction hash: SHA-512Half over `"TXN\0" + <signed encoded blob>` (the same value the `submit` endpoint echoes back as `tx_json.hash`). It is computable from the compiled signing output with zero signing-logic changes, and that single change arms all three existing nets for XRP.

Note: the broadcast-layer dedup shipped for the tesSUCCESS gate uses the *echoed* `tx_json.hash` from the submit response precisely because the signing helper doesn't populate this value; once this issue is fixed the two hashes should be identical and the RPC-level-error case (network blip during submit) also becomes recoverable via the generic nets.

## Scope / caution

This touches the **signing path** (`Blockchain/Ripple/Signing/Ripple.swift`), unlike the broadcast-layer work — it needs its own focused review even though the change is mechanical.

Found while implementing #4743 (broadcast gating), where the dedup had to live inside `RippleService` specifically because these ViewModel nets are disarmed for XRP.

Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gastonm5:
Closes #4747

## What was inert & why

`RippleHelper.getSignedTransaction` returned the signed XRP transaction with an **empty** hash — `SignedTransactionResult(rawTransaction: …, transactionHash: "")`. Three chain-agnostic keysign safety nets in `KeysignViewModel` key off `transactionType.transactionHash` and short-circuit on an empty string, so all three were **disarmed for XRP**:

1. **Verify-by-hash net** — `isAlreadyOnChain(...)` returns `false` immediately when the hash is empty, so an ambiguous/failed broadcast could never be rescued by confirming the tx already landed.
2. **Pre-broadcast idempotency guard** — `hasPriorBroadcastAttempt(...)` returns `false` on the empty string, so a re-entered keysign flow couldn't detect a prior broadcast of this exact tx.
3. **Cancelled-broadcast net** — falls through to the neutral "couldn't confirm" outcome with an empty txid instead of verifying whether the signed tx landed.

Populating the hash with the deterministic XRPL transaction identifier arms all three at once.

## Algorithm (pinned against mainnet)

XRPL transaction identifying hash = **SHA-512Half** (first 32 bytes of SHA-512) over the 4-byte prefix `0x54584E00` (`"TXN\0"`, `HashPrefix::transactionID`) concatenated with the serialized **signed** transaction blob (`output.encoded`), rendered as canonical **uppercase** hex. SHA-512 comes from WalletCore's `Hash.sha512` (not hand-rolled). This mirrors the existing **Polkadot** precedent, which computes its hash the same way with blake2b over `output.encoded`.

The algorithm is pinned by offline fixture tests against two **real mainnet** transactions from ledger `105416247` (fetched via the `tx` JSON-RPC method with `binary: true` from a public XRPL endpoint), asserting the local computation reproduces each tx's own on-chain hash:

| type | on-chain hash (== computed) |
|---|---|
| Payment (secp256k1) | `00D3BAABAE790E1B9F772172CB5AD486161294E7F25916CC4E52423227768A14` |
| OfferCreate (ed25519) | `06E809490BE5C3341F9F11ED2067F168094E0BCD7AE047D37334113FE559D732` |

The OfferCreate vector demonstrates the hash is agnostic to transaction type and signature scheme — it is a pure digest of the serialized signed bytes.

## Security boundary (HIGH-tier repo)

The hash is derived **only** from the already-signed `output.encoded` bytes. **No** signing input, `SigningInput` proto, pre-image hashing, or TSS/keysign flow is touched — the bytes that get signed and broadcast are byte-for-byte unchanged. The diff is the hash computation + populating `transactionHash` + tests. An empty blob yields an empty string, so if the compiled output were ever empty the nets stay disarmed rather than keying off a meaningless prefix-only digest (no behavior regression vs. today).

## Which nets arm & composition with #4746

Once `transactionHash` is populated, the three generic nets above become active for XRP with zero further wiring. This composes with the broadcast-layer dedup in #4746 (broadcast gating): that work leaned on the `tx_json.hash` **echoed** by the `submit` response precisely because the signing helper didn't populate this value. With this fix the locally-computed hash and the echoed hash are identical, and the RPC-level-error broadcast case (network blip during submit) also becomes recoverable via the generic nets. Note: #4746 may not be merged yet; the tests here do **not** depend on its code.

## Parity

- **vultisig-sdk** already computes this locally from the signed bytes: `getRippleTxHash({ encoded }) => hashes.hashSignedTx(hex(encoded))` (`packages/core/chain/tx/hash/resolvers/ripple.ts`), delegating to the canonical `xrpl` lib (same `0x54584E00` prefix + SHA-512Half). This iOS change ports that exact behavior.
- **vultisig-android** has the **same** latent gap — `RippleHelper.getSignedTransaction` returns `transactionHash = ""` (`RippleHelpter.kt:153-155`). Parity note only; not changed here.

## Gate

- `make test` on a dedicated iPhone 17 Pro Max (iOS 26.5, en_US) simulator: **2372 tests, 0 failures** (7 new `RippleTransactionHashTests`, including the two mainnet vectors).
- SwiftLint (repo config): changed files clean, no new warnings.
- Cross-model review: Codex (gpt-5.5 high, read-only) gated each commit + a whole-branch pass — **no findings** in any round.

Part of the XRP chain-hardening set (tracker vultisig-sdk#979).

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction signing now returns a real Ripple transaction hash instead of an empty value.
  * Hashes are now formatted consistently in uppercase and match expected XRPL length and behavior.
  
* **Tests**
  * Added coverage for Ripple transaction hash generation using real signed transaction examples.
  * Verified deterministic hashing, empty-input handling, and sensitivity to byte changes.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
@@ -214,4 +214,24 @@ enum RippleHelper {
         return result
     }
 
+    /// XRPL "transaction identifying hash": SHA-512Half (the first 32 bytes of
+    /// SHA-512) over the 4-byte `TXN\0` prefix (`HashPrefix::transactionID`,
+    /// `0x54584E00`) concatenated with the serialized *signed* transaction blob.
+    /// Rendered as uppercase hex — the canonical XRPL rendering, and the same
+    /// value the `submit`/`tx` endpoints echo back as `tx_json.hash`.
+    ///
+    /// Derived purely from the already-signed output bytes; no signing input,
+    /// pre-image, or TSS state is involved.
+    ///
+    /// An empty blob yields an empty string so the hash-keyed keysign safety
+    /// nets stay disarmed rather than keying off a meaningless prefix-only hash.
+    ///
+    /// Reference: https://xrpl.org/docs/references/protocol/data-types/hash-prefixes
+    static func signedTransactionHash(signedBlob: Data) -> String {
+        guard !signedBlob.isEmpty else { return "" }
+        let transactionIDPrefix = Data([0x54, 0x58, 0x4E, 0x00]) // "TXN\0"
+        let digest = Data(Hash.sha512(data: transactionIDPrefix + signedBlob).prefix(32))
+        return digest.toHexString().uppercased()
+    }
+
 }
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
@@ -209,7 +209,7 @@ enum RippleHelper {
 
         let result = SignedTransactionResult(
             rawTransaction: output.encoded.hexString,
-            transactionHash: "")
+            transactionHash: signedTransactionHash(signedBlob: output.encoded))
 
         return result
     }
--- a/VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleTransactionHashTests.swift
+++ b/VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleTransactionHashTests.swift
@@ -1,0 +1,86 @@
+//
+//  RippleTransactionHashTests.swift
+//  VultisigAppTests
+//
+//  Pins the deterministic XRPL transaction identifying hash that
+//  `RippleHelper.signedTransactionHash(signedBlob:)` computes for a signed XRP
+//  transaction: SHA-512Half of the 4-byte `TXN\0` prefix (0x54584E00) + the
+//  serialized signed blob, rendered as uppercase hex.
+//
+//  The fixtures are real mainnet transactions pulled from ledger 105416247 via
+//  the `tx` JSON-RPC method (`binary: true`) against a public XRPL endpoint. The
+//  expected value is each transaction's own on-chain hash — the canonical proof
+//  that the local computation reproduces exactly what the ledger assigns. The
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4754, opened 2026-07-06
fix(ripple): retry XRPL requests on transient node errors (amendmentBlocked)

### LINKED ISSUE #4752: fix(ripple): retry XRPL requests on transient node errors (amendmentBlocked) — same-host retry, no fallback list
## What
XRP requests hit a single host (`xrplcluster.com` by default, or a custom override) with **no retry**. `xrplcluster.com` is a load-balanced **pool**, and some of its backend nodes run outdated `rippled` and answer with `amendmentBlocked` (error_code 14, HTTP 200 with `{"result":{"error":"amendmentBlocked",...}}`). One request lands on a stale backend and the error bubbles straight up — intermittently, "from time to time."

Observed on a `tx` status-lookup after a send: the transaction had **already succeeded on-chain** (`tesSUCCESS`, validated), but the status check hit a bad backend and surfaced `amendmentBlocked`, so the app showed an error / unconfirmed state for a send that actually landed. The same pool-node issue can transiently affect broadcast, status, and account/fee/server_state reads.

## Fix (decided)
**Retry the same resolved host on transient node errors — NO fallback host list.** Because the host is a pool, a retry hits a different backend, which is almost always healthy. Do NOT add a `publicFallbackHosts`-style list (unlike Solana); keep it to a bounded same-host retry.

- Detect the retryable condition from the **decoded JSON-RPC body** (`result.error`), not the HTTP status — `amendmentBlocked` comes back as HTTP 200. Retryable set: `amendmentBlocked` plus the node-unavailable family (`noNetwork`, `noCurrent`, `noClosed`, `tooBusy`/`slowDown` if judged safe) and transient transport failures (timeout / 5xx). Non-retryable JSON-RPC errors (e.g. `actNotFound`, `invalidParams`) pass through unchanged.
- Apply at the `RippleService` request layer so broadcast, status, and reads all benefit. Small, bounded backoff and a low max-retry count (keep added latency small — the broadcast happens in the time-sensitive post-signing window).
- **Broadcast retry is safe/idempotent:** resubmitting the same signed blob yields the same tx hash; an already-applied tx returns `tefALREADY` (handled by the broadcast result-code work). No double-spend.
- No signing-path changes — this is request/transport resiliency only.

## Acceptance
- [ ] A single `amendmentBlocked` response is retried on the same host and succeeds without surfacing an error (pool routes the retry to a healthy backend)
- [ ] Retryable set is a defined, documented list; non-retryable JSON-RPC errors (`actNotFound`, etc.) are NOT retried
- [ ] Retries are bounded (small count + backoff); broadcast latency stays acceptable
- [ ] Broadcast/status/reads all go through the retrying request path
- [ ] Unit tests: retryable error → retried then succeeds; non-retryable → surfaced immediately; retries exhausted → surfaces the last error; broadcast idempotency preserved

Part of the XRP chain-hardening set (tracker: vultisig-sdk#979). Independent of the other XRP branches (off main).

gastonm5:
Closes #4752

## Problem

XRP requests hit a single host (`xrplcluster.com` by default, or a custom override) with **no retry**. `xrplcluster.com` is a load-balanced **pool**; some backend nodes run outdated `rippled` and answer `amendmentBlocked` (error_code 14, returned as **HTTP 200** with `{"result":{"error":"amendmentBlocked",...}}`). One request lands on a stale backend and the error bubbles straight up — intermittently, "from time to time."

Observed on a `tx` status lookup after a send: the transaction had **already succeeded on-chain** (`tesSUCCESS`, validated), but the status check hit a bad backend and surfaced `amendmentBlocked`, so the app showed an error / unconfirmed state for a send that actually landed. The same pool-node issue can transiently affect broadcast, status, and account/fee/server_state reads.

## Fix — bounded SAME-HOST retry, no fallback list

**Retry the same resolved host on transient node errors — NO fallback host list.** Because the host is a pool, a retry hits a different backend, which is almost always healthy. Unlike Solana, we deliberately do **not** add a `publicFallbackHosts`-style list; this stays a bounded same-host retry.

- Detection reads the **decoded JSON-RPC body** (`result.error`), not the HTTP status — `amendmentBlocked` comes back as HTTP 200.
- Applied at the `RippleService` request layer via a shared `RippleRequestRetrier`, so **broadcast (`submit`), status (`tx`), and reads (`account_info`, `server_state`, fee)** all benefit.
- **Bounded:** 3 attempts total (1 try + up to 2 retries), linear backoff (0.25s, 0.5s). Added latency stays low in the time-sensitive post-signing window.
- No signing-path changes — request/transport resiliency only.

### Retryable set (from the decoded body's `result.error`)

`amendmentBlocked`, `noNetwork`, `noCurrent`, `noClosed`, `tooBusy`, `slowDown`, and `failedToForward`.

- `tooBusy` / `slowDown` are per-backend overload signals emitted by the node (not the pool edge), so a pool retry lands on a different node; bounded by the attempt cap + backoff, so including them is safe.
- `failedToForward` (Clio's forwarding-failure error on exactly the forwarded commands — `submit`/`server_state`/`fee`) was added on cross-model review: it's squarely a transient per-node fault the same-host pool retry routes around.

Transient **transport** failures are also retried: `HTTPError.timeout`, `HTTPError.networkError`, `HTTPError.statusCode(5xx)`. Task cancellation is **never** retried (even wrapped in `.networkError`).

**Non-retryable** JSON-RPC errors pass through UNCHANGED: `actNotFound`, `txnNotFound`, `invalidParams`, engine `tef*`/`tec*`/`tem*`/`ter*`/`tel*` codes, etc. They're real outcomes, not transient node faults. When the retryable set is exhausted, the last transient error is surfaced as `RippleRetryError.exhausted` rather than silently returned as data (which would e.g. make a persistently-`amendmentBlocked` balance read report `0`).

### Also (whole-branch review)

`RippleTransactionStatusAPI` hard-coded `xrplcluster.com`, so a custom Ripple RPC override sent submit/reads to the configured node but the status lookup (and its retry) to the default pool. The status API now takes the resolver-resolved host, so the "same **resolved** host" guarantee holds for status too (default `xrplcluster.com` when no override → behavior-preserving).

## Broadcast idempotency (safety)

Resubmitting the same signed blob is safe: XRPL `submit` is deterministic on the blob → same tx hash; an already-applied tx returns `tefALREADY`/`tesSUCCESS` (non-retryable, handled downstream). The retry re-sends identical bytes, never generates a new signature, and is bounded by the attempt cap — no double-spend. `amendmentBlocked` on `submit` means the blocked backend didn't apply the tx, so the retry is the first real apply. This **composes with, but is independent of,** the broadcast result-code classification work (#4746) — this PR is off `main`.

## Test coverage

New `RippleRetryTests` (24 tests, injected no-op clock — no real sleeps; sequenced `HTTPClientProtocol` stub):
- Policy: each retryable rpc code → retried; `actNotFound`/`txnNotFound`/`invalidParams`/`tef*` → not retried; transport timeout/networkError/5xx → retried; 4xx/decode/cancellation (incl. wrapped) → not; backoff monotonic + bounded.
- Retrier: retryable → retried → success; non-retryable surfaced without retry; exhaustion throws the **last** transient error (distinct errors queued to prove it) and is bounded to the attempt cap; transient transport retried; every retry stays on the **same host** (no fallback).
- Broadcast: `amendmentBlocked` → retried → returns the hash; exhausted → throws, no submit beyond the bound.
- Reads: `fetchServerState` / `fetchAccountsInfo` retry then succeed; exhausted `account_info` throws rather than returning an empty body.
- Status: `amendmentBlocked` → retried → confirmed; preserved behaviors pinned — HTTP 404 → notFound, `txnNotFound` → notFound, validated non-`tesSUCCESS` → failed, each without retrying.

## Gate

Full `make test` on a dedicated iPhone 17 Pro Max (iOS 26.5, en_US): **2389 tests, 0 failures — TEST SUCCEEDED.** SwiftLint clean on all changed files. Cross-model (Codex) review loop: per-commit + whole-branch passes, all findings triaged (fixed/justified) with a final clean pass.

## Not done

- On-device / two-device broadcast + status pass on XRP (simulate a stale-backend hit).

## Summary by CodeRabbit

* **New Features**
  * Added bounded same-host retries for Ripple JSON-RPC calls, with linear backoff and improved resilience for transient node/network issues.
  * Ripple transaction status and service requests now route through the shared retry mechanism.

* **Bug Fixes**
  * Enhanced decoding of node-level RPC errors from successful (HTTP-200) responses and surfaced them consistently across Ripple responses.
  * Improved status lookup parsing to read the correct result field and handle missing results as “not found”.

* **Tests**
  * Added retry policy/backoff coverage, retrier exhaustion behavior, cancellation non-retry rules, and end-to-end Ripple service/provider flow tests.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleAPI.swift
@@ -99,15 +99,23 @@ struct RippleSubmitResponse: Decodable {
         let engineResult: String?
         let engineResultMessage: String?
         let txJson: TxJson?
+        /// Node-level error (e.g. `amendmentBlocked`) returned in an HTTP-200
+        /// body when the backend can't process the request at all.
+        let error: String?
 
         enum CodingKeys: String, CodingKey {
             case engineResult = "engine_result"
             case engineResultMessage = "engine_result_message"
             case txJson = "tx_json"
+            case error
         }
 
         struct TxJson: Decodable {
             let hash: String?
         }
     }
+}
+
+extension RippleSubmitResponse: RippleRPCResponse {
+    var rpcError: String? { result?.error }
 }
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleRetryPolicy.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleRetryPolicy.swift
@@ -1,0 +1,167 @@
+//
+//  RippleRetryPolicy.swift
+//  VultisigApp
+//
+
+import Foundation
+import OSLog
+
+/// A decoded XRPL JSON-RPC response body that can surface a node-level error.
+///
+/// XRPL returns node/server errors (e.g. `amendmentBlocked`) as **HTTP 200**
+/// with the error nested in the body (`result.error`), so retry detection reads
+/// the decoded body rather than the HTTP status.
+protocol RippleRPCResponse {
+    /// The JSON-RPC error string the node returned, or `nil` on success.
+    var rpcError: String? { get }
+}
+
+/// Pure classifier for "should this XRPL request be retried on the same host?".
+///
+/// The default XRPL host (`xrplcluster.com`) is a load-balanced **pool**, so a
+/// retry against the same host is routed to a different backend — which is
+/// almost always healthy. That is why the fix is a bounded same-host retry and
+/// deliberately does NOT add a fallback host list.
+enum RippleRetryPolicy {
+    /// Total attempts for a request = one initial try plus up to
+    /// `maxAttempts - 1` retries. Kept small so the post-signing broadcast
+    /// window stays responsive.
+    static let maxAttempts = 3
+
+    /// JSON-RPC `result.error` values that indicate a transient/stale backend in
+    /// the pool. A same-host retry routes to a different (healthy) node.
+    ///
+    /// `tooBusy` / `slowDown` are per-backend overload signals emitted by the
+    /// node (not the pool edge), so a pool retry lands on a different node; they
+    /// stay bounded by `maxAttempts` and the backoff, so including them is safe.
+    static let retryableRPCErrors: Set<String> = [
+        "amendmentBlocked",
+        "noNetwork",
+        "noCurrent",
+        "noClosed",
+        "tooBusy",
+        "slowDown",
+        // Clio forwards write/server commands (submit, server_state, fee) to a
+        // rippled backend; a forwarding failure is a transient per-node fault a
+        // same-host pool retry can route around.
+        "failedToForward"
+    ]
+
+    /// Whether a decoded node error warrants a same-host retry. Business errors
+    /// (`actNotFound`, `txnNotFound`, `invalidParams`, engine `tef*`/`tec*`
+    /// codes, …) are not transient and pass through unchanged.
+    static func isRetryable(rpcError: String?) -> Bool {
+        guard let rpcError else { return false }
+        return retryableRPCErrors.contains(rpcError)
+    }
+
+    /// Whether a thrown transport error is transient and worth a retry. A
+    /// cancellation is never retried, even when wrapped in `.networkError`.
+    static func isRetryable(transportError error: Error) -> Bool {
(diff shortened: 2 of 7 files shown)


johnnyluo (approved):
LGTM

PR #4763, opened 2026-07-07
feat(reshare): redesign the reshare flow to the new Figma (pre-flight warning, success, backup guide)

### LINKED ISSUE #4739: Update reshare flow to new Figma redesign (pre-flight warning + device-count + success + backup guide)
## Context
From Figma review 2026-07-05. The reshare flow is still the pre-redesign single "Reshare your vault" screen (`Features/Reshare/Views/ReshareView.swift`, ~3 files). Verified: three net-new screens from the redesign do not exist in code.

## What (new Figma reshare flow)
1. Redesigned **Start Reshare / Join Reshare** entry cards with new subtitles ("You are changing this vault's device setup." / "You received a reshare request from another device.")
2. **"Before you reshare"** pre-flight warning — "Resharing regenerates your vault shares…", with: *Old backups will stop working* (existing .vult files become invalid) and *All co-signers must be online*.
3. **"How many devices do you have?"** device-count selection.
4. **"Vault reshared successfully"** success screen.
5. Post-reshare **backup guide** ("Backups, your new recovery method" / "Back up each device" / "Store backups separately" / "Old backups won't work" / "Backup in the cloud"). Note: this copy exists in the repo but is currently wired to onboarding/import, not reshare.

Already matching (shared keygen components, no work needed): QR pairing + local mode.

## Acceptance Criteria
- [ ] All five new/updated screens match the Figma
- [ ] The "Before you reshare" warning surfaces the invalidated-backup + all-co-signers-online requirements before a reshare starts
- [ ] Post-reshare backup guide runs as a reshare step (not just onboarding)

## Figma
https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=65858-81261

gastonm5:
Redesigns the reshare flow to the new Figma — replacing the single pre-redesign "Reshare your vault" screen.

## Flow
1. **Entry redesign** — Start / Join Reshare option cards over the `review_devices` animation, with the new subtitles.
2. **Before you reshare** — pre-flight warning sheet with three cards: *New backups will be created*, *Old backups will only work together*, *All co-signers must be online* — gating Start Reshare behind "I understand".
3. **Peer discovery** — Start Reshare goes **straight to peer discovery**; there is no device-count selection step (a reshare's committee is decided by who joins pairing, not chosen up front). Server-backed (fast) vaults collect their FastVault password first so the server is registered and joins the session.
4. **Success** — "Vault reshared" copy on the shared keygen success state.
5. **Backup guide** — the backup-guide content surfaced as a reshare step (extracted from onboarding for reuse), incl. the "Old backups won't work" row.

## Notes
- The device-count / "how many devices" selection step (and its Rive animation) was intentionally **dropped** for reshare — see the flow above. The shared `DevicesSelectionView` component remains (still used by onboarding).
- `BeforeYouReshareBottomSheet` and the backup-guide copy match the updated Figma (node `65858-81261`); new `file-tree` icon added. Strings updated across all 8 locales.
- No TSS/keygen signing logic touched; `PeerDiscoveryScreen` / local-mode unchanged.
- Cross-platform parity note: Android (#5222) and Windows (#4298) shipped the redesign **with** a device-count step; iOS intentionally drops it (aligns with the newer Figma, which replaces the slider with a "Manage your devices" screen). Threshold is still enforced at pairing, so this is a UX simplification, not a safety change.
- Gate: `make test` green (2441 tests, 0 failures), SwiftLint clean, cross-model (codex) review clean.
- **On-device verified** — 2-device reshare walk-through passed (entry → 3-card "Before you reshare" → peer discovery → success → backup guide).

Closes #4739


VultisigApp/VultisigApp/Features/Reshare/Views/ReshareDevicesSelectionScreen.swift
```diff
@@ -0,0 +1,105 @@
...
+            tipText: "seedPhraseImportTip".localized,
+            buttonTitle: "getStarted".localized,
+            isLoading: viewModel.isLoading,
+            isButtonDisabled: !viewModel.isThresholdMet || viewModel.isLoading,
+            onContinue: onContinue
+        ) {
+            if !viewModel.isThresholdMet {
+                thresholdWarningCard
+            }
+        }
```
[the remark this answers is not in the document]
gastonm5: Addressed the CTA: this reshare device-count step is mid-flow (the user has already tapped **Start Reshare** and confirmed the *Before you reshare* sheet), so the primary button now uses the existing `continue` ("Continue") string instead of the onboarding `getStarted` copy.

I left the tip text unchanged on purpose: `DevicesSelectionView` is the selector shared by both onboarding and reshare, and the displayed tip ("You can use a browser as a device") is generic device-selection guidance that applies equally to reshare — only the underlying key *name* (`seedPhraseImportTip`) is onboarding-flavored. Renaming that shared key would touch the onboarding screen plus all 7 `Localizable.strings` files for identical copy, which is disproportionate to a minor nit; happy to do that key rename as a follow-up if preferred.

Fixed in e44090620.

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareScreen.swift
```diff
@@ -0,0 +1,169 @@
...
+        EllipticalGradient(
+            stops: [
+                Gradient.Stop(color: Color(hex: "084BFF").opacity(0.5), location: 0.00),
+                Gradient.Stop(color: Color(hex: "084BFF").opacity(0), location: 1.00)
+            ],
+            center: UnitPoint(x: 0.5, y: 0.5)
+        )
+        .frame(width: 360, height: 360)
+        .blur(radius: 36)
+    }
```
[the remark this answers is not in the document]
gastonm5: Good catch on the hardcoded color. I did not apply the suggested `Theme.colors.primaryAccent4` because that token resolves to `4879FD`, a visibly different blue than the `084BFF` glow — using it would be a color regression. Instead I added a `Theme.colors.devicesSelectionGlow` token (= `084BFF`) to the design system and used it in `ReshareScreen.glow` and in the shared `DevicesSelectionBackground`, which used the same raw hex. The exact glow color is preserved and the magic value is now gone from both call sites.

Fixed in e44090620.

johnnyluo (commented):
A few points on the reshare model and copy — see inline comments. The two structural ones (device-count step and the single-device path) are worth resolving together, since reshare device count is really determined by which existing devices join rather than by an up-front selection.

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareDevicesSelectionScreen.swift
```diff
@@ -0,0 +1,105 @@
...
+//
+//  Created by Gaston Mazzeo on 06/07/2026.
+//
+
+import SwiftUI
+
+/// Reshare step asking how many devices the new vault setup will use.
+/// Reuses the shared devices selector and gates selections that would drop
+/// below the vault's required number of active signers.
+struct ReshareDevicesSelectionScreen: View {
```
johnnyluo: I don't think we should be asking the user how many devices to use during a reshare. Reshare works by having 2/3 of the existing devices join, and every device that joins becomes part of the new committee — so the resulting device count is determined by who participates, not by a target chosen up front. Presenting a device-count selector here misrepresents how reshare actually works.

Anyone reaching for reshare should already understand this model; if they don't, reshare isn't the right tool for them. Can we drop the device-count step and keep the flow driven by which devices join the session?

VultisigApp/VultisigApp/Features/Reshare/ViewModels/ReshareDevicesSelectionViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+        selectedIndex + 1
+    }
+
+    var isThresholdMet: Bool {
+        selectedDeviceCount >= requiredActiveSigners
+    }
+
+    var destination: Destination {
+        guard selectedIndex > 0 else {
+            return .fastVaultPassword(isExistingVault: isFastVaultEligible)
```
johnnyluo: Resharing to a single device isn't a supported outcome — a reshare can't produce a 1-of-1 (FastVault) vault. This `selectedIndex == 0` branch routes a "1 device" selection into the FastVault password flow, which shouldn't be reachable from reshare. If the device-count step stays, the single-device option should be removed so it can't be selected.

VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
```diff
@@ -753,6 +758,10 @@
 "of" = "of";
 "off" = "Off";
 "ok" = "OK";
+"oldBackupsStopWorkingDescription" = "Any existing .vult files become invalid. You'll need to create fresh backups after resharing.";
+"oldBackupsStopWorkingTitle" = "Old backups will stop working";
```
johnnyluo: "Old backups will stop working" isn't accurate. Old backups keep working — they remain valid as a set among themselves. What actually changes is that old and new backups can't be mixed: an old backup won't interoperate with a backup created after the reshare.

Let's reword this (and the related `oldBackupsWontWork` key on the next line, plus their descriptions) to reflect that — e.g. framing it as "old and new backups can't be combined" rather than implying the old backups become unusable. This also needs updating across all locales.

VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
```diff
@@ -693,6 +698,8 @@
 "networkFee" = "Network Fee";
 "new" = "NEW";
 "newAddress" = "New Address";
+"newBackupsCreatedDescription" = "Each device will create a new .vult backup which is different to the old backups";
```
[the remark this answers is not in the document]
gastonm5: Good catch — updated to "…a new .vult backup that differs from the old backups" in `en` (the other locales already express "differs from"). Fixed in 8e033e417.

johnnyluo (commented):
Inline review findings from the reshare flow audit.

VultisigApp/VultisigApp/Features/Onboarding/ImportVault/KeyImport/OnboardingOverviewScreen.swift
```diff
@@ -173,6 +165,14 @@ struct OnboardingOverviewScreen: View {
                 subtitle: row2Subtitle,
                 icon: "arrow-split"
             )
+
+            if content.showsOldBackupsRow {
```
johnnyluo, line 169: **[P1] Incorrect old-backup guidance**

Resharing preserves the vault public key, so a threshold set of old-generation backups remains usable together. Saying these backups cannot be used contradicts the pre-flight warning and may encourage users to discard valid recovery material. Please explain that old and new generations cannot be mixed, but each complete threshold set remains valid, and update the localized copy consistently.

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareScreen.swift
```diff
@@ -0,0 +1,207 @@
...
+    private func onStartReshare() {
+        showBeforeYouReshareSheet = true
+    }
+
+    private func onStartReshareConfirmed() {
+        // A server-backed vault must collect its password first so the server
+        // is registered and joins the session; otherwise peer discovery would
+        // wait forever for a device that never connects. Every other vault
+        // goes straight to peer discovery.
+        if vault.hasServerSigner {
```
johnnyluo, line 132: **[P1] Route using canonical FastVault eligibility**

`hasServerSigner` is only a structural helper and does not verify that the backend vault exists. A restored or stale vault containing a `server-*` signer can therefore be forced into password validation that can never succeed. Please await canonical FastVault eligibility and cover backend-present, missing, and network-error states.

VultisigApp/VultisigApp/Features/Reshare/Views/BeforeYouReshareBottomSheet.swift
```diff
@@ -0,0 +1,106 @@
...
+
+/// Pre-flight warning shown before starting a reshare: old backups become
+/// invalid and every current co-signer must be online for the session.
+struct BeforeYouReshareBottomSheet: View, BottomSheetProperties {
+    let onContinue: () -> Void
+
+    var bgColor: Color? { Theme.colors.bgPrimary }
+
+    var body: some View {
+        VStack(spacing: 10) {
```
johnnyluo, line 18: **[P2] Bottom sheet can clip its only action**

This non-scrollable stack contains a header, three expanding cards, and the only continue button. With Accessibility Dynamic Type or limited screen height, the intrinsic-height detent can exceed the available area and clip the action. Please put the expandable content in a `ScrollView` while keeping the primary action reachable.

VultisigApp/VultisigApp/Features/Onboarding/ImportVault/KeyImport/OnboardingOverviewScreen.swift
johnnyluo: I think this one should be fine

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareScreen.swift
johnnyluo: going to close this one , because it should be fine , we are not going to allow fastvault to reshare it

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareScreen.swift
gastonm5: Fixed in 263e3eb20. `onStartReshareConfirmed` no longer routes on the structural `vault.hasServerSigner`; it now awaits the canonical eligibility check `FastVaultService.isEligibleForFastSign(vault:)` (= `hasServerSigner && exist(pubKeyECDSA:)`, the same probe the eligibility refresher uses) and routes to the FastVault password screen **only when the backend vault is confirmed present** (a 2xx from `/vault/exist`). Every other outcome falls back to peer discovery instead of forcing password entry — so a restored/stale vault carrying a `server-*` signer with no backend vault now goes to peer discovery and can recover, rather than dead-ending on a password that can never validate.

On the three states (present / missing / network-error): I looked for a canonical async API that distinguishes them and there isn't one that's reliable. The backend's `ExistVault` handler (vultiserver `api/server.go:647-658`) returns **400 for both "vault absent" and a storage-lookup error** (there's no 404), and rate-limiting can return 429 — so "missing" can't be told apart from a transient backend/network failure from the client. Given that, I implemented the safest behavior you suggested: route to the password flow only on confirmed presence, and treat everything unconfirmed (missing / throttled / storage error / unreachable) as "don't force password" → peer discovery. The residual case (a genuine FastVault that hits a transient error at the exact moment of tapping Start) lands on peer discovery and is recoverable by retrying once the backend is healthy; forcing the password screen there would risk re-creating the dead-end for genuinely-missing vaults, which is the bug this addresses.

Also added a re-entry latch + a loading overlay around the probe, and extracted a pure `ReshareScreen.startReshareRoute(isBackendConfirmedPresent:)` with unit tests. Happy to add an explicit tri-state + retry UX if the backend gains a distinct not-found status.

VultisigApp/VultisigApp/Features/Reshare/Views/BeforeYouReshareBottomSheet.swift
gastonm5: Fixed in 263e3eb20. The header + three warning cards are now wrapped in a `ScrollView` with the primary "I understand" button pinned outside it, so the action stays reachable under large Accessibility Dynamic Type / on short screens.

One wrinkle worth noting: this bottom sheet auto-sizes its detent from a one-shot measurement of the content's natural height (`BottomSheetModifier` + `InnerHeightPreferenceKey`, which locks on the first non-zero height). A naive greedy `ScrollView` would get measured at the initial `.fraction(0.3)` detent and lock the sheet at ~30% height. So the content renders unscrolled on the first pass (letting the sheet measure and size the detent correctly), then swaps to `ScrollView { … }.frame(maxHeight: measuredContentHeight)`. Result: normal sizing is unchanged (the sheet still hugs its content, no scroll), and only when the content exceeds the available height do the cards scroll while the button stays pinned.

VultisigApp/VultisigApp/Features/Reshare/Views/ReshareScreen.swift
gastonm5: Understood — thanks for the note that FastVault reshare won't be allowed; in that light the original `hasServerSigner` routing is fine and our messages crossed (I'd already pushed 263e3eb20). My change is a strictly-safer interim version: it only routes to the password screen on a confirmed `/vault/exist` 2xx and otherwise falls back to peer discovery, so the stale/restored-vault dead-end can't happen while this path is still reachable — and it doesn't get in the way of hard-blocking FastVault reshare later.

Happy to revert the #2 routing change (keeping just the copy and scroll fixes) if you'd rather ship the simpler original and close this. Your call — leaving the thread for you to resolve.

PR #4764, opened 2026-07-07
feat(ripple): inline send-form validation of XRP amount vs destination base reserve

### LINKED ISSUE #4751: feat(ripple): inline send-form validation of XRP amount vs destination base reserve (unfunded accounts)
## What
When sending XRP to a **new / unfunded** destination, the send **form** (amount entry) does not warn that the amount must be at least the network **base reserve** (currently 1 XRP). A user can enter e.g. 0.1 XRP and only discover the problem later — or, on builds without the activation guard, not at all (the on-chain result would be `tecNO_DST_INSUF_XRP`, which burns the fee).

The reserve/activation guard added in #4744 (PR #4748) catches this, but only at the **Verify screen** (pre-signing). This asks for the same check **inline on the send form**, for immediate feedback while typing the amount.

## Rule
- Destination is an **unfunded** XRP account (`account_info` → `actNotFound`) **and** amount `< base reserve` → show an inline form error (e.g. "New XRP accounts need at least N XRP to activate").
- Destination is **funded** → no minimum (any amount valid).
- Native XRP only; reserve value comes **live from `server_state`** (owner-aware machinery already in #4744), never hardcoded.

## Reuse (do not duplicate)
Build on #4744's `RippleReserve` + the destination `account_info` lookup used by `validateDestinationActivation`; expose a non-throwing, form-friendly variant that returns the minimum for inline display rather than throwing at Verify. Keep the Verify-screen guard as the last-line safety net.

## Acceptance
- [ ] Entering an amount `< base reserve` for an **unfunded** XRP destination shows an inline form error naming the minimum
- [ ] A **funded** destination shows no such error for any amount
- [ ] The minimum reflects the live `server_state` base reserve, not a constant
- [ ] Continue is blocked (or the error clears) consistently with the Verify-screen guard — no divergence between the two
- [ ] Unit test: unfunded + below-reserve → error; funded → no error; unfunded + ≥ reserve → no error

Stacks on #4744 (PR #4748) to reuse its reserve logic. Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

gastonm5:
Inline send-**form** validation of the XRP amount against the destination's base reserve, for **unfunded** destinations — so the user gets immediate feedback while typing the amount instead of only discovering it at the Verify screen.

## What
- New non-throwing `RippleService.destinationReserveShortfall(address:amountDrops:forceRefresh:)` returning `RippleReserveCheck` (`.satisfied` / `.belowMinimum(minimumXRP:)` / `.unknown`). It **reuses** #4748's `account_info` lookup + `RippleReserve.reservedDrops(ownerCount: 0, …)`, so the form's minimum can **never diverge** from the Verify guard's (both now share the `minimumActivationXRP` helper).
- Funded/unfunded verdict is **cached per destination** (host-scoped), so typing the amount doesn't re-hit the node on every keystroke.
- **Fails open on the form**: any lookup it can't complete shows no inline error — the #4748 Verify-screen guard stays the fail-closed backstop. Only a definitive `actNotFound` yields `.belowMinimum`.
- Wired into `SendDetailsViewModel` (while-typing advisory + a forced-live re-check at Continue so it can't act on a stale verdict).

## Notes
- **Stacked on #4748** (base `fix/xrp-owner-reserve-4744`) to reuse its reserve/lookup code; merges after #4748, then rebases to `main`.
- This branch merges the latest #4748 (which now includes the broadcast/retry work + johnnyluo's review fixes). The shared `RippleService` code was reconciled: `fetchAccountsInfo` routes through the retrier **with** host-pinning (transient-error resilience + the funding-cache key), and the min-XRP computation is unified in `minimumActivationXRP` using `toDecimal` (preserving johnnyluo's finding #5).
- Gate: **2481 tests, 0 failures**.

Closes #4751

## Summary by CodeRabbit

* **New Features**
  * Added inline XRP destination reserve warnings during send flow validation.
  * Improved amount screen feedback so reserve issues appear directly in the form.

* **Bug Fixes**
  * Updated XRP destination checks to use consistent minimum activation thresholds.
  * Reduced repeated network checks while typing by reusing recent results.
  * Kept validation responsive by ignoring stale lookup results and clearing warnings when inputs change or reset.

* **Tests**
  * Added coverage for XRP reserve validation, warning behavior, caching, and fail-open handling.


johnnyluo (approved):
LGTM

PR #4765, opened 2026-07-08
feat(send): generic SendAmountValidator + animate inline warning + gate Continue

gastonm5:
## Description

Follow-up polish + refactor on top of the inline XRP base-reserve check shipped in #4764.

1. **Animate the inline error/warning.** The amount-field error and the destination base-reserve warning now appear/disappear with the shared `verticalGrowAndFade` transition (driven off the value via `.animation(.interpolatingSpring, value:)`), matching the rest of the app instead of popping in.
2. **Disable Continue while a blocking warning is shown.** `continueButtonDisabled` now consults the validation state, so a below-reserve XRP send greys out Continue the moment the warning appears (and re-enables when the amount/destination is fixed).
3. **Extract the check behind a chain-agnostic protocol.** The XRP-specific logic (and its `RippleService` dependency + copy) moved out of `SendDetailsViewModel` into a generic `SendAmountValidator` protocol, implemented by `RippleDestinationReserveValidator` under `Blockchain/Ripple/`. The VM holds an injected `[any SendAmountValidator]` and a general `amountValidation` state (`message` + `blocksContinue`), so adding another chain is a new validator rather than another branch in the VM.

Design:
- `SendAmountValidator` — `isApplicable(to:)` (pure/sync) + `validate(_:forceRefresh:)` (async) → `.ok` / `.invalid(message:blocksContinue:)`.
- `SendAmountValidationInput` — value-type snapshot (no `@Model Coin` crosses the async boundary).
- The VM keeps the generic orchestration only: debounce, cancellation, stale-input guard, first-applicable-wins, publish-if-not-moved-on.

No new issue — this is a quality/UX follow-up to #4764.

## Which feature is affected?
- [x] Sending — native XRP send to an unfunded destination below the base reserve (no on-chain tx needed; the warning + Continue gate are pre-flight)

## Checklist

_Animation is the shared `verticalGrowAndFade` transition; behavior verified via unit tests._

## Additional context

- Unifies the while-typing and Continue-time preconditions: both now require a **format-valid** destination (at Continue the address is already resolved, so this is strictly more correct and matches the while-typing path).
- Verification: SwiftLint clean on all touched files; full unit suite green — **2486 tests, 0 failures**. Added `SendAmountValidatorTests` proving the generic wiring blocks Continue independent of any specific chain.

- **Agent**: Claude Code (Opus 4.8)
- **Issue**: follow-up to #4764
- **Confidence**: high
- **Needs human review for**: SwiftUI transition feel on device; the tightened Continue-time address-format precondition

## Summary by CodeRabbit

* **New Features**
  * Introduced a chain-agnostic send amount validation pipeline that drives inline warnings and Continue gating.
  * Added XRP destination reserve validation for native XRP sends where the recipient reserve may not yet be satisfied.

* **Bug Fixes**
  * Improved validation refresh/clearing when changing coins, updating recipient address, typing amounts, or resetting the form.
  * Continue button now correctly blocks only when a validation verdict requires it.

* **Tests**
  * Added validator behavior tests (blocking vs. non-blocking, inapplicable validators, force-refresh, and first-objection short-circuit).
  * Updated existing send flow tests to assert the new amount-validation state.


johnnyluo (approved):
LGTM

PR #4784, opened 2026-07-09
[Fix] Harden BlockChainSpecific decoding of untrusted keysign payload numeric fields

gastonm5:
## Summary

`BlockChainSpecific(from:)` decoded chain-specific numeric string fields from an inbound keysign payload with the **trapping** `BigInt(stringLiteral:)` initializer, which crashes on any non-numeric or empty string. A keysign payload is untrusted data deserialized from a co-signer at the start of a ceremony, so a buggy or malicious peer sending a garbled `compute_limit`/`priority_fee` (Solana), `max_fee_per_gas_wei`/`priority_fee`/`gas_limit` (EVM) or `current_block_number` (Polkadot) could hard-crash the **receiving** device while decoding the payload.

This replaces the trapping initializer with the failable `BigInt(_:) ?? 0` for the peer-supplied numeric fields, mirroring the existing Sui precedent in the same function. Android shipped the equivalent fix in [vultisig-android#5244](https://github.com/vultisig/vultisig-android/pull/5244).

`closes #4782`

## Why `?? 0` is safe

- **Solana:** the downstream signer substitutes defaults for non-positive values — `SolanaHelper.getPreSignedInputData` uses `priorityFee > 0 ? … : defaultPriorityFeePrice` and `priorityLimit > 0 ? … : priorityFeeLimit` (= 100_000). So `0` == "use default" == absent field, matching Android's "fall back to default."
- **EVM / Polkadot:** pure crash-hardening. If a peer sends a garbled numeric field, decoding to `0` makes this device build a different tx than the initiator, so the MPC ceremony **fails safe** on a signature mismatch (no fund loss) instead of crashing the app. The intent here is crash-avoidance, not semantic recovery.

Every well-formed value — including an explicit `"0"` — still maps through **verbatim**, so the signed transaction is unchanged for valid payloads. The signing path, default constants, and amount fields (`toAmount`/`amount`/`fromAmount`) are untouched.

## Changes

- `Features/Keysign/Services/KeysignMessage+ProtoMappable.swift` — Solana (`priorityFee`, `priorityLimit`), EVM (`maxFeePerGasWei`, `priorityFeeWei`, `gasLimit`), and Polkadot (`currentBlockNumber`) cases now parse with the failable `BigInt(_:) ?? 0`.
- New `VultisigAppTests/Keysign/KeysignPayloadMalformedChainSpecificTests.swift`.

## Test coverage

13 focused unit tests decoding a proto directly through `BlockChainSpecific(proto:)`:
[...]

PR #4785, opened 2026-07-10
fix(keysign): harden BlockChainSpecific numeric decoding against malformed peer payloads

### LINKED ISSUE #4782: [Fix] Malformed Solana keysign compute_limit/priority_fee crashes the co-signer [Keysign]
# [Fix] Malformed Solana keysign compute_limit/priority_fee crashes the co-signer [Keysign]

## Problem
When decoding an inbound keysign payload from a peer, `BlockChainSpecific`'s Solana case parses the `priority_fee` and `compute_limit` fields with the **trapping** `BigInt(stringLiteral:)` initializer (`Features/Keysign/Services/KeysignMessage+ProtoMappable.swift:455-456`). That initializer crashes on any non-numeric or empty string. Because a keysign payload is untrusted data deserialized from a co-signer at the start of a ceremony, a buggy or malicious peer sending a garbled `compute_limit`/`priority_fee` hard-crashes the **receiving** device. The app's own code already documents this hazard for the Sui case (`:485` — "`BigInt(stringLiteral:)` traps on ''") and parses defensively there.

Android shipped the equivalent fix in [vultisig-android#5244](https://github.com/vultisig/vultisig-android/pull/5244) (falls back to the default on malformed `compute_limit`). iOS is more exposed: `priority_fee` is parsed unconditionally with the trapping initializer, and the sibling EVM (`maxFeePerGasWei`/`priorityFee`/`gasLimit`, `:424-427`) and Polkadot (`currentBlockNumber`, `:465`) chain-specific numeric fields share the identical crash class.

## Expected Behavior
A malformed/empty numeric field in an inbound keysign payload falls back to a safe default (mirroring an absent field) instead of crashing. Every valid value — including an explicit `"0"` — still maps through verbatim, so the fix does not change the signed transaction for well-formed payloads.

## Steps to Reproduce
1. Act as co-signer; receive a keysign payload whose `solanaSpecific.compute_limit` (or `priority_fee`) is `"abc"` or `""`.
2. The receiving device crashes with a `BigInt` precondition failure while decoding `BlockChainSpecific`.

## Solution
Replace the trapping `BigInt(stringLiteral:)` with the failable `BigInt(_:) ?? <fallback>` for the peer-supplied numeric fields in `BlockChainSpecific(from:)`, mirroring the existing Sui precedent (`:487`,`:489`). The downstream Solana signer already substitutes defaults for non-positive values (`Blockchain/Solana/Signing/Solana.swift:32,35` → `defaultPriorityFeePrice` / `priorityFeeLimit` = 100_000), so a `0` fallback yields the same behavior as an absent field — matching Android's "fall back to default." Harden the sibling EVM and Polkadot chain-specific numeric fields in the same pass (same function, same crash class). Keep it a pure decoding-hardening change — no signing-path or default-value changes.

### Must Do
- [ ] `KeysignMessage+ProtoMappable.swift` Solana case: `priorityFee: BigInt(value.priorityFee) ?? 0` and `priorityLimit: value.hasComputeLimit ? (BigInt(value.computeLimit) ?? 0) : BigInt(0)`.
- [ ] Same defensive parse for the EVM (`maxFeePerGasWei`, `priorityFee`, `gasLimit`) and Polkadot (`currentBlockNumber`) chain-specific fields decoded from the peer proto.
- [ ] Unit test (camelCase) covering malformed / empty / absent / well-formed `compute_limit` and `priority_fee`, asserting no crash and correct fallback, and that a valid explicit `"0"` still maps to `0`.

### Must NOT Do
- Don't change unrelated code
- Don't alter the signing path, default constants, or the amount fields' behavior beyond crash-hardening

## Acceptance Criteria
- [ ] Build succeeds
- [ ] Malformed/empty `compute_limit`/`priority_fee` in an inbound payload no longer crashes; falls back to the signer's default
- [ ] Well-formed values (including `"0"`) map through unchanged — no change to the signed tx

johnnyluo:
## Summary
- A keysign payload is untrusted data deserialized from a co-signer. `BlockChainSpecific(proto:)` parsed several numeric fields with the **trapping** `BigInt(stringLiteral:)`, which crashes on any non-numeric/empty string — so a buggy or malicious peer could hard-crash the receiving device at the start of a ceremony.
- Swapped to the failable `BigInt(_:) ?? 0` for the peer-supplied fields: Solana `priority_fee`/`compute_limit`, EVM `maxFeePerGasWei`/`priorityFee`/`gasLimit`, Polkadot `currentBlockNumber`. Mirrors the existing Sui precedent and [Android's fix](https://github.com/vultisig/vultisig-android/pull/5244).
- The `0` fallback matches the signer's default for non-positive values (`defaultPriorityFeePrice`/`priorityFeeLimit`), so **well-formed payloads — including an explicit `"0"` — sign identically**. Pure decode-hardening; signing path, defaults, and amount fields untouched.

## Issue
Closes #4782

## Test Plan
- [x] New `BlockChainSpecificDecodingHardeningTests` — 8 cases (malformed / empty / absent / well-formed incl. explicit `"0"`) across Solana, EVM, Polkadot; assert no crash + correct fallback. All pass.
- [x] SwiftLint passes
- [x] Build succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or empty blockchain transaction values during key signing.
  * Invalid numeric fields now safely default to zero instead of causing decoding failures.
  * Preserved support for valid large numeric values and explicit zero values across Ethereum, Solana, and Polkadot.

* **Tests**
  * Added coverage for malformed, missing, zero, and valid blockchain-specific values.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
@@ -421,10 +421,10 @@ extension BlockChainSpecific {
             )
         case .ethereumSpecific(let value):
             self = .Ethereum(
-                maxFeePerGasWei: BigInt(stringLiteral: value.maxFeePerGasWei),
-                priorityFeeWei: BigInt(stringLiteral: value.priorityFee),
+                maxFeePerGasWei: BigInt(value.maxFeePerGasWei) ?? 0,
+                priorityFeeWei: BigInt(value.priorityFee) ?? 0,
                 nonce: value.nonce,
-                gasLimit: BigInt(stringLiteral: value.gasLimit)
+                gasLimit: BigInt(value.gasLimit) ?? 0
             )
         case .thorchainSpecific(let value):
             self = .THORChain(
@@ -452,8 +452,8 @@ extension BlockChainSpecific {
         case .solanaSpecific(let value):
             self = .Solana(
                 recentBlockHash: value.recentBlockHash,
-                priorityFee: BigInt(stringLiteral: value.priorityFee),
-                priorityLimit: value.hasComputeLimit ? BigInt(stringLiteral: value.computeLimit) : BigInt(0),
+                priorityFee: BigInt(value.priorityFee) ?? 0,
+                priorityLimit: value.hasComputeLimit ? (BigInt(value.computeLimit) ?? 0) : BigInt(0),
                 fromAddressPubKey: value.fromTokenAssociatedAddress.isEmpty ? nil : value.fromTokenAssociatedAddress,
                 toAddressPubKey: value.toTokenAssociatedAddress.isEmpty ? nil : value.toTokenAssociatedAddress,
                 hasProgramId: value.programID
@@ -462,7 +462,7 @@ extension BlockChainSpecific {
             self = .Polkadot(
                 recentBlockHash: value.recentBlockHash,
                 nonce: value.nonce,
-                currentBlockNumber: BigInt(stringLiteral: value.currentBlockNumber),
+                currentBlockNumber: BigInt(value.currentBlockNumber) ?? 0,
                 specVersion: value.specVersion,
                 transactionVersion: value.transactionVersion,
                 genesisHash: value.genesisHash,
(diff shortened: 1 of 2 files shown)


johnnyluo:
looks like it is duplicated of https://github.com/vultisig/vultisig-ios/pull/4784 , let me close it

PR #4791, opened 2026-07-10
[Fix] Solana SPL ATA: fall back to derivation on a transient token-account lookup error

### LINKED ISSUE #4783: [Fix] Solana SPL token-account lookup fails the send on transient RPC error instead of deriving the ATA [Send/Solana]
# [Fix] Solana SPL token-account lookup fails the send on a transient RPC error instead of deriving the ATA [Send/Solana]

## Problem
`SolanaService.fetchTokenAssociatedAccountByOwner` (`Blockchain/Solana/Service/SolanaService.swift:297`) already falls back to deriving the classic-SPL and Token-2022 associated-token addresses (and confirming them via `getAccountInfo`) when `getTokenAccountsByOwner` returns an **empty** result — this covers the common RPC indexer-lag case right after an ATA is created. But when the primary `getTokenAccountsByOwner` call **throws** (a transient RPC/node error rather than an empty index), the error propagates from the `try await` at `:299` and the derivation fallback is skipped, so the send fails even though the ATA is derivable/exists.

Android's [vultisig-android#5245](https://github.com/vultisig/vultisig-android/pull/5245) covers both the empty-result and the primary-call-error cases. This brings iOS to parity on the error path.

## Expected Behavior
If the primary `getTokenAccountsByOwner` lookup errors, fall through to the same deterministic ATA derivation + `getAccountInfo` existence probe that the empty-result path already uses, before concluding the account is missing. A successful primary lookup behaves exactly as today.

## Steps to Reproduce
1. Send an SPL token while the Solana RPC intermittently errors on `getTokenAccountsByOwner` (transient node error, not an empty result).
2. The send fails with the RPC error instead of recovering via ATA derivation.

## Solution
Wrap the primary `getTokenAccountsByOwner` call so a thrown error is treated like an empty result — i.e. flow into the existing classic-SPL / Token-2022 derivation + `getAccountInfo` probe — rather than propagating immediately. Preserve current behavior on success and keep the final "not found" result (`("", false)`) when neither derived ATA exists. Do not swallow the error silently: log it (OSLog `Logger.warning`) before falling back, per the repo error-handling rules. Reuse the existing `checkAccountExists`/derivation helpers — no new lookup logic.

### Must Do
- [ ] In `fetchTokenAssociatedAccountByOwner`, catch a thrown error from `getTokenAccountsByOwner` and continue into the existing ATA-derivation probe (same as the empty branch).
- [ ] Log the caught error via OSLog before falling back.
- [ ] Unit test (camelCase) asserting that when the primary lookup throws, a derivable/existing ATA is still returned, and that a genuinely-missing account still yields the not-found result.

### Must NOT Do
- Don't change unrelated code
- Don't refactor the surrounding Solana service or signing logic
- Don't change the success path or caching behavior

## Acceptance Criteria
- [ ] Build succeeds
- [ ] A transient `getTokenAccountsByOwner` error no longer fails the send when the ATA is derivable/exists
- [ ] Successful lookups and the genuine "account missing" result are unchanged

gastonm5:
## Summary
`SolanaService.fetchTokenAssociatedAccountByOwner` already derives the classic-SPL and Token-2022 ATAs and confirms them via `getAccountInfo` when `getTokenAccountsByOwner` returns an **empty** result (the common indexer-lag case right after an ATA is created). But when the primary `getTokenAccountsByOwner` call **threw** (a transient RPC/node error rather than an empty index), the error propagated and the derivation fallback was skipped — failing the send even though the ATA is deterministically derivable/exists.

This wraps the primary lookup in `do/catch` so a thrown error flows into the **same existing** ATA-derivation + `checkAccountExists` probe as the empty branch, logging the error via OSLog first (never swallowed). The success path and caching are unchanged; the final not-found result `("", false)` is kept when neither derived ATA exists.

Brings iOS to parity with Android's error-path handling ([vultisig-android#5245](https://github.com/vultisig/vultisig-android/pull/5245)).

`closes #4783`

## Changes
- `Blockchain/Solana/Service/SolanaService.swift` (+19/−5) — `do/catch` around the primary lookup; OSLog warning on the error path before falling through to the existing derivation probe.
- `VultisigAppTests/Blockchain/Solana/SolanaServiceAtaFallbackTests.swift` (new, 4 tests).

## Tests
New `SolanaServiceAtaFallbackTests` (all pass):
- `testSuccessfulPrimaryLookupReturnsAccountWithoutProbing` — success path unchanged.
- `testThrowingPrimaryLookupReturnsDerivedExistingAta` — a thrown lookup still returns the derivable classic-SPL ATA.
- `testThrowingPrimaryLookupReturnsDerivedToken2022Ata` — same for a Token-2022 ATA.
- `testThrowingPrimaryLookupWithNoExistingAtaReturnsNotFound` — a genuinely-missing account still yields not-found.

## Gate
- Full suite (`make test`, iPhone 16 Plus, sim pinned to en_US): **`** TEST SUCCEEDED **` — Executed 2660 tests, 1 skipped, 0 failures.**
- SwiftLint: clean on the changed files (0 new warnings).
- Codex read-only review (whole-branch, single commit): **0 findings** — confirmed the success path and caching are unchanged and the error is logged, not swallowed.

## Notes
Out of scope: no change to the empty-result path, the success path, or caching. This only adds the error → derivation fallback (previously the error propagated).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana token account lookup reliability when the primary account service is unavailable.
  * Automatically checks standard and Token-2022 associated accounts as a fallback.
  * Returns a clear not-found result when no associated account exists.

* **Tests**
  * Added coverage for fallback discovery, Token-2022 accounts, missing accounts, and successful primary lookups.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
@@ -295,14 +295,23 @@ class SolanaService {
     }
 
     func fetchTokenAssociatedAccountByOwner(for ownerAddress: String, mintAddress: String) async throws -> (String, Bool) {
-        // First try getTokenAccountsByOwner
-        let (tokenAccounts, isToken2022) = try await getTokenAccountsByOwner(walletAddress: ownerAddress, mintAddress: mintAddress)
+        // First try getTokenAccountsByOwner. A transient RPC/node error is treated
+        // like an empty result: the account index can lag right after an ATA is
+        // created, or the node may momentarily fail, yet the ATA is still
+        // deterministically derivable. Falling through to the derivation probe
+        // instead of propagating keeps the send from failing when the account
+        // actually exists. A successful lookup keeps today's behavior exactly.
+        do {
+            let (tokenAccounts, isToken2022) = try await getTokenAccountsByOwner(walletAddress: ownerAddress, mintAddress: mintAddress)
 
-        if !tokenAccounts.isEmpty {
-            return (tokenAccounts, isToken2022)
+            if !tokenAccounts.isEmpty {
+                return (tokenAccounts, isToken2022)
+            }
+        } catch {
+            logger.warning("getTokenAccountsByOwner lookup failed; falling back to ATA derivation: \(error.localizedDescription, privacy: .public)")
         }
 
-        // If getTokenAccountsByOwner returns empty, probe the deterministic ATAs directly
+        // If getTokenAccountsByOwner returns empty (or errored), probe the deterministic ATAs directly
         guard let walletCoreAddress = WalletCore.SolanaAddress(string: ownerAddress) else {
             return ("", false)
         }
--- a/VultisigApp/VultisigAppTests/Blockchain/Solana/SolanaServiceAtaFallbackTests.swift
+++ b/VultisigApp/VultisigAppTests/Blockchain/Solana/SolanaServiceAtaFallbackTests.swift
@@ -1,0 +1,249 @@
+//
+//  SolanaServiceAtaFallbackTests.swift
+//  VultisigAppTests
+//
+//  Covers `SolanaService.fetchTokenAssociatedAccountByOwner` recovering from a
+//  transient `getTokenAccountsByOwner` failure. When the primary indexer lookup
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4792, opened 2026-07-10
Fix TerraClassic IBC token display: resolve to symbol/icon/fiat, truncate unknown vouchers

### LINKED ISSUE #4761: IBC tokens on TerraClassic display in poor shape
## Description
On the **TerraClassic** chain, IBC tokens (tokens identified by an `ibc/<hash>` denom) render poorly in the token list. The full `ibc/` denom hash is shown as the token name — a long, wrapping string that breaks the layout and reads as raw data instead of a friendly token symbol.

## Actual Behavior
- The token name shows the raw `ibc/0BB9D8513E8E8E9AE6A9D211D9136E6DA42288DDE6CFAA453A150A4566054DC5` hash, wrapping across multiple lines
- The balance column repeats the same full `ibc/…` hash next to the amount
- No token symbol, no fiat value (shows `—` / $0.00), and no proper icon

## Expected Behavior
- Resolve IBC denoms to their human-readable symbol/metadata where possible
- When metadata is unavailable, truncate the `ibc/` hash (e.g. `ibc/0BB9…4DC5`) instead of showing the full hash
- Keep the row layout consistent with other tokens (single-line name, aligned balance)

## Environment
- Platform: macOS
- Chain: TerraClassic

## Screenshot
TerraClassic IBC token display

gastonm5:
closes #4761

## Problem

On **TerraClassic**, a held `ibc/<hash>` token rendered with the raw 64-char denom as its name — a long wrapping string, no symbol, no icon, and no fiat value (`—` / $0.00). iOS was the only client showing the raw hash; Android and Windows/SDK both degrade gracefully.

## Root cause

A held `ibc/<hash>` denom flows through `CosmosCoinFinder.resolve`'s tiers and lands on the last one:

1. **Curated `TokensStore`** — no TerraClassic `ibc/` entries → miss.
2. **`denoms_metadata`** — TerraClassic returns a malformed record (`decimalsFromMeta` → nil) → rejected.
3. **IBC trace recursion** — gated to `.terra` via `ibcTraceCapableChains`, skipped for TerraClassic because the deprecated `/denom_traces/{hash}` endpoint returns `code 12 Not Implemented` there.
4. **Fallback `deriveTicker(denom:meta:nil)`** — had **no `ibc/` branch**, so it hit `return denom` → the raw `ibc/<hash>` became `Coin.ticker`.

That single ticker value drove all four symptoms: the name, the repeated balance-column suffix, and (via empty `logo`/`priceProviderId`) the missing icon and fiat.

## Key finding: the modern ibc-go endpoint works on TerraClassic

The deprecated `/ibc/apps/transfer/v1/denom_traces/{hash}` is genuinely dead on TerraClassic (publicnode `-32701`, hexxagon `code 12`) — the existing code correctly avoids it. But the **modern** ibc-go path is implemented:

```
GET /ibc/apps/transfer/v1/denoms/{hash}
→ {"denom":{"base":"uusdc","trace":[{"port_id":"transfer","channel_id":"channel-113"}]}}
```

So the reported token is **Axelar USDC** (base `uusdc`, 6 decimals). Neither Android nor Windows/SDK uses this endpoint yet — iOS exceeds parity here.

## The fix (layered — highest fidelity first, guaranteed-safe fallback last)

1. **`CosmosAPI`** — new `.ibcDenom(hash:)` endpoint + a `CosmosIbcDenom` model whose fields are optional so the gRPC-gateway NotFound shape (`{code,message}`) decodes to a nil base rather than throwing.
2. **`CosmosTokenMetadataResolver`** —
   - `ibcDenom(chain:denom:)` resolves the base denom via `/denoms`, cache-coalesced exactly like `ibcDenomTrace`;
   - `ibcTicker(baseDenom:)` strips a single leading `u`/`a` micro-unit prefix and uppercases (`uusdc` → `USDC`), mirroring the SDK `deriveIbcTraceTicker` / Android `stripDenomUnitPrefix`;
   - a new `ibc/` branch in `deriveTicker` returns **`IBC-<6hex>`** (e.g. `IBC-0BB9D8`) — the guaranteed layout fix even with zero network, matching Android's `String.toCosmosTicker`.
3. **`CosmosCoinFinder`** — `ibcModernDenomChains = [.terraClassic]` plus a new resolution tier between direct-metadata and the `.terra`-only trace recursion: an `ibc/<hash>` on a modern-endpoint chain resolves its base denom → derived ticker + metadata/fee-coin decimals, `isHidden = true` (SDK semantics), with curated logo/`priceProviderId` backfilled by ticker. `.terra`/phoenix-1 stays on its working `denom_traces` path (no regression).
4. **`TokensStore`** — one curated `.terraClassic` Axelar-USDC entry keyed by the full `ibc/0BB9…4DC5` contract (`USDC`, 6 decimals, `usd-coin`, `usdc` logo), so the reported token resolves at Tier 0 → symbol + icon + fiat with **no network call**.

The display layer (`TokenCellView`) and models (`Coin`/`CoinMeta`) are unchanged — they already render `ticker`.

### Follow-up hardening (from on-device testing against a TerraClassic address holding many vouchers)

The `u`-based IBC bases resolved cleanly (ATOM, DGN, USDT, SCRT, WHALE, USDC). Two more display gaps were then closed:

- **IBC vouchers whose resolved base is not a `u`/`a` micro-denom.** `ibcTicker(baseDenom:)` now returns `String?`: `factory/<minter>/<sub>` bases use the last path segment (`…/alloyed/allBTC` → `ALLBTC`), an already-clean short ticker uppercases (`wbnb-wei` → `WBNB-WEI`), and an opaque base (an EVM `0x…` address, namespaced, or over-long) returns `nil`. At the modern-IBC tier the call site then falls back to the voucher's `IBC-<6hex>` (`0x4580…` base → `IBC-AF91F7`), so a raw address never reaches the UI.
- **Native Terra fiat micro-denoms.** Non-IBC denoms like `ucny`/`uhkd`/`ukrw` reached the plain-denom Tier-4 fallback and rendered as raw `ucny`. `deriveTicker`'s final fallback now strips a leading `u`/`a` micro-unit prefix (only when the 2nd char is a letter, so `u123`-style ids are untouched) and uppercases → `CNY`. This tier only runs for the Cosmos discovery path (`CosmosCoinFinder`, allowlisted to Terra/TerraClassic), so the change is Terra-scoped. The `u`/`a` strip is factored into a shared `stripMicroUnitPrefix` helper reused by `ibcTicker` and `deriveTicker`.

## Tests

`CosmosCoinFinderTests` gains: the reported token resolving via curated Tier 0 (no LCD call); a non-curated voucher resolving via `/denoms` (`uusdc` → `USDC`, trace endpoint untouched); an unknown voucher degrading to `IBC-0BB9D8` instead of the raw hash; `deriveTicker` `ibc/` + `ibcTicker` `u`/`a`-strip edge cases (incl. the "second char must be a letter" guard); `ibcDenom` cache short-circuit/coalesce/NotFound; and `CosmosIbcDenom` decode for the resolved + NotFound shapes.

## On-device status

Tested on-device against a TerraClassic address holding many vouchers: `u`-based bases resolve to symbols (ATOM, DGN, USDT, SCRT, WHALE, USDC); factory/alloyed, EVM `0x…`, and native fiat micro-denoms are handled by the follow-up above. Remaining manual check: confirm the curated Axelar-USDC token shows `USDC` + icon + fiat, and an unknown voucher shows `IBC-<6hex>`.

## Known limitations / recommended follow-ups

These are pre-existing behaviors this display fix does not change; both surfaced during cross-model review and are deliberately out of scope here (they touch persisted `@Model` state the plan excluded):

- **Already-affected vaults are not auto-refreshed.** A vault that auto-added this token *before* this fix keeps the raw-hash ticker/empty logo, because `CoinService.insertVisibleDiscoveredToken` skips existing coins (`vault.coin(for:)` matches by contract) without refreshing them. New discoveries and re-added tokens display correctly. A safe refresh must also update the persisted `Coin.id` (it embeds the ticker and is the balance-cache key, the `Coin` equality key, and keysign identity) — a migration worth its own change. **Workaround for the reporter:** remove and re-add the token.
- **Hidden-token ticker match.** A user who explicitly hid the raw-ticker token could see it reappear (correctly labeled) once the ticker resolves, because `HiddenToken.matches` requires a ticker match. A fix would broaden hidden-token matching to chain+contract for non-native tokens.

## Summary by CodeRabbit

* **New Features**
  * Added modern IBC voucher denomination resolution for Terra Classic (via the modern denom endpoint).
  * Added a built-in Terra Classic USDC token entry.
  * Improved ticker derivation for unresolved IBC assets using deterministic `IBC-` labels.

* **Bug Fixes**
  * Unknown/unresolvable IBC vouchers now decode/fail gracefully and produce hidden-token fallbacks (no crashes).
  * Improved concurrent IBC denom resolution with request coalescing and safer cache retry behavior.
  * Refined ticker derivation to prefer resolved base-denom metadata when available.

* **Tests**
  * Expanded coverage for modern IBC resolution, caching behavior, and ticker/metadata fallbacks.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Cosmos/Service/CosmosAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Cosmos/Service/CosmosAPI.swift
@@ -24,6 +24,7 @@ struct CosmosAPI: TargetType {
         case wasmTokenBalance(contractAddress: String, base64Payload: String)
         case wasmTokenInfo(contractAddress: String)
         case ibcDenomTrace(hash: String)
+        case ibcDenom(hash: String)
         case denomMetadata(denom: String)
         case allDenomsMetadata
         case latestBlock
@@ -51,6 +52,12 @@ struct CosmosAPI: TargetType {
             return Self.wasmSmartQueryPath(contractAddress: contractAddress, base64Payload: base64Payload)
         case .ibcDenomTrace(let hash):
             return "/ibc/apps/transfer/v1/denom_traces/\(hash)"
+        case .ibcDenom(let hash):
+            // Modern ibc-go path. Unlike the deprecated `denom_traces/{hash}`
+            // endpoint (which Terra Classic LCDs answer with `code 12 Not
+            // Implemented`), Terra Classic implements this one — it returns the
+            // base denom + hop trace for a hashed IBC voucher.
+            return "/ibc/apps/transfer/v1/denoms/\(hash)"
         case .denomMetadata(let denom):
             // Denom strings can contain `/` (factory/..., ibc/HASH) which the
             // URL parser would otherwise treat as path separators. Percent-
@@ -143,6 +150,32 @@ struct CosmosCw20TokenInfoResponse: Decodable {
         let name: String?
         let symbol: String?
         let decimals: Int?
+    }
+}
+
+/// Response for the modern `/ibc/apps/transfer/v1/denoms/{hash}` endpoint.
+///
+/// A resolved voucher answers `{"denom":{"base":"uusdc","trace":[{...}]}}`;
+/// the base denom is what the wallet derives a ticker + decimals from. Every
+/// field is optional so the gRPC-gateway NotFound shape (`{"code":5,
+/// "message":"...not found..."}`) decodes to a `nil` base — treated as
+/// unresolved — instead of throwing a decode error.
+struct CosmosIbcDenom: Decodable {
+    let denom: Denom?
+
+    struct Denom: Decodable {
+        let base: String?
+        let trace: [Hop]?
+
+        struct Hop: Decodable {
+            let portId: String?
+            let channelId: String?
+
+            enum CodingKeys: String, CodingKey {
+                case portId = "port_id"
+                case channelId = "channel_id"
+            }
+        }
     }
 }
 
--- a/VultisigApp/VultisigApp/Blockchain/Cosmos/Service/CosmosCoinFinder.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Cosmos/Service/CosmosCoinFinder.swift
@@ -168,7 +168,11 @@ actor CosmosCoinFinder {
                 let baseMeta = await metadataResolver.denomMetadata(chain: chain, denom: baseDenom)
                 let decimals = baseMeta.flatMap(CosmosTokenMetadataResolver.decimalsFromMeta)
                     ?? feeDecimals(for: chain)
+                // An opaque base (EVM `0x…` address, namespaced/over-long denom)
+                // has no readable symbol, so degrade the voucher to its short
+                // `IBC-<6hex>` label rather than showing the raw base string.
                 let ticker = CosmosTokenMetadataResolver.ibcTicker(baseDenom: baseDenom)
+                    ?? CosmosTokenMetadataResolver.deriveTicker(denom: denom, meta: nil)
                 return makeDiscovered(
                     chain: chain,
                     denom: denom,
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4798, opened 2026-07-13
fix(solana): gate signAllTransactions N>1 before the keysign ceremony (#4796)

### LINKED ISSUE #4796: fix(solana): gate signAllTransactions N>1 before the keysign ceremony (sibling of android#5238)
## Problem
iOS has the same late-failure shape as vultisig-android#5238. `SolanaHelper.getPreSignedImageHash` loops over ALL raw transactions (`Solana.swift:306-322` — `for base64Tx in signSolana.rawTransactions`), so the full multi-device keysign ceremony — including the user's physical approval on the paired phone — completes. But `getSignedTransaction` guards `signSolana.rawTransactions.count == 1` (`Solana.swift:337`) and only assembles the first, so an N>1 batch is discarded with an opaque error AFTER the user has already approved.

## Expected
Detect `rawTransactions.count > 1` at request-acceptance time (before peer discovery / the ceremony) and show a clear "batch signing not yet supported" message. Never start a ceremony that cannot deliver a result.

## Fix
Move/duplicate the `count == 1` guard so it fires before the ceremony starts (the `KeysignMessageFactory` / `KeysignViewModel` accept path), with a clear error, rather than only inside `getSignedTransaction` after the ceremony completes.

### Must do
- [ ] Reject `signSolana.rawTransactions.count > 1` before starting the ceremony, with a clear message
- [ ] Test asserting the guard fires before ceremony start for a multi-transaction payload
### Must NOT do
- Don't attempt full batch support here
- Don't change the single-transaction signing path

## Acceptance criteria
- [ ] An N>1 `signSolana` request is rejected before the ceremony, not after user approval

## References
- Android source: vultisig-android#5238 · SDK sibling: vultisig-sdk#1205

gastonm5:
## Problem

`SolanaHelper.getPreSignedImageHash` looped over **all** `signSolana.rawTransactions`, so a multi-transaction Solana `signAllTransactions` batch (N>1) relayed to an iOS **co-signer** ran the FULL multi-device keysign ceremony — including the user's physical approval on the paired phone — before `getSignedTransaction` rejected N>1 (`rawTransactions.count == 1`) with an opaque post-ceremony error. Sibling of android#5238.

This does **not** implement batch support; it fails fast, before the ceremony, with a clear localized message. The single-transaction path is unchanged; the existing post-ceremony guard stays as defense-in-depth.

## Fix

1. **Pre-ceremony reject** — `VultisigApp/Blockchain/Solana/Signing/Solana.swift:317` — guard `signSolana.rawTransactions.count <= 1` at the top of `getPreSignedImageHash`, throwing `HelperError.runtimeError("solanaBatchSigningNotSupported".localized)`. This runs while keysign messages are built (`KeysignMessageFactory.getKeysignMessages()`), before peer discovery.

2. **Surface the failure on the co-signer** — `VultisigApp/Features/Keysign/ViewModels/JoinKeysignViewModel.swift:385` — the guard makes `prepareKeysignMessages` set `.FailedToStart`, but the joiner deeplink handler then always ran `manageQrCodeStates()`, which unconditionally advanced a relay payload to `.JoinKeysign` — clobbering the failure and letting the co-signer tap "Join transaction signing" into the ceremony with **no messages to sign**. `manageQrCodeStates()` now preserves an already-set `.FailedToStart`, so the error screen shows and the ceremony can't be entered. (This also correctly fixes every other `prepareKeysignMessages` failure the transition previously masked.)

### Why the initiator path needs no change
A cross-model review flagged that the initiator's `refreshSolanaBlockhash` strips `signData` (only staking is preserved), so the guard can't see an N>1 batch there. Verified this is a non-issue: iOS **never initiates** a non-staking / N>1 `signSolana` payload — the only in-app `.signSolana` producers are single-transaction staking (`BlockChainService`/`SolanaStakingSignDataResolver`) and the proto-ingress joiner deserialize path (`Cosmos+ProtoMappable`). An N>1 batch only ever arrives as a co-signer.

## Localization
Added `solanaBatchSigningNotSupported` (EN: "Signing multiple Solana transactions at once isn't supported yet.") to all 7 maintained locales (en, de, es, hr, it, pt, zh-Hans), alphabetically sorted via `sort_localizable.py`.

## Tests
- `SolanaHelperTests.testGetPreSignedImageHashRejectsMultipleRawTransactions` — asserts N>1 throws.
- `SolanaHelperTests.testGetPreSignedImageHashAllowsSingleRawTransaction` — regression: single-tx path still returns one non-empty hash.
- `JoinKeysignFailedStateGuardTests.testManageQrCodeStatesKeepsFailedToStart` — co-signer stays on the failure screen (can't enter the ceremony).
- `JoinKeysignFailedStateGuardTests.testManageQrCodeStatesStillAdvancesHealthyRelayPayload` — healthy relay payload still transitions to `.JoinKeysign`.

## Gate receipts
- **Full `make test`** on sim `vk-4796` (iPhone 17 Pro Max, iOS 26.5): **2677 tests, 1 skipped, 1 failure** — the sole failure is the known pre-existing decimal-comma locale flake (`StakingValidatorProjectionTests.testCosmosEmptyMonikerFallsBackToTruncatedAddress`, "12,3%" vs "12.3%"), unrelated to this change. All 4 new tests pass.
- **SwiftLint**: clean on all changed files (0 violations).
- **Cross-model review (Codex, read-only)**: round 1 raised the joiner-state overwrite (fixed here) + the initiator concern (verified non-exploitable); round 2 = "No functional findings".

## Cross-platform
- android#5238 (sibling fail-fast guard) still open.
- sdk#1205 closed as already-fixed via full batch support on main.

## Pending manual check
- [ ] On-device 2-device Solana `signAllTransactions` (N>1) batch is rejected before the ceremony with the localized message (no physical approval prompt on the paired phone).

Closes #4796

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana signing now clearly rejects attempts to sign multiple transactions at once.
  * Prevented failed key-signing sessions from incorrectly progressing toward confirmation.
  * Preserved the normal key-signing flow for valid relay sessions.

* **Localization**
  * Added messages explaining that Solana batch signing is unavailable in English, German, Spanish, Croatian, Italian, Portuguese, and Chinese.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
@@ -306,10 +306,12 @@ enum SolanaHelper {
     static func getPreSignedImageHash(keysignPayload: KeysignPayload) throws -> [String] {
         // Handle SignSolana (raw transactions)
         if let signSolana = keysignPayload.signSolana {
-            // Reject a multi-transaction batch (signAllTransactions, N>1) here —
-            // the single pre-ceremony chokepoint reached on BOTH the initiator
-            // and the co-signer before peer discovery. getSignedTransaction only
-            // supports a single raw transaction, so without this the user would
+            // Reject a multi-transaction batch (signAllTransactions, N>1) here.
+            // This runs while keysign messages are built, before peer discovery.
+            // A relayed N>1 batch reaches this device only as a co-signer (the
+            // sole in-app signSolana producers are single-tx staking and the
+            // proto-ingress joiner path), and getSignedTransaction supports just
+            // one raw transaction — so without this the co-signer would
             // physically approve the full multi-device keysign and only then hit
             // an opaque post-ceremony failure. Fail fast with a clear message.
             guard signSolana.rawTransactions.count <= 1 else {
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Signing/Solana.swift
@@ -306,6 +306,15 @@ enum SolanaHelper {
     static func getPreSignedImageHash(keysignPayload: KeysignPayload) throws -> [String] {
         // Handle SignSolana (raw transactions)
         if let signSolana = keysignPayload.signSolana {
+            // Reject a multi-transaction batch (signAllTransactions, N>1) here —
+            // the single pre-ceremony chokepoint reached on BOTH the initiator
+            // and the co-signer before peer discovery. getSignedTransaction only
+            // supports a single raw transaction, so without this the user would
+            // physically approve the full multi-device keysign and only then hit
+            // an opaque post-ceremony failure. Fail fast with a clear message.
+            guard signSolana.rawTransactions.count <= 1 else {
+                throw HelperError.runtimeError("solanaBatchSigningNotSupported".localized)
+            }
             var allHashes: [String] = []
             for base64Tx in signSolana.rawTransactions {
                 let hashes = try getPreSignedImageHashForRaw(base64Transaction: base64Tx)
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -1079,6 +1079,7 @@
 "slippageHelperText" = "Begrenzt, wie stark sich der Preis ändern darf, bevor die Ausführung abgebrochen wird.";
 "slippageMaxNote" = "Slippage ist auf %@ begrenzt.";
 "slippageTolerance" = "Slippage-Toleranz";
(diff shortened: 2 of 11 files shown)


gastonm5:
Re: CodeRabbit's "out-of-scope" pre-merge warning on the `JoinKeysignViewModel` change — keeping it here is intentional, because the two changes are causally coupled and the guard alone does not fully fix the issue:

- The new `getPreSignedImageHash` guard throws on `rawTransactions.count > 1` **before** the ceremony, and the join path sets `status = .FailedToStart` in the `catch`.
- But `manageQrCodeStates()` was then overwriting `.FailedToStart` back to `.JoinKeysign`, so on the co-signer/joiner a user could still tap "Join transaction signing" into a ceremony with empty messages — the exact late-failure walkthrough this issue is about.

So the `.FailedToStart` preservation is the second half of the same fix (make the pre-ceremony rejection actually stick on the joiner), not an unrelated change. Splitting it out would ship a guard that doesn't hold on one of the two paths. Both are covered by the new `JoinKeysignFailedStateGuardTests`. Happy to split if a maintainer prefers, but flagging the coupling so it isn't bounced as scope creep.

johnnyluo (approved):
LGTM

Issue #4232, opened 2026-04-23
feat: THORChain limit swap — Place flow (Phase 1)


realpaaao:
## Overview

Add the Place flow for THORChain limit swaps. Users pick Market/Limit on the swap screen, set a target price + expiry, review, and sign. Post-sign, the inbound TX is visible in TX History; no in-app "open orders" list in Phase 1 (coming in Phase 2 inside TX History).

## In scope (matches Figma)

- Market / Limit tab on Swap screen
- Asset sell/buy rows (reuse existing Market swap asset picker)
- "Execute when" expandable price entry with $/asset unit toggle
- Preset pills: `Market` / `+1%` / `+5%` / `+10%` — set price = market × (1 + pct)
- Expiry pills: `12h` / `24h` / `3d` (default: 24h)
- Live "% from market" label under price
- Warnings: price ≤ market → "Price already available, consider market swap"; price > market × 1.2 → "May not fill within expiry"
- Confirmation sheet with Blockaid scan, target price, expiry, fees, required "The swap amount is correct" checkbox before Sign enables
- Memo construction mirrors SDK spec; native test suite asserts against SDK-published fixtures
- Local-storage record on successful send: `{inbound_tx_hash, source_asset, source_amount, target_asset, dest_addr, target_price, expiry_blocks, created_at, status: 'pending'}`
- Limit tab only enabled for THORChain-routable pairs

## Out of scope

- Open Orders list (Phase 2, inside TX History)
- Modify / Cancel flows
- Partial-fill UI
- Notifications
- MAYAChain
- Cross-device sync

## Acceptance

- [ ] Swap screen shows Market/Limit tabs; selecting Limit swaps the body to the Limit layout (matches Figma nodes)
- [ ] Asset picker behavior identical to Market swap; Limit tab hides/disables for non-THORChain-routable pairs
- [ ] "Execute when" collapsed row expands into price entry on tap
- [ ] $/asset toggle switches the displayed unit; underlying LIM math is unit-agnostic
- [ ] Preset pills (`Market`/`+1`/`+5`/`+10%`) set price correctly; "% from market" updates live
- [ ] Expiry pills set `expiry_hours ∈ {12, 24, 72}`; default = 24h
- [ ] Price-below-market and price-far-above-market warnings render exactly per spec
- [ ] Place Limit Order opens confirmation sheet matching Figma's sign screen; Sign disabled until "The swap amount is correct" checkbox is ticked
[...]

PR #4814, opened 2026-07-14
spike(limit-swap): all 4 UX variants behind a runtime picker — NOT FOR MERGE

gastonm5:
## Description

> ⚠️ **SPIKE — NOT FOR MERGE.** This branch exists only so the team can compare the four candidate limit-swap layouts in a single build. Once a direction is picked, the winner gets implemented properly on `feat/4232-thorchain-limit-swap-place-flow` (#4319) and this branch is closed and deleted.

Carries all four candidate UX layouts for the THORChain limit-swap form in one build, switchable at runtime.

Related: #4232 / #4319 (deliberately not `Fixes` — nothing to close).

### How to try it

Settings → Advanced → enable **Limit Swap** → Swap → **Limit** tab. The small **A** dropdown to the right of the Market/Limit tabs switches layouts (the menu spells out the full names); your choice persists across relaunch.

The VM is owned by the entry view and each body rehydrates from the draft on load, so **flipping variants preserves the order you typed** — you can compare like-for-like without re-entering an amount.

### The variants

|  | Variant | Source |
|---|---|---|
| **A** | **Accordion (current)** — Execute-when first, Asset second | the current #4319 layout, byte-identical |
| **B** | **Assets first** — Asset first, Execute-when second | `feat/4232-limit-ux-assets-first` |
| **C** | **Uniswap flat** — no accordion, horizontal unit toggle, chipped assets, unified radii | `feat/4232-limit-ux-uniswap` |
| **D** | **Three sections** — Asset → Execute-when → Amount with reflected output | `feat/4232-limit-ux-3section` |

### Notes for reviewers

- **Presentation only.** All four share the same `LimitSwapFormViewModel`, memo math, validation, routability gate and place-order flow. No behavior differs between them.
- **The duplication is deliberate — please don't factor it out.** Each variant is a self-contained file whose sub-views are file-private. Variant A is byte-identical to the base; B/C/D differ from their source commits by exactly the two rename lines. Keeping them unshared is what makes the comparison honest rather than homogenized.
- **Please don't "clean up" the single-child `VStack` in `LimitSwapEntryView`.** It anchors `.task`/`.onChange` to a stable identity. Attached directly to the variant switch they'd live on a `_ConditionalContent` whose type changes on every flip, so `.task` would re-run and reset the amount each time you switch — destroying the comparison. It's documented in-code.
- **Market mode is visually unchanged.** The picker renders only in Limit mode, in the slot the quote countdown already leaves free.

## Which feature is affected?
- [x] Swap — no tx link: this is a UI-comparison spike. Nothing is signed or broadcast here that isn't already covered by #4319.

## Checklist

- [x] My changes generate no new warnings — SwiftLint 0 violations; macOS Debug `** BUILD SUCCEEDED **`
- [ ] I have added tests — **N/A by design**: throwaway spike, presentation-only, no logic changed. The full suite is green on the base (3018 tests, 0 failures) and this branch adds no behavior to test.

## Screenshots (if applicable)

Not attached on purpose — the whole point is to run it and flip between the four live.

## Additional context

Based on `feat/4232-thorchain-limit-swap-place-flow` rather than `main`, so the diff shows only the spike (~+3980/−9) instead of burying it under the entire limit-swap feature.

## Agent Metadata

- **Agent**: Claude Code
- **Issue**: #4232 / PR #4319 (design-review aid)
- **Confidence**: high — presentation-only; build + lint green; each variant verified byte-faithful to its source commit
- **Needs human review for**: nothing to merge. What's wanted here is **design feedback: which layout wins.**

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -661,6 +661,8 @@
 "limitSwap.error.title" = "Order konnte nicht platziert werden";
 "limitSwap.error.unsupportedAsset" = "Dieses Asset wird für Limit-Orders nicht unterstützt. Wähle ein anderes Asset.";
 "limitSwap.executeWhen" = "Ausführen wenn";
+"limitSwap.executeWhen.headerIsWorth" = "wert ist";
+"limitSwap.executeWhen.headerWhenOne" = "Wenn 1";
 "limitSwap.executeWhen.marketReference" = "Markt %@";
 "limitSwap.executeWhen.oneUnit" = "1 %@";
 "limitSwap.expiry" = "Ablauf";
@@ -677,6 +679,7 @@
 "limitSwap.verify.title" = "Du platzierst eine Limit-Order";
 "limitSwap.warning.priceAtOrBelowMarket" = "Kurs bereits verfügbar, erwäge einen Markt-Swap";
 "limitSwap.warning.priceFarAboveMarket" = "Wird möglicherweise nicht vor Ablauf ausgeführt";
+"limitSwap.youReceiveAtLeast" = "Du erhältst mindestens";
 "limitSwapToggle" = "Limit-Swap";
 "liquidityPools" = "Liquiditätspools";
 "loading" = "Loading...";
--- a/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
@@ -659,6 +659,8 @@
 "limitSwap.error.title" = "Couldn't place order";
 "limitSwap.error.unsupportedAsset" = "This asset isn't supported for limit orders. Pick a different asset.";
 "limitSwap.executeWhen" = "Execute when";
+"limitSwap.executeWhen.headerIsWorth" = "is worth";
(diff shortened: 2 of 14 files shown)


johnnyluo:
CleanShot 2026-07-15 at 10 38 21@2x
Let's use the flat view , but also assets first

gastonm5:
**Decision made: Variant C — Uniswap flat — wins.** Closing this spike; it did its job.

The flat layout is being applied to the real feature branch on #4319, so this comparison build has served its purpose and shouldn't linger as a merge candidate.

For the record, the four variants were:
- **A** — Accordion (the layout #4319 shipped with)
- **B** — Assets first
- **C** — **Uniswap flat** ← selected
- **D** — Three sections with reflected output

Notes carried forward from the exploration:
- C's flat layout means every input is visible at once — no collapse. The tradeoff to watch is vertical space on small screens when the queue-disabled row, a routability notice and a price warning all render at the same time; the accordion's compact summaries had been buying that back.
- C deliberately did **not** restore the old inline `sourceAmount × targetPrice` buy-preview from `f773ca85e` — that was view-side business logic which can display *more* than the truncated LIM actually guarantees. It keeps `vm.expectedBuyAmount`.
- C's header needs the split `headerWhenOne` / `headerIsWorth` locale keys, which are a mild i18n regression vs. `oneUnit`'s single format string (word order can't be reordered per locale — Korean already needs a bare `"1"` prefix). Worth revisiting.

Branch `spike/4232-limit-ux-variants` will be deleted. Thanks to anyone who took it for a spin.

PR #4819, opened 2026-07-15
fix(keygen): point the peer-discovery info pointer at "Scan QR" and make it render on macOS

### LINKED ISSUE #4804: [iOS] Keygen peer-discovery info pop-up: pointer animation points at the wrong location
## Summary
In the keygen **"Scan the QR on your other devices"** info pop-up (`PeerDiscoveryInfoBanner`), the animated **pointer** (white dot) sits at the wrong spot on the tutorial illustration — it doesn't point at the element it's meant to highlight.

## Cause
`Components/New Wallet/PeerDiscoveryInfoBanner.swift`

The pointer is a Rive animation overlaid on the static tutorial image, positioned by a hardcoded offset:
```swift
var display: some View {
    ZStack {
        image        // Image("secure-qr-tutorial"), 290×230   (:48-52)
        animation    // Rive "peer_discovery_info_banner", 80×80
    }
}
var animation: some View {
    animationVM?.view()
        .frame(width: 80, height: 80)
        .offset(x: -16, y: -30)     // :57  ← lands the pointer in the wrong place
}
```
The `.offset(x: -16, y: -30)` doesn't align the pointer with its intended target in the `secure-qr-tutorial` image, so it appears near top-center over empty space instead of on the element being explained.

## Proposed fix
Adjust the pointer's offset (or anchor it relative to the image content) so it lands on the intended element per the Figma/design. Verify against the actual `secure-qr-tutorial` asset and Rive animation bounds.

## Scope
iOS (native). File: `PeerDiscoveryInfoBanner.swift:48-58`.

gastonm5:
closes #4804

## Problem

The animated pointer in the keygen **"Scan the QR on your other devices"** info pop-up (`PeerDiscoveryInfoBanner`) sat over empty space instead of the **"Scan QR" button** it explains. Separately, it **never appeared at all on macOS**.

## Root cause

Two independent defects:

**1. Wrong position** — the hardcoded `.offset(x: -16, y: -30)` landed near top-center of the `secure-qr-tutorial` illustration.

**2. Invisible on macOS — a RiveRuntime limitation, not a layout bug.** A `RiveView` hosted inside a macOS **native sheet** loads, advances and draws but never becomes visible. Proven on macOS 26.5.1 / RiveRuntime 6.15.1 with two probes of the *same* `.riv` in one build:

| probe | context | result |
|---|---|---|
| control | main window (`AppKitWindow`) | pointer **renders** |
| banner | native sheet (`SheetPresentationWindow`) | **blank** |

Both reported identical `advances=311 draws=286 playing=true`, artboard `108×108` — Rive genuinely renders ~286 frames into a correctly-placed view and none of it reaches the screen.

Each ruled out by test, not assumption: asset membership (one unified target; `.riv` confirmed in `VultisigApp.app/Contents/Resources/`); model loading; geometry (`RiveView frame=(0,0,40,40)`, `inWindow=(199,283,40,40)`, inside the 470×423 sheet); `edgesIgnoringSafeArea`; Rive's `isOnscreen()` draw gate (`offscreenBehavior = .playAndDraw` changed nothing); the display link; and z-order against both the image and the opaque `Background()`. **Confirming A/B:** forcing the pre-macOS-26 inline `customSheet` path makes the pointer appear — the sheet window is the variable.

> ⚠️ `CrossPlatformSheet` uses the inline `customSheet` for macOS < 26 and a native sheet for macOS ≥ 26. Rive renders in the former, not the latter — so **macOS 26 silently broke Rive inside any sheet**. This banner was the only such site, so nothing else regressed, but the trap remains for future work.

## Fix

Replace the Rive pointer with an equivalent SwiftUI `TapPointerIndicator` (solid core + halo that expands and fades), which renders on both platforms and sidesteps Rive-in-sheet entirely. The pointer is anchored with `.overlay` on the **image** so the offset resolves against the rigid 290×230 box rather than a ZStack sibling. Geometry is the maintainer-validated `frame 40×40` / `offset(x: -16, y: 5)`. Halo is deliberately more prominent than the original Rive. Honors Reduce Motion, matching `ErrorView`/`DoneScreen`. Deletes the now-unused `peer_discovery_info_banner.riv`.

## Verification

- **macOS: visually confirmed** — pointer renders on the "Scan QR" button in the native sheet (the exact spot that was blank), pulse cycles correctly.
- **iOS: visually confirmed** — pointer renders on the "Scan QR" button in the iPhone 16 Pro simulator, halo pulsing, same placement as macOS.
- **Gate green**: `make test` → **2769 tests, 1 skipped, 0 failures** (re-run after the safe-area restore). SwiftLint clean (43 violations = unchanged pre-existing baseline; 0 in touched files).
- Builds green on **both** macOS and iOS Simulator.
- Codex read-only whole-branch review: 1 Low (Reduce Motion) — **fixed**, no findings outstanding.

Second commit restores `.edgesIgnoringSafeArea(.top)` on `display`, which was dropped as debris while bisecting the macOS bug — keeping it scopes the change to the pointer alone.

## Pending manual pass

- [ ] **On-device pass** via the real secure-keygen flow (both platforms verified in simulator/desktop, not on hardware).
- [ ] Design sign-off on the pulse **timing**: the original Rive motion was reconstructed from sampled frames, not extracted from its state machine.

## Not in scope

RiveRuntime is 6 minors behind (6.15.1 vs 6.21.1). Bumping is **source-breaking** — 6.21.1 exports a public `Image` that collides with SwiftUI's `Image`. Whether it fixes the macOS sheet bug is **untested**. Worth a separate change, and worth reporting upstream (no matching issue found on `rive-app/rive-ios`).

## Summary by CodeRabbit

* **New Features**
  * Added a new accessibility-hidden “Tap Here” pointer indicator with a gentle pulse animation.
  * The pointer automatically remains steady when Reduce Motion is enabled.

* **UI Improvements**
  * Updated the peer discovery info banner to replace the prior animated element with the tap pointer overlay.
  * Adjusted the banner layout and positioning so the pointer appears over the intended image area for clearer guidance.


johnnyluo (approved):
LGTM

PR #4821, opened 2026-07-16
fix(solana): re-read stake accounts via getAccountInfo to stop flipping Defi state

### LINKED ISSUE #4822: [Fix] Solana Defi staked list shows flipping/unstable stake state [Defi / Solana staking]
## Problem
The Solana Defi staked list fetches stake accounts with a single `getProgramAccounts` (jsonParsed) call and binds its parsed state straight to the UI. Many RPC providers serve `getProgramAccounts` from a lagging secondary index, so `delegation`/`deactivationEpoch` flip between refreshes right after a stake/unstake. Users see unstable/stale activation state (active ↔ deactivating ↔ inactive) on the Defi screen.

## Expected Behavior
The staked list reflects each account's true on-chain state and stays stable across refreshes.

## Steps to Reproduce
1. Open the Solana Defi screen for a vault with a stake account.
2. Stake, unstake, or deactivate an account.
3. Refresh the list repeatedly — the activation state flip-flops between refreshes.

## Solution
Use `getProgramAccounts` only to DISCOVER the owner's stake-account pubkeys, then re-read each account's live state via `getAccountInfo` (a direct bank read with no index lag). Return results in discovery order so the list is stable. `getAccountInfo` is a known-address direct read, which is not subject to the gPA index lag.

### Must Do
- [x] Discover pubkeys via `getProgramAccounts`, re-read state via `getAccountInfo` in `SolanaService.fetchSolanaStakeAccounts`
- [x] Preserve discovery order for a stable list

### Must NOT Do
- Don't change unrelated code
- Don't refactor surrounding staking logic

## Acceptance Criteria
- [x] Build succeeds
- [x] Activation state stays stable across refreshes after stake/unstake/deactivate
- [x] Vault with no stake accounts returns an empty list (no error)

## Fixed by
PR #4821

johnnyluo:
## Issue
Closes #4822

## Summary
The Solana Defi staked list (`SolanaStakeDefiViewModel` → `SolanaStakingService.fetchStakeAccounts` → `SolanaService.fetchSolanaStakeAccounts`) fetched stake accounts with a single `getProgramAccounts` (jsonParsed) call and bound its parsed state directly to the UI.

Many RPC providers serve `getProgramAccounts` from a **lagging secondary index**, so its `delegation`/`deactivationEpoch` fields flip between refreshes right after a stake/unstake — the staked list showed unstable/stale state (reproduced in testing).

This change uses `getProgramAccounts` only to **discover** the owner's stake-account pubkeys, then re-reads each account's **live** state via `getAccountInfo` — a direct bank read with no index lag:

- Discovery call unchanged (existing jsonParsed decoding); only the `pubkey` of each row is taken.
- Each account re-read concurrently via the existing `fetchSolanaStakeAccount(address:)` (`getAccountInfo`).
- Results returned in **discovery order** for a stable list across refreshes.

`getAccountInfo` is a known-address, direct read — the same fix pattern that resolves the flip-flop when querying a single stake account manually.

## Notes
- Cost goes from 1 → 1 + N requests (N = stake accounts held); the re-reads run in parallel. Fine for the handful a vault typically holds.
- The unused `getStakeAccountsByOwner(pubkeyOnly: true)` branch was left untouched — its base64 `data` array can't decode into `SolanaStakeAccountData` (expects an object), so it's dead as-is. Out of scope here; can be removed/fixed separately.

## Test Plan
- [x] Open the Solana Defi screen with an active stake account; refresh repeatedly — state (active/deactivating/inactive) stays stable.
- [x] Stake, then verify the new account appears with correct state.
- [x] Deactivate, then verify the row reflects deactivating/cooldown correctly across refreshes.
- [x] Vault with no stake accounts returns an empty list (no error).
- [x] SwiftLint passes (0 violations on changed file)
- [x] xcodebuild build succeeds

## Summary of changes

* **Bug Fixes**
  * Improved Solana stake account retrieval by first discovering stake account addresses, then fetching each account’s latest live state for more accurate details.
  * Preserves the original discovery order in the returned results.
  * Returns an empty list when no stake accounts are found, and avoids failing when individual accounts can’t be refreshed.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
@@ -577,26 +577,22 @@ class SolanaService {
     /// Not cached — must reflect a just-submitted stake/unstake and freshly
     /// accrued rewards; the UI refreshes on appear.
     func fetchSolanaStakeAccounts(owner: String) async throws -> [SolanaStakeAccount] {
-        // Discover the owner's stake accounts via getProgramAccounts. We prefer to
-        // re-read each account's live state with getAccountInfo below: many RPC
-        // providers serve getProgramAccounts from a lagging secondary index, so its
-        // parsed delegation/deactivation fields flip between refreshes right after a
-        // stake/unstake (observed on the Defi screen). getAccountInfo is a direct
-        // bank read with no index lag.
-        //
-        // But we KEEP the discovery row as a fallback: a just-created stake account
-        // may not be visible to getAccountInfo yet (it hasn't propagated to the node
-        // the follow-up read hits), and must not drop off the list in that window.
+        // Discover the owner's stake-account pubkeys via getProgramAccounts, but
+        // take ONLY the addresses from it. Many RPC providers serve
+        // getProgramAccounts from a lagging secondary index, so its parsed
+        // delegation/deactivation fields flip between refreshes right after a
+        // stake/unstake (observed on the Defi screen). Each account's live state is
+        // re-read below with getAccountInfo — a direct bank read with no index lag.
         let discovery = try await httpClient.request(
             api(.getStakeAccountsByOwner(staker: owner, pubkeyOnly: false)),
             responseType: SolanaGetProgramAccountsResponse.self
         )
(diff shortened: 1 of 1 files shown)


johnnyluo:
hold on , we might not need this , I think the solana node we are using is having issues

PR #4500, opened 2026-06-05
fix(wallet): reactive + cheap vault chain list via value-type row projection

gastonm5:
Closes #4494
Closes #4495

## Why

The vault main chain list renders per-row computed properties over the live `Vault`/`Coin` SwiftData models. Because `Vault`/`Coin` are `@Model` + `ObservableObject` with **no** `@Published` members, `@ObservedObject var vault` is inert for relationship/balance mutations — the list only repaints when a `@Published` on `VaultDetailViewModel` changes. That single fact coupled both issues:

- **#4494 (not reactive):** after a *same-vault* add/remove, the synchronous seed in `updateBalance` was skipped (its guard fired only on a vault switch — PR #4405), so a newly added chain (e.g. ZEC) only surfaced via the async balance tail, which is gated behind `BalanceService.updateBalances`' all-or-nothing `withTaskGroup` (waits for the slowest coin RPC). Remove+re-add appeared to need a view switch.
- **#4495 (scroll jank):** each visible row recomputed `vault.coins(for:)` ~3× + a `RateProvider` lookup per coin, ran a per-cell `withAnimation(.interpolatingSpring)` + `.contentTransition(.numericText())` + a per-cell `DispatchQueue.main.asyncAfter(+1s)` timer, and the cells weren't `Equatable`, so any `viewModel.chains` republish re-evaluated the whole `ForEach` and scrolling re-instantiated expensive cells.

## The fix — `ChainRowModel` value-type projection

Precompute one `ChainRowModel` (Identifiable + **Equatable**) per chain in the view model and render dumb, cheap rows off a new `@Published private(set) var rows: [ChainRowModel]`.

- **Membership becomes the reactive source → fixes #4494.** The synchronous seed guard is made *membership-aware*: it now also fires when `Set(vault.chainsWithCoins) != Set(chains)`. A freshly added/removed chain appears in the **same runloop as the save**, no longer waiting on the network-gated async tail.
- **Rows are cheap + Equatable → fixes #4495.** A single `Dictionary(grouping: vault.coins, by: \.chain)` pass replaces the per-row `vault.coins(for:)` (kills the O(rows×coins) work); `GroupedChainCellView` renders precomputed strings with the per-cell `@State`/spring/`asyncAfter` churn removed.
[...]

PR #4824, opened 2026-07-16
fix(wallet): refresh per-chain fiat when rates land, not only after a balance refresh

### LINKED ISSUE #4830: [Fix] Per-chain fiat total is stale or never appears [Wallet / Main screen]
# [Fix] Per-chain fiat total is stale or never appears [Wallet / Main screen]

## Problem

The **per-chain fiat total** on the wallet main screen takes far too long to appear, or never appears at all — the rows sit at `$0.00`. The vault header total updates fine; only the per-chain rows are stuck. Affects all chains equally and all entry paths (cold launch, revisit, pull-to-refresh, currency change).

The rates needed to render the correct value are normally **already cached in memory** while the rows show `$0.00`. Instrumented on a real 21-coin vault, the list sat at `$0.00` for **60.6 seconds with the right answer already in RAM**:

```
t=  0.003  VM.SEED rows -> Ethereum=$0.00, TerraClassic=$0.00
t=  0.066  RateProvider.WARMUP loaded 34 PERSISTED rates from SwiftData
t= 60.632  fetchRatesSafely THREW after 60.37s: timeout -- REMAINING RATES NEVER FETCHED
t= 60.657  VM.ROWS REBUILT -> TerraClassic=$44.32, Ethereum=$27.00
```

The price fetch *threw* — zero new rates arrived that cycle. The correct figures at t=60.657 came entirely from the t=0.066 cache.

## Expected Behavior

Per-chain fiat renders from cached rates as soon as those rates are available (~100 ms on cold launch), independent of whether a vault-wide network refresh completes. A slow or failing RPC on one chain must not gate every chain's fiat display.

## Steps to Reproduce

1. Cold-launch the app on a vault with several chains, on a **degraded or throttled network** (the defect is network-dependent — ~1.7 s gap on a healthy connection, 60 s+ or never on a bad one).
2. Open the wallet main screen.
3. Observe: the vault header total populates, while the per-chain rows stay at `$0.00`.
4. Alternatively, trigger a refresh and navigate away/back quickly to cancel the in-flight cycle — the rows can then stay `$0.00` indefinitely.

## Root Cause

`VaultMainChainListView` renders `ChainRowModel.fiatBalance`, a **preformatted `String`** baked when `VaultDetailViewModel.rows` is assigned. Three facts combine:

1. **`RateProvider` is a silent cache** — no `ObservableObject`, no `@Published`, no subject. A rate landing is unobservable by design.
2. **`BalanceService.refreshFiat`'s `objectWillChange.send()` doesn't help the rows** — it repaints the views, but the rows' *data* is the stale snapshot. It does work for the header total, which reads `vault.coins` live via `@ObservedObject var vault` — hence the asymmetry.
3. **The only rebuild is the tail of `updateBalance`'s network refresh**, behind a 250 ms debounce, the full round trip, and `if !Task.isCancelled`.

"Slow" and "never" are one bug with two triggers: **slow** = rows wait on the slowest balance RPC in the whole vault; **never** = `refresh()` fires from ~6 entry points each cancelling the previous task, and a cycle cancelled after `updateBalances` returns skips the rebuild entirely, with nothing retrying until the next appear/pull.

### Origin

`2dee43844` ("reactive + cheap vault chain list via value-type row projection", #4500, first in `v1.39.60`) — it replaced a live-computed `fiatBalance` (recomputed on every `vault.objectWillChange`) with the baked snapshot. The defect is **latent and network-dependent**, so it surfaces intermittently rather than at a clean release boundary.

`30fb07391` (#4693, price/balance decoupling) was the prime suspect and is **innocent** — `VaultDetailViewModel.updateBalance` is byte-identical between `v1.40.63` and `origin/main`, and an A/B experiment put the two orderings within 1 ms of each other.

## Solution

Give the rows a reactive link to rate arrival so they rebuild when rates land, rather than only at the tail of an uncancelled vault-wide network refresh. Render from the already-warm `RateProvider` cache on first load instead of waiting on the network round trip.

### Must Do
- [ ] Make rate arrival observable so `VaultDetailViewModel.rows` rebuilds from cached rates
- [ ] Ensure per-chain fiat renders on cold launch from the persisted-rate warmup, without a completed network refresh
- [ ] Cover the regression with unit tests at the view-model boundary

### Must NOT Do
- Don't change unrelated code
- Don't revert or undo the `#4693` price/balance decoupling — it is not the cause and it improves the worst case

## Acceptance Criteria
- [ ] Build succeeds
- [ ] Per-chain fiat renders from cache within ~100 ms of cold launch on a degraded network
- [ ] A cancelled refresh cycle no longer leaves rows at `$0.00` indefinitely
- [ ] One chain's slow/failing RPC does not gate other chains' fiat display
- [ ] Vault header total behavior is unchanged

## Related

Three follow-ups found during this investigation, tracked separately — **not** in scope here:
- Multicall3 sub-call failure zeroes a token balance where the per-coin path preserves the last known value (fixed in a separate PR)
- Cancelled refresh cycles don't cancel their in-flight requests (observed ETH fetched 6×)
- Chain rows never adopted `#4693`'s `—` placeholder, so token cells show `—` while rows show a misleading `$0.00`

gastonm5:
Closes #4830

## Problem

The **per-chain fiat total** on the wallet main screen takes far too long to appear, or never appears at all. The vault header total updates fine — only the per-chain rows are stuck.

## Root cause (proven at runtime, not inferred)

`VaultMainChainListView` renders `ChainRowModel.fiatBalance`, a **preformatted `String`** baked when `VaultDetailViewModel.rows` is assigned. `VaultChainCellView` holds `vault` as a plain `let`, so nothing about a rate landing can repaint it.

Three facts combine:

1. **`RateProvider` is a silent cache** — no `ObservableObject`, no `@Published`, no subject. A rate landing is unobservable by design.
2. **`BalanceService.refreshFiat`'s `objectWillChange.send()` is useless for the rows.** It repaints the views, but the rows' *data* is still the stale snapshot. It does work for the header total, which reads `vault.coins` live via `@ObservedObject var vault` — hence the asymmetry.
3. **The only rebuild is the tail of `updateBalance`'s network refresh**, behind a 250 ms debounce, the full balance+price round trip, and `if !Task.isCancelled`.

So the per-chain fiat can only change when a *full, uncancelled, vault-wide network refresh* completes — even though the rates it needs are normally already cached.

### Log timeline

Instrumented `origin/main`, real 21-coin vault (Ethereum + TerraClassic), dedicated en_US sim. Cold launch:

```
t=  0.001  chainRows() BUILD: RateProvider has rates for 0/21 coins
t=  0.003  VM.SEED rows -> Ethereum=$0.00, TerraClassic=$0.00
t=  0.003  VM.TASK aborted during 250ms debounce (cancelled) -> rows NEVER rebuilt this cycle
t=  0.066  RateProvider.WARMUP loaded 34 PERSISTED rates from SwiftData -- fiat renderable from here on
t=  0.258  BalanceService.updateBalances ENTER coins=21
t= 60.632  BalanceService.BALANCES fetched (21 updates)
t= 60.632  fetchRatesSafely THREW after 60.37s: timeout -- REMAINING RATES NEVER FETCHED
t= 60.655  chainRows() BUILD: RateProvider has rates for 15/21 coins
t= 60.657  VM.ROWS REBUILT -> TerraClassic=$44.32, Ethereum=$27.00
```

The decisive detail: **the price fetch threw** — zero new rates were saved that cycle. The correct fiat shown at t=60.657 came entirely from rates cached at **t=0.066**. The list sat at `$0.00` for **60.6 seconds with the right answer already in RAM**.

On a healthy network the same gap is ~1.7 s (rates t=2.345, rows t=4.071) — which is why this reads as intermittent rather than always-broken.

### "Slow" and "never" are one bug, two triggers

- **Slow** — rows wait for the slowest balance RPC in the entire vault (`withTaskGroup` waits for all) plus `await pricesDone`. One dead chain gates every chain's fiat.
- **Never** — `refresh()` fires from ~6 entry points, each cancelling the previous task. A cycle cancelled after `updateBalances` returns skips the rebuild entirely (the `t=0.003` line above is that path firing). Nothing retries until the next appear/pull, so it can stay `$0.00` indefinitely.

**Affects all chains equally** — a view-model binding defect, not chain-specific, and not the Multicall3/EVM path. All entry paths (cold launch, revisit, pull-to-refresh, currency change) share the same tail rebuild.

## Which commit

**`2dee43844`** ("reactive + cheap vault chain list via value-type row projection", #4500, first in `v1.39.60`) — it replaced a live-computed `fiatBalance` (recomputed on every `vault.objectWillChange`) with the baked snapshot.

The defect is **latent and network-dependent** (~1.7 s good network, 60 s+/never on a bad one), so it surfaces intermittently rather than at a clean release boundary.

### `30fb07391` (#4693, price/balance decoupling) is innocent

It was the prime suspect. Cleared two ways:

1. **Static** — `VaultDetailViewModel.updateBalance` is *byte-identical* between `v1.40.63` (last good) and `origin/main`. The row-rebuild trigger never changed.
2. **A/B experiment** — one binary, env-gated ordering, same vault/sim:

   | Run | Ordering | Rates in RAM | Rows correct |
   |---|---|---|---|
   | A | POST-#4693 (concurrent) | t=0.065 | **t=60.657** |
   | B | PRE-#4693 (prices-first, as v1.40.63) | t=0.065 | **t=60.656** |

   Identical to within 1 ms. If anything #4693 *improves* the worst case: rows land at `max(T_price, T_balance)` instead of `T_price + T_balance`.

## Fix

Fix the broken binding, not the timing — no timers, no forced refresh, no re-coupling:

1. **`RateProvider.ratesDidChange`** — a `PassthroughSubject` fired on the persisted-rate warmup and every `save(rates:)`. The missing reactive link.
2. **`VaultDetailViewModel`** subscribes and rebuilds its projection's fiat **in place, preserving row order**. Pure, network-free, no-ops when unchanged.
3. **Send from a `defer`** right after the cache mutation in `save`, so a throwing persist still announces the already-changed cache.

Balance/price fetching is untouched — #4693's Multicall3 batching and decoupling keep their wins.

**Why order-preserving:** `chainRows(vault:)` sorts by fiat descending and rates land once per chain group, so a re-sorting rebuild would reshuffle the list several times per refresh — the "stale-then-fresh double reorder" this view model deliberately avoids. The rate rebuild refreshes values only; the authoritative re-sort stays on `updateBalance`'s tail.

## Tests

`BalancePriceDecouplingTests` passes while the UI is broken because it tests the wrong seam: it mocks `CryptoPriceService` and asserts on `Coin.balanceInFiatForDisplay` — the **per-coin lazy** path used by `TokenCellView`/`CoinDetailHeaderView`, which the chain list does not use. It never touches `VaultDetailViewModel`, `ChainRowModel`, or the projection.

New tests pin the seam that actually failed:
- `testRowsRefreshWhenRateLandsAfterProjectionWasBuilt`
- `testRowsRefreshOnRateArrivalWithoutAnyBalanceRefresh` — guards against "fixing" this by re-coupling to balances
- `testRateArrivalRefreshesFiatWithoutReorderingRows`
- `testUnrelatedRateDoesNotRepublishRows`

## Follow-ups (not in this PR)

- **`$0.00` vs `—` inconsistency.** #4693 added `String.fiatPlaceholder` + `Coin.balanceInFiatForDisplay` for the cold-start "misleading $0.00 frame", but chain rows use `totalBalanceInFiatDecimal` and never adopted it — token cells show `—`, chain rows show `$0.00`. Adopting it for rows is a behaviour change beyond restoring refresh semantics, so it's left for a decision. Largely moot once rates render at t≈0.066.
- **Multicall3 zeroes on sub-call failure.** `decodeAggregate3Results` maps a failed sub-call to `nil` → `?? 0` → balance persisted as 0, whereas the per-coin path returns `nil` and *preserves* the last known balance.
- **Cancelled cycles keep their network work** (ETH ×6, LUNC ×3 duplicate fetches observed in one cycle).
- **`VaultDetailViewModel` actor isolation is implicit** — touches SwiftData `@Model`s from non-isolated methods, relying on `.receive(on: .main)`/call-site discipline rather than `@MainActor`. Pre-existing across the whole VM; annotating it is a separate refactor.

## Verification

- Gate: `make test` — see comment below for the literal marker + counts.
- Codex read-only review, per-commit + whole-branch: 1 Low + 1 Minor fixed (order preservation, `defer` send); 1 Medium deferred with reason (VM-wide `@MainActor` = separate refactor). Codex independently confirmed the first three new tests fail without the fix.
- Diagnostic logging stripped before commit.

**Pending on-device:** cold launch on a real multi-chain vault under a degraded network — per-chain fiat should render from cache within ~100 ms instead of waiting on the balance round trip.

## Summary by CodeRabbit

* **New Features**
  * Wallet details now update fiat values and projections automatically when exchange rates become available or change.
  * Existing chain rows retain their order while refreshed rate information is applied.

* **Bug Fixes**
  * Improved wallet balance displays when rates load after the initial screen render.

* **Tests**
  * Added coverage for rate-driven updates, projected fiat values, row ordering, and unrelated rate changes.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
+++ b/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
@@ -27,6 +27,15 @@ final class RateProvider {
 
     static let shared = RateProvider()
 
+    /// Fires on the main actor whenever the cached rate set changes — the initial
+    /// load of persisted rates and every `save(rates:)`.
+    ///
+    /// Rates are an otherwise silent cache: callers that render a *precomputed*
+    /// fiat value (rather than reading `rate(for:)` inside `body`) have no way to
+    /// learn that a rate arrived, so their fiat stays frozen at whatever was
+    /// cached when they last recomputed. This signal is that missing link.
+    let ratesDidChange = PassthroughSubject<Void, Never>()
+
     static func cryptoId(for coin: CoinMeta) -> CryptoId {
         switch coin.chain.chainType {
         case .EVM, .Solana, .Sui, .THORChain:
@@ -66,6 +75,10 @@ final class RateProvider {
             do {
                 let objects = try Storage.shared.modelContext.fetch(descriptor)
                 self.rates = Set(objects.map { Rate(object: $0) })
+                // Persisted rates make fiat renderable immediately on a warm
+                // start; announce them so views that cached a pre-load fiat
+                // value recompute instead of waiting on a network refresh.
+                self.ratesDidChange.send()
             } catch {
                 print("Failed to load rates: \(error.localizedDescription)")
             }
@@ -155,5 +168,6 @@ final class RateProvider {
         }
 
         try Storage.shared.modelContext.save()
+        ratesDidChange.send()
     }
 }
--- a/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
+++ b/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
@@ -149,6 +149,12 @@ final class RateProvider {
         let newRateIds = Set(newRates.map { $0.id })
         rates = rates.filter { !newRateIds.contains($0.id) }.union(newRates)
 
+        // The in-memory cache is what rendering reads, and it has already
+        // changed — announce it even if the persistence below throws, or a
+        // failed write would leave the UI showing fiat that contradicts the
+        // cache until the next refresh.
+        defer { ratesDidChange.send() }
+
         // Update existing or insert new rates
         for rate in newRates {
             let rateId = rate.id // Capture the value outside the predicate
@@ -168,6 +174,5 @@ final class RateProvider {
         }
 
         try Storage.shared.modelContext.save()
-        ratesDidChange.send()
     }
 }
--- a/VultisigApp/VultisigApp/Features/Wallet/ViewModels/VaultDetailViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Wallet/ViewModels/VaultDetailViewModel.swift
@@ -56,13 +56,23 @@ class VaultDetailViewModel: ObservableObject {
             .store(in: &cancellables)
     }
 
-    /// Recomputes the projection against the currently cached rates. No-ops when
-    /// nothing changed, so the repeated `save(rates:)` calls of a single refresh
-    /// collapse into at most one publish.
+    /// Recomputes the projection's fiat against the currently cached rates,
+    /// **keeping the current row order**.
+    ///
+    /// `chainRows(vault:)` orders by fiat descending, and rates land once per
+    /// chain group, so re-sorting here would reshuffle the list under the user
+    /// several times per refresh. A rate arriving should refresh the values, not
+    /// reorder — the authoritative fiat-desc re-sort stays on `updateBalance`'s
+    /// tail, which keeps the deliberate "no stale-then-fresh double reorder"
+    /// behaviour intact. No-ops when nothing changed.
     private func rebuildRowsForRateChange() {
-        guard let vault = rowsVault, !chains.isEmpty else { return }
-        let rebuilt = logic.chainRows(vault: vault)
-        guard rebuilt != rows else { return }
+        guard let vault = rowsVault, !rows.isEmpty else { return }
+        let byChain = Dictionary(
+            logic.chainRows(vault: vault).map { ($0.chain, $0) },
+            uniquingKeysWith: { _, latest in latest }
+        )
+        let rebuilt = rows.compactMap { byChain[$0.chain] }
+        guard rebuilt.count == rows.count, rebuilt != rows else { return }
         rows = rebuilt
     }
 
--- a/VultisigApp/VultisigApp/Features/Wallet/ViewModels/VaultDetailViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Wallet/ViewModels/VaultDetailViewModel.swift
@@ -5,6 +5,7 @@
 //  Created by Amol Kumar on 2024-03-09.
 //
 
+import Combine
 import Foundation
 import SwiftUI
 
@@ -32,8 +33,37 @@ class VaultDetailViewModel: ObservableObject {
 
     private let bannerStore: PromoBannerDismissalStoring
 
-    init(bannerStore: PromoBannerDismissalStoring = PromoBannerDismissalStore.shared) {
+    /// The vault `rows` currently projects. Held so a rate arrival can rebuild the
+    /// projection without waiting for the next `updateBalance(vault:)` call.
+    private weak var rowsVault: Vault?
+    private var cancellables = Set<AnyCancellable>()
+
+    init(
+        bannerStore: PromoBannerDismissalStoring = PromoBannerDismissalStore.shared,
+        rateProvider: RateProvider = .shared
+    ) {
         self.bannerStore = bannerStore
+
+        // `rows` bakes fiat into a formatted string, so it is inert to a rate
+        // landing afterwards. Without this subscription the only rebuild happens
+        // at the tail of `updateBalance`'s network refresh, so the per-chain fiat
+        // stays at "$0.00" for the whole balance round trip — and never updates at
+        // all when that task is cancelled by a re-entrant refresh. Rebuilding here
+        // is pure and network-free.
+        rateProvider.ratesDidChange
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] in self?.rebuildRowsForRateChange() }
+            .store(in: &cancellables)
+    }
+
+    /// Recomputes the projection against the currently cached rates. No-ops when
+    /// nothing changed, so the repeated `save(rates:)` calls of a single refresh
+    /// collapse into at most one publish.
+    private func rebuildRowsForRateChange() {
+        guard let vault = rowsVault, !chains.isEmpty else { return }
+        let rebuilt = logic.chainRows(vault: vault)
+        guard rebuilt != rows else { return }
+        rows = rebuilt
     }
 
     func filteredChains(in vault: Vault) -> [Chain] {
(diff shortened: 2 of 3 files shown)


gastonm5:
## Gate

```
** TEST SUCCEEDED **
Executed 3060 tests, with 1 test skipped and 0 failures (0 unexpected) in 62.975 seconds
```

Full suite (`xcodebuild test`, the exact invocation `make test` runs — scheme `VultisigApp`, `-skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO`), dedicated en_US sim (iPhone 16 Plus, `E20764E2-9076-4DF9-B23E-1E6A079C22BF`). No failures — the known `JoinKeysignDoubleTapGuardTests` flake did not fire.

New tests:
```
testRateArrivalRefreshesFiatWithoutReorderingRows        passed (0.005s)
testRowsRefreshOnRateArrivalWithoutAnyBalanceRefresh     passed (0.004s)
testRowsRefreshWhenRateLandsAfterProjectionWasBuilt      passed (0.004s)
testUnrelatedRateDoesNotRepublishRows                    passed (0.504s)
```

SwiftLint: 0 violations on all three changed files.

### Codex review (read-only, per-commit + whole-branch)
| Finding | Severity | Triage |
|---|---|---|
| Rate-driven rebuild re-sorts rows → list reshuffles once per chain group mid-refresh | Low | **Fixed** (`8ce9459d`) — rebuild in place, preserve order; authoritative fiat-desc re-sort stays on `updateBalance`'s tail |
| `save(rates:)` mutates cache before persisting but only signals after → throwing write leaves cache updated + unannounced | Minor | **Fixed** (`735d4d35`) — send from a `defer` right after the cache mutation |
| `VaultDetailViewModel` actor isolation implicit (relies on `.receive(on: .main)`, not `@MainActor`) | Medium | **Deferred** — pre-existing across the whole VM (`groupChains`, `filteredChains`, the existing seed all touch SwiftData models from non-isolated methods); annotating it is a separate refactor, out of scope for restoring refresh semantics |

Codex independently confirmed the first three new tests fail without the fix, and cleared: no retain cycle, no missed-warmup race, no lock held across `send`, no row drop/duplication in the order-preserving rebuild.

johnnyluo (approved):
LGTM

PR #4825, opened 2026-07-16
fix(limit-swap): honor the entered source chain + put asset selection before the execution condition

gastonm5:
Two reviewer suggestions on the THORChain limit-swap Place flow.

## 1. The Limit tab ignored the source chain you entered with

**Repro:** THORChain chain detail → "Swap" → the Market tab correctly preselects THORChain. Switching to "Limit" silently switched the source to **BTC**.

### Why the override existed (the heuristic is NOT deleted)

`preferredLimitSourceChain` deliberately prefers a high-value, THORChain-routable source (BTC, then ETH). The shared market default sorts alphabetically (`SwapCoinsResolver` picks the first held coin), which lands on a cheap source like RUNE and presents an untradeable-looking `RUNE→BTC` default — the preference exists to fix exactly that.

The bug is ordering. Step 2 of the function already said *"keep the market default when it's THORChain-routable and not a self-pair"* — which would have honored THORChain — but the BTC/ETH preference ran **first**, so for any vault holding BTC an explicitly-chosen THORChain source was **always** overridden.

The root defect: the heuristic couldn't distinguish two situations that arrive as the same non-optional `marketDefault`:
- **(a)** an arbitrary alphabetical fallback — no user intent — where the BTC/ETH preference is genuinely wanted;
- **(b)** an **explicit user selection** (arriving from chain detail via Swap) — real intent that must be honored.

### The fix: teach it the difference, keep the preference

`SwapDetailsScreen.fromCoin` is `Coin?` and is non-nil exactly when the user entered from a specific chain/coin. That optional is already the intent signal — `SwapDetailsViewModel.load` resolves it as `initialFromCoin ?? defaultFromCoin` ("explicit, else alphabetical"). `limitInitialFromCoin` was throwing that signal away by passing the already-resolved non-optional coin.

It's now threaded through as `isSourceExplicit`, and step 2's existing rule simply runs **first** when intent is real. **The BTC/ETH preference is unchanged and still governs the no-intent path.**

Edge cases fall out of the same gate rather than needing special-casing — an explicit source is honored **only** when THORChain-routable and not a self-pair:
- **explicit but unroutable** (SOL/TON) → falls through to the preference, so it can never seed an unplaceable order;
- **explicit source == target** (self-pair) → falls through. This preserves the **buy-VULT** entry (explicit ETH source + VULT-on-Ethereum target is same-chain), which still resolves to BTC exactly as before.

Pure functions stay pure; resolution stays in the view — extending the existing design rather than restructuring it.

## 2. Asset selection now precedes the execution condition

`LimitAssetSwapForm` (Sell / swap / Buy) now renders **above** `LimitPriceCard` (target price / "Execute When"); expiry stays last. The user picks what they're trading before stating the condition it executes under.

Verified couplings, not just the two lines:
- **Keyboard accessory:** the single `focusedField`-driven `.keyboard` toolbar is attached to the **outer VStack**, not to the cards, so the previously-fixed merged-accessory bug cannot resurrect from a sibling reorder.
- **Price-card chips:** kept. "When 1 BTC is worth X ETH" *is* the condition sentence, so it has to name both assets — the chips label the price's units and double as picker shortcuts; they don't duplicate the Sell/Buy controls.
- **Focus order** is tap-driven and follows the new visual order.

## Before / after (item 2 + item 1 in one shot)

| Before | After |
|---|---|
| Price card first; **Sell = Bitcoin** despite entering from THORChain | Sell = **THORChain / RUNE** first, then the condition |

## Verification

- **Runtime, iOS simulator** — drove the actual repro (THORChain chain detail → Swap → Limit) on a dedicated en_US sim, before **and** after:
  - **before:** source seeded **BTC**, price card above the asset form;
  - **after:** source stays **RUNE**, asset form above the price card.
- **Gate:** `** TEST SUCCEEDED **` — Executed **3065 tests, 1 skipped, 0 failures**.
- **macOS:** `** BUILD SUCCEEDED **` with the change. ⚠️ A macOS **runtime** UI pass was **not** completed — the UI-test target doesn't compile for macOS on `main` (pre-existing: `XCUIElement+Helpers.swift`), and it also lacks `AccessibilityID` in its sources. Item 1 is platform-independent pure logic, and the only platform-conditional code near item 2 (the `#if os(iOS)` keyboard toolbar) is untouched — but a manual macOS look at the Limit form is still worth doing before merge.
- **Tests:** `preferredLimitSourceChain`'s contract changed, so existing coverage was updated and new cases added for both paths — explicit THORChain honored, explicit non-preferred (LTC) honored, no-intent still prefers BTC/ETH, explicit-but-unroutable (SOL/TON) falls back, explicit self-pair avoided. Also updated `LimitSwapPayloadGasTests`, which covers `limitDefaultSourceCoin` over real coins.
- Cross-model reviewed (per-commit + whole-branch); no blocking findings.

## Notes for the reviewer

- Honoring intent deliberately restores `RUNE→BTC` **for an explicit THORChain entry** — that's the requested behavior. The no-intent default still prefers BTC/ETH, so the original fix is preserved where it applies.
- **Known limitation (deliberate, out of scope):** the signal is *entry* intent. A user who arrives with no intent and then manually changes the source in the Market tab still gets the heuristic on the Limit tab — the picker binds `$vm.fromCoin` directly, so capturing "user touched the source" needs a market-form binding change. Same defect class, much rarer; left out to keep this diff clear of PR #4816's conflict surface.

## Summary by CodeRabbit

* **New Features**
  * Limit swaps now preserve an explicitly selected source asset when it is routable.
  * Improved fallback selection chooses a suitable native asset when the selected source cannot be used.
  * Reordered the limit-swap layout to show asset selection before the price card.

* **Bug Fixes**
  * Prevented self-pair swaps and unsupported source assets from producing invalid limit-swap entries.

* **Tests**
  * Expanded coverage for explicit source selection, routing fallbacks, and source-asset resolution.

### DIFF
--- a/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitMath.swift
+++ b/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitMath.swift
@@ -147,13 +147,21 @@ func marketProbeAmount(
     return probe
 }
 
-/// Preferred default SOURCE chain for the **limit-swap entry only**. The shared
-/// market default sorts alphabetically (`SwapCoinsResolver` picks the first held
-/// coin), which lands on a cheap source like RUNE and presents an
-/// untradeable-looking RUNE→BTC default. Prefer a high-value, liquid,
-/// THORChain-routable native source the vault actually holds — BTC, then ETH —
-/// skipping any candidate that collides with the target chain (which would be a
-/// self-pair).
+/// Preferred default SOURCE chain for the **limit-swap entry only**.
+///
+/// Two different situations arrive here as the same `marketDefaultChain`, and
+/// `isSourceExplicit` is what tells them apart:
+///
+/// - **No intent** (`false`) — the shared market default sorts alphabetically
+///   (`SwapCoinsResolver` picks the first held coin), which lands on a cheap
+///   source like RUNE and presents an untradeable-looking RUNE→BTC default. Here
+///   we prefer a high-value, liquid, THORChain-routable native source the vault
+///   actually holds — BTC, then ETH — skipping any candidate that collides with
+///   the target chain (which would be a self-pair).
+/// - **Explicit intent** (`true`) — the user entered the swap from a specific
+///   chain/coin, so that source is a real choice, not an alphabetical accident.
+///   Honor it whenever it is usable, and only fall back to the preference above
+///   when it isn't (unroutable, or a self-pair with the target).
 ///
 /// **Never returns `targetChain` while any held, THORChain-routable alternative
 /// native source exists** (a same-chain source→target is not THORChain-routable).
@@ -163,9 +171,17 @@ func marketProbeAmount(
 /// vault `Coin`. Does NOT change the shared market default.
 func preferredLimitSourceChain(
     marketDefaultChain: Chain,
+    isSourceExplicit: Bool,
     targetChain: Chain,
     availableNativeChains: Set<Chain>
 ) -> Chain {
+    // 0. An explicitly chosen source is real user intent — keep it when it's
+    //    usable (THORChain-routable and not a self-pair). Same rule as step 2;
+    //    intent is what promotes it above the BTC/ETH preference. An unusable
+    //    explicit source falls through rather than seeding an unplaceable order.
+    if isSourceExplicit, marketDefaultChain != targetChain, isThorchainRoutable(chain: marketDefaultChain) {
+        return marketDefaultChain
+    }
     // 1. High-value routable sources the vault holds (BTC → ETH), excluding a
     //    self-pair with the target. (BTC/ETH are always THORChain-routable.)
     for candidate in [Chain.bitcoin, .ethereum]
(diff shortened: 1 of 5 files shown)


gastonm5:
### Runtime evidence (iOS simulator, dedicated en_US sim)

The same automated pass was driven against **pre-fix** and **post-fix** code, entering via THORChain chain detail → Swap → Limit. Accessibility-tree geometry from the Limit tab (y-coordinates show the card order):

**BEFORE** — price card on top, and the source was overridden to BTC despite entering from THORChain:
```
StaticText  y=189.8  'When 1'      <- price card FIRST
Button      y=184.0  'BTC'         <- condition names BTC
StaticText  y=370.3  'Sell'        <- asset form SECOND
StaticText  y=370.3  '0 BTC'       <- source = BTC  (the bug)
Button      y=402.3  'BTC'
StaticText  y=491.3  'Buy'
```

**AFTER** — asset form on top, and the entered THORChain source is honored:
```
StaticText  y=180.0  'Sell'        <- asset form FIRST
StaticText  y=180.0  '1 RUNE'      <- source = RUNE  (entered source honored)
Button      y=212.0  'RUNE'
StaticText  y=301.0  'Buy'
StaticText  y=434.8  'When 1'      <- price card SECOND
Button      y=429.0  'RUNE'
```

The before/after screenshots are on the machine that produced this run:
- `BEFORE-limit-tab.png` — price card first, `Sell: Bitcoin / BTC`
- `AFTER-limit-tab.png` — `Sell: THORChain / RUNE` first, then `When 1 RUNE is worth … USDC`, then Expiry

(GitHub image upload isn't available from the CLI, so they need to be dragged in manually.)

Gate: `** TEST SUCCEEDED **` — Executed 3065 tests, 1 skipped, 0 failures. macOS: `** BUILD SUCCEEDED **` (macOS runtime UI pass still pending — see the PR body).

johnnyluo (approved):
LGTM

PR #4829, opened 2026-07-16
fix(balances): stop a partial Multicall3 failure silently zeroing an EVM balance

gastonm5:
## Problem

A **partial** Multicall3 failure silently persisted a `0` balance over a coin that has funds.

`aggregate3` is called with `allowFailure = true`. When an individual sub-call reverts, the batch **still succeeds at the top level** — the sub-call just comes back with `success = false`. `decodeAggregate3Results` correctly maps that to `nil`, and then the callers collapsed it with `?? 0`:

- `EvmServiceStruct.swift:254` — `native = decoded[index] ?? 0`
- `EvmServiceStruct.swift:260` — `balances[contractAddress] = decoded[index] ?? 0`
- `BalanceService.swift:230` — `let value = result.native ?? 0`
- `BalanceService.swift:234` — `let value = result.balances[coin.coinMeta.contractAddress] ?? 0`

A reverting or erroring token contract therefore wrote a real `0` to a funded coin. No throw, no fallback, no log.

Introduced by `30fb07391` (#4693), first shipped in **v1.41.64**.

## The non-obvious part: a partial failure is not a thrown failure

`BalanceService` already documents that a batch failure "never zeroes balances — it just reverts to today's behaviour", and that is **true for a _thrown_ error**: a network blip, an unexpected decode, or a failed runtime gate degrades the whole group to the per-coin path. That comment is why this area reads as if the case were already handled.

But a **partial** failure never throws. The top-level `eth_call` succeeded, so the `catch` never fires and the per-coin fallback never runs. The existing guard covers the case that *does* throw — which reads as complete coverage right up until the case that doesn't.

## The doc comment that justified it was wrong

`Multicall3.swift:87-91` justified the `nil → 0` mapping:

> …or `nil` when the sub-call failed or returned fewer than 32 bytes — the caller maps `nil` to a `0` balance, **reproducing today's per-call "balance check failed → 0" fallback**.

**There is no such fallback.** The per-coin path preserves the last known balance:

1. `EvmServiceStruct.getBalance` → `RpcServiceStruct.intRpcCall` **throws** on a JSON-RPC error. It does not substitute zero.
2. `fetchBalanceUpdate` catches it and leaves `rawBalance == nil`.
3. `CoinBalanceUpdate.hasUpdates` goes `false` (EVM carries no staked/bonded balance).
4. `applyBalanceUpdates`' `for update in updates where update.hasUpdates` **skips the coin entirely** — the stored balance survives. The write is additionally guarded by `if let rawBalance`.

So the `?? 0` was a **regression**, not parity. The comment is corrected here.

The phrase does exist in the codebase — `EvmServiceStruct.getTokensFallback` carries *"If balance check fails, assume zero balance"* — but that is the **token-discovery** walk, which merely omits a token from the discovered list. It never persists a balance. The phrase exists; the behaviour does not.

## Fix

Preserve the `nil` all the way to the write. The plumbing already existed — only the callers threw it away.

- **`Multicall3.mapBalances`** (new, pure) — maps results back onto the call plan while keeping the distinction: a failed sub-call is **omitted**; a sub-call that genuinely returned `0` **is recorded**. Returns `nil` on a count mismatch (whole-batch failure). Extracted so the mapping has a unit-testable seam — `fetchERC20Balances` does I/O and `EvmService` is a non-injectable enum factory, so there was nowhere to test this before.
- **`EvmServiceStruct.fetchERC20Balances`** — delegates; both `?? 0` gone.
- **`BalanceService.fetchEvmBatchBalances`** — partitions into ready updates and failed coins; failed coins go through the **existing** `fallbackPerCoin`, and the failure is now **logged**. If the retry fails too, the per-coin path's own `nil` preservation keeps the last known value.

The batch path is not restructured — the I/O flow is untouched. A permanently reverting token now costs one extra per-coin `eth_call` per refresh, which is exactly its pre-#4693 cost.

### A genuine zero is still written

An empty wallet is a legitimate `0` and must still render as `0` — the whole point is that "failed" and "actually zero" stop being the same value. `testGenuineZeroBalanceStillWritesZero` and `testZeroAndFailureAreDistinguishableInTheSameBatch` pin both halves in a single batch.

## Verification

**Runtime — proven against live Ethereum mainnet.** A temporary probe drove the real production path end-to-end (`updateBalances` → `encodeAggregate3` → real `eth_call` → `decodeAggregate3Results` → `mapBalances` → `partitionBatchResult` → `fallbackPerCoin` → `applyBalanceUpdates` → SwiftData write) against a real funded mainnet address, with a **reverting contract injected into the batch** as the failing sub-call (`balanceOf` on the Multicall3 contract itself reverts → `success = false` for that entry only, while the batch still succeeds at the top level — exactly the partial failure that never throws).

```
PROBE ---- BEFORE ---- ETH=0 USDC=0 BOGUS=123456789
PROBE ---- AFTER  ---- ETH=226351975215885383039011 USDC=49052076175 BOGUS=123456789
PROBE ---- VERDICT ---- BOGUS preserved: true | BOGUS zeroed: false
** TEST SUCCEEDED **
```

The affected token's balance is **preserved**, while its siblings update with real mainnet values. The run also emitted `Multicall3 sub-call failed for BOGUS on Ethereum; retrying per-coin`, which confirms the partial-failure path genuinely executed — so the pass is not vacuous. The probe was removed before commit; only the intentional failure log ships.

**Honest gap — the pre-fix A/B control could not be captured.** I attempted 7 runs of the same harness against the pre-fix code to show `BOGUS → 0`; every one timed out at exactly 60s (URLSession timeout) with the whole batch failing, which exercises the *thrown* path, not the partial one. This is **environmental, not behavioural**: the identical post-fix build that succeeded above also began timing out the same way once the endpoint started throttling repeated runs, and the two builds issue byte-identical network calls (only the result *mapping* differs). So the pre-fix zeroing rests on the static trace — `decoded[index] ?? 0` → `String(BigInt(0))` → `"0"` → `hasUpdates == true` → written — which Codex independently confirmed, rather than on an observed run.

**Gate.** `** TEST SUCCEEDED **` — `Executed 3070 tests, with 1 test skipped and 0 failures (0 unexpected)`. SwiftLint clean. No `Package.resolved` churn.

## Interaction with the `—` placeholder follow-up (noted, not fixed)

This fix makes "never successfully fetched" a *distinguishable* state for the first time: a coin whose sub-call fails on its very first refresh now keeps `rawBalance` unset rather than being written to `0`. What the chain rows should **display** for a never-fetched balance is the separate placeholder question raised in the sibling investigation (chain rows never adopted #4693's `String.fiatPlaceholder`, so token cells show `—` while chain rows show `$0.00`). Left for a human decision.

## Scope

Deliberately independent of #4824 (per-chain fiat staleness) — disjoint file sets, no overlap.

Out of scope: `—` placeholder adoption in chain rows; cancelled refresh cycles not cancelling in-flight requests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed EVM balance reads are no longer reported as zero balances.
  * Failed token reads are omitted and retried individually, preserving existing balances.
  * Genuine zero balances continue to display correctly.
  * Unexpected batch results now trigger an error instead of producing incorrect data.

* **Tests**
  * Added coverage for mixed success and failure scenarios, native balances, token balances, retries, and result ordering.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/EVM/Multicall3.swift
+++ b/VultisigApp/VultisigApp/Blockchain/EVM/Multicall3.swift
@@ -87,8 +87,12 @@ enum Multicall3 {
     /// Decodes the `(bool success, bytes returnData)[]` returned by `aggregate3`,
     /// in the same order the calls were built. Each entry is the `uint256` parsed
     /// from `returnData`, or `nil` when the sub-call failed or returned fewer than
-    /// 32 bytes — the caller maps `nil` to a `0` balance, reproducing today's
-    /// per-call "balance check failed → 0" fallback.
+    /// 32 bytes.
+    ///
+    /// `nil` means "this read failed", NOT "the wallet holds zero" — the two must
+    /// stay distinguishable all the way to the write. A failed read has to preserve
+    /// the last known balance (what the per-coin path does), whereas a genuine zero
+    /// is a legitimate balance to persist. See `mapBalances`.
     static func decodeAggregate3Results(hex: String) -> [BigInt?] {
         guard let data = Data(hexString: hex.stripHexPrefix()) else { return [] }
         let bytes = [UInt8](data)
@@ -152,6 +156,46 @@ enum Multicall3 {
         }
 
         return results
+    }
+
+    // MARK: - Result mapping
+
+    /// Maps `decodeAggregate3Results` output back onto the inputs that produced it,
+    /// in the order `fetchERC20Balances` builds the calls: the optional native
+    /// `getEthBalance` first, then one `balanceOf` per contract address.
+    ///
+    /// Preserves the success/failure distinction the decoder encodes as `nil`: a
+    /// failed sub-call is **omitted** from `balances` (and leaves `native` nil)
+    /// rather than being recorded as `0`, so the caller can retry just that coin
+    /// instead of persisting a bogus zero over a funded one. A sub-call that
+    /// genuinely returned `0` **is** recorded — an empty wallet is a real balance.
+    ///
+    /// Returns `nil` when `decoded` doesn't match the call plan, which the caller
+    /// treats as a whole-batch failure.
+    static func mapBalances(
+        decoded: [BigInt?],
+        includeNative: Bool,
+        contractAddresses: [String]
+    ) -> (native: BigInt?, balances: [String: BigInt])? {
+        let expectedCount = (includeNative ? 1 : 0) + contractAddresses.count
+        guard decoded.count == expectedCount else { return nil }
+
+        var index = 0
+        var native: BigInt?
+        if includeNative {
+            native = decoded[index]
+            index += 1
+        }
+
+        var balances: [String: BigInt] = [:]
+        for contractAddress in contractAddresses {
+            if let value = decoded[index] {
+                balances[contractAddress] = value
+            }
+            index += 1
+        }
+
+        return (native, balances)
     }
 
     // MARK: - Hex helpers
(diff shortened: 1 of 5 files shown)


johnnyluo (approved):
LGTM

Issue #4827, opened 2026-07-16
[Add] Adopt Icons V3 across all screens [Design System / Icons]


realpaaao:
Adopt [Icons V3](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78527-123296) across all screens. Chevrons stay on the old pack (per design) — existing chevron assets are untouched.

> **Blocked by #4826** — the Figma screens currently mix three icon sources, and the name-mapping table lands there. Don't start the swap until it's answered.

## How this lands (read first)

iOS icons live in `VultisigApp/VultisigApp/Assets.xcassets/Icons/` — **150 UI icons**, SVG imagesets, kebab-case. They're referenced by **string literal**: `Icon(named: "chevron-down")`. There is no registry; the imageset *name* is the API. `Components/Icons/Icon.swift` centralizes rendering (size, tint), not naming.

**Keep the names identical and this is a pure asset swap: replace the `.svg` inside each `.imageset`, keep `Contents.json`, zero Swift edits.**

Precedent — `e2d419bb2 "fix: updated settings icons"`: three icons redrawn in place = **0 Swift changes**; the one icon that was *renamed* = 1 Swift edit (`SettingsOption.swift`). Any rename costs an edit per referencing site, and ~114 Swift files carry icon-name literals.

⚠️ **Renames fail silently.** `Icon(named: "typo")` compiles fine and renders a blank rectangle at runtime. Only 3 snapshot tests exist, so CI will not catch it either. This is exactly why #4826 pins names.

## Screens to verify

Verify rendered icons match design; check tint and optical size on each flow.

- [ ] [Main View (iOS 26)](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=48948-108982)
- [ ] [Main View (Android / iOS older versions)](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=49057-156999)
- [ ] [Onboarding Flow V3 (Main device)](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=60739-32143)
- [ ] [Vault Setup](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=59111-92965)
[...]

Issue #4826, opened 2026-07-16
[Design] Resolve icon-source conflicts in Mobile App screens before Icons V3 adoption [Design System / Icons]


realpaaao:
Tracking issue for a **design decision**, filed here for visibility — it is cross-platform and blocks the iOS / Android / Desktop icon adoption issues.

## Context

The [Mobile App page](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=3-1510) currently places icons from **three different sources**: the new [Icons V3](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78527-123296) library, the old pack, and a third set that is not in the V3 library at all (`Arrows/…`, `Files/…`, `Finance/…`, `CaretLeft`).

Design has confirmed the intent: **chevrons stay on the old pack on purpose; everything else moves to V3.** The screens don't reflect that yet, so implementing "what's in Figma" today would bake in a mixed state.

Engineering can't resolve these — they're design decisions. Each answer below unblocks a mechanical change.

## 1. Chevrons — which component is the keeper, per direction?

"Chevrons stay old pack" is unambiguous as a rule, but the file has **11 distinct chevron/caret labels across ~287 placements**. Left-facing alone has 5 candidates, and the most-used one is **not in the V3 library**:

| label | uses | in V3 library? |
|---|---:|---|
| `IconChevronLeftSmall 2` | 82 | no |
| `CaretLeft` | 67 | no |
| `IconChevronLeftSmall 1` | 49 | yes (old pack) |
| `IconChevronRightSmall 1` | 22 | yes (old pack) |
| `IconChevronDownSmall 1` | 20 | yes (old pack) |
| `Arrows/chevron-right` | 18 | no |
| `chevron-right-small` | 11 | no |
| `chevron-left` | 10 | **yes — V3** |
| `IconChevronLeft 1` | 5 | no |
| `Arrows/chevron-down` | 2 | no |
| `IconChevronTopSmall 1` | 1 | yes (old pack) |

Note `IconChevronLeftSmall` **1** vs **2** — Figma's ` 1`/` 2` suffix marks duplicate names, so there appear to be two components sharing one name, and the one used more isn't in the library.

[...]

PR #4834, opened 2026-07-17
feat(icons): rebuild the icon set on Icons V3, make the compiler the guarantee

gastonm5:
## What & why

The app's icon set is now **Icons V3**, and `Icon` is **type-safe**: it takes an `ImageResource`, not a `String`. A missing or misspelled icon is now a **compile error** instead of a blank rectangle discovered in production.

This replaces five earlier partial attempts. The reason they kept leaving icons un-updated: nobody had established *which V3 icon each of ours becomes*. Names lie in both directions, the Figma screens are a mixed state, and geometry can prove "same drawing" but can't choose a *replacement metaphor* — our set is Lucide, V3 is a different pack that redraws the same concepts. So the 125 ambiguous icons were decided **by a human against rendered art**; the rest fell out of geometric identity.

Because `Icon(named: String)` is deleted (not overloaded), **the build cannot succeed until every call site points at a real V3 symbol.** Green build = complete coverage, by construction.

## Scope

- `Icon(named: String)` → `Icon(_ image: ImageResource)`. `isSystem` removed — `Icon` no longer draws SF Symbols. `DefiButton.isSystemIcon` removed likewise.
- **154 old imagesets → 138.** 120 V3 icons imported under V3 names, 15 brand-keep (`vult-*` tiers, `atom-shield`, `phantom`, glossies), 3 keep-ours (`stars`, `seedphrase`, `active-chain`).
- Every call site repointed to a compile-time symbol: ~17 wrapper components, ~11 computed members, 3 protocol requirements, and the `vult-\(rawValue)` interpolation, all migrated `String → ImageResource`.
- `Coin.logo` / `Chain.logo` stay `String` — decoded from remote JSON, runtime data not compile-time symbols.
- 9 former SF-Symbol sites moved to real assets (`clock`, `clock-rotate-clockwise`, `minus` imported from V3 for this).

## Two live bugs fixed in passing

Both rendered **nothing** and neither the linter nor CI caught them, because a stringly-typed icon name that resolves to no asset just draws blank:
- `Icon(named: "info")` at 4 Cosmos sites — no `info` asset has ever existed → `.circleInfo`.
- QBTC claim banner passed the SF Symbol `exclamationmark.triangle.fill` to `InfoBannerView`, an asset-only wrapper — blank since #4271 → `.triangleWarning`.

## ⚠️ Reviewer: the icon set shrank — 14 merges, some lose a semantic element

V3 has no equivalent for a few of our composed glyphs, so several old icons collapse onto one V3 imageset. **The dropped elements are forced — verified absent from the 1,930-icon V3 catalogue, not an oversight:**

| now renders | was | note |
|---|---|---|
| `cloud` | `cloud`, **`cloud-check`**, **`cloud-check-2`** | V3 has no cloud-with-check |
| `shield-filled` | `shield-check-filled` | V3 has no shield-with-check |
| `cloud-upload` | `cloud-upload`, `cloud-upload-filled`, `arrow-up-from-dot` | filled/outline collapse |
| `check` | `check`, `checkmark-2-small` | |
| `circle-dashed` | `crypto`, `crypto-outline` | |
| `clone-2` | `clipboard-paste`, `external-recipient` | |
| `coins` | `coins`, `ice-hockey` | |
| `folder` | `folder`, `folder-hexagon` | |
| `folder-key` | `folder-key`, `rec-keyframe` | |
| `menu` | `MenuIcon`, `line-3-horizontal` | |
| `qrcode` | `qr-code`, `InitiatingDeviceIcon` | |
| `shield` | `shield`, `secure` | |
| `bolt` | `lightning`, `slippage` | |
| `arrows-rotate-center` | `arrow-bottom-top`, `arrow-left-right` | |

The **cloud-check** and **shield-check** losses are the only two that drop a *meaning* (a "confirmed/backed-up" checkmark) rather than merging near-duplicates — worth a design call on whether those surfaces need the check back some other way.

## Rendering intent

`template-rendering-intent` was set **per icon from the art** — single-colour glyphs are `template` (tintable), multi-colour ones omit the intent (render as original). An unset intent renders as original, which is why the `vult-tiers` badges stay non-template.

## Gate

- `** TEST SUCCEEDED **` — **3118 tests, 0 failures**.
- `make lint-icons` green — retargeted at the ~47 bare `Image("…")` literals that stay String (chain logos/banners), plus dead-art and duplicate-name checks.
- No `Package.resolved` churn.

## Verification honesty

Build-green proves every symbol resolves. It does **not** prove the right glyph renders — spot-checked the home screen, tab bar, settings and the merge sites visually; the DeFi/Tron screens need a funded vault and are compiled-only. Reviewers should eyeball their own most-used flows.

Refs #4826, #4827.


johnnyluo (approved):
LGTM

Issue #4838, opened 2026-07-18
Swap slippage: iOS/Android send `tolerance_bps` (spot-anchored, unusable) while Windows sends `liquidity_tolerance_bps` (movement-anchored) — investigate + unify


realpaaao:
## Summary — investigation, not a fix proposal

Three clients, one protocol, three different levels of on-chain slippage protection. The divergence traces to **two different THORChain quote parameters that are easy to confuse**, and the evidence suggests we may have concluded "an achievable floor is impossible" after testing only the one that can't work.

**Nothing here should be implemented before the live quote diff in step 1 below.** Re-adding a floor on mobile has broken swaps before (#4556 → reverted by #4639) and the current no-floor default is deliberate.

## The two parameters

From thornode's own `openapi.yaml`:

| param | description |
|---|---|
| `tolerance_bps` | *"the maximum basis points **from the current feeless swap price** to set the limit in the generated memo"* |
| `liquidity_tolerance_bps` | *"the maximum basis points of tolerance **for pool price movements** to set the limit in the generated memo"* |

`tolerance_bps` is anchored to the **feeless** price. A swap's own fees and slip eat the budget before pool movement is even considered — which is exactly why a 1% default rejected any swap with real price impact, at quote time, with `emit asset X less than price limit Y`.

`liquidity_tolerance_bps` is anchored to **movement**, so it tolerates the swap's own impact.

## Current state across the clients

| | param sent | default | on-chain floor |
|---|---|---|---|
| **iOS** | `tolerance_bps` (`ThorchainMainnetAPI.swift:180`) | 0 → omitted (`SwapService.defaultThorchainToleranceBps`) | **none** |
| **Android** | `tolerance_bps` (`ThorChainApi.kt:238`) | 0 → omitted (`SwapQuoteSource.kt:80-82`) | **none** |
| **Windows / SDK** | `liquidity_tolerance_bps` (`@vultisig/core-chain` `getNativeSwapQuote.js:24`) | **100 bps** (`DEFAULT_NATIVE_SWAP_SLIPPAGE_TOLERANCE_BPS`) | **1%** |

Neither mobile app sends `liquidity_tolerance_bps` anywhere (grep: zero hits in both repos).

## Why this is worth a look

#4639 concluded:

> no protective-but-achievable value can be derived from the quote fields alone

[...]

PR #4851, opened 2026-07-20
fix(swap): derive toAmountLimit from the signed memo, not expected output

### LINKED ISSUE #4841: `toAmountLimit` in the swap proto always reports 100% of expected output — no caller passes toleranceBps, so it never matches the signed memo
## Summary

`buildThorchainSwapPayload` derives `toAmountLimit` from a `toleranceBps` parameter that **no caller ever passes**, so it always uses the `Auto` default of 0. With tolerance 0 the formula returns the full expected output:

```swift
// SwapPayloadBuilder.swift:128-135
let limit = expected * BigInt(10_000 - toleranceBps) / BigInt(10_000)   // toleranceBps == 0 → limit == expected
```

Two consequences:

**1. It claims a floor that doesn't exist.** On the default path the signed memo carries no `LIM` at all, while the proto says the minimum is 100% of expected. The doc comment at `:86-89` states these stay *"consistent with the signed memo"* — they don't.

**2. It's wrong in the other direction when the user sets slippage.** `transaction.advancedSettings.slippage.bps` *is* threaded into the quote request (`SwapVerifyViewModel.swift:109`, `SwapDetailsViewModel.swift:781`), so the node bakes a real `LIM` at e.g. 97% of expected — but `SwapPayloadBuilder.swift` contains no reference to `slippage` at all (`:97`, `:171-179`, `:192-200`, `:213-221`, `:234-242`), so the proto still carries 100%.

## Impact today: low, but the field is a trap

Nothing on iOS renders `toAmountLimit` — it exists only in the proto shipped to the co-signer. So the current blast radius is cross-device display.

But Windows renders its equivalent field, and gets it wrong in the mirror-image way (vultisig-windows#4392: it labels `expected_amount_out` as "min. payout"). This is precisely the field a future "minimum received" row would reach for, and it's wrong on both defaults.

## Proposed fix

Either:
- **(a)** thread the real `slippageBps` through to `buildThorchainSwapPayload` so `toAmountLimit` matches the memo `LIM` the node actually produced, or
- **(b)** set `toAmountLimit` to `"0"` when we send no tolerance, so it honestly reports "no floor" — Android already hardcodes `"0"` (`SwapTransactionBuilder.kt:115`), and its comment notes the proto field is unused at sign time.

(b) is the smaller change and is honest on the default path; (a) is needed if we ever display it.

Whichever way, fix or remove the `:86-89` doc comment, which currently asserts a consistency that doesn't hold.

## Acceptance
- [ ] `toAmountLimit` either matches the signed memo's `LIM` or is `"0"`
- [ ] Correct when the user sets a custom slippage
- [ ] Doc comment matches behavior
- [ ] Test covering both default and custom-slippage paths

Related: #4838 (the tolerance-parameter question this sits downstream of), vultisig-windows#4392.

gastonm5:
Fixes #4841.

## The bug

`SwapCryptoLogic.buildThorchainSwapPayload` derived the payload's minimum-output floor as:

```swift
let limit = expected * BigInt(10_000 - toleranceBps) / BigInt(10_000)
```

`toleranceBps` defaulted to `SwapService.defaultThorchainToleranceBps`, which is `0`, and **no production caller ever passed it** — all four call sites (maya / thorchain / chainnet / stagenet) omitted the argument. So `toAmountLimit == quote.expectedAmountOut` verbatim: every THORChain and MayaChain swap payload claimed a guaranteed floor of 100% of the expected output.

Only one test passed a nonzero tolerance, exercising a path production never took. That is what kept this invisible.

## Why not just thread the real `slippageBps` through

That was the obvious fix, and live measurement against thornode rules it out. At 0.1 BTC with `tolerance_bps=100`:

| | value |
|---|---|
| real memo `LIM` | `340077095` |
| what the formula gives | `338179686` |

The formula matches **`liquidity_tolerance_bps`** semantics exactly, but `tolerance_bps` is what iOS actually sends — so threading the real slippage through would still report a wrong floor on the custom-slippage path. And on Auto the memo carries no `LIM` at all while the formula keeps returning 100%. Full measurements: https://github.com/vultisig/vultisig-ios/issues/4838#issuecomment-5021569752

## The fix — parse `LIM` out of `quote.memo`

The memo is signed and broadcast verbatim, which makes its `LIM/INTERVAL/QUANTITY` triple the only authoritative statement of the floor and the streaming plan. Parsing it is correct by construction on Auto (`LIM` `0`), on custom slippage (the node's exact number), and under either tolerance parameter — so it also survives whatever #4838 settles on, with no rework.

Parsing is deliberately **all-or-nothing**. A non-swap action, a term that isn't a canonical bare decimal integer, or more terms than the grammar has all yield `"0"` across the board. `"0"` honestly means *no floor asserted*; salvaging a number out of a memo we failed to parse is the same class of false claim this change exists to remove. Legitimately absent trailing terms still read `"0"` — a rapid-swap memo may stop after `LIM`, which is valid grammar rather than a parse failure.

`thorchainToAmountLimit` and its five tests are deleted (four covered the path no production caller took).

## `streamingQuantity` and `streamingInterval` had the same bug

Both are fixed here, from the same parse:

- `streamingQuantity` reported `max_streaming_quantity` — the node's *capacity ceiling*, not the quantity it baked into the memo. `SwapService` already notes that "the chosen quantity is baked into the returned memo".
- `streamingInterval` reported a static per-provider constant. It diverges whenever the anti-rekt path refetches at `interval: 1` and picks the streaming quote: memo says `1`, payload said `0`.

Fixing quantity from the memo while leaving interval on a provider constant would have left the two halves of one memo field disagreeing, so both now come from the same source. This made the `provider` parameter dead, so it is removed from `buildThorchainSwapPayload`.

## Scope note — these fields are inert on iOS today

Nothing renders `toAmountLimit` and no signer reads it; it only travels in the keysign proto to the co-signer. So this is a correctness fix for the cross-device payload and for any future "minimum received" display, not a user-visible change. **Tests are the only verification surface — no runtime verification was performed, and none is possible without a display.**

Two false doc comments are corrected: `SwapPayloadBuilder` claimed these fields "mirror the node's derivation", and `LimitSwapPayloadAssembler` claimed `toAmountLimit` drives the WalletCore native-RUNE min-output arg (it does not, at HEAD).

## Tests

13 tests replace the 5 deleted ones. Parser: Auto memo (`LIM` `0`), custom slippage parsed exactly, trailing affiliate/fee fields, unabbreviated `SWAP:` prefix with no streaming spec, secured-asset memo, Maya memo, every swap-action spelling (`SWAP`/`swap`/`=`/`s`/`S`), non-swap actions rejected, whole-triple rejection on an unreadable term, and a table of unusable memos (empty, truncated, non-numeric, signed, fractional, padded, non-ASCII numerals, too many terms). Payload wiring: limit + streaming read from the memo, Auto reports no floor, `max_streaming_quantity` ceiling ignored.

The real mainnet memo `=:e:0x742d35Cc6634C0532925a3b844Bc454e4438f44e:0/1/0` is used as a fixture.

## Gate

`make generate && make test` on iPhone 16e — `** TEST SUCCEEDED **`, **3126 tests executed, 1 skipped, 0 failures**.

Reviewed across two `codex exec --sandbox read-only` rounds: round 1 raised one MINOR finding on parser leniency (three parts — non-swap actions, whitespace normalization, per-field vs whole-triple rejection), all three accepted and fixed; round 2 on the amended commit returned no findings.

## Summary by CodeRabbit

* **Updates**
  * THORChain and MayaChain swaps now use limit and streaming settings directly from the swap memo.
  * Improved handling of auto-slippage, custom minimum outputs, and incomplete or invalid memo terms.
  * Limit-swap documentation now clarifies how minimum-output limits are represented.

* **Tests**
  * Added coverage for memo parsing, streaming parameters, secured assets, MayaChain swaps, and invalid memo formats.

### DIFF
--- a/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitSwapPayloadAssembler.swift
+++ b/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitSwapPayloadAssembler.swift
@@ -267,10 +267,11 @@ func buildLimitSwapKeysignPayload(
 /// only cross-device SWAP display — the co-signer's "you receive" row — so a
 /// resting limit order shows its LIM floor rather than 0. Signing and every
 /// fund-safety gate ignore it (they read `transaction` / `toCoin.address`), so
-/// its exact value can never affect what is signed. `toAmountLimit` stays "0":
-/// it drives only the WalletCore native-RUNE `THORChainSwap.build` min-output
-/// arg, which an ERC20 `depositWithExpiry` never reaches, and nothing displays
-/// it.
+/// its exact value can never affect what is signed. `toAmountLimit` and the two
+/// streaming fields stay "0": nothing on this device reads them and no signer
+/// consumes them — they exist only to travel in the cross-device proto. For a
+/// resting limit order that is also the honest value, since the order's floor
+/// lives in the `=<` memo on `KeysignPayload.memo`, not here.
 ///
 /// `expirationTime` is the router's ON-CHAIN tx-execution deadline
 /// (`require(block.timestamp < expiry)` inside `depositWithExpiry`) — a guard
--- a/VultisigApp/VultisigApp/Features/Swap/Logic/SwapPayloadBuilder.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/Logic/SwapPayloadBuilder.swift
@@ -78,32 +78,106 @@ extension SwapCryptoLogic {
         return ERC20ApprovePayload(amount: amount, spender: spender)
     }
 
+    /// The `LIM/INTERVAL/QUANTITY` triple a THORChain/MayaChain swap memo
+    /// carries. Every field is a decimal string in the node's own units and
+    /// semantics; `"0"` is meaningful in all three positions (no output floor,
+    /// no streaming, protocol-chosen sub-swap count) rather than a sentinel.
+    struct ThorchainMemoSwapTerms: Equatable {
+        /// Minimum acceptable output, in the node's base units (THORChain 1e8 —
+        /// the same units as `expected_amount_out`). `"0"` = no floor asserted.
+        let limit: String
+        /// Streaming-swap block interval. `"0"` = a single (rapid) swap.
+        let streamingInterval: String
+        /// Streaming-swap sub-swap count. `"0"` = the protocol picks it.
+        let streamingQuantity: String
+
+        /// What an absent, truncated, or unparseable memo yields: a swap that
+        /// asserts nothing. Every field claims the weakest thing it can.
+        static let unspecified = ThorchainMemoSwapTerms(
+            limit: "0",
+            streamingInterval: "0",
+            streamingQuantity: "0"
+        )
+    }
+
+    /// Read the `LIM/INTERVAL/QUANTITY` triple out of a node-returned
+    /// THORChain/MayaChain swap memo.
+    ///
+    /// Memo grammar: `SWAP:ASSET:DESTADDR:LIM/INTERVAL/QUANTITY[:AFFILIATE:FEE]`,
+    /// with `SWAP` abbreviating to `=` or `s` (e.g. `=:e:0x742d…:0/1/0`). Neither
+    /// ASSET nor DESTADDR can contain a colon, so the triple is always the 4th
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4852, opened 2026-07-20
fix(swap): map the node's price-limit rejection to an actionable slippage error

gastonm5:
Part of #4838.

## What this does

Maps THORChain/Maya's price-limit quote rejection to a new localized `SwapError.slippageToleranceTooTight`, instead of leaking the raw node string to the user.

When the node simulates a swap whose output lands under the `LIM` floor derived from the slippage tolerance, it rejects the quote. That body fell through the classifier to `.serverError(message:)`, so the untranslated node text was shown verbatim — with no hint that the slippage setting is the knob to turn.

## Split out of #4849

This is **commit 2 of #4849, lifted onto `main` on its own**.

#4849 bundles this UX fix with a contested change: switching quotes to `liquidity_tolerance_bps` with a 100 bps default, which re-introduces an on-chain minimum-output floor in the same territory as the previously-reverted #4556. That one needs a cross-client product decision and an Android plan before it can land.

This error mapping is a **pure UX fix that is correct regardless of how the floor decision goes** — the node can reject on a price limit whenever a user sets a tight slippage, floor default or not. There's no reason to hold it behind that decision, so it ships separately. #4849 now contains only the floor change.

This branch contains **no** tolerance-related changes: no `liquidity_tolerance_bps`, no constant rename, no parameter switch.

## Real captured node strings

Both mapped, verbatim from live rejections:

- **THORChain:** `failed to simulate swap: failed to simulate handler: emit asset 42579895573 less than price limit 340083366993: invalid request`
- **Maya:** `failed to simulate swap: emit asset 336029740 less than price limit 340582434`

The two node shapes differ, so the match is on the **shared `less than price limit` substring**, placed inside `classifyNativeQuoteError` — the helper both `mapThorchainSwapError` and `mapMayachainSwapError` run through. One match therefore covers both nodes.

It sits **after** the trading-halt check, so a halted pool still surfaces the retryable halt message rather than the slippage one. `testHaltOutranksPriceLimit` pins that ordering.

Localized in all 8 locales (de, en, es, hr, it, ko, pt, zh-Hans).

## Gate

`make test` on iOS Simulator: `** TEST SUCCEEDED **` — **3126 tests, 1 skipped, 0 failures**. No `BUILD FAILED` / `TEST FAILED` markers in the log.

New tests (5), all passing:
- `testThorchainPriceLimitRejectionMapsToSlippageTooTight`
- `testMayachainPriceLimitRejectionMapsToSlippageTooTight`
- `testPriceLimitRejectionIsCaseInsensitive`
- `testPriceLimitRejectionNoLongerLeaksRawNodeString`
- `testHaltOutranksPriceLimit`

## Diff

11 files, +74/-0 — `SwapError.swift` (new case), `SwapService.swift` (one classifier branch), 8 locale files, `SwapErrorMappingTests.swift`.

Draft pending review of the split.

## Summary by CodeRabbit

- **New Features**
  - Added a clear swap error when the price exceeds the configured slippage tolerance.
  - Users are advised to increase slippage or try a smaller swap amount.
- **Bug Fixes**
  - Price-limit rejections now display a helpful localized message instead of raw node error text.
  - Existing trading-halted errors continue to take precedence when applicable.
- **Localization**
  - Added translations for the new error message across supported languages.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapError.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapError.swift
@@ -17,6 +17,10 @@ enum SwapError: Error, LocalizedError, Equatable {
     case noLiquidityPool
     case tradingHalted
     case swapAmountTooSmall
+    /// The node refused the quote because its simulated output fell below the
+    /// minimum-output floor derived from the requested slippage tolerance.
+    /// Actionable by the user: raise the slippage setting or reduce the amount.
+    case slippageToleranceTooTight
     case lessThenMinSwapAmount(amount: String)
     case serverError(message: String)
 
@@ -34,6 +38,8 @@ enum SwapError: Error, LocalizedError, Equatable {
             return "swapTradingHalted".localized
         case .swapAmountTooSmall:
             return "swapAmountTooSmall".localized
+        case .slippageToleranceTooTight:
+            return "swapSlippageToleranceTooTight".localized
         case .lessThenMinSwapAmount(let amount):
             return String(format: "swapAmountTooSmallRecommended".localized, amount)
         case .serverError(let msg):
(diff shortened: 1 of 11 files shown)


johnnyluo (approved):
LGTM

Issue #4422, opened 2026-05-26
Unify swap fee display between initiating + co-signing devices (cross-platform)


gastonm5:
## Background

For the same swap, the initiating device and the co-signing (peer) device do NOT render the same fee breakdown.

**Concrete divergence** (SwapKit BCH→ETH used as the trace, but applies to every multi-fee swap):

| Surface | File | Fees rendered |
|---|---|---|
| Initiator `SwapVerifyScreen` | `Features/Swap/Views/Screens/SwapVerifyScreen.swift:123-147` | `networkFee` + `swapFee` + `maxTotalFee` (**3 rows**) |
| Cosigner `KeysignSwapConfirmView` | `Features/Keysign/Views/KeysignSwapConfirmView.swift:30-46` | **`networkFee` only (1 row)** |

**Why**: `SwapKitSwapPayload`, `THORChainSwapPayload`, and `GenericSwapPayload` carry **zero fee fields**. The peer's `JoinKeysignGasViewModel` reads only `chainSpecific.fee` / `gas`. So `swapFee` + `maxTotalFee` literally never cross the relay.

## Sibling-platform state

- **Android**: solves the same-platform case by re-deriving the breakdown peer-side (`vultisig-android` `JoinKeysignViewModel.loadTransaction:483-697`). Heuristic; doesn't generalize to SwapKit non-EVM (opaque tx bytes, no quote on the wire); accepts a VULT-tier-discount gap.
- **Windows**: same gap as iOS.

## Goal

Every device renders the same fee breakdown (`network` + `swap/provider` + `affiliate` + `total` + VULT-discount-applied) regardless of which initiates or co-signs. Source of truth is data threaded explicitly through `KeysignPayload`, not peer-side re-derivation.

## Approach — direct cross-platform parity (no iOS-only intermediate)

Add an optional `SwapFeeBreakdown` message to `commondata`, populate on the initiator, render on both devices through one shared formatter. Five PRs, strict sequence, all draft against `main` of each repo.

### Proto schema (attachment point: parent `KeysignPayload`)

```protobuf
message SwapFeeBreakdown {
  // Network gas fee in source chain's native units (base-10 string).
  string network_fee = 1;
  // Provider's swap fee (THORChain inbound, SwapKit affiliate, etc).
  string swap_fee = 2;
  // Denom the swap_fee is denominated in.
  string swap_fee_denom = 3;
  // POST-discount swap_fee is what's in field 2 above.
[...]

PR #4864, opened 2026-07-21
fix(swap): itemize the Verify fee rows — affiliate-only Vultisig Fee, % from bps, reconciled total

### LINKED ISSUE #4859: Swap verify screen drops the itemized affiliate fee and shows THORChain's composite as "Swap Fee"
## Problem

The quote screen itemizes correctly — `SwapDetailsSummary.swift:157-168` renders `"Swap Fee (0.50%)"` from `fees.affiliate` (`SwapCryptoLogic.swift:427-443`), with separate Outbound Fee and VULT tier rows including a fiat saving.

**The verify screen throws all of that away.** `SwapVerifyScreen.swift:145-153` renders `"swapFee"` (no percentage) using `swapFeeString`, which on native routes resolves `SwapCryptoLogic.swift:332-343` → `SwapQuote.swift:141-143` → **`quote.fees.total`** — the composite (affiliate + outbound + liquidity). No discount or tier row exists on verify at all.

The screen the user actually approves before signing is the least transparent one in the flow.

## Route-level inconsistencies in the same path

1. **Solana/Jupiter is inverted.** The affiliate row is hidden on the form — `evmSwapFeeBigInt` returns nil for `.jupiter` (`SwapQuote.swift:169-171`), so `baseAffiliateFee` falls through to `default: return .empty` (`SwapCryptoLogic.swift:440-441`) — yet the fee *does* appear on verify via `inboundFeeDecimal` → `platformFee` (`SwapQuote.swift:149-152`). Opposite of every other route.
2. **SwapKit shows no fee on any screen** (`SwapQuote.swift:153-155`, `:164-168`).
3. **Ultimate tier hides the row entirely.** `SwapCryptoLogic.swift:437` guards `feeAmount > 0`, so a user paying 0% sees no fee row rather than an explicit zero. Android already handles this correctly by keeping the row at `$0.00` — that's the target behavior.

## Blocked by

vultisig/vultisig-sdk#1450 — affiliate and protocol fees need to be separate fields on the SDK fee surface.

## Canonical fee-display spec (all clients)

Rule: **the row labeled with Vultisig's name contains only money Vultisig receives.**

| # | Label | Value | Shown when |
|---|---|---|---|
| 1 | `Network Fee` | source-chain gas, native + fiat | always |
| 2 | `Vultisig Fee (X.XX%)` | affiliate component only — never a composite | always, including when 0 |
| 3 | `Protocol Fee` | native: `outbound` (+ `asset`/liquidity) | native routes |
| 4 | `Price Impact` | `slippage_bps` only | when provider reports slippage |
| 5 | `Total Fee` | must equal rows 1 + 2 + 3 | always |
| 6 | `VULT {Tier} · −{bps} bps · saved ~$X` | tier + fiat saving | when a tier applies |

## Acceptance criteria

- [ ] Verify screen renders the same rows as the details screen: Network Fee, `Vultisig Fee (X.XX%)` (affiliate only), Protocol/Outbound Fee, Price Impact, Total Fee, VULT tier row
- [ ] `swapFeeString` no longer sources `fees.total` for the row labeled as the swap fee
- [ ] Total Fee reconciles to the itemized rows (Android's approach at `SwapQuoteManager.kt:360-367` is the reference)
- [ ] Percentage derived from the `affiliate_bps` sent on the quote request, not from a fiat division
- [ ] Jupiter/Solana shows the affiliate row on **both** form and verify
- [ ] At Ultimate, row stays visible as `Vultisig Fee (0.00%) — $0.00` + waiver line — remove the `feeAmount > 0` guard
- [ ] SwapKit routes render `Vultisig Fee (0.50%) — included in quoted rate`

## Supersedes

#4683 (`[Verify] Swap fee display correctness across all providers`) was closed 2026-07-09 with zero comments, despite acceptance criteria requiring a results table — reopening the substance here as a build spec rather than a verification checklist. Prior art for the cross-device slice: #4422.

gastonm5:
Closes #4859

## Problem

The Verify screen — the one the user approves before signing — collapsed every swap fee into THORChain's composite `fees.total` and labelled it "Swap Fee", throwing away the affiliate itemization, the percentage, the outbound row and the VULT tier saving that the Details screen shows. The least transparent screen in the flow was the one the user signs from.

## iOS was not actually blocked by SDK#1450

The issue flags vultisig-sdk#1450 as a blocker. That binds windows/extension (they read the TS SDK's fee surface). **iOS has its own `ThorchainSwapQuote.Fees` where `affiliate` is already a separate field from `total`** — the Details screen itemizes it today. Every acceptance criterion here is met with data iOS already has locally; SDK#1450 gates cross-client unification, not this fix. (Confirmed with the maintainer.)

Root cause was a one-field divergence: Verify rendered `fees.total` via `swapFeeString`; Details renders `fees.affiliate`. And Verify already had `baseAffiliateFee` / `swapFeeLabel` / `vultDiscount` on the transaction — it just never referenced them.

## What changed

**Canonical rows on Verify (and Details / Done, kept consistent):** `Network Fee`, `Vultisig Fee (X.XX%)` (affiliate only), `Protocol Fee` (outbound), `Price Impact` (when the provider reports slippage), `Total Fee`, and the `VULT {tier}` saving row.

**Percentage from the effective affiliate bps, not a fiat division.** Extracted `THORChainSwaps.effectiveAffiliateFeeBps` (and the `discountedAffiliateBps` clamp) and had **both** the request builders (`ThorchainService`/`MayaChainService`/`Endpoint`) **and** the display label consume it, so the shown % equals the bps the node was actually asked to charge — by construction, not by inference. It also reconciles with the `fees.affiliate` amount because the node computes that amount from that bps. (Android's own reference uses a fiat-division %; this does better on that criterion.)

**Total reconciles to the itemized rows, dropping liquidity.** `Total = Network + Vultisig(affiliate) + Protocol(outbound)`, not `fees.total` — the `asset`/liquidity component is already reflected in the expected output and must not be double-counted (Android `SwapQuoteManager.kt:360-367` is the reference).

**Per-route behaviour fixed:**
- **Jupiter/Solana** now shows the affiliate row on **both** form and verify (was inverted — hidden on form, shown on verify).
- **Ultimate tier** keeps the row visible as `Vultisig Fee (0.00%) — $0.00` (removed the `feeAmount > 0` guard).
- **SwapKit** renders `Vultisig Fee (0.50%) — included in quoted rate` (no per-tx fee field; 50 bps is a known constant).
- **LiFi-Solana** sources the fee from `integratorFee` (it reports `swapFee "0"` and charges via the integrator fraction), so it no longer shows $0.00 or drops from Total.

## Distinguishing the two Jupiter zero-fee cases

`platformFee == nil` previously conflated two situations. The `.jupiter` case is now `jupiter(EVMQuote, fee: BigInt?, platformFee: Decimal, feeOnInput: Bool)`, where `feeOnInput = (feeMint != outputMint)` computed in `JupiterService` where both mints are known:
- **native-SOL output** (`feeOnInput = true`) — fee is on the input mint, unsurfaceable in `toCoin` units → row suppressed (correct).
- **Ultimate token output** (`feeOnInput = false`, `platformFee = 0`) — a real fee-on-output of zero → row shows `0.00% / $0.00` (correct, per AC).

## Signed-payload invariant (this is the screen the user signs from)

**All changes are display-only.** `SwapPayloadBuilder` is untouched; the added transaction fields and the enum-shape change do not feed the memo, signed amount, `toAmount`, or `toAmountLimit`; the request-builder refactor is mathematically identical; route ranking still reads only `outAmount`. Verified by a whole-branch cross-model (Codex) review focused on this invariant, which confirmed it holds, plus that native displayed-bps match sent-bps across every tier and the referral split.

## Tests

`baseAffiliateFee` / `swapFeeString` / `swapFeeLabel` / `vultDiscount` / `outboundFeeString` were **entirely untested** (fixtures built `affiliate: "0"`, so affiliate itemization was never exercised). New `SwapAffiliateFeeDisplayTests` (28 tests) covers: the bps-% equalling the sent bps per tier + referral, the Total reconciliation identity (excludes liquidity), Ultimate `0.00%/$0.00`, SwapKit static label, LiFi-Solana integrator fee in row + Total, and Jupiter native-SOL-suppressed vs Ultimate-token-shown. The pre-existing Done breakdown-gate test was updated (native swaps now always carry the $0.00 affiliate row, so the breakdown chevron is meaningful).

## Gate

`make test` on iPhone 16 Pro: **`** TEST SUCCEEDED **` — 3148 executed, 0 failures, 1 skipped**. SwiftLint clean on the touched files. Reviewed with a per-commit + whole-branch cross-model (Codex) loop; two P1 route bugs (LiFi-Solana, Ultimate-Jupiter) it caught were fixed in the final commit and re-reviewed clean.

## Out of scope

- **Co-signer parity** (`KeysignSwapConfirmView` / `JoinKeysignSwapFeeViewModel`) — #4422's territory; none of this issue's ACs touch it.
- Cross-client SDK unification — vultisig-sdk#1450 (windows/extension).

## Manual check worth doing before merge

A visual pass of the itemized Verify rows across the four route families (native THOR/Maya, EVM aggregator, Jupiter token-output, SwapKit) on device — the row math is unit-tested, but the layout/blur/loading states are only exercised by the UI layer.

## Summary by CodeRabbit

- **New Features**
  - Added clearer swap fee breakdowns for affiliate, protocol, network, referral, and VULT discounts.
  - Added price-impact and “included in quoted rate” details during swap verification.
  - Improved Jupiter fee handling, including fees charged in the input asset.
  - Added referral-aware fee calculations and display logic.

- **Localization**
  - Added translated labels for protocol fees, included rates, and Vultisig fees across supported languages.

- **Bug Fixes**
  - Improved total-fee reconciliation and prevented misleading zero-value fee rows.
  - Corrected affiliate fee calculations for discounted and referred swaps.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapQuote+Ranking.swift
@@ -38,7 +38,7 @@ extension SwapQuote {
             guard let raw = BigInt(quote.dstAmount), raw > 0 else { return nil }
             return toCoin.decimal(for: raw)
 
-        case .jupiter(let quote, _, _):
+        case .jupiter(let quote, _, _, _):
             // Jupiter `outAmount` is already net of the affiliate platform fee
             // (Jupiter deducts the fee from the AMM output and reports it
             // separately in `platformFee`), so it's the amount the user
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -636,14 +636,14 @@ private extension SwapService {
         slippageBps: Int?
     ) async throws -> SwapQuote {
         let fromAmount = fromCoin.raw(for: amount)
-        let (quote, fee, platformFee) = try await service.fetchQuote(
+        let (quote, fee, platformFee, feeOnInput) = try await service.fetchQuote(
             fromCoin: fromCoin,
             toCoin: toCoin,
             fromAmount: fromAmount,
             vultTierDiscount: vultTierDiscount,
             slippageBps: slippageBps
         )
-        return .jupiter(quote, fee: fee, platformFee: platformFee)
+        return .jupiter(quote, fee: fee, platformFee: platformFee, feeOnInput: feeOnInput)
     }
 }
 
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Jupiter/JupiterService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Jupiter/JupiterService.swift
@@ -62,7 +62,7 @@ struct JupiterService {
         fromAmount: BigInt,
         vultTierDiscount: Int,
         slippageBps: Int?
-    ) async throws -> (quote: EVMQuote, fee: BigInt?, platformFee: Decimal?) {
+    ) async throws -> (quote: EVMQuote, fee: BigInt?, platformFee: Decimal, feeOnInput: Bool) {
         let inputMint = jupiterMint(for: fromCoin)
         let outputMint = jupiterMint(for: toCoin)
 
@@ -104,14 +104,15 @@ struct JupiterService {
             feeAccount: feeAccount
         )
 
-        // `platformFee` is denominated in the fee mint. Surface it for display
-        // only when the fee mint is the output mint (so it's in `toCoin` units);
-        // for the input-mint case (native-SOL outputs) it would be
-        // mis-denominated, so omit it. Ranking uses `outAmount` (already net of
-        // the fee) regardless of which mint the fee is taken in.
-        let platformFee = feeMint == outputMint
-            ? platformFeeDecimal(from: quoteResponse, toCoin: toCoin)
-            : nil
+        // `platformFee` is denominated in the fee mint. It's surfaceable in
+        // `toCoin` units only when the fee mint IS the output mint. For the
+        // input-mint case (native-SOL outputs) it would be mis-denominated, so
+        // it's flagged `feeOnInput` and reported as 0 — display suppresses the
+        // affiliate row there rather than show a misleading amount. A zero
+        // output-mint fee (Ultimate tier) is a real 0, shown as a $0.00 row.
+        // Ranking uses `outAmount` (already net of the fee) regardless.
+        let feeOnInput = feeMint != outputMint
+        let platformFee: Decimal = feeOnInput ? .zero : (platformFeeDecimal(from: quoteResponse, toCoin: toCoin) ?? .zero)
 
         let evmQuote = EVMQuote(
             dstAmount: quoteResponse.outAmount,
@@ -124,7 +125,7 @@ struct JupiterService {
                 gas: 0
             )
         )
-        return (evmQuote, nil, platformFee)
+        return (evmQuote, nil, platformFee, feeOnInput)
     }
 
     /// The mint Jupiter expects for a coin: the SPL contract address, or wrapped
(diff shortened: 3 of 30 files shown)


johnnyluo (approved):
LGTM

Issue #4843, opened 2026-07-18
Send: scanning a `terra1…` address can silently switch the form to the wrong Terra chain (Terra vs Terra Classic share the bech32 prefix)


realpaaao:
## Summary

`AddressService.detectChain` resolves a scanned address by iterating `CoinType.allCases` and taking the **first** CoinType that validates it. Terra and Terra Classic both use the bech32 prefix `terra`, so a `terra1…` address validates under **both** — and which one wins is decided by enum ordering, not by anything about the address.

The result feeds `SendDetailsViewModel.detectAndSwitchChain`, which silently rewrites the send form's chain, coin and from-address.

## The code

```swift
// AddressService.swift:53-62
for coinType in CoinType.allCases {
    if coinType.validate(address: address) {
        if let chain = Chain.allCases.first(where: { $0.coinType == coinType }) {
            return vault.coins.contains(where: { $0.chain == chain && $0.isNativeToken }) ? chain : nil
        }
    }
}
```

The two chains map to distinct CoinTypes — `Chain.swift:469-472`, `.terra → CoinType.terraV2`, `.terraClassic → CoinType.terra` — but both CoinTypes accept the same address string.

Caller (`SendDetailsViewModel.swift:205-216`), used by the QR scanner sheet on Send Details:

```swift
selectedChain = detectedChain
coin = detectedCoin
fromAddress = detectedCoin.address
```

## Impact

A user with both LUNA and LUNC in their vault scans a Terra Classic receive address. The form can silently switch to **Terra** and send LUNA instead — to an address that is syntactically valid on both networks but on the wrong chain. Terra and Terra Classic are separate networks; the recipient may not control that address there, and there is no recovery path.

Direction depends on `CoinType.allCases` ordering, which is WalletCore's, not ours — so this can flip on a WalletCore bump with no change on our side.

Mitigating: the switch only fires when the wrongly-detected chain's native coin is in the vault. A user holding only one of the two is safe (they get `nil`, no switch). The exposure is users holding both.

## The fix pattern is already in this function

`detectChain` already special-cases three other ambiguous prefixes ahead of the loop — `sthor` (`:32-35`), `qbtc` (`:38-41`), Bittensor SS58 (`:44-47`). Terra just never got one.

Terra vs Terra Classic cannot be told apart from the address at all — they genuinely share the address space. So the honest options are:

[...]

PR #4899, opened 2026-07-24
fix(send): gate invalid recipient with a visible inline error (#4861)

### LINKED ISSUE #4861: fix(send): reject an invalid recipient with a visible error across all chains
## Problem

A send flow can accept an invalid recipient into its state and leave the next action unavailable without explaining why. Recipient validation must be a first-class, chain-general user-facing gate, not an implicit failure later in the flow.

This surfaced while preparing a **USDC send on Ethereum**: a transaction ID was entered where the recipient address belongs. The app should explicitly say the value is not a valid Ethereum recipient address.

## Required behaviour

For every supported chain and send asset:

- validate the recipient as soon as it is captured or selected;
- show a clear inline error when invalid: `That recipient is not a valid address for this chain.` Include the expected format where useful;
- keep the next action unavailable only with that visible reason and a direct correction path;
- allow the user to replace the value and continue without restarting;
- never build, sign, or broadcast from an invalid recipient.

## Acceptance criteria

- [ ] Invalid-recipient UI and recovery are shared across native send flows, not limited to a chain-specific screen.
- [ ] Coverage includes an Ethereum/USDC send with a txid substituted for the address, plus representative Cosmos, UTXO, Solana and XRP-family invalid inputs.
- [ ] Each case renders a human-readable error and prevents transaction construction/signing.
- [ ] Valid replacement enables the flow normally.

— ⚒️ _Mjölnir, Paaao's agent — posting on his behalf, not Paaao directly._

gastonm5:
Closes #4861

## Problem

The Send flow accepted an invalid recipient (e.g. a **transaction id pasted into the recipient field** on an ETH/USDC send) and merely **disabled the Next button with no explanation**. The user was stuck with no idea why they couldn't continue.

## Root cause

The recipient field only renders an error when `showAddressAlert == true`, which was set **exclusively** inside `validateForm()`'s sub-validators. But `validateForm()` runs on the final Continue from the amount tab — a step that is **unreachable while the address is invalid**, because every address-commit path sets `addressSetupDone = false` and stops, without ever setting the error:

1. `SendDetailsScreen.onChange(of: isAddressResolved)` — async resolution failed (`resolved == false`).
2. `SendDetailsAddressTab.handleClose` — collapsing the tab on an unresolved recipient.
3. `SendDetailsAddressFields.handle(addressResult:)` — paste/QR/address-book no-detect path.

Latent bug: the key `invalidAddressError` referenced by `validateForm`'s address sub-validators **did not exist in any `.strings` file**, so `CommonTextField` (which renders `Text(error.localized)`) would have shown the raw key.

## Fix (chain-general — the Send screen is shared across all chains)

- Added `SendDetailsViewModel.markInvalidRecipient()` and a guarded `markInvalidRecipientIfUnresolvable()` (for the paste/QR path) that surface a clear inline error under the destination field while keeping Next disabled **with that visible reason**.
- Wired all three address-commit failure paths to show the reason.
- The error **clears on correction**: valid format, cleared field, or a fresh resolution attempt. While async resolution is pending (`isAddressResolved == nil`) the error is kept cleared so it never flashes while the user is still typing — it only appears on a **definitive** failure.
- Repointed both `validateForm` address sub-validators off the non-existent `invalidAddressError` to the new key, so no path can ever render a raw key.
- Hardened the debounced resolver against a stale/late verdict clobbering a newer recipient: resolution now commits (both the verdict and the address rewrite) only when a monotonic generation token still matches, so an uncancelled in-flight resolver can't overwrite what the user has since typed or flash an error mid-resolution.

### Regressions guarded
- **ENS/TNS names** (`.eth`/`.sol`, THORChain-family TNS) legitimately fail the sync format check and resolve async — they are deliberately left to the async path and never flash an error prematurely.
- **Chain-switch-on-paste** (Terra, #4843) still succeeds silently — detection runs before any invalid-marking.

## Localization

New key `invalidRecipientAddressError` = "That recipient is not a valid address for this chain." added to **all 8 locales** (en, de, es, hr, it, **ko**, pt, zh-Hans), translated per locale and sorted via `scripts/sort_localizable.py`.

## Tests

Added 15 camelCase tests in `SendDetailsViewModelTests` covering: the ETH-txid case from the issue, invalid Cosmos / UTXO / Solana / XRP recipients through the full `validateForm()` pipeline, valid-replacement clearing + flow-enable, and the ENS / paste-guard regressions. Updated the 5 stale `invalidAddressError` assertions in `SendDetailsViewModelValidationTests` to the new key.

## Gate

`make test` (full suite, dedicated en_US simulator): **`** TEST SUCCEEDED **`** — `Executed 3831 tests, with 1 test skipped and 0 failures`.

Cross-model review: two rounds of `codex exec` (read-only) — round 1 caught the stale test-key assertions + a resolver-race edge; round 2 caught that the generation guard needed to also protect the in-`validateToAddress` address mutation. Both triaged and fixed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recipient validation for invalid, unsupported, or unresolvable addresses.
  - Displays a clear inline error when a recipient is invalid for the selected chain.
  - Prevents outdated address-resolution results from overwriting newer input.
  - Keeps resolvable name-service addresses eligible for asynchronous resolution.
  - Returns users to the recipient field when validation fails, with “Next” disabled.

- **Localization**
  - Added translated invalid-recipient messages across supported languages.

- **Tests**
  - Expanded coverage for invalid recipients, valid replacements, unresolved names, and multiple blockchain address formats.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Send/SendDetailsAddressFields.swift
+++ b/VultisigApp/VultisigApp/Components/Send/SendDetailsAddressFields.swift
@@ -93,6 +93,13 @@ struct SendDetailsAddressFields: View {
                 viewModel.errorMessage = nil
                 viewModel.addressSetupDone = true
                 viewModel.onSelect(tab: .amount)
+            } else {
+                // Paste/QR/address-book value that neither matches the current
+                // chain nor one we can switch to. A same-chain-valid value
+                // advances via the shared `toAddress` onChange; an ENS/TNS name
+                // still needs async resolution. Only a value that can never
+                // resolve is flagged now, so Next is disabled *with a reason*.
+                viewModel.markInvalidRecipientIfUnresolvable()
             }
         }
     }
--- a/VultisigApp/VultisigApp/Components/Send/SendDetailsAddressTab.swift
+++ b/VultisigApp/VultisigApp/Components/Send/SendDetailsAddressTab.swift
@@ -78,6 +78,10 @@ struct SendDetailsAddressTab: View {
         }
         if !viewModel.toAddress.isEmpty {
             guard await viewModel.validateToAddress() else {
+                // Collapsing the address tab on an unresolved recipient is a
+                // definitive failure — surface the inline reason rather than
+                // just leaving Next disabled.
+                viewModel.markInvalidRecipient()
                 viewModel.addressSetupDone = false
                 if viewModel.selectedTab == .amount {
                     viewModel.onSelect(tab: .address)
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -583,6 +583,7 @@
 "invalidIntegerError" = "Ungültige Ganzzahl";
 "invalidLPUnits" = "Ungültige LP-Einheiten";
 "invalidOperatorFee" = "Ungültige Betreibergebühr";
+"invalidRecipientAddressError" = "Dieser Empfänger ist keine gültige Adresse für diese Kette.";
 "invalidVaultData" = "Ungültige Tresordaten";
 "Italian" = "Italienisch";
 "iUnderstand" = "Ich verstehe";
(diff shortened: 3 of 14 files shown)


johnnyluo (approved):
LGTM

Issue #4902, opened 2026-07-24
[Fix] Swap error handling relabels unknown errors as user-facing money errors (3 sites) [Swap]


gastonm5:
# [Fix] Swap error handling relabels unknown errors as user-facing money errors (3 sites) [Swap]

## Problem
Three catch/`default` sites in the swap flow coerce **any unclassified error** (URLSession timeout, TLS failure, JSON decode, OS cancellation slipping typing, a new error case) into a user-facing **money error**. The user is told to add funds / that the amount is too small when the real cause is a network/decoding failure — so they change their amount against a route that is actually just down, or abandon a pair that works. The real error is also discarded, not logged, destroying diagnosability. These fire on the live happy path whenever the backend hiccups.

## Locations
1. `Blockchain/Swaps/Common/SwapService.swift:491-492` — `catch { throw SwapError.swapAmountTooSmall }` (native cross-chain quote).
2. `Features/Swap/ViewModels/SwapDetailsViewModel.swift:861-868` — fee-fetch `default:` → `SwapCryptoLogic.Errors.insufficientGas`.
3. `Features/Swap/Logic/SwapPayloadBuilder.swift:42-61` — UTXO/Cardano fee `catch` → `Errors.insufficientFunds` (also infers underfunded from `planFee <= 0`).

## Solution
At each site: pass through the genuinely-typed money errors, and for the unknown `default`/`catch`, **log the real error and rethrow it** (or a neutral non-money error). Money verdicts must come only from the explicit min-amount / balance checks that already exist (`SwapService` min-amount checks + `classifyNativeQuoteError`; the swap balance gate). Also split "planner returned fee ≤ 0" from "user underfunded" in site 3.

### Must Do
- [ ] Site 1: log + rethrow the real error instead of `swapAmountTooSmall`
- [ ] Site 2: pass typed UTXO errors through; log + surface the real error in `default:` (not `insufficientGas`)
- [ ] Site 3: log + rethrow non-`KeysignPayloadFactory` errors; stop inferring `insufficientFunds` from `planFee <= 0`

### Must NOT Do
- Don't remove the legitimate typed money errors (min-amount, explicit UTXO-too-small)
- Don't change fee values that feed signed payloads (UTXO/Cardano) — behavior-preserving

## Acceptance Criteria
- [ ] SwiftLint passes; build + `make test` green
- [ ] A simulated network failure during quote/fee shows a network/route error, not "amount too small" / "insufficient gas" / "insufficient funds"
[...]

PR #4907, opened 2026-07-24
fix(swap): tag native quotes via a per-service Strategy, not a type-probe

### LINKED ISSUE #4901: [Fix] Swap provider quote-tagging uses a type-probe with a mislabeling default [Swap]
# [Fix] Swap provider quote-tagging uses a type-probe with a mislabeling `default` [Swap]

## Problem
`SwapService.fetchCrossChainQuote` tags a fetched native quote to a `SwapQuote` case by **type-probing the service instance** (`case _ as ThorchainService: …`) ending in `default: return .thorchain(quote)`. The authoritative discriminant — `provider: SwapProvider` — is already in scope and goes unused. A new native provider (or making one service subclass another) is then silently tagged `.thorchain`, which drives the wrong branch in `SwapPayloadBuilder` (wrong `toAddress`/memo/`swapPayload` variant) → a corrupted signed transfer while the UI shows a plausible quote.

Latent today (the 4 existing services are sibling protocol-conformers, so the tag is byte-identical), but the `default:` is a live mislabel-on-extension trap and it is signing-critical.

## Location
`VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift:471-482`

## Solution
Replace the type-probe with an **exhaustive switch over the `provider` enum** (no `default:`), so the compiler forces every new provider to declare its tag. Optionally have each `ThorchainSwapProvider` vend its own tag (Strategy) so the mapping can't drift from the service.

### Must Do
- [ ] Tag the quote from `provider` (exhaustive; aggregator cases that can't reach this path throw)
- [ ] Remove the `as?` type-probe and the `default:` fallback

### Must NOT Do
- Don't change the produced tag for the 4 existing services (must stay byte-identical)
- Don't refactor the surrounding quote/streaming logic

## Acceptance Criteria
- [ ] Existing THORChain / THORChainnet / THORStagenet / Maya swaps produce the identical `SwapQuote` case as before
- [ ] Adding a new `SwapProvider` case fails to compile until its tag is declared

gastonm5:
Closes #4901

## Root cause
`SwapService.fetchCrossChainQuote` tagged a fetched THORChain/Maya quote to a `SwapQuote` case by **type-probing the service instance** inside a `switch service { … }` that ended with a mislabeling catch-all:

```swift
default: return .thorchain(quote)   // silent mislabel
```

Any conforming service that wasn't one of the four explicitly matched cases (e.g. a newly added network) would be silently tagged as **mainnet THORChain** — the exact three-network distinction (`.thorchain` / `.thorchainChainnet` / `.thorchainStagenet`) the result depends on, quietly collapsed.

## Why switching on `provider` is insufficient
The obvious "switch on the `provider:` param instead" does **not** work: `SwapProvider` only has `.thorchain` / `.maya` (plus the aggregators) — it has **no chainnet/stagenet distinction**, so it is too coarse to produce the three separate `SwapQuote` cases. The concrete service **type** is the only thing that carries the network.

## Fix — Strategy on the service
Add a requirement to the `ThorchainSwapProvider` protocol so each concrete service declares its own tag, then call it at the site (no type-probe, no `default:`):

```swift
func makeSwapQuote(_ quote: ThorchainSwapQuote) -> SwapQuote
```

- `ThorchainService` → `.thorchain`
- `ThorchainChainnetService` → `.thorchainChainnet`
- `ThorchainStagenetService` → `.thorchainStagenet`
- `MayachainService` → `.mayachain`

The call site becomes `return service.makeSwapQuote(quote)`. The four mappings are **byte-identical** to what the old type-probe produced. Because it is now a protocol requirement, adding a new conforming service is a **compile error** until it declares its tag — that's the point: the silent `default:` mislabel is impossible now.

## Test
Adds `testMakeSwapQuoteTagsEachServiceWithItsNetworkCase`, which pins each of the 4 concrete services to its correct `SwapQuote` case (and verifies the payload passes through unchanged) so a future edit can't silently regress the mapping.

## Out of scope
The catch-all `catch { throw SwapError.swapAmountTooSmall }` in `fetchCrossChainQuote` is a separate issue (#4902) and is intentionally left untouched.

## Gate
`make test` → `** TEST SUCCEEDED **` — **3819 tests, 1 skipped, 0 failures**. SwiftLint clean. Cross-model Codex read-only review: no concrete issues (confirmed byte-identical mappings, removing `default:` cannot drop/mislabel a case, MockSwapProvider/tests correct).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-chain swap quote handling to ensure quotes are correctly tagged for their corresponding network.
  * Added support for consistent quote identification across THORChain, stagenet, chainnet, and MayaChain swaps.

* **Tests**
  * Added coverage verifying correct network tagging for each supported swap service.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -468,18 +468,7 @@ private extension SwapService {
                 liquidityToleranceBps: liquidityToleranceBps
             )
 
-            switch service {
-            case _ as ThorchainService:
-                return .thorchain(quote)
-            case _ as ThorchainChainnetService:
-                return .thorchainChainnet(quote)
-            case _ as ThorchainStagenetService:
-                return .thorchainStagenet(quote)
-            case _ as MayachainService:
-                return .mayachain(quote)
-            default:
-                return .thorchain(quote)
-            }
+            return service.makeSwapQuote(quote)
         } catch let error as ThorchainSwapError {
             logger.error("THORChain swap error: code=\(error.code, privacy: .public), message=\(error.message, privacy: .public)")
             throw Self.mapThorchainSwapError(error)
--- a/VultisigApp/VultisigApp/Blockchain/THORChain/Service/MayaChainService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/THORChain/Service/MayaChainService.swift
@@ -29,6 +29,10 @@ class MayachainService: ThorchainSwapProvider {
         self.resolver = resolver
     }
 
+    func makeSwapQuote(_ quote: ThorchainSwapQuote) -> SwapQuote {
+        .mayachain(quote)
+    }
+
     /// The override-aware Mayanode host. Falls back to the default host when no
     /// override is set.
     private var resolvedHost: URL {
--- a/VultisigApp/VultisigApp/Blockchain/THORChain/Service/ThorchainService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/THORChain/Service/ThorchainService.swift
@@ -39,6 +39,10 @@ class ThorchainService: ThorchainSwapProvider {
         self.httpClient = httpClient
     }
 
+    func makeSwapQuote(_ quote: ThorchainSwapQuote) -> SwapQuote {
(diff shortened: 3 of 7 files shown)


johnnyluo (approved):
LGTM

PR #4935, opened 2026-07-27
refactor(coin-service): resolve token discovery via TokenDiscoverer strategy registry

### LINKED ISSUE #4917: [Refactor] CoinService.fetchDiscoveredTokens nested switch → TokenDiscoverer strategy [Discovery]
# [Refactor] CoinService.fetchDiscoveredTokens nested switch → TokenDiscoverer strategy [Discovery]

## Problem (code smell — Tier 1, S10)
`Core/Services/CoinService.swift` `fetchDiscoveredTokens` (~L224-261) is a nested `switch chainType` → `switch chain` (THORChain arm) with a `.Cosmos where allowlisted` guard, terminating in silent `default: tokens = []` (and a nested THORChain `default: []`). A new chain/family whose discovery isn't wired silently discovers **zero tokens** — no error, no compile failure.

## Solution — Strategy per chain/family
```swift
protocol TokenDiscoverer { func discover(native: CoinMeta, address: String) async throws -> [CoinMeta] }
let discoverer = TokenDiscovererRegistry.discoverer(for: nativeCoin.chain)  // chains w/o discovery register an explicit no-op
```
"No discovery" becomes a declared choice, not a `default`. Keep the current per-chain behavior (EVM/Solana/Sui service factories, THORChain variants, Cosmos allowlist) exactly.

### Must Do
- [ ] Extract each discovery branch into a `TokenDiscoverer`; the mega-switch → a registry lookup; explicit no-op for non-discovering chains
### Must NOT Do
- Don't change which tokens any chain currently discovers

## Acceptance Criteria
- [ ] Existing discovery behavior preserved (add a routing test); no chain silently regresses to empty

gastonm5:
## Problem
`CoinService.fetchDiscoveredTokens` resolved token discovery through a nested `switch nativeCoin.chain.chainType -> switch nativeCoin.chain`, with a `where`-guarded Cosmos branch and a silent `default: tokens = []` fall-through. Adding a chain family meant editing the mega-switch, and a chain that fell through the `default` silently discovered nothing with no compile-time signal.

## Approach
Introduce a `TokenDiscoverer` strategy protocol with one concrete conformer per discovery branch — `EVMTokenDiscoverer`, `SolanaTokenDiscoverer`, `SuiTokenDiscoverer`, `ThorchainTokenDiscoverer`, `MayachainTokenDiscoverer`, `CosmosBankDenomTokenDiscoverer` — plus an explicit `NoTokenDiscoverer` no-op. `TokenDiscovererRegistry.discoverer(for:)` resolves the strategy via a switch that is **exhaustive over `ChainType` with no `default`**, so a new chain family must declare its discovery behavior (a real discoverer or `NoTokenDiscoverer`) or the build fails. `fetchDiscoveredTokens` collapses to a two-line lookup + delegate.

The `TokenDiscoverer` protocol and the registry are `@MainActor`. This is not a behavior change: `CoinService` is already `@MainActor`, so `fetchDiscoveredTokens` and its service calls already ran on the main actor — the annotation just makes the existing isolation explicit (and is required so the registry can construct the discoverer values).

## Behavior-preservation note
Every original branch is preserved exactly: EVM, Solana, Sui, the three THORChain networks, and MayaChain invoke the identical services with identical arguments; Cosmos keeps the `.terra`/`.terraClassic` allowlist, the `!isHidden` filter, and the `CoinMeta` conversion. Every formerly-defaulted case (`UTXO, Cardano, Polkadot, Ton, Ripple, Tron`, non-allowlisted Cosmos, and any non-thor/maya `.THORChain` chain) resolves to `NoTokenDiscoverer` returning `[]`, matching the original `default: tokens = []`. No functional change.

## Verification
- Full test suite on iPhone 16 simulator: `Executed 3894 tests, with 1 test skipped and 1 failure`. The full suite passes **except** the pre-existing `CreateVaultSnapshotTests.testCreateVaultScreen_iPhone16Pro` snapshot artifact — a sub-pixel anti-aliasing delta on the Create Vault screen, unrelated to this diff (which touches only `CoinService`/`TokenDiscoverer`) and failing identically on `main` and other branches. No false-green claim.
- Cross-model Codex review (read-only, two rounds): verdict "behavior-preserving; no regression found" — mapping exhaustive over all 11 `ChainType` cases, no chain case missed, `@MainActor` introduces no actor-hop change for this path.

## Risks
- The exhaustive `ChainType` switch is intentional: adding a chain family will fail to compile until its discovery behavior is declared — a safety property, not a defect.
- `TokenDiscoverer` strategies are `@MainActor`-only; harmless for the sole current caller, but future reuse from a background actor would need to account for it.

Closes #4917

## Summary by CodeRabbit

* **New Features**
  * Improved automatic token discovery across supported EVM, Solana, Sui, THORChain, MayaChain, and Cosmos networks.
  * Cosmos token discovery now displays only visible bank-denominated assets.
  * Added consistent handling for networks without token-discovery support, returning no results instead of attempting unsupported lookups.
* **Bug Fixes**
  * Improved reliability and consistency when discovering tokens across different blockchain networks.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/CoinService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/CoinService.swift
@@ -222,42 +222,8 @@ struct CoinService {
     }
 
     static func fetchDiscoveredTokens(nativeCoin: CoinMeta, address: String) async throws -> [CoinMeta] {
-        var tokens: [CoinMeta] = []
-        switch nativeCoin.chain.chainType {
-        case .EVM:
-            let service = try EvmService.getService(forChain: nativeCoin.chain)
-            tokens = try await service.getTokens(nativeToken: nativeCoin, address: address)
-        case .Solana:
-            tokens = try await SolanaService.shared.fetchTokens(for: address)
-        case .Sui:
-            tokens = try await SuiService.shared.getAllTokensWithMetadata(address: address)
-        case .THORChain:
-            switch nativeCoin.chain {
-            case .thorChain, .thorChainChainnet, .thorChainStagenet:
-                let service = ThorchainServiceFactory.getService(for: nativeCoin.chain)
-                tokens = try await service.fetchTokens(address)
-            case .mayaChain:
-                tokens = try await MayachainService.shared.fetchTokens(address)
-            default:
-                tokens = []
-            }
-        case .Cosmos where CosmosCoinFinder.allowlistedChains.contains(nativeCoin.chain):
-            // Terra + TerraClassic auto-discover held bank denoms via the
-            // Cosmos finder. Hidden-by-default denoms are surfaced separately
-            // through `addDiscoveredHiddenTokens` so they don't pollute the
-            // visible coin list — only `!isHidden` discoveries land here.
-            let discovered = try await CosmosCoinFinder.shared.discoverBankDenoms(
-                chain: nativeCoin.chain,
-                address: address
-            )
-            tokens = discovered
-                .filter { !$0.isHidden }
-                .map { $0.toCoinMeta(chain: nativeCoin.chain) }
-        default:
-            tokens = []
-        }
-
-        return tokens
+        let discoverer = TokenDiscovererRegistry.discoverer(for: nativeCoin.chain)
+        return try await discoverer.discoverTokens(nativeCoin: nativeCoin, address: address)
     }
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4937, opened 2026-07-27
fix(send): re-solve Terra Classic burn-tax fixed point on Verify Max (no 1-uluna underfunding)

### LINKED ISSUE #4854: [Fix] Terra Classic Verify re-derives native Max without re-solving the burn-tax fixed point — can underfund by 1 uluna [Send / Verify]
# [Fix] Terra Classic Verify re-derives native Max without re-solving the burn-tax fixed point — can underfund by 1 uluna [Send / Verify]

## Problem

For Terra Classic, `SendCryptoVerifyViewModel` re-derives a native Max as `candidate = balance − fee − ED` at Verify time. But `fee` embeds a **burn tax computed from the previous amount**, so the re-derived candidate does not satisfy the tax fixed point and can land 1 uluna above what the balance actually supports.

Writing `A` for the amount the Details screen produced and `A*` for the exact fixed point `(balance − baseGas) / 1.005`:

```
candidate − A = (balance − baseGas) − tax(A) − A = 1.005·A* − 1.005·A = 1.005·(A* − A)
```

So the overshoot is driven entirely by the truncation residue `A* − A`. The Verify re-derivation itself never truncates — it formats `candidate / 10^decimals` straight to a string — so it inherits whatever residue the Details screen left.

**Impact is small but non-zero:** the worst case is a 1 uluna deficit (0.000001 LUNC), which surfaces as a broadcast failure on a Max send rather than a wrong amount.

## Expected Behavior

A Terra Classic native Max should always be spendable — the Verify re-derivation should solve the burn-tax fixed point rather than reusing a fee derived from a different amount.

## Steps to Reproduce

Not reliably hand-reproducible — it depends on where the balance falls relative to the truncation residue. In simulation over 4000 balances (baseGas 30000 uluna, tax rate 0.5%) it occurs in ~0.5% of balances after #4848; see below.

1. Hold LUNC in a vault
2. Tap Max on a native LUNC send
3. For a balance whose residue lands unfavourably, the broadcast fails as underfunded

## Solution

Re-solve the fixed point at Verify instead of subtracting a stale `fee`: derive the amount from `(balance − baseGas) / (1 + taxRate)` and truncate **down**, so the resulting tax is consistent with the amount actually being sent. Alternatively clamp the re-derived candidate to the Details-screen amount when the latter is already valid.

### Must Do
- [ ] Make the Terra Classic Verify re-derivation solve the burn-tax fixed point (or clamp so it can never exceed a spendable amount)
- [ ] Add a regression test over the balance range that currently produces the 1-uluna deficit

### Must NOT Do
- Don't change the Details-screen Max path — it is correct as of #4848
- Don't change non-Terra-Classic Verify re-derivation

## Acceptance Criteria
- [ ] Build succeeds
- [ ] A Terra Classic native Max is spendable across a swept balance range (no underfunded cases)
- [ ] Non-Terra chains' Verify behaviour unchanged

## Context

Pre-existing; **not** introduced by #4848 (`fix(send): fill the full balance on Max instead of stranding a digit of dust`). Raised by Codex during that PR's review as a claimed regression and rejected after empirical checking — the causal direction is inverted. Because the deficit scales with the truncation residue, finer truncation *shrinks* it:

| truncation | underfunded cases | worst deficit |
|---|---|---|
| old, 5 dp (`decimals − 1`) | 200 / 4000 | 1 uluna |
| new, 6 dp (`decimals`) | **20 / 4000** | 1 uluna |

So #4848 improved frequency 10× and left the worst case unchanged. This issue tracks the residual case, which exists independently of that PR.

gastonm5:
## Problem

On a Terra Classic (LUNC) native **Max** send, `SendCryptoVerifyViewModel.loadGasInfoForSending()` re-derives the amount at Verify time as `candidate = balance − fee − ED`. The freshly-refetched `fee` embeds a **burn tax computed from the Details-screen amount** (`baseGas + tax(A)`), so `balance − fee` re-adds the balance the tax already consumed and can land **1 uluna above** what the balance actually supports. Once the chain recomputes the tax on the larger candidate at broadcast, the send is underfunded — a **broadcast failure on a Max send** (fund-safety, not a wrong displayed amount).

## Root cause (the fixed-point math)

Writing `A` for the Details amount and `A* = (balance − baseGas)/1.005` for the exact fixed point, with `A = floor(A*)`:

```
candidate − A = (balance − baseGas) − tax(A) − A ≈ 1.005·(A* − A)
```

The overshoot is driven entirely by the truncation residue `A* − A ∈ [0,1)`. `candidate` is therefore `A` or `A+1`; when it is `A+1`, `tax(A+1)` can exceed `tax(A)` by 1 uluna and the send underfunds. The Details-screen Max path (`SendCryptoLogic.terraClassicMaxValue`, correct as of #4848) already solves the fixed point and is spendable; only the Verify re-derivation overshoots.

## Fix

Clamp the Terra Classic Verify re-derivation to the Details amount, which already solves the fixed point:

```swift
static func verifyMaxCandidateRaw(coin:, fee:, previousAmountRaw:) -> BigInt {
    let candidate = balance - fee - existentialDeposit(for: coin)
    guard coin.chain == .terraClassic else { return candidate }
    return Swift.min(candidate, previousAmountRaw)
}
```

The clamp only ever *lowers* the re-derived amount, so it stays spendable in both directions: if the fee is unchanged/lower, it yields the spendable Details amount `A`; if the fee rose so `candidate < A`, the smaller amount owes no more tax than `A` did (`tax(candidate) ≤ tax(A)`), so `candidate + baseGas + tax(candidate) ≤ balance`. **Non-Terra-Classic re-derivation is byte-identical (`balance − fee − ED`); the Details-screen Max path is untouched.**

## Verification

- **Swept-balance regression test** (`SendMaxAmountTests.testTerraClassicVerifyMaxNeverUnderfundsAcrossSweptBalances`): sweeps 4000 contiguous balances at `baseGas = ulunaBaseGas`, rate 0.5%, and asserts (a) the pre-fix `balance − fee` formula **does** overshoot (bug is real, ~20/4000), and (b) the clamped re-derivation is spendable (`amount + baseGas + burnTax(amount) ≤ balance`) for **every** balance — **0 underfunded**. Plus 3 focused unit tests (non-Terra passthrough, Terra clamp-down, fee-grew keeps smaller candidate). **All 4 passed.**
- **Gate** (`make test`, iPhone 16 Plus): `** TEST FAILED **` / `Executed 3898 tests, 1 skipped, 1 failure`. The **sole** failure is the pre-existing, unrelated `CreateVaultSnapshotTests.testCreateVaultScreen_iPhone16Pro` sub-pixel snapshot artifact (fails identically on `main`, nothing to do with Send/Verify). All Send/Verify/Max suites are green.
- **Codex read-only review:** no correctness or fund-safety blocker; confirmed the monotonicity safety argument and that non-Terra behavior is unchanged. Non-blocking suggestions (add an explicit DOT/ED assertion, a wording nuance) deferred.
- ⚠️ **On-device Max-send broadcast verification is a manual step not performed here** — the fix is validated by the swept-range math test, not an on-chain LUNC Max send.

## Risks

- Live burn-tax rate could exceed the fallback 0.5%; the clamp only lowers the Verify amount, so a higher live rate cannot make it overshoot. Worst case is a slightly conservative Max (strands < 1 uluna), never underfunding.

## Review flags

🛑 **Fund-safety change** (prevents underfunding real Max sends) → requesting **human review + on-device Max-send verification** before merge.

Closes #4854

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved maximum native-token send amount calculations during transaction verification.
  * Prevented Terra Classic send amounts from exceeding the previously confirmed spendable amount.
  * Reduced rounding and fee-related discrepancies that could cause verified transactions to be underfunded.

* **Tests**
  * Added coverage for maximum-send calculations across supported native-token chains, including Terra Classic edge cases.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Send/Logic/SendCryptoLogic.swift
+++ b/VultisigApp/VultisigApp/Features/Send/Logic/SendCryptoLogic.swift
@@ -196,6 +196,29 @@ enum SendCryptoLogic {
         return scaled < .zero ? 0 : scaled.truncated(toPlaces: coin.decimals)
     }
 
+    /// Re-derive the native-Max raw amount at Verify time from a freshly
+    /// resolved `fee`. The Verify screen refetches the fee after the Details
+    /// screen produced the amount, so a native Max is re-derived here as
+    /// `balance − fee − ED`.
+    ///
+    /// Terra Classic needs a clamp: its fee embeds a proportional burn tax
+    /// computed from the amount the Details screen produced, so `balance − fee`
+    /// re-adds the balance the tax already consumed and overshoots the tax
+    /// fixed point by up to one unit. Sending that candidate underfunds by up
+    /// to 1 uluna once the chain recomputes the tax on the larger amount at
+    /// broadcast. The Details amount already solves the fixed point
+    /// (`terraClassicMaxValue`) and is spendable, so the re-derived candidate
+    /// must never exceed it — clamping down also stays spendable when the fee
+    /// rose (a smaller amount owes no more tax than the Details amount did).
+    ///
+    /// Every other chain keeps `balance − fee − ED` unchanged.
+    static func verifyMaxCandidateRaw(coin: Coin, fee: BigInt, previousAmountRaw: BigInt) -> BigInt {
+        let balance = coin.rawBalance.toBigInt(decimals: coin.decimals)
+        let candidate = balance - fee - existentialDeposit(for: coin)
+        guard coin.chain == .terraClassic else { return candidate }
+        return Swift.min(candidate, previousAmountRaw)
+    }
+
     /// Apply a percentage (0–100) to a max amount string. Used by the
     /// "25% / 50% / 75%" preset buttons in the Details screen.
     ///
--- a/VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyViewModel.swift
@@ -110,14 +110,19 @@ class SendCryptoVerifyViewModel: ObservableObject {
             var newAmount = transaction.amount
             // Adjust amount for max send if fee changed (only for native tokens where fee is deducted from balance)
             if transaction.sendMaxAmount && transaction.coin.isNativeToken {
-                let balance = transaction.coin.rawBalance.toBigInt(decimals: transaction.coin.decimals)
-                // Reserve the existential deposit so a DOT max-send settles at
-                // `balance − fee − ED`; `transfer_keep_alive` rejects a transfer
-                // that would reap the sender. Zero for non-ED chains — including
-                // TAO (`transfer_allow_death`) and XRP, whose rawBalance is
-                // already reserve-net, so its max settles at `balance − fee`.
-                let existentialDeposit = SendCryptoLogic.existentialDeposit(for: transaction.coin)
-                let candidate = balance - feeResult.fee - existentialDeposit
+                // `balance − fee − ED`. The ED reservation lets a DOT max-send
+                // settle at `balance − fee − ED` (`transfer_keep_alive` rejects
+                // a transfer that would reap the sender); zero for non-ED chains
+                // — including TAO (`transfer_allow_death`) and XRP, whose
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4939, opened 2026-07-27
fix(swap): stop relabeling unknown/internal failures as money errors (#4902, #4903)

### LINKED ISSUE #4902: [Fix] Swap error handling relabels unknown errors as user-facing money errors (3 sites) [Swap]

### LINKED ISSUE #4903: [Fix] Swap payload builder disguises internal errors as insufficient funds; defaults signed provider tag to 1inch [Swap]
# [Fix] Swap payload builder disguises internal errors as "insufficient funds" and defaults the signed provider tag to 1inch [Swap]

## Problem
Two escape hatches in `buildSwapKeysignPayload`:
1. `guard let quote = transaction.quote else { throw Errors.insufficientFunds }` — a nil quote is an internal/state bug ("limit orders never reach this builder"), yet the user sees a **money error** ("insufficient funds") for what is actually a routing bug — undiagnosable.
2. `provider: quote.swapProviderId ?? .oneInch` — a `?? .oneInch` fallback **stamps the signed `GenericSwapPayload.provider`**. Dead today (all grouped EVM cases have non-nil `swapProviderId`), but a latent mislabel: a future EVM aggregator case whose `swapProviderId` is nil would be signed/attributed as 1inch.

## Location
`VultisigApp/VultisigApp/Features/Swap/Logic/SwapPayloadBuilder.swift:218-220` and `:342`

## Solution
- Throw a non-money internal error for the nil quote (`Errors.unexpectedError` already exists) instead of `insufficientFunds`.
- Make the provider non-optional at the type level (each aggregator case carries its `SwapProviderId`, or split the grouped case) so no `?? .oneInch` guess is possible.

### Must Do
- [ ] nil-quote → internal/unexpected error, not a money error
- [ ] Remove the `?? .oneInch` fallback; provider is statically known per case

### Must NOT Do
- Don't change the signed provider tag for any route that works today (byte-preserving)

## Acceptance Criteria
- [ ] A nil quote surfaces an internal error, not "insufficient funds"
- [ ] EVM aggregator swaps stamp the same `provider` as before

gastonm5:
## Problem

Five sites in the swap flow turned **unclassified or internal-state failures** (request timeout, TLS error, JSON decode failure, node 5xx, a `nil` market quote, a guessed provider) into confident, user-facing **"money" verdicts** — "amount too small", "insufficient gas", "insufficient funds". This masks the real cause and actively misleads users about their balances on a money path.

Closes two issues:
- **#4902** — unknown failures relabeled as money errors.
- **#4903** — an internal bug (`nil` quote) reported as a funding problem, and a grouped-EVM code path that defaulted a `nil` provider to `.oneInch` (which would sign a LI.FI / KyberSwap payload as 1inch).

## Root cause

Each site substituted a hard-coded `SwapError` / `SwapCryptoLogic.Errors` money case for *any* error it didn't otherwise handle, rather than surfacing the real error or an honest internal-error signal.

## Fix

**#4902 — surface the real error instead of inventing a money error:**
- `SwapService.fetchCrossChainQuote`: the final catch-all mapped every unclassified error to `SwapError.swapAmountTooSmall`; now logs and rethrows the real error. The typed `ThorchainSwapError` / `MayachainSwapError` / `SwapError` catches above are untouched.
- `SwapDetailsViewModel.updateFees`: the `default:` relabeled every non-UTXO fee failure as `insufficientGas`; now surfaces the real error. Typed UTXO cases already flowed through, so the switch collapses to `self.error = error`.
- `SwapPayloadBuilder.thorchainFee` (UTXO/Cardano): dropped the catch-all `insufficientFunds` relabel; typed `KeysignPayloadFactory.Errors` still pass through, other errors log + rethrow. A **non-positive planned fee** — previously inferred as `insufficientFunds` — is now reclassified as the internal `unexpectedError` and **fails closed** (it never propagates a zero network-fee reservation downstream), rather than either a false money verdict or silent acceptance.

**#4903 — internal bug / guessed provider:**
- `SwapPayloadBuilder.buildSwapKeysignPayload`: a `nil` market quote is an internal/state inconsistency (limit orders build via `LimitSwapPayloadAssembler`), so it now throws the internal `unexpectedError` instead of `insufficientFunds`.
- The grouped EVM case (`.oneinch` / `.lifi` / `.kyberswap`) no longer defaults `provider` to `.oneInch` when `swapProviderId` is `nil`; `swapProviderId` is non-nil for all three by construction, so a `guard let` + throw fails closed instead of mis-attributing the signed payload. **Fail-closed, never mis-signs.**

## Reconciliation with #4901 / #4907

This branch was cut before **#4907** ("tag native quotes via a per-service Strategy, not a type-probe", issue #4901) merged. #4907 touches the **same method**, `SwapService.fetchCrossChainQuote`, but a different, non-overlapping hunk: it replaced the top-of-function `switch service { … default: .thorchain }` quote-tagging block with `service.makeSwapQuote(quote)`. This branch replaces the **final catch-all**. #4907's own commit message states the catch-all "is a separate issue and is intentionally left alone", confirming they are complementary. The rebase auto-merged cleanly; the reconciled function keeps **both** changes with the three typed catches preserved between them. All 5 sites were re-verified post-merge — none were pre-empted by #4901.

## Verification

- **Full test gate** (`xcodebuild test`, whole VultisigApp suite, iPhone 16 Pro Max sim): `Executed 3896 tests, with 1 test skipped and 1 failure`. **No false-green claim:** the run reports `** TEST FAILED **`, but the *sole* failing test is `CreateVaultSnapshotTests.testCreateVaultScreen_iPhone16Pro` — a pre-existing ~sub-pixel anti-aliasing snapshot artifact on the CreateVault screen that fails identically on `main`, unrelated to swap code (a sibling PR is already tracking it). **Zero swap tests failed** — `SwapService` / `SwapPayloadBuilder` / `SwapDetailsViewModel` all green, and both new tests pass.
- **Cross-model Codex review (2 rounds):** round 1 raised one P1 — dropping the non-positive-fee guard removed a fail-closed invariant; addressed by reintroducing it as an internal `unexpectedError` (with a typed `catch let error as Errors` so it rethrows intact, not relabeled). Round 2: **no findings** — confirmed fail-closed restored, catch-ordering correct, provider guard never mis-signs, `nil`-quote change breaks no caller.
- New unit tests: `buildSwapKeysignPayload` throws `unexpectedError` (not `insufficientFunds`) on a missing quote; `updateFees` surfaces an injected generic fee error instead of `insufficientGas`.

## Risks

- **Behavior change on a money/signing path.** Errors previously swallowed into a friendly money message now surface as their real error (generic title + raw localized description). This is the intent, but the arbitrary-error UI copy should be sanity-checked on device.
- **Deferred:** a unit test for a *zero planned fee* (vs. a thrown planner failure) — `thorchainFee` is a static function with no injection seam for the UTXO/Cardano planner, so forcing a zero fee needs harness work beyond this fix's scope. The new guard's logic is covered by the Codex review.
- Touches the swap **payload/signing** path → **requires human review + on-device review** before merge.

Closes #4902
Closes #4903

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/Common/SwapService.swift
@@ -478,7 +478,11 @@ private extension SwapService {
         } catch let error as SwapError {
             throw error
         } catch {
-            throw SwapError.swapAmountTooSmall
+            // Any error not classified above (timeout, TLS, decode, node 5xx)
+            // is unknown here — surface the real error instead of a confident
+            // "amount too small" money verdict the app cannot actually justify.
+            logger.error("Cross-chain quote failed with unclassified error: \(error.localizedDescription, privacy: .public)")
+            throw error
         }
     }
 
--- a/VultisigApp/VultisigApp/Features/Swap/Logic/SwapPayloadBuilder.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/Logic/SwapPayloadBuilder.swift
@@ -9,6 +9,9 @@
 
 import BigInt
 import Foundation
+import OSLog
+
+private let logger = Logger(subsystem: "com.vultisig.app", category: "swap-payload-builder")
 
 extension SwapCryptoLogic {
 
@@ -49,15 +52,31 @@ extension SwapCryptoLogic {
                     planFee = BigInt(plan.fee)
                 }
 
-                if planFee <= 0 && fromAmount > 0 {
-                    throw Errors.insufficientFunds
+                // A non-positive planned fee is an anomalous planner result, not
+                // a funding verdict: fail closed with an internal error rather
+                // than reserving a zero network fee downstream (or, as before,
+                // inferring "insufficient funds" from it).
+                guard planFee > 0 || fromAmount <= 0 else {
+                    logger.error("UTXO/Cardano fee planner returned a non-positive fee")
+                    throw Errors.unexpectedError
                 }
                 return planFee
+            } catch let error as KeysignPayloadFactory.Errors {
+                // The UTXO/Cardano transfer builder raises typed funding errors
+                // (notEnoughUTXO / utxoTooSmall / utxoSelectionFailed) — trust
+                // them as the authoritative "insufficient funds" signal.
+                throw error
+            } catch let error as Errors {
+                // Our own internal verdict (e.g. the non-positive-fee guard
+                // above) is already correctly typed — rethrow without relabeling.
+                throw error
             } catch {
-                if error is KeysignPayloadFactory.Errors {
-                    throw error
-                }
-                throw Errors.insufficientFunds
+                // Any other failure (fee-plan build, decode, transport) is not a
+                // funding verdict — log and surface the real error instead of
+                // inferring "insufficient funds" (previously also inferred from a
+                // non-positive planned fee).
+                logger.error("UTXO/Cardano swap-fee planning failed: \(error.localizedDescription, privacy: .public)")
+                throw error
             }
 
         case .Cosmos, .THORChain, .Polkadot, .MayaChain, .Solana, .Sui, .Ton, .Ripple, .Tron:
@@ -214,9 +233,11 @@ extension SwapCryptoLogic {
         )
 
         // Limit orders never reach this builder — they construct their
-        // payload via `LimitSwapPayloadAssembler`. Guard for symmetry.
+        // payload via `LimitSwapPayloadAssembler`. A missing quote here is an
+        // internal/state inconsistency, not a funding problem, so surface it as
+        // an internal error rather than a confident "insufficient funds" verdict.
         guard let quote = transaction.quote else {
-            throw Errors.insufficientFunds
+            throw Errors.unexpectedError
         }
 
         switch quote {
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

Issue #4758, opened 2026-07-06
feat(ripple): send XRP issued-currency tokens


realpaaao:
## What
**Send** XRP issued-currency tokens (via the SDK CurrencyAmount path). Destination-tag rules from the hardening set still apply.

iOS: Ripple.swift send flow.

Consumes the SDK logic (vultisig-sdk#996). Part of the **XRP token support** feature set (all platforms, low-prio backlog); 1:1 with the SDK issue.

## Acceptance
- [ ] Send an XRP token end-to-end (broadcasts + confirms)
- [ ] Amount uses issued-currency (currency/issuer/value), not drops
- [ ] Destination-tag handling applies

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Partial payments (delivered_amount): https://xrpl.org/docs/concepts/payment-types/partial-payments

johnnyluo:
Already implemented in PR #4944 (feat(ripple): XRP trust lines + issued-currency token sends), merged into main via the #4756 stack. GitHub's `Closes #4758` keyword didn't fire because that PR's base was a stacked feature branch, not `main`, at merge time — the code itself is confirmed live on `main` and proven on XRPL mainnet (see the PR body for the signed Payment tx with a currency/issuer/value Amount).

Issue #4757, opened 2026-07-06
feat(ripple): add XRP token / open trust line UI (+ 0.2 XRP reserve warning)


realpaaao:
## What
**Add token / open trust line** UI: let the user add an XRP token by issuer + currency (curated entry for RLUSD + custom-add). Opening a trust line issues a TrustSet and **must warn about the 0.2 XRP owner-reserve cost**.

iOS: send/asset UI.

Consumes the SDK logic (vultisig-sdk#995). Part of the **XRP token support** feature set (all platforms, low-prio backlog); 1:1 with the SDK issue.

## Acceptance
- [ ] Add-token UI (RLUSD + custom by issuer/currency)
- [ ] Opening a trust line surfaces the 0.2 XRP reserve warning before signing
- [ ] Trust line opens (TrustSet) end-to-end

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Trust lines: https://xrpl.org/docs/concepts/tokens/fungible-tokens/authorized-trust-lines
- Reserves (0.2 XRP/trust line): https://xrpl.org/docs/concepts/accounts/reserves

johnnyluo:
Already implemented in PR #4944 (feat(ripple): XRP trust lines + issued-currency token sends), merged into main via the #4756 stack. GitHub's `Closes #4757` keyword didn't fire because that PR's base was a stacked feature branch, not `main`, at merge time — the code itself is confirmed live on `main`.

PR #4940, opened 2026-07-27
feat(ripple): show XRP trust-line token balances in the asset list

### LINKED ISSUE #4756: feat(ripple): show XRP token balances in the asset list
## What
Show XRP trust-line **token balances** in the asset list (reuse the existing token-discovery/asset-list pattern from EVM/Solana/Cosmos — no new UI states).

iOS: Ripple.swift / asset list.

Consumes the SDK logic (vultisig-sdk#997). Part of the **XRP token support** feature set (all platforms, low-prio backlog); 1:1 with the SDK issue.

## Acceptance
- [ ] XRP issued-currency tokens appear in the asset list with correct balances (issuer + currency)
- [ ] Reuses existing token/asset-list patterns; no bespoke states

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Reserves: https://xrpl.org/docs/concepts/accounts/reserves
- account_lines: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines

gastonm5:
Closes #4756

https://xrpscan.com/tx/BF0C41AC7A54DB3673D5C05F0C7510E50A71E9BFD37200D1E2E78074E0E17042

XRP trust-line (issued-currency) support: balances in the asset list, **and the signing surface that goes with them**.

> ### ⚠️ Scope correction
>
> This description previously read *"Read-only: this PR touches no signing path and makes no wire change — RippleHelper / Ripple.swift are untouched, and no commondata field is added"*, and described this as PR 1 of three with #4757/#4758 to follow. **None of that is true at HEAD**, and a reviewer trusting it would skip the part of this PR that most needs review. Corrected here after @NeOMakinG flagged it.
>
> What this branch actually carries:
>
> - **`Ripple.swift` implements signing** — TrustSet, issued-currency Payment, and the dApp rawJson gate.
> - **A new commondata field**, `RippleSpecific.transaction_type`, discriminates the signing arm (`vultisig/commondata#100`).
> - **~20 new localizable keys**, across all locale files.
> - **wallet-core bumped 4.3.19 → 4.7.1** — load-bearing, since 4.7.1 is the first release whose `Ripple.proto` carries `RippleOperationTrustSet` / `CurrencyAmount`.
> - The #4757 / #4758 work described elsewhere as separate follow-ups is absorbed here.
>
> **The fund-safety signing surface is the part to review**, not the balance display.

> **Base was repointed to `main`** after #4935 merged, so the `Closes #4756` above now resolves normally.

Plan: `wiki/pages/projects/vultisig/xrp-hardening/xrp-token-support-4756-4757-4758-plan.md`. Originally scoped as PR 1 of three; the #4757 (add-token / TrustSet / Activate CTA) and #4758 (issued-currency Payment) work is now included in this branch rather than following separately.

## What changed

**`account_lines` RPC** (`RippleAPI`, `RippleTrustLine`) — the XRP Ledger returns trust lines in pages, so `RippleService.fetchAccountLines(for:host:)` follows the `marker` cursor to completion. Reading one page would silently truncate an account with many lines and *hide* balances rather than fail. The walk is pinned to one resolved host (a custom-RPC change mid-walk must not stitch two networks' pages together) and each page goes through the existing same-host retrier. `limit` / `no_ripple` / `freeze` are decoded now because #4757 needs them and decoding costs nothing.

**Error semantics.** `actNotFound` on the first page is an empty set — an unfunded account has no AccountRoot and so holds no trust lines. Everything else fails loud rather than degrading to "no lines": a non-retryable node error throws, and so does a success body missing `result.lines` (a proxy error page decoding to all-nils is not an empty wallet, and reading it as one would render a real balance as zero — same fail-closed posture as `validateDestinationActivation`).

**Token balance** — `RippleService.getTokenBalance(coin:address:)` parses the coin's composite `<currencyCode>.<issuer>` token id and matches the line whose counterparty is the issuer and whose currency normalizes to the same on-ledger code. Both sides go through `toXrplCurrencyCode`, so a node spelling a non-standard code in lowercase hex still resolves to the coin whose id holds it uppercased; the issuer comparison stays case-sensitive because XRPL issuer addresses are base58.

**A negative trust-line balance reports zero.** A negative balance means the account is the token's *issuer* and owes the counterparty — an issuance liability, not a holding. It must never render as a negative asset. It is also excluded from discovery.

**`BalanceService.case .ripple`** now branches on `coin.isNativeToken`, the way `.ton` branches to `getJettonBalance`. It previously returned the native XRP balance unconditionally, so a token row would have rendered the account's XRP balance. `RippleService` is now injected into `BalanceService.init` (alongside the already-injected `cryptoPriceService`) so that routing is testable without a network.

**Discovery** — `RippleTrustLineTokenDiscoverer` replaces `NoTokenDiscoverer()` on the registry's `.Ripple` arm. The filtering lives in a pure `RippleTrustLineTokens.discoverableTokens` so the rules are testable without a network: only **strictly positive** balances surface (negative = issuance liability; zero = open-but-empty line that would only clutter the list), and a curated `TokensStore` entry wins over line-derived metadata for the logo and price provider, neither of which the ledger can supply.

**Ticker decoding** — `RippleIssuedCurrency.toIssuedCurrencyTicker` ports the SDK's helper: a 40-char hex currency decodes to its ASCII ticker when every decoded byte is printable (`524C555344…00` → `RLUSD`), else the hex is kept verbatim. Unlike the SDK's Node `ascii` decode, a high-bit byte is *not* masked into the ASCII range — an arbitrary 160-bit code stays hex instead of being dressed up as a plausible ticker.

**Curated RLUSD** — `chain: .ripple`, `decimals: 15`, `priceProviderId: "ripple-usd"`, contract `524C555344000000000000000000000000000000.rMxCKbEDwqr76QuheSUMdEGf4B9xJ8m5De`. The hex is *derived* in a test from `"RLUSD"`'s ASCII bytes rather than trusted as a transcription. This also makes RLUSD selectable in the XRP token picker for free.

Ported from the SDK's `chains/ripple/account/lines.ts`, `coin/balance/resolvers/ripple.ts`, `coin/find/resolvers/ripple.ts`, and `rippleKnownIssuedTokens`.

## Two things worth a reviewer's attention

**1. `isSpamToken` no longer rejects an empty logo on Ripple.** Not cosmetic — without it the feature is a no-op. `CoinService.addDiscoveredTokens` runs every discovered token through `isSpamToken`, which rejected any token with an empty logo. XRPL has **no on-ledger token metadata registry**, so every non-curated issued currency has an empty logo, and the rule would have silently dropped 100% of discovered XRP tokens. It also has nothing to catch there: a trust line exists only because the account holder opened one, so an issuer cannot push an unsolicited token into the asset list the way an ERC-20 airdrop can. The exemption is scoped to that single condition; the ticker/URL/non-ASCII spam heuristics still apply.

**2. No `rlusd` logo asset — the logo is left empty.** None exists in `Assets.xcassets`, and none exists in `vultisig-windows`' shared coin SVGs, `vultisig-sdk`, or `vultisig-android` either, so there was nothing to port and nothing was invented. `AsyncImageView` falls back to the ticker initial ("R" in a circle), so the row renders correctly — but dropping an `rlusd` imageset in is a clean follow-up.

## Known limitations (all pre-existing or accepted; called out so they aren't read as new defects)

- **Pagination is not pinned to one ledger.** Each page requests `ledger_index: "current"` (plan-specified, SDK parity), so the ledger can advance mid-walk. A marker invalidated by that advance comes back as a node error and is surfaced (last known balance preserved), and a line repeated across pages is collapsed — the ledger holds at most one line per (counterparty, currency). Only reachable past the 200-lines-per-page threshold. Pinning properly needs a `validated` read plus `ledger_hash`, which is a behavioural change worth its own PR.
- **Pagination markers must be strings.** rippled and Clio both serialize the `account_lines` marker as one. A structured marker fails the decode, which throws and preserves the last known balance rather than truncating — never a wrong balance.
- **Coin identity lowercases the whole contract address**, while XRPL issuer addresses are case-sensitive base58 — `TokensStore.findTokenMeta`, `CoinMeta.uniqueId`/`==`, `Vault.coin(for:)`, and `CoinService.migrateOldContractAddresses` all do it. Two issuers differing only in letter case would be conflated, and the second one's discovery silently skipped. Both would have to be valid checksummed base58 addresses that differ only by case, so this is astronomically improbable; the consequence is a skipped row, not a wrong balance. Pre-existing and shared by all 40+ chains, so changing coin-identity semantics belongs in its own PR rather than here.
- **`TokenSelectionLogic` matches held tokens by ticker, not contract address.** XRPL lets *any* issuer mint `USD`, so two issuers' `USD` collide in the picker. Not introduced here — XRP is just the first chain that makes it reachable. Worth a separate fix keyed on `uniqueId`.
- **A token added from the picker before its trust line exists sits at zero balance.** That is the accepted trade-off from the plan's Decision 2; the "Activate" CTA that makes it legible lands with #4757.
- **`decimals: 15` is a modelling choice, not an on-chain fact.** XRPL issued currencies carry 15 *significant* digits with a floating exponent, not a fixed scale, and `parseIssuedCurrencyValue` truncates below it. Inherited from the SDK and required for parity.
- **Frozen / unauthorized lines are decoded but not acted on.** A send against a frozen line fails on-ledger with a real engine code rather than a friendly pre-check.

## Tests

`VultisigAppTests/Blockchain/Ripple/RippleTrustLineTests.swift` — canned-JSON HTTP stub, no network, no sleeps, in the style of `RippleBroadcastGatingTests` / `RippleRetryTests`:

- multi-page `marker` pagination assembles the full set, and each page resumes from the previous marker
- the first request carries `ledger_index: "current"` + `ignore_default: true` and **no** `marker` key
- every page goes to the host resolved when the walk started, even when the custom-RPC override changes between pages
- a line repeated across pages is collapsed
- `actNotFound` on page 1 ⇒ `[]`; `actNotFound` mid-pagination throws rather than truncating
- a non-retryable node error, and a success body without `lines`, both throw rather than reporting "no lines"
- a negative balance reports zero, and is excluded from discovery
- a zero balance is excluded from discovery; a dust balance is kept; a value below 15 decimals truncates to zero and is excluded
- same-issuer/different-currency and same-currency/different-issuer do not match, and issuer matching is pinned as case-*sensitive* (base58)
- lowercase hex from the node matches an uppercase token id; a standard 3-char code matches verbatim
- `BalanceService` routes a non-native coin to its trust-line balance (never the XRP balance) and the native coin to `account_info` without reading trust lines
- hex ⇄ ticker round trip; a non-printable hex code stays hex; an all-zero code stays hex
- the ASCII-packed hex of a 3-character ticker is pinned as a *distinct* currency from the standard code (XRPL encodes standard codes at bytes 12–14 behind a `0x00` first byte) so nobody "normalizes" 3-char codes to hex later
- the curated `TokensStore` entry wins over line-derived metadata
- the curated RLUSD currency code equals `toXrplCurrencyCode("RLUSD")`, derived in-test from the ASCII bytes
- `TokenDiscovererRegistry.discoverer(for: .ripple)` is the trust-line discoverer, not a no-op

No new user-facing strings, so no localization change. The two new error cases are technical, developer-facing copy in the same style as `RippleBroadcastError` and only ever reach logs.

## Summary by CodeRabbit

* **New Features**
  * Added XRP Ledger issued-currency trust-line discovery (paginated) and issued-token discovery for curated RLUSD/SOLO/USDC/EQ.
  * Added trust-line activation (TrustSet) flow with reserve/limit-aware verification and updated send/verify UI.
  * Enhanced custom XRP token input support for `currency.issuer`, including token ID resolution.
* **Bug Fixes**
  * Correctly distinguishes native XRP vs issued-currency trust-line balances; negative balances clamp to zero and missing/unavailable trust lines surface as “unknown”.
* **Tests**
  * Expanded end-to-end trust-line, balance, activation, custom-token, and issued-currency signing coverage; updated RPC/http stubs.


johnnyluo:
PR in commondata has been approved and merged , need to update the packages

johnnyluo (commented):
## ⚠️ Scope note

The PR description says this is *"Read-only: this PR touches no signing path and makes no wire change."* That's now stale: #4944 (`XRP trust lines + issued-currency token sends`, #4757/#4758) was merged **into this branch** on 2026-07-28, so the diff against `main` now includes the full `TrustSet`/issued-currency-`Payment` signing path, the new `RippleSpecific.transaction_type` wire field, the activation UI, and the Verify/Done-screen changes — none of which the body describes. Worth updating the description (or splitting the PR) so reviewers aren't scoped to "read-only" by mistake.

## Findings

**1. Possible nested-button conflict — `TokenCellView.swift` + `ChainDetailListView.swift`**
`ChainDetailListView` wraps each row in `Button { onPress(token) } label: { TokenCellView(...) }`. For a token needing a trust line, `TokenCellView` now renders its own `PrimaryButton("Activate", action: onActivate)` *inside* that same label. Nested `Button`s have unreliable SwiftUI hit-testing — tapping "Activate" could also (or instead) fire the outer row's `onPress`, navigating to coin detail rather than opening the activation sheet. Worth confirming on-device/simulator; if it misbehaves, isolate the hit-test region (e.g. `.buttonStyle(.plain)` plus explicit contentShape framing) rather than nesting tappable controls.

**2. Stale "UNMERGED PIN" warning — `project.yml`**
The `VultisigCommonData` pin comment reads `⚠️ UNMERGED PIN … Must move to the merged main commit before this can ship`. Checked via the GitHub API: `vultisig/commondata#100` merged 2026-07-28, and the pinned SHA (`a16c710f...`) is confirmed an ancestor of commondata's `main` (0 ahead / 1 behind). The pin is valid now, but the comment reads as an active ship-blocker — should be updated/removed before merge so it doesn't mislead the next reader.

**3. Readability nit — `SendCryptoVerifyLogic.validateTrustLineReserveIfNeeded`**
```swift
guard spendable < required else { return }
throw HelperError.runtimeError(...)
```
Correct (insufficient funds falls through to the throw), but inverted from the `guard sufficientFunds else { throw }` style used elsewhere in this file, which reads backwards on a skim. Consider `guard spendable >= required else { throw ... }`.

## What's strong

Fail-closed posture is consistent and well-reasoned throughout: contradictory coin metadata (`isNativeToken` vs. `contractAddress`) is refused before signing, an unsupported `transaction_type` is refused rather than silently downgraded to `.unspecified`, `rawJson`/dApp payloads now refuse non-native coins (closing a real silent token-as-drops fund-loss path), and base58 issuer comparisons stay case-sensitive everywhere they need to. Mixed-version MPC byte-parity is explicitly reasoned about and pinned in tests. Pagination is host-pinned per walk and fails closed on ambiguous/error bodies rather than truncating silently. Locale coverage is complete across all 8 files, alphabetically sorted. Test coverage (canned-JSON stubs, no network/sleeps, byte-level signing parity, pagination/error/dedup cases) is above this codebase's typical bar.

Known trade-offs already disclosed in the PR body (unpinned ledger during pagination, marker-must-be-string assumption, case-sensitive contract-address collision risk, ticker-based token-selection collision, decimals-as-modelling-choice) all look reasonable and appropriately scoped — no new issues found there.

## Bottom line

Solid, well-tested change for a security-sensitive signing path. Before merge: sync the PR description with the actual diff, confirm the nested-button interaction in `TokenCellView`/`ChainDetailListView` on-device, and clean up the stale "UNMERGED PIN" comment.

VultisigApp/VultisigApp/Features/Wallet/ChainDetail/View/TokenCellView.swift
```diff
@@ -38,17 +45,26 @@ struct TokenCellView: View {
...
-                        .font(Theme.fonts.priceCaption)
-                        .foregroundStyle(Theme.colors.textTertiary)
-                        .contentTransition(.numericText())
-                        .animation(.interpolatingSpring, value: coin.balanceStringWithTicker)
+                if let onActivate {
+                    PrimaryButton(
+                        title: "rippleTrustLineActivateAction".localized,
+                        size: .mini,
+                        action: onActivate
+                    )
```
johnnyluo, line 57: **Possible nested-button conflict.** `ChainDetailListView` wraps each row in `Button { onPress(token) } label: { TokenCellView(...) }`. When `onActivate` is set, this `PrimaryButton` renders *inside* that same label — a `Button` nested inside another `Button`'s label. SwiftUI's hit-testing for nested tappable controls is unreliable; tapping "Activate" could also (or instead) fire the outer row's `onPress`, navigating to coin detail instead of opening the activation sheet. Worth confirming on-device/simulator. If it misbehaves, isolate the hit-test region (e.g. `.buttonStyle(.plain)` + explicit `contentShape` framing) rather than nesting tappable controls.

VultisigApp/project.yml
```diff
@@ -57,15 +57,30 @@ packages:
...
-    # Pinned to the merged `main` commit that adds `SignRipple { string raw_json }`
-    # at `sign_ripple = 46` in `KeysignPayload.sign_data` — the XRPL dApp
-    # co-signer passthrough field. This commit is a linear descendant of the
-    # earlier `RippleSpecific.destination_tag` (optional uint32) pin, so it keeps
-    # that first-class XRPL DestinationTag field: the initiator dual-writes the
-    # tag into BOTH the field AND the legacy memo carrier, and signers prefer the
-    # field, falling back to the memo, so mixed-version device pairs stay
-    # byte-identical in MPC keysign.
-    revision: c91f2ea7db5eb03cf6bb637da19e4ed93b07dc46
+    # ⚠️ UNMERGED PIN — points at the open commondata branch that adds
```
johnnyluo: **Stale warning.** This reads `⚠️ UNMERGED PIN … Must move to the merged main commit before this can ship`. I checked via the GitHub API: `vultisig/commondata#100` merged 2026-07-28, and the pinned SHA (`a16c710f...`) is confirmed an ancestor of commondata's `main` (0 ahead / 1 behind). The pin is valid now, but this comment still reads as an active ship-blocker — worth updating/removing before merge so it doesn't mislead the next reader.

johnnyluo (commented):
Inline follow-ups to my earlier review comment (https://github.com/vultisig/vultisig-ios/pull/4940#pullrequestreview-4802424037) — same three findings, anchored to the lines they're about.

VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyLogic.swift
```diff
@@ -205,6 +235,81 @@ struct SendCryptoVerifyLogic {
...
+        let ownerReserve = RippleReserve.reservedDrops(
+            ownerCount: 1,
+            reserveBase: 0,
+            reserveInc: reserveValues?.reserveInc
+        )
+        // The XRP balance is already reserve-net, so it is what can actually be
+        // locked up and spent.
+        let spendable = nativeToken.rawBalance.toBigInt(decimals: nativeToken.decimals)
+        let required = ownerReserve + tx.fee
+        guard spendable < required else { return }
```
johnnyluo: **Readability nit.** `guard spendable < required else { return }` followed by the throw below is correct (insufficient funds falls through to the throw), but it's inverted from the `guard sufficientFunds else { throw }` style used elsewhere in this file, so it reads backwards on a skim. Consider `guard spendable >= required else { throw ... }` for consistency.

johnnyluo:
On MacOS  , the `activate` button is not clickable , the parent row hijacked it
CleanShot 2026-07-29 at 08 49 34@2x

VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleTrustLineTests.swift
```diff
@@ -0,0 +1,679 @@
...
+        case "account_info":
+            guard let accountInfoJSON else { throw StubError.unexpectedMethod(method) }
+            return HTTPResponse(data: Data(accountInfoJSON.utf8), response: Self.ok)
+        case "server_state":
+            guard let serverStateJSON else { throw StubError.unexpectedMethod(method) }
+            return HTTPResponse(data: Data(serverStateJSON.utf8), response: Self.ok)
+        default:
+            throw StubError.unexpectedMethod(method)
+        }
+    }
```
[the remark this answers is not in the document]
gastonm5: Valid, and fixed in `9b4ca194b`.

Confirmed the race first: `RippleService.getBalance` fans `fetchAccountsInfo` and `fetchReserveValues` out with `async let` (`RippleService.swift:333-334`), so in `testBalanceServiceRoutesTheNativeCoinToTheAccountBalance` two tasks really do land in `request(_:)` at once, and the `await Task.yield()` at the top widens the window exactly as you describe.

Recording now goes through an `NSLock`, matching the discipline `ReserveCheckScriptedHTTPClient` / `RippleScriptedHTTPClient` / `TrippingHTTPClient` already use.

Two deliberate differences from the suggested diff:

- The **writes** are locked, not just the reads. `requestedMethods`, `requestedHosts`, `requestedParams` and `callCount` are now private, mutated only inside one critical section, and exposed through locked-snapshot accessors — locking `requestedMarkers` alone would still have left the concurrent `append`s undefined.
- The `account_lines` page index is claimed **inside the same critical section** as the append, so two concurrent calls can never be handed the same page (`recordedCallCount += 1` and the index read are not two separate lock acquisitions).

`pages` / `accountInfoJSON` / `serverStateJSON` stay plain `var`s on purpose: tests write them before the exercise phase and only read them once requests start, exactly as the sibling stubs treat their own canned results.

Full suite green on the rebased branch: **4084 tests, 1 skipped, 0 failures**.

VultisigApp/VultisigApp/Features/Wallet/ChainDetail/View/TokenCellView.swift
gastonm5: Good catch — that was a real nesting problem. Fixed in `f3d5565bf` + `2e0ebe942`.

`ChainDetailListView` wrapped every row in `Button { onPress(token) } label: { TokenCellView(...) }`, so whenever the cell rendered the Activate `PrimaryButton` — itself a real `Button` — it landed inside another `Button`'s label.

**First attempt** replaced the wrapping `Button` with `.contentShape(Rectangle())` + `.onTapGesture` on the cell, leaving the CTA as a sibling. A cross-model review pushed back on it, correctly: that un-nests the two but only moves the ambiguity into SwiftUI's gesture arbitration — a row-wide tap gesture still overlaps the CTA's hit region, and the resolution isn't guaranteed to be the same on iOS and macOS. It also cost *every* row the `Button`'s press feedback, keyboard activation and focus semantics, which an `.isButton` accessibility trait does not restore.

**Final shape** avoids arbitration entirely. `ChainDetailListView.row(for:)` branches on whether the row has an activation CTA:

- **no CTA** — exactly the plain `Button { onPress(token) } label: { TokenCellView(coin: token) }` it always had, unchanged;
- **CTA present** — the cell renders bare, with no row-level tap of any kind, so the CTA is the only control in that row.

Nesting is now structurally impossible rather than arbitrated, and the common case keeps its original `Button` behaviour. `.commonListItemContainer` moved outside the branch (it was inside the label before); it only applies the corner clip and the separator, so rendering is unchanged — the separator simply is no longer part of the tap area.

The CTA row also **drops its chevron**: it was advertising a navigation tap that no longer exists, and coin detail has nothing to show for a token that can neither hold nor receive a balance until the line is open. If you'd rather keep detail navigation reachable for a not-yet-activated token, say so — the only way to keep it without re-introducing an overlap is two visibly separate controls in the row, and I didn't want to make that product call unilaterally.

**What I verified, and what I did not.** Verified: full build + full unit suite green (**4084 tests, 1 skipped, 0 failures**), and by reading the composed hierarchy that no `Button` is a descendant of another `Button`'s label in either branch. **Not verified on device or in the simulator UI** — I did not reproduce a tap on the Activate CTA of a real XRPL row with no trust line, so the hit-testing claim rests on the structure being unambiguous, not on an observed tap.

VultisigApp/project.yml
gastonm5: Right — the comment was stale. Updated in `1e21f4845`.

Verified the ancestry before rewording rather than taking it on trust: `vultisig/commondata#100` (the PR adding `RippleSpecific.transaction_type = 5` and `TRANSACTION_TYPE_RIPPLE_TRUST_SET = 10`) merged on 2026-07-28, and

```
gh api repos/vultisig/commondata/compare/main...a16c710f0b0dfb6e2c8debde54b72414a04feb9a
→ {"status":"behind","ahead_by":0,"behind_by":1,"total_commits":0}
```

`ahead_by: 0` + `behind_by: 1` means the pinned SHA is an **ancestor** of `main` — it is the merged head, and `main`'s only extra commit is the merge commit `98df9d9f…`. So the pin itself was already valid and shippable; only the warning was wrong.

The `⚠️ UNMERGED PIN … Must move to the merged main commit before this can ship` block is gone. The comment now records what the commit adds and that it is merged and an ancestor of commondata `main`; the rationale for why the field has to be on the wire at all is kept, since that is the part a future reader actually needs.

VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyLogic.swift
gastonm5: Agreed — it read backwards. Flipped in `866cc299b`:

```swift
guard spendable >= required else {
    throw HelperError.runtimeError(
        String(
            format: "rippleTrustLineInsufficientXrpError".localized,
            RippleReserve.xrpAmount(drops: required)
        )
    )
}
```

Readability only, and the inversion is exact: `>=` is the negation of `<` over `BigInt`'s total order, so the boundary is unchanged — `spendable == required` still passes, one drop short still throws, and the thrown error, localization key and format argument are untouched.

One thing that turned up while checking "existing tests still pin the same outcomes": **nothing pinned them.** `validateTrustLineReserveIfNeeded` had no coverage at all — `RippleTrustLineActivationTests.testQuoteBlocksWhenSpendableXrpCannotCoverReservePlusFee` covers the sheet's `RippleTrustLineActivationQuote`, not this Verify-stage guard. So the same commit adds it, in `SendRippleDestinationGuardTests`:

- exact cover (`reserve_inc + fee`) passes — the boundary this flip had to preserve;
- one drop short throws, with the required total in the copy;
- a vault with no XRP coin fails closed, before any lookup;
- a non-TrustSet transaction never reaches the network.

A follow-up (`a53aca791`) hardened those: the stub reports `reserve_inc = 500_000` rather than `200_000`, because 200,000 is also `RippleReserve.seedReserveIncDrops` — production reads the value through `try? await fetchReserveValues()`, which degrades to that seed, so the tests would have passed even if the `server_state` request had never happened. Every expected figure is now derived from the stubbed value and both boundary tests assert exactly one node request.

Full suite green: **4084 tests, 1 skipped, 0 failures**.

VultisigApp/VultisigApp/Features/Send/ViewModels/SendCryptoVerifyLogic.swift
```diff
@@ -205,6 +235,81 @@ struct SendCryptoVerifyLogic {
...
+
+        throw HelperError.runtimeError(
+            String(
+                format: "rippleTrustLineInsufficientXrpError".localized,
+                RippleReserve.xrpAmount(drops: required)
+            )
+        )
+    }
```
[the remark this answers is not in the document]
gastonm5: Valid — fixed in `c8f3d6720`.

Confirmed against the code before changing anything: `RippleService.fetchReserveValues()` collapses every recoverable failure into `nil` through its own live → cache → seed chain and rethrows exactly one error, `CancellationError` (`RippleService.swift:393-398`). `try?` erased that one distinction, so a cancelled pass carried on to a seeded verdict and either returned success or threw `rippleTrustLineInsufficientXrpError`. Both are wrong once the pass has been superseded: the caller's generic `catch` writes `errorMessage` / `hasBalanceError` (`SendCryptoVerifyViewModel.swift:193-198`), which are shared `@Published` state, so a stale answer can leave a spurious banner with Sign disabled.

Adopted your do/catch shape — it is the convention already used two functions up in the same file (`validateDestinationIfNeeded`), and it keeps the general `catch → nil` arm so the non-cancellation fallback is unchanged. That arm is unreachable today given the service contract above; it is kept so the contract stays stated at the call site rather than assumed.

Three tests added to `SendRippleDestinationGuardTests`, since the previous round found this function had no coverage at all:

- a balance affordable under the **seed** but not under the stubbed live increment — swallowing the cancel would have returned success, the silent half of the bug that a "did it throw?" test misses;
- a balance below both — swallowing would have raised the spurious `HelperError`;
- an arithmetic test asserting the seeded fallback really would have flipped each verdict, so neither of the first two can pass vacuously.

VultisigApp/VultisigApp/Blockchain/Ripple/Service/RippleCustomTokenResolver.swift
```diff
@@ -0,0 +1,163 @@
...
+            // The standard form excludes `XRP` itself — that spelling is
+            // reserved for the native asset and can never be an issued currency.
+            return currency.uppercased() != "XRP"
+        }
+        if currency.utf16.count == hexCurrencyCodeLength,
+           currency.allSatisfy({ $0.isHexDigit && $0.isASCII }) {
+            return true
+        }
+        return isPackableTicker(currency)
+    }
```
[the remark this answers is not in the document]
gastonm5: The defect is real and is fixed in `09d951d33` — but "enforce ASCII like the hex path" is the wrong predicate, so the fix implements the XRPL rule instead.

**Where the rule was confirmed.** rippled defines the standard-code repertoire itself, in [`src/libxrpl/protocol/UintTypes.cpp`](https://github.com/XRPLF/rippled/blob/develop/src/libxrpl/protocol/UintTypes.cpp):

```cpp
constexpr std::string_view kIsoCharSet =
    "abcdefghijklmnopqrstuvwxyz"
    "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
    "0123456789"
    "<>(){}[]|?!@#$%^&*";
```

`toCurrency` rejects a 3-character code containing anything outside that set, and `to_string` renders a 160-bit value as a 3-character code only when every byte is in it — otherwise it falls back to hex. That matches [xrpl.org's currency-formats page](https://xrpl.org/docs/references/protocol/data-types/currency-formats) ("exactly 3 ASCII characters", "case-sensitive", "`XRP` all-uppercase is disallowed") and xrpl.js's `ripple-binary-codec` `ISO_REGEX` = `/^[A-Z0-9a-z?!@#$%^&*(){}[\]|]{3}$/`, which is the same set minus `<` and `>`. The ledger is the authority for whether a TrustSet succeeds, so the implementation follows rippled's set; the xrpl.js divergence is noted but does not gate this signing path.

**Why plain ASCII is both too broad and too narrow.**

- Too broad: `A-B` is 3 bytes of printable ASCII and would pass an `isASCII` check, but `-` is not in `kIsoCharSet`. rippled would render such a line's currency back as *hex*, so the coin's `A-B` token id could never match its own `account_lines` row.
- The real hazard your finding points at is confirmed, though: a 3-UTF16-unit non-ASCII string is 4+ UTF-8 bytes. wallet-core 4.7.0 (the pinned version) branches on **byte** length — `rust/chains/tw_ripple/src/types/currency.rs`, `if s.len() == 3 { Currency::ISO(s.to_uppercase()) }` — so `ÁBC` matches neither the 3-byte standard form nor the 40-char hex form and fails at `Currency::from_str`. The token was addable, listed, and unsignable.

**One extra narrowing, deliberately stricter than the ledger.** Lowercase letters are excluded. XRPL currency codes are case-*sensitive* (rippled copies the code verbatim into bytes 12–14), but that same wallet-core line calls `.to_uppercase()` before encoding — so `usd` would **sign as `USD`**, a different on-ledger currency from the one added, displayed, and matched against `account_lines`. Refusing a spelling we cannot carry through to signing beats signing a token the user did not choose. The currency stays reachable through its 40-character hex spelling, which the signer encodes byte for byte; there is a test for exactly that.

**The 3-character branch is now terminal**, which matters more than the allowlist: a rejected 3-unit string must not fall through to the packable-ticker branch, which packs at byte 0 and would silently resolve to a different currency than the standard code the user typed. Tests pin that for both the non-ASCII cases and the outside-the-repertoire ones (all 3 bytes of printable ASCII, so they *would* otherwise pack), plus the accepted symbol set, the lowercase rejection, and the hex escape hatch.

No curated `TokensStore` XRPL entry uses a standard 3-character code — all four are the hex form — so nothing shipped changes shape.

**The resolver was not enough, and chasing that down found more.** Validating the add-token field only closes one of the ways a coin reaches the signer — trust-line discovery builds one straight from `account_lines`, and a restored vault carries whatever it was saved with. Since the mangling happens at the WalletCore boundary rather than at input, the predicate moved to `RippleIssuedCurrency.isSignableCurrencyCode`, shared by the resolver and by `issuedCurrencyAmount`, which now fails closed for both TrustSet and issued-currency Payment (`437b6634d`). `7e7caa9e2` applies the same test to a dApp-supplied Payment's `Amount`, where the raw JSON reaches WalletCore unrewritten and two matching lowercase spellings would otherwise have licensed signing a third currency.

Discovery is deliberately **not** filtered: hiding a real holding is a worse failure for a wallet than listing one that refuses to sign, and the signing guard makes that refusal guaranteed rather than hoped for.

Still open, and pre-existing on `main` rather than introduced here, so left for its own change: a dApp `Payment`'s `SendMax` / `DeliverMin`, and the `OfferCreate` / `TrustSet` arms of the co-signer, are bound on `Account` alone and can still carry a manglable code to WalletCore.

VultisigApp/VultisigApp/Features/Wallet/ChainDetail/ViewModel/ChainDetailViewModel.swift
```diff
@@ -59,6 +75,55 @@ final class ChainDetailViewModel: ObservableObject {
...
+            // A newer pass has already answered — discard this one rather than
+            // overwrite it.
+            guard generation == trustLineRefreshGeneration else { return }
+            tokensNeedingTrustLine = Set(
+                states
+                    .filter { $0.value == .absent }
+                    .compactMap { identifiers[$0.key] }
+            )
+        }
+    }
```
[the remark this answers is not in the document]
gastonm5: Valid — fixed in `5dd5f0d77`. It is a hard rule here (`CLAUDE.md` rule 2: never access `@Model` classes off the MainActor), and the reads you name — `self.tokens`, `nativeCoin.address`, `contractAddress`, `uniqueId`, plus `toCoinMeta()` — all happen before the `Task` hop, on whatever context the caller was on.

Took the "mark the view model" option rather than moving those reads behind a boundary, because the narrower fix would not have satisfied the rule: `computeTokens` walks `vault.coins` and reads each coin's balance, `filteredTokens` reads tickers, `needsTrustLine` reads `uniqueId`, and `init` reads the native coin's chain and address. Isolating one method would leave the rule broken everywhere else in the same type. Whole-type isolation is also what this type already is in practice — an `ObservableObject` driving a SwiftUI screen — and it matches the ~40 other view models in the app; the now-redundant `@MainActor` on `qbtcClaimBitcoinCoin()` came off.

The only friction was the two Combine sinks: both already receive on the main queue, but a Combine closure carries no isolation of its own. They now state the scheduler's guarantee with `MainActor.assumeIsolated` rather than adding a `Task` hop, which would have let a publish land after the state it was meant to describe. No timing change, and the 100 ms debounce is untouched.

`ChainDetailScreen.init`, the `ChainDetailListView` preview and `ChainDetailViewModelTests` (already `@MainActor`) all build clean against the widened isolation.

johnnyluo:
@gastonm5 on MacOS , I am seeing this  , close button and back button both exist on the prompt dialog
CleanShot 2026-07-29 at 11 10 24@2x

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
```diff
@@ -307,7 +640,16 @@ enum RippleHelper {
                       let iouValue = iou["value"] as? String else {
                     throw amountMismatch
                 }
-                let currencyMatches = try RippleIssuedCurrency.toXrplCurrencyCode(iouCurrency)
+                let normalizedIou = try RippleIssuedCurrency.toXrplCurrencyCode(iouCurrency)
+                // The rawJson is handed to WalletCore VERBATIM, and it uppercases
```
NeOMakinG, line 822: `q:` Confirm WalletCore actually uppercases 3-byte currency codes on the rawJson path (not just the typed path) — the whole gate design hinges on it — If rawJson currency is preserved case-exact by WalletCore, the L649 Payment gate is harmlessly over-cautious and the non-Payment gap above is moot; if it's mangled, the gap is real. Worth pinning with a WalletCore-level test rather than inferring from the typed path's behavior.

NeOMakinG (changes requested):
gm — multi-lane pass: . conceptual-only (no runtime this pass). requesting changes.

## blocking
### 1. dApp non-Payment rawJson (TrustSet/OfferCreate/OfferCancel) signs an issued-currency code with ZERO validation — confirmed as pinned-intended by the PR's own test — `VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift:562`
`dappSigningInput` (L543-606) checks only `tx["Account"]` (L562) and calls the currency-validating `bindPaymentToReviewedValues` ONLY when `TransactionType == "Payment"` (L570-572). TrustSet/OfferCreate/OfferCancel — the three types `RippleDAppTransaction.supportedTransactionTypes` explicitly supports rendering (RippleDAppTransaction.swift:67-69) — are signed verbatim via `$0.rawJson = rawJson` (L600) with no currency check at all, even though TrustSet's `LimitAmount` and OfferCreate's `TakerGets`/`TakerPays` carry issued-currency codes. This is not an oversight the tests missed — it's pinned as intended: `RippleSignRippleTests.testOfferCreatePassesOnAccountCheckAlone` (L248-260) asserts `XCTAssertNoThrow` specifically for this case, with a comment confirming OfferCreate 'passes on the Account check alone.' The fixture happens to use an already-uppercase 'USD', so the test never exercises the mangling case the PR's own fixup commits were written to catch. Per the code's own documented WalletCore behavior (RippleIssuedCurrency.swift:106-112), a lowercase 3-byte code gets uppercased before encoding — so a dApp TrustSet reviewed as `LimitAmount:{currency:"usd"}` signs as `USD`, opening/reserving a different trust line than what was shown; a dApp OfferCreate places a DEX offer against a different currency pair than reviewed. Reachable via the standard dApp keysign-relay path already wired for signSolana/signTon/signBitcoin/signSui (Cosmos+ProtoMappable.swift) and rendered live in SendCryptoVerifySummaryView — this is not a dead/disabled code path.
action: Extend the currency-code check to every issued-currency-bearing field reachable via signRipple.rawJson (LimitAmount for TrustSet, TakerGets/TakerPays for OfferCreate), not just Payment.Amount — walk the parsed tx dict for any object with a `currency` key and run it through `isSignableCurrencyCode`, failing closed for every TransactionType. Add a mangled-currency test for TrustSet/OfferCreate (a negative-control companion to testOfferCreatePassesOnAccountCheckAlone, not a replacement for it).

### 2. Even the 'fixed' Payment path leaves SendMax/DeliverMin unbound and unchecked for currency mangling — `VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift:614`
`bindPaymentToReviewedValues` (L614-676) binds and currency-gates only `Destination` and `Amount`. It never inspects `SendMax` or `DeliverMin` — both legitimate Payment fields for cross-currency/partial payments, both able to carry an issued-currency object with its own currency code, and both explicitly decoded and shown to the user via `RippleDAppTransaction.amountFields` (L51-58, which includes SendMax/DeliverMin). Grep confirms SendMax/DeliverMin appear nowhere in Ripple.swift's signing code or in RippleSignRippleTests.swift. So a Payment can display 'SendMax: 100 usd' on the Verify screen and sign 'SendMax: 100 USD' — a different, potentially much larger, asset the sender is authorizing as spend — completely bypassing the isSignableCurrencyCode check this PR introduces at L649 (scoped to the Amount IOU branch only).
action: Bind and currency-gate SendMax and DeliverMin the same way Amount is gated in bindPaymentToReviewedValues. Add a test mirroring testIssuedCurrencyPaymentWhoseCodeTheSignerWouldMangleThrows but for a mangled SendMax/DeliverMin currency code.

## preferably-blocking
### 1. WalletCore SPM bump 4.3.19 → 4.7.1 (plus commondata) has no on-device signing verification, per this repo's own dependency-bump rule — `VultisigApp/project.yml`
The bump is load-bearing, not incidental: 4.7.1 is the first WalletCore release whose Ripple.proto carries `RippleOperationTrustSet`/`CurrencyAmount`, so the feature likely doesn't build against 4.3.19. This box has no Xcode/swift toolchain, so no build, unit test, or signing round-trip against the bumped WalletCore was possible this pass. The frozen-golden-byte tests (`testTrustSetSerializesToFrozenBytes`, `testTokenPaymentSerializesToFrozenBytes`) pin bytes produced by whatever WalletCore build ran them — they'd stay green even if 4.7.1 quietly changed unrelated wire encoding elsewhere in the same binary, so they don't substitute for a real signing/broadcast test.
action: Per CLAUDE.md's dependency-bump rule (guilty until proven safe; CI green ≠ safe), get a real device/simulator signing round-trip against WalletCore 4.7.1 before merge — sign and broadcast (or at minimum decode) a TrustSet and a token Payment on testnet, not just run the frozen-byte unit tests.

## should-fix
### 1. PR description materially understates scope — claims read-only/no-signing while the branch carries the full issued-currency signing surface — `PR body`
The description states "Read-only: this PR touches no signing path and makes no wire change — RippleHelper / Ripple.swift are untouched, and no commondata field is added" and "No new user-facing strings." None of that holds at HEAD: Ripple.swift now implements TrustSet + token Payment + dApp-gate signing, `RippleSpecific.transaction_type` (a new commondata field) discriminates the signing arm, and ~20 new localizable keys were added. The branch's own commit log shows a squashed feat commit (absorbing the #4757/#4758 signing-path work described elsewhere as separate follow-ups) plus ~10 fixup commits closing signing-boundary bugs found via cross-model review. Both the reviewer and QA lanes independently flagged this: a reviewer who trusts the description would skip the fund-safety audit this PR most needs.
action: Update the PR description to reflect the true scope (issued-currency signing, not just balance display) before merge, so future reviewers don't under-scope off the stale framing.

## q
- Confirm WalletCore actually uppercases 3-byte currency codes on the rawJson path (not just the typed path) — the whole gate design hinges on it
- Does TTLCache preserve a CancellationError's identity when thrown inside the cached fetch closure, and is it ever reachable through the coalescing wrapper?

## risk
Medium-high. Fund-safety signing PR with strong, consistent defense-in-depth on the read paths and the app's own typed-signing path, and correctly-fixed, genuinely-tested cancellation/stale-quote concurrency for the one call site that exists today. But two independent review passes confirm the dApp signing boundary — the exact thing this PR's fixup commits target — still has two live gaps: TrustSet/OfferCreate/OfferCancel sign an unchecked currency code (pinned as intended by the PR's own test), and Payment's SendMax/DeliverMin bypass the new gate entirely. Both are reachable through the already-shipped dApp keysign-relay path, not a hypothetical future feature. Worst realistic case is a wrong-currency trust line, DEX offer, or authorized spend amount. Add the unverified WalletCore 4.3.19→4.7.1 signing-library bump (no device test possible on this box) and this needs another round before merge.

## verdict
REQUEST CHANGES

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `n/a — codex subagent unavailable; substituted general-purpose adversarial + Claude correctness lanes (GLM third-eye errored on model access)` conceptual pass.

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
gastonm5: Measured rather than inferred — and the premise holds on the rawJson path.

`RippleCurrencyCodeCaseTests` signs through wallet-core and compares the **entire encoded transaction** for two inputs differing only in the case of the currency code:

- `usd` → encodes byte-identically to `USD`
- `UsD` → encodes byte-identically to `USD`
- `USD` vs `EUR` → differ, which is the control proving the comparison is sensitive to the currency at all

Byte equality rather than searching the blob for `"USD"`: three ASCII bytes can occur coincidentally in a signature, so a substring hit would have been suggestive rather than conclusive — and this measurement is what the gate design rests on. Two different inputs producing identical bytes has no explanation other than the code being uppercased before encoding.

The typed path behaves the same way, but that is now a separate assertion rather than the basis for the claim.

So the gate is load-bearing, not over-cautious — and the two blocking findings in your review are **live rather than theoretical**. That is what motivated the fix in `78f04276b`.

gastonm5:
@NeOMakinG thanks — this was a genuinely useful pass. Both blocking findings were real, and measuring your `q` turned them from "possible" into "confirmed". Fixed in `78f04276b`.

## q1 — measured, premise holds

`RippleCurrencyCodeCaseTests` signs through wallet-core on the **rawJson** path and compares the entire encoded transaction for inputs differing only in currency case. `usd` and `UsD` each encode byte-identically to `USD`; a `USD` vs `EUR` control confirms the encoding is sensitive to the currency at all. Byte equality rather than an ASCII search, since three bytes can appear coincidentally in a signature and this measurement is what the gate rests on.

So it is not inferred from the typed path — **rawJson mangles too, and both blocking findings were live.** Detail in the thread reply.

## Blocking 1 — fixed

`assertEveryCurrencyCodeIsSignable` now walks the whole decoded transaction for any object carrying a `currency` key, for **every** `TransactionType`, before any type-specific binding. Structural rather than a field allowlist, so a term this signer does not know about is covered by default — there is a test for a mangled code nested inside an unknown field.

`testOfferCreatePassesOnAccountCheckAlone` is kept and still passes, exactly as you suggested: an already-uppercase code is not false-rejected. Its `USD` fixture is precisely why it never caught this, so the new cases are its negative controls — TrustSet `LimitAmount`, OfferCreate `TakerPays`, lowercase and mixed case.

## Blocking 2 — currency-gated, but *not* value-bound

Worth being precise about, because it is half of what you asked for.

`SendMax` / `DeliverMin` are now covered by the same walk, so a mangled code in either is refused. I did **not** bind them to reviewed values: `KeysignPayload` carries a single `toAmount`, and `SendMax` is legitimately a *different* currency in a cross-currency payment, so there is nothing to bind against.

What this guarantees is **the code you saw is the code that signs** — not *the amount you saw is the amount that signs*. Closing that second half needs a payload change carrying the extra reviewed values. Flagging rather than letting it read as fully closed.

## WalletCore bump — partially addressed

Added `RippleWalletCoreWireContractTests`, which signs through 4.7.1 and decodes the output against the XRPL layout (field padding, TrustSet vs Payment divergence, native still encoding as drops).

Your point about the frozen-golden-byte tests is right and worth restating: their fixtures were generated by 4.7.1 too, so they cannot detect 4.7.1 encoding something differently — they only detect our own drift.

**A testnet broadcast has still not been done.** A decode round-trip is not a substitute and I agree it should happen before merge.

## q2 — yes on both counts

Identity is preserved, and it is reachable through the coalescing wrapper — but for a non-local reason worth knowing: `HTTPClient` converts `URLError.cancelled` into a real `CancellationError()` before any cache sees it, so `if error is CancellationError` in `TTLCache` matches. Joiners await the same `Task`, so they receive the same error instance; the exemption cannot be bypassed by arriving second.

`TTLCacheCancellationTests` pins the cache side — cancellation never answered from the last-good entry, an ordinary failure still is — with a gate actor asserting the joiner really joins rather than starting a second fetch. The cross-file dependency is called out in the file header, since nothing in `TTLCache` would notice if that conversion in `HTTPClient` changed.

## Scope

PR description corrected. You were right that it materially understated things — it claimed read-only/no-signing/no-commondata-field/no-new-strings, none of which holds at HEAD.

---

Full suite green: **4186 tests, 1 skipped, 0 failures.** Leaving all threads for you and @johnnyluo to resolve.

Outstanding and not claimed as done: the `SendMax`/`DeliverMin` value-binding above, and the testnet broadcast.

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
```diff
@@ -272,6 +617,35 @@ enum RippleHelper {
...
+    /// term — is covered by default instead of silently exempt. Fail-closed on
+    /// both branches: a code that cannot even be normalised is refused too.
+    private static func assertEveryCurrencyCodeIsSignable(in value: Any) throws {
+        let mangled = HelperError.runtimeError(
+            "signRipple rawJson currency code would be altered by the signer"
+        )
+
+        if let object = value as? [String: Any] {
+            if let currency = object["currency"] as? String {
+                guard let normalized = try? RippleIssuedCurrency.toXrplCurrencyCode(currency),
```
NeOMakinG: `preferably-blocking:` Whitespace-trim asymmetry re-opens the currency-mangling class the walker was just added to close — assertEveryCurrencyCodeIsSignable validates the output of RippleIssuedCurrency.toXrplCurrencyCode(currency), which trims .whitespacesAndNewlines before classifying the code (RippleIssuedCurrency.swift:85), but WalletCore signs the raw, untrimmed rawJson string verbatim. A 3-raw-byte code with edge whitespace, e.g. "Ab ", trims to 2 chars and is misrouted to the 40-hex-code branch (passes unchanged), while WalletCore treats the raw 3-byte string as an ISO code and uppercases it (b->B) before encoding. This lands exactly on the unbound types (OfferCreate TakerGets/TakerPays, TrustSet LimitAmount) where the walker is the sole gate — the Payment IOU branch re-trims and re-validates in bindPaymentToReviewedValues so it's not exposed there. A trailing space is invisible on the Verify screen, so the co-signer has no way to notice the input that trips this.

fix: Validate the raw, untrimmed currency string (or explicitly reject any code containing non-canonical whitespace) instead of trimming before classification, so the walker's decision matches what WalletCore actually signs.

VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleSignRippleTests.swift
```diff
@@ -258,6 +258,109 @@ final class RippleSignRippleTests: XCTestCase {
...
+    private static func paymentJson(field: String, currency: String) -> String {
+        """
+        {"TransactionType":"Payment","Account":"\(account)","Destination":"\(destination)","Amount":"1000000","\(field)":{"currency":"\(currency)","issuer":"rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh","value":"100"},"Fee":"10","Sequence":99,"LastLedgerSequence":12345678}
+        """
+    }
+
+    /// `SendMax` is a spend authorisation shown on the Verify screen. Displaying
+    /// `100 usd` and signing `100 USD` authorises a different asset entirely, and
+    /// it sits outside the `Amount` branch the original gate covered.
+    func testPaymentWithAMangledSendMaxCurrencyIsRefused() {
```
NeOMakinG, line 448: `should-fix:` Dedicated SendMax/DeliverMin mangled-currency tests are inert — testPaymentWithAMangledSendMaxCurrencyIsRefused and testPaymentWithAMangledDeliverMinCurrencyIsRefused (L341-353) call assertRefused with toAddress:"" while the JSON fixture sets a real Destination, so bindPaymentToReviewedValues' Destination-equality guard throws before the currency walker is ever reached — these two tests would stay green even if the walker's SendMax/DeliverMin coverage were deleted. The underlying code fix is real and is proven by a separate load-bearing nested-unknown-field test, so this is a test-quality gap, not a functional one.

fix: Fix the fixtures so toAddress matches the JSON's Destination (and Amount matches where relevant) so these two tests actually isolate and pin the walker's SendMax/DeliverMin coverage.

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
```diff
@@ -272,6 +617,35 @@ enum RippleHelper {
...
+    /// Walks the whole decoded object rather than a fixed field list, so a field
+    /// this signer does not know about — a future transaction type, or a nested
+    /// term — is covered by default instead of silently exempt. Fail-closed on
+    /// both branches: a code that cannot even be normalised is refused too.
+    private static func assertEveryCurrencyCodeIsSignable(in value: Any) throws {
+        let mangled = HelperError.runtimeError(
+            "signRipple rawJson currency code would be altered by the signer"
+        )
+
+        if let object = value as? [String: Any] {
```
NeOMakinG, line 771: `q:` Duplicate "currency" keys at the same JSON level rely on unpinned parser agreement — JSONSerialization and WalletCore's own JSON parser both currently keep the last-seen value on a duplicate key, so the walker and the signer stay in agreement today — but that agreement is unpinned by any test. Silent divergence between the two parsers would reopen the exact mismatch class this PR fixes.

fix: Add a comment or a test pinning that both parsers keep the same (last) value on a duplicate key, so a future parser change surfaces here instead of silently.

NeOMakinG (dismissed):
gm — multi-lane pass: . conceptual-only (no runtime this pass). requesting changes.

## prior issues — status
| round | issue | status |
|---|---|---|
| 1 | null | fixed |
| 1 | null | partial |
| 1 | null | still_blocking |
| 1 | null | fixed |

## preferably-blocking
### 1. Whitespace-trim asymmetry re-opens the currency-mangling class the walker was just added to close — `VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift:634`
assertEveryCurrencyCodeIsSignable validates the output of RippleIssuedCurrency.toXrplCurrencyCode(currency), which trims .whitespacesAndNewlines before classifying the code (RippleIssuedCurrency.swift:85), but WalletCore signs the raw, untrimmed rawJson string verbatim. A 3-raw-byte code with edge whitespace, e.g. "Ab ", trims to 2 chars and is misrouted to the 40-hex-code branch (passes unchanged), while WalletCore treats the raw 3-byte string as an ISO code and uppercases it (b->B) before encoding. This lands exactly on the unbound types (OfferCreate TakerGets/TakerPays, TrustSet LimitAmount) where the walker is the sole gate — the Payment IOU branch re-trims and re-validates in bindPaymentToReviewedValues so it's not exposed there. A trailing space is invisible on the Verify screen, so the co-signer has no way to notice the input that trips this.
action: Validate the raw, untrimmed currency string (or explicitly reject any code containing non-canonical whitespace) instead of trimming before classification, so the walker's decision matches what WalletCore actually signs.

### 2. WalletCore 4.3.19 -> 4.7.1 still has no real on-device/testnet signing round-trip — `VultisigApp/project.yml`
Carried over from round 1. RippleWalletCoreWireContractTests.swift is a genuine wire-contract decode test against the real linked 4.7.1 binary (not tautological — verified padding/offset/native-drops facts external to this repo's own logic), but it is still a decode round-trip, not a testnet broadcast. The author's own PR comment concedes the testnet broadcast has not been done and is not claimed as done.
action: Get a real testnet submit-and-validate round for a TrustSet and an issued-currency Payment signed with WalletCore 4.7.1 before merge, or have a human with device/testnet access explicitly accept the residual risk given this is a fund-safety signing-library bump.

## should-fix
### 1. Dedicated SendMax/DeliverMin mangled-currency tests are inert — `VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleSignRippleTests.swift:341`
testPaymentWithAMangledSendMaxCurrencyIsRefused and testPaymentWithAMangledDeliverMinCurrencyIsRefused (L341-353) call assertRefused with toAddress:"" while the JSON fixture sets a real Destination, so bindPaymentToReviewedValues' Destination-equality guard throws before the currency walker is ever reached — these two tests would stay green even if the walker's SendMax/DeliverMin coverage were deleted. The underlying code fix is real and is proven by a separate load-bearing nested-unknown-field test, so this is a test-quality gap, not a functional one.
action: Fix the fixtures so toAddress matches the JSON's Destination (and Amount matches where relevant) so these two tests actually isolate and pin the walker's SendMax/DeliverMin coverage.

## q
- Duplicate "currency" keys at the same JSON level rely on unpinned parser agreement

## risk
Medium-high, unchanged in class from round 1 but narrower. This is a fund-safety Ripple/XRP signing PR. Both round-1 blockers are now closed by a well-designed generic walker with load-bearing tests against the real WalletCore 4.7.1 binary. What remains: a fail-open whitespace edge case in the same walker that was just added to close this exact bug class (small blast radius — needs an invisible trailing space in a dApp-supplied currency code, but the mechanism is proven, not theoretical), and a fund-safety dependency bump (WalletCore 4.3.19->4.7.1) still without any real testnet signing evidence beyond the new decode-only wire-contract test. Neither is a full regression to round-1's severity, but both are real, unresolved gaps on a signing path handling user funds.

## verdict
REQUEST CHANGES

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `n/a — codex subagent unavailable this pass; substituted general-purpose adversarial + Claude correctness lanes (GLM third-eye errored on model access)` conceptual pass.

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
gastonm5: Valid, and confirmed by measurement rather than by reading — fixed in `fb4a2e32b`.

**What I measured.** I signed each candidate through wallet-core 4.7.1 on this exact rawJson path and read the encoded currency field back out (a 3-byte standard code sits at bytes 12–14 of the 20-byte currency field; the hex form starts at offset 0):

| rawJson `currency` | wallet-core encodes | `toXrplCurrencyCode` produces |
|---|---|---|
| `"Ab "` | `41 42 20` → **`AB `** (read as ISO, uppercased) | `4162…` (hex-packed `Ab`) |
| `"AB "` | `41 42 20` — **byte-identical** to `"Ab "` | `4142…` |
| `" AB"` | `20 41 42` → `" AB"` verbatim | `4142…` |
| `"€"` (1 char, 3 bytes) | `E2 82 AC` at offset 12 (ISO) | `E282AC…` at offset 0 |
| `"RLUSD"` | refused outright: `Some input field cannot be parsed` | `524C555344…` |
| 40-hex upper vs lower | **identical bytes** | uppercased |

So your reading is exactly right, and the root cause is slightly wider than whitespace: **wallet-core classifies on RAW BYTE length, the normaliser classifies on trimmed character count and then ASCII-packs.** `"€"` splits the two with no whitespace involved at all.

**The fix.** `RippleIssuedCurrency.isVerbatimSignableCurrencyCode(_:)` classifies the raw string the way the ledger does — 3 bytes of standard code, or 40 hex digits, and nothing else — and `assertEveryCurrencyCodeIsSignable` now uses it instead of normalising first. Three consequences worth stating:

- **Refusing the unpackable lengths loses nothing.** `"RLUSD"` was never signable; the old gate ASCII-packed it into a code wallet-core never received, and wallet-core then refused the input anyway. Refusing it here just moves the same failure earlier, with a message that says why.
- **40-character hex stays accepted in either case**, because wallet-core decodes hex case-insensitively (last row). I did not want the tightening to remove a legitimate spelling.
- **`AB ` and ` AB` are refused even though wallet-core encodes them unchanged.** A space is not in rippled's `kIsoCharSet`, so the ledger renders such a line's currency back as hex and the code could never match its own `account_lines` row — the same repertoire argument as `09d951d33`.

`toXrplCurrencyCode` is untouched: trimming is right for a user-entered ticker, and only the signing-path gate needed the raw form. `bindPaymentToReviewedValues` adopts the same raw judgement — redundant while the walk runs first, kept so the binding does not depend on that ordering.

One correction to the thread for the record: the walk runs **before** `bindPaymentToReviewedValues` (`Ripple.swift:576` vs `:584`), not after. So the Payment IOU branch was not protected by re-validation — it was covered by the walk, which had the same hole. Your conclusion about the unbound types holds either way, and it is the unbound types that were exposed.

**Pinned, and verified load-bearing.** Five new tests in `RippleSignRippleTests` on the unbound types — TrustSet `LimitAmount` with `Ab `, `€` and `RLUSD`; OfferCreate `TakerPays` with `Ab ` and ` AB` — plus a positive control that the 40-hex form still signs in both cases. Against the pre-fix walk **exactly those five fail** (`Executed 43 tests, with 5 failures`) and they pass after; the ablation was a temporary revert of the two call sites, reverted before committing. The six measurements above are pinned in `RippleCurrencyCodeCaseTests` as byte-equality assertions over whole encoded transactions, not substring searches.

**What I did not verify:** nothing here was broadcast to a ledger. The claim is about what wallet-core encodes, which is measured, not about how rippled would then respond to a `AB ` line — that part rests on `kIsoCharSet`, read from rippled's source.

Full suite green on this branch: **4212 tests, 1 skipped, 0 failures.**

VultisigApp/VultisigAppTests/Blockchain/Ripple/RippleSignRippleTests.swift
gastonm5: Valid — fixed in `561a21c78`.

Confirmed the tests were inert, though by a slightly different mechanism than described: the currency walk actually runs **first** (`Ripple.swift:576`), *before* `bindPaymentToReviewedValues` (`:584`), so the walk did fire. The finding is nonetheless exactly right — with `toAddress: ""` against a fixture that sets a real `Destination`, the Destination guard would have rejected the payload regardless, so **deleting the walk's SendMax/DeliverMin coverage would not have failed either test.** A refusal proved only that *something* objected.

Both fixtures now bind cleanly — reviewed `toAddress` equal to the JSON's `Destination`, reviewed `toAmount` equal to the drops `Amount` (`BigInt(1_000_000)` against `"1000000"` on the native 6-decimal coin, which the native branch compares as `toAmount.description` with no decimal conversion) — so the currency walk is the only thing left that can reject. Two **controls** were added asserting the same fixtures with an already-uppercase code actually SIGN, because without them a refusal still would not identify its own cause. Plus a whitespace-edged `SendMax` case, from the sibling thread.

**Proved load-bearing, as asked.** I ablated the walk by pruning `SendMax` / `DeliverMin` out of the object handed to `assertEveryCurrencyCodeIsSignable` and re-ran:

```
testPaymentWithAMangledDeliverMinCurrencyIsRefused        failed: XCTAssertThrowsError failed: did not throw an error
testPaymentWithAMangledSendMaxCurrencyIsRefused           failed: XCTAssertThrowsError failed: did not throw an error
testPaymentWithAWhitespaceEdgedSendMaxCurrencyIsRefused   failed: XCTAssertThrowsError failed: did not throw an error

Executed 46 tests, with 3 failures (0 unexpected)
```

Exactly those three and nothing else — including the two new uppercase controls, which stayed green, so the ablation removed only the coverage it was meant to. The ablation was reverted; the committed walk is unchanged.

Full suite green on this branch: **4212 tests, 1 skipped, 0 failures.**

VultisigApp/VultisigApp/Blockchain/Ripple/Signing/Ripple.swift
gastonm5: Good question — and the premise turns out to be false. **The two parsers do not agree**, so this is a live fail-open rather than an unpinned assumption. Fixed in `953429dba`.

**What I measured**, on the two things that can actually settle it — the decoded object for one parser, the encoded transaction bytes for the other:

```
{"currency":"USD", …, "currency":"EUR"}

JSONSerialization  →  "USD"    keeps the FIRST occurrence
wallet-core        →  EUR      encodes byte-identically to a single "currency":"EUR"
```

Every check in `dappSigningInput` reads the object `JSONSerialization` produced; wallet-core then signs the same text through its own parser. So a dApp that writes a key twice gets its **first** value reviewed and its **second** value signed.

I measured the sharp end directly rather than reasoning about it. A Payment carrying two `Destination` fields encodes **byte-identically to one naming only the second address**, while `bindPaymentToReviewedValues` checks the first: bind against the address the co-signer approved, pay a different one. `currency`, `issuer` and `value` substitute the same way.

Neither parser is wrong — duplicate keys are legal JSON and RFC 8259 leaves the winner unspecified. There is simply no agreement to pin.

**So the fix is a refusal, not a comment.** The collapsed object cannot say what wallet-core will keep, so `assertNoDuplicateJsonKeys` scans the rawJson **text** and refuses any key repeated inside one object, before any other check reads the object. Nothing legitimate is lost: a duplicate key expresses nothing a single-keyed object cannot, and a transaction that means two different things to its two readers is not one a co-signer can review.

Scanner details that matter:
- keys are compared **decoded**, so `Destination` is `Destination` — both parsers compare them that way;
- whole string literals are consumed, so a brace or colon inside a memo blob is text, not structure;
- each open object gets its own key set, so `currency` in both halves of an offer is fine;
- the set is mutated in place through its subscript — rebuilding it per key would be quadratic in a field count the dApp controls (a cross-model review caught that; the first draft used `Set.union`).

**Scope, stated plainly:** this hazard predates this branch. It lives in the dApp rawJson co-signer path from the `signRipple` work, not in the trust-line changes. I closed it here because this thread surfaced it and it is a signing-path fail-open; happy to move it to its own PR if you'd rather keep this one to balances.

**Pinned** by `RippleDuplicateJsonKeyTests` (11 tests): the two parser behaviours as measurements, the `Destination` byte-substitution, four refusals (top-level, nested in an `Amount`, inside an array element, escaped spelling), and four controls that the scan does not false-reject (the same payload without the duplicate, the same key in sibling objects, a JSON-looking string value, an object with 500 distinct keys). Verified load-bearing: with the guard removed, **exactly the four refusal tests fail** and the seven measurement/control tests stay green.

**What I did not verify:** the last-wins behaviour is measured against wallet-core 4.7.1 only, and the `JSONSerialization` behaviour against this OS's Foundation. That is precisely why the refusal does not depend on either — it refuses the input whichever way the two resolve it, and the two measurement tests will fail loudly if either side ever changes.

Full suite green on this branch: **4212 tests, 1 skipped, 0 failures.**

gastonm5:
@johnnyluo you're right, and thanks for the screenshot — the pixel positions in it are what pinned the cause down. Fixed in `d8177a112`.

**What was happening.** `Screen` already renders a `crossPlatformToolbar` of its own on macOS, configured through the screen environment:

```swift
// Screen.swift:69-78
#if os(macOS)
    VStack {
        contentContainer
            .crossPlatformToolbar(
                title,
                ignoresTopEdge: ignoresTopEdge,
                showsBackButton: !backButtonHidden,   // defaults to true
                items: toolbarItems
            )
    }
#endif
```

The sheet layered a **second** `.crossPlatformToolbar` on top of that `Screen`. So `showsBackButton: false` was applied to the outer layer — the one carrying the close button — while `Screen`'s own inner toolbar never saw it and kept its default chevron. Two toolbars, not one toolbar configured twice.

Your screenshot confirms it rather than just being consistent with it: `MacOSToolbarView.toolbarContent` is `.frame(height: 60)`, so a second stacked row predicts the two buttons sitting exactly 60pt apart. In the capture (2x) the X centres at y≈56px and the chevron at y≈176px — 60pt apart. The chevron is `Screen`'s, dismissing via `@Environment(\.dismiss)`; the X is the sheet's, clearing the presentation binding.

**The fix** routes both settings through `Screen`'s own toolbar instead of stacking a second one:

```swift
.screenBackButtonHidden()
.screenToolbar {
    #if os(macOS)
        CustomToolbarItem(placement: .leading) {
            ToolbarButton(image: .xmark) { isPresented = false }
        }
    #endif
}
```

That's the pattern the rest of the app already uses (`ImportVaultSelectionSheet`, `SwapVerifyScreen`, `AddressBookScreen` and 15 others). The close button stays macOS-only, so iOS keeps swipe-to-dismiss plus the drag indicator, and it also drops a redundant empty `IOSToolbarView` layer there.

**Scope check.** I swept for the same shape elsewhere and this sheet was the only instance — `CoinDetailScreen` and `DefiChainMainScreen` attach `.crossPlatformToolbar` to a plain `ScrollView`, not to a `Screen`, so they're fine. Worth knowing that `Screen {} .crossPlatformToolbar {}` silently produces two toolbars on macOS rather than overriding one; it reads like configuration and behaves like stacking.

**Verification.** macOS `BUILD SUCCEEDED`; full iOS suite 4212 tests / 0 failures. To be straight about the limit: I verified this structurally (one toolbar layer instead of two, plus the 60pt geometry above) rather than by reproducing the sheet on a macOS run — reaching it needs a vault holding an unactivated XRPL trust-line token, which I don't have set up. If you still see two affordances on your build, that would mean the cause is something other than the stacked toolbars and I'd want to know.

Your other two — the nested Activate button and the stale commondata pin comment — were fixed earlier in `f3d5565bf` + `2e0ebe942` and `1e21f4845`.

VultisigApp/VultisigApp.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -32,7 +32,7 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/vultisig/commondata",
       "state" : {
-        "revision" : "c91f2ea7db5eb03cf6bb637da19e4ed93b07dc46"
+        "revision" : "a16c710f0b0dfb6e2c8debde54b72414a04feb9a"
```
johnnyluo: Please pin the latest commit on main branch , it should be `63a65c37c50f6a944f6791d36eb905810af82760`

johnnyluo (approved):
LGTM

PR #4945, opened 2026-07-27
[Logging] Base: Log facade + Feature×Layer taxonomy + scheme/xcconfig enable-disable + gate tests

gastonm5:
## Description

Base / PR 0 of the logging-facade stack (epic #4943). Adds a `Log` facade so OSLog
can be enabled/disabled **per feature and per layer** via scheme/build config —
**without** losing OSLog's per-argument `privacy:` redaction, Console category
filtering, or log levels. **ADDITIVE: no existing `Logger` call site changed.**

Fixes #4942

### What the base adds
- **`Core/Logging/`** — `LogFeature`, `LogLayer`, `LogConfig` (gate), `Log` facade.
- **A1 facade**: `Log.<feature>.<layer>` returns a *configured* `os.Logger` — the
  real `Logger(subsystem: "com.vultisig.app", category: "<feature>.<layer>")` when the
  category is enabled, `Logger(.disabled)` when off. Call sites stay 100% native
  OSLog, so `\(…, privacy:)`, levels, Console/`log stream` filtering and signposts
  are all preserved (no message wrapping — that would flatten the 246 `privacy:`
  annotations). Usage at migration time is a one-line declaration swap:
  `private let logger = Log.swap.service`.
- **Composed category** `"<feature>.<layer>"` → Console filters either axis
  (`category BEGINSWITH "swap"` for one feature, `category ENDSWITH ".network"` for
  one layer).

### How the gate / toggle works
- **Compile-time master switch `LOGGING`** in `SWIFT_ACTIVE_COMPILATION_CONDITIONS`:
  present in Debug, absent in Release → `Log.*` compiles down to `Logger(.disabled)`
  (logging stripped in Release).
- **Debug-only runtime override** via a `VULTI_LOG` string, resolved (last wins):
  Info.plist `VultiLogConfig` default (`$(VULTI_LOG_CONFIG)` per config) → scheme env
  var `VULTI_LOG` → `UserDefaults["VULTI_LOG"]`.
- **`VULTI_LOG` grammar**: comma-separated `selector:level`; selector ∈ `*` |
  `feature` | `.layer` | `feature.layer`; level ∈ `debug|info|notice|warning|error|off`.
  Resolution is **most-specific-wins** (exact > feature > layer > wildcard), ties
  broken by last-declared.

  Example — `VULTI_LOG="swap:debug,.network:off,*:info"`:
  swap (all layers) → debug; the network layer of every feature → off; everything
  else → info.

[...]

PR #4914, opened 2026-07-24
refactor(chain): table-driven ChainConfig replacing parallel per-chain switches

gastonm5:
Closes #4911

## Summary

Code-quality Tier-1 (S1): `Model/Chain.swift` carried ~9 parallel 42-case `switch self` statements that each returned static per-chain **data** (`name`, `ticker`, `feeUnit`, `swapAsset`, `logo`, `chainID`, `coinType`, `chainType`), plus two more in `Model/Chain+extensions.swift` (`banxaBlockchainCode`, `minimumSendAmount`). Adding a chain meant editing ~11 switches; a value could silently disagree between them.

This replaces all of them with a single **table**:

- `struct ChainConfig` — one row per chain holding every migrated field.
- `private static let configs: [Chain: ChainConfig]` — the lookup table, seeded once from an **exhaustive `makeConfig()` switch** (`Chain.allCases.map`), so a **new `Chain` case fails to compile** until its row is defined.
- Each accessor becomes a keypath read (`var name: String { config.name }`). **No public signature changed**, so every call site is untouched.

New file: `VultisigApp/VultisigApp/Model/ChainConfig.swift` (added to the project via `make generate`; `project.pbxproj` is XcodeGen-generated, not hand-edited).

### `type` vs `chainType` unified
`Chain` had two identical 42-case switches — `chainType` and `type` — returning the same `ChainType` for every chain. Resolved to a **single source of truth**: `chainType` reads the table; `type` is now a thin alias (`var type: ChainType { chainType }`). The 4 existing `.type` call sites keep compiling unchanged.

### Also
- Dropped the now-unused `import WalletCore` from `Chain.swift` (the only `CoinType` user moved out). The pre-existing unused `import SwiftData` is left untouched (out of scope).

## coinType parity verified ✅
`coinType` feeds address/signing derivation, so it was checked chain-by-chain against the removed switch. The non-obvious mappings are reproduced exactly:
- `terra → .terraV2`, `terraClassic → .terra` (reversed vs. the names)
- `mayaChain → .thorchain`, `avalanche → .avalancheCChain`, `bscChain → .smartChain`
- `hyperliquid → .ethereum`, `sei → .ethereum`, `qbtc → .cosmos`, `bittensor → .polkadot`

## Tests
[...]

PR #500, opened 2024-05-22
feat: add shapeshift fox token

firebomb1:
Adds support for the ShapeShift FOX token on the following chains: Ethereum, Optimism, Polygon and Arbitrum.

Addresses can be verified on Coingecko, execept for Arbitrum which is recent, but it is the automated bridged token contract.

Creating this PR on behalf of @hpayne1, the current Marketing Workstream Leader of ShapeShift DAO.

PR #4954, opened 2026-07-27
[Add] Dynamic token search — TokenCatalogRepository + verified/unverified add-token surface [Wallet / Tokens]

### LINKED ISSUE #4941: [Add] Dynamic token search — unify token sourcing behind a TokenCatalogRepository (reuse swap datasources) [Wallet / Tokens]
# [Add] Dynamic token search — unify token sourcing behind a `TokenCatalogRepository` (reuse swap datasources) [Wallet / Tokens]

## Problem
Token search / add-token / pickers are anchored to the static, hand-maintained `TokensStore` (`Core/Stores/TokensStore.swift:5`) as both the catalog's floor and ceiling: a token can't be searched or held until a release ships a new entry, breadth is a tiny fraction of what exists per chain, and every ticker/logo/`priceProviderId` is hand-verified per platform and rots.

Crucially this is **not greenfield** — the app already fetches, caches, merges and trust-filters dynamic token lists in two disconnected places:
- `TokenSearchService` + `SwapTokenListCache` (1inch on EVM, Jupiter on Solana), and
- the swap-destination `DestinationTokenProvider` / `DestinationTokenRegistry` (SwapKit, THOR/Maya pools, secured assets) — an actual pluggable multi-source registry whose own doc-comment invites more providers.

They don't share a contract, and neither feeds the wallet token search or auto-discovery. Full analysis + file:line evidence: wiki `pages/projects/vultisig/dynamic-token-datasource/analysis.md`.

## Goal
Generalize the swap-only `DestinationTokenProvider`/`DestinationTokenRegistry` into an app-wide `TokenCatalogProvider` / `TokenCatalogRepository`, make **both** the static `TokensStore` **and** every dynamic source conform, and route token search + pickers through it — so search returns dynamic + curated tokens (deduped by `CoinMeta.uniqueId`), with the bundled curated list as the always-available offline floor, and **verification-to-surface** so scam/lookalike tokens don't auto-appear.

## Scope — Must Do (phase 1; one reviewable commit per step, codex-review loop)
1. **Trust field.** Add `TokenVerification` (`.curated` / `.verified(source)` / `.unverified`) threaded via a `CatalogToken { meta: CoinMeta, verification, sourceKind }` wrapper. Do NOT widen `CoinMeta`'s `uniqueId`/`==` identity — unit-test it's unaffected.
2. **Promote the contract.** `DestinationTokenProvider` → `TokenCatalogProvider`, `DestinationTokenRegistry` → `TokenCatalogRepository`, moved to `Core/Services/TokenCatalog/`; add `precedence`; port the bucket dedup into the repository. Keep thin typealias/shims so the swap picker compiles. Reuse `DestinationTokenRegistryTests`.
3. **`BundledTokensProvider`.** Wrap `TokensStore.TokenSelectionAssets` — highest `precedence`, `.curated`, no network, honoring the existing `sepolia`/`thorchainChainnet`/QBTC gates → the offline floor. This makes everything else additive.
4. **Network providers from the swap sources.** `OneInchCatalogProvider` (**FIX** `OneInchToken.toCoinMeta`, which currently drops `providers`→verification and `coingeckoId`→`priceProviderId`), `JupiterCatalogProvider`, and register the existing SwapKit / native-pool / secured-asset conformers under the new protocol. Each keeps its TTL cache; add a Codable per-chain **disk snapshot** for offline cold-start.
5. **Route search through the repository.** Replace `TokenSelectionLogic`/`TokenSearchService`'s hardcoded preset+external merge with `TokenCatalogRepository.catalog(for:)`; the swap source/destination pickers switch to the same call. Preserve the merged-list-not-cached rule and fail-open-to-last-good semantics.
6. **Verification-to-surface.** Only `.curated`/`.verified` candidates auto-appear in search/discovery; `.unverified` are reachable but badged + explicit opt-in. Keep `CoinService.isSpamToken` as a second gate.

**Merge rule:** dedup by `CoinMeta.uniqueId` (existing app-wide key); the highest-`precedence` provider wins the `CoinMeta` (curated logo/`priceProviderId`), but the strongest `verification` seen is kept.

## Must NOT Do
- Don't delete `TokensStore` — it becomes a provider; leave the ~30 direct named-static readers (`cacao`, `sruji`, VULT…) alone.
- Don't change the SwiftData enable/disable model — candidate set = repository output; enabled set stays `vault.coins`; tombstone stays `HiddenToken`. A dynamic token is a *candidate* until the user adds it.
- Don't redo the in-flight #4935 (TokenDiscoverer) / #4936 (CustomToken) refactors — build on their seams; branch off `origin/main`.

## Security — top risk: scam / impersonation tokens
Dynamic sourcing admits lookalike-USDC, homoglyph tickers, airdrop dust, and fee-on-transfer traps (especially SwapKit long-tail). Defense in depth: (a) verification-to-surface (above); (b) keep the `EvmCoinFinder` CoinGecko `providers` allowlist **and** fix the `toCoinMeta` signal-drop; (c) keep `CoinService.isSpamToken` heuristics; (d) candidate ≠ enabled — nothing enters `vault.coins` without explicit user action.

## Out of scope / follow-up
- Migrating the ~30 direct `TokensStore` named-static readers.
- Optional phase-2 Uniswap-Token-Lists / Trust-Wallet allow/deny provider.
- Android/Windows/SDK implementation (parity notes only; align the contract shape with the SDK resolver/`isHidden` model rather than diverging).

## Acceptance Criteria
- SwiftLint + `make test` green (bar the known-flaky/snapshot CI cases).
- Token search / add-token / swap pickers surface dynamic (1inch/Jupiter/SwapKit) + curated tokens, deduped, curated winning logo/price.
- Airplane-mode cold launch still shows the bundled catalog (offline floor); online overlays the dynamic sources.
- `.unverified` tokens don't auto-surface — opt-in + badged; `isSpamToken` still applies.
- Enable/disable round-trips: a dynamically-sourced token added persists to `vault.coins`; removing it writes a `HiddenToken` so re-discovery doesn't re-add it.
- Unit tests: repository merge/precedence/dedup (curated wins logo, verified beats unverified); provider fail-open + disk snapshot offline; verification gating; `OneInchToken.toCoinMeta` now carries verification + `priceProviderId`.

### LINKED ISSUE #4955: [Add] Dynamic token search: surface unverified tokens behind an opt-in toggle + warning badge [Wallet / Tokens]
# [Add] Dynamic token search: surface unverified tokens behind an opt-in toggle + warning badge [Wallet / Tokens]

## Context
Phase 1 (#4941, PR #4954) added the `TokenCatalogRepository` + `TokenVerification` model and **withholds `.unverified` tokens** from search/auto-surface (`TokenSearchService.surfaceableTokens`). That's the safe default — but it also means the new dynamic breadth is currently unreachable in the UI. This issue completes verification-to-surface: let users **opt in** to see unverified results, clearly **badged** as a scam/impersonation risk. Builds on #4954 — branch off `feat/dynamic-token-catalog-4941` and open this PR with `base: feat/dynamic-token-catalog-4941` (stacked; retarget to `main` after #4954 merges).

Analysis: wiki `pages/projects/vultisig/dynamic-token-datasource/analysis.md` (risks §, MetaMask/Rabby posture).

## Goal
Thread token **verification** through the wallet token-search / add-token picker pipeline (today it passes bare `[CoinMeta]`, which drops the `CatalogToken` verification), so the search UI can:
- show a **warning badge** on `.unverified` (and visually distinguish `.verified(source)` vs `.curated`), and
- expose an **opt-in "Show unverified results" toggle** (default OFF) that reveals the withheld candidates.
Curated/verified tokens keep auto-surfacing exactly as today.

## Scope — Must Do
1. Thread verification to the picker: surface `CatalogToken` (or a parallel `uniqueId → TokenVerification` map) through `TokenSelectionViewModel`/`TokenSelectionLogic` to the row views — without regressing the swap picker (keep the `.shared` vs `.appCatalog` split from #4954, and the merged-list-not-cached rule).
2. **Warning badge** on unverified rows (design-system-consistent — use the existing `Screen`/design tokens; a small "Unverified" pill + risk affordance). Distinguish curated/verified visually if cheap.
3. **Opt-in toggle** in the search UI, default OFF; when ON, `surfaceableTokens` also includes `.unverified`. Decide persistence (per-session vs `UserDefaults`) — prefer per-session so it re-arms safely.
4. Consider an explicit **confirmation** when *adding* an unverified token (a short "this token isn't verified — make sure the contract is correct" step), reusing existing confirm patterns.
5. 8-locale strings (de,en,es,hr,it,ko,pt,zh-Hans) for all new labels (badge, toggle, confirm). Keep `.strings` sorted.

## Must NOT Do
- Don't auto-surface or auto-enable `.unverified` without the toggle. Keep `CoinService.isSpamToken` as a hard second gate (spam is never shown, toggle or not).
- Don't change the enable/disable model (`vault.coins`/`HiddenToken`) — this is display/surfacing only.
- Don't regress the swap source/destination pickers.

## Acceptance Criteria
- Default: unverified tokens are hidden; curated/verified surface as before.
- Toggling "Show unverified" reveals the withheld tokens, each badged; `isSpamToken` results still never appear.
- Adding an unverified token still persists via the normal `addToChain` path (+ optional confirm step).
- New/updated tests: verification threads to the picker; toggle on/off gating; badge shown only for unverified; spam still filtered regardless of toggle.
- SwiftLint + `make test` green (bar the known snapshot/flaky cases).
- No Figma yet → implement a clean minimal design and attach/describe a simulator screenshot for review.

### LINKED ISSUE #4966: [Refactor] Dynamic token search: migrate direct TokensStore list/discovery readers to TokenCatalogRepository [Wallet / Tokens]
# [Refactor] Dynamic token search: migrate direct `TokensStore` list/discovery readers to `TokenCatalogRepository` [Wallet / Tokens]

## Context
#4941 / PR #4954 introduced `TokenCatalogRepository` (bundled `TokensStore` + 1inch/Jupiter/SwapKit, verification-to-surface) and routed the token **search / manage-tokens** flow through it. But `TokensStore.TokenSelectionAssets` is still read **directly** in ~30 sites, so those surfaces don't get the dynamic catalog. This is **step 7** of the analysis (`wiki/pages/projects/vultisig/dynamic-token-datasource/analysis.md`). Stacks on #4954 — branch off `feat/dynamic-token-catalog-4941`, open the PR with `base: feat/dynamic-token-catalog-4941` (retargets to `main` after #4954 merges).

## The key distinction (do NOT migrate everything)
The ~30 direct readers split into two kinds — migrate one, leave the other:
- **MIGRATE — list / picker / discovery surfaces** that read the whole preset array to build a browsable list (e.g. `CoinSelectionViewModel.groupAssets` `CoinSelectionViewModel.swift:81`, any list/discovery surface still merging `TokensStore` presets). These should read the **candidate catalog** from `TokenCatalogRepository`, honoring verification-to-surface.
- **LEAVE — named-static curated-coin readers** that want a SPECIFIC curated coin, not the catalog: `TokensStore.rune`/`cacao`/`tcy`/`sruji`/`qbtc`/… statics and `findTokenMeta(chain:contractAddress:)` lookups in `VaultDefaultCoinService`, `CryptoPriceService` (`:254,283,363`), referral / VultTier / keysign / DeFi paths. `TokensStore` statics remain the curated source of truth for named coins.

## Scope — Must Do (codex-review loop, one reviewable commit per logical group)
1. **Enumerate + classify** every direct `TokensStore` reader (`rg -n 'TokensStore' VultisigApp/VultisigApp`). Produce a short ledger in the PR body: each site → migrate / leave (with the one-line reason). If a site is genuinely ambiguous or high-risk, LEAVE it and note it — bias toward conservative.
2. **Migrate the list/picker/discovery sites** to `TokenCatalogRepository` (use the `.appCatalog` instance the search path already uses; keep the merged-list-not-cached rule and fail-open semantics). Behavior-preserving for curated chains: the same coins that surface today still surface; `.unverified` stay gated per verification-to-surface; spam still filtered.
3. **Document** that `TokensStore` statics remain the curated named-coin source of truth (a short comment / PR note), so future readers know which path to use.

## Must NOT Do
- Don't migrate the named-static curated-coin readers or `findTokenMeta`-for-a-specific-coin sites.
- Don't change the enable/disable model (`vault.coins` / `HiddenToken`).
- Don't regress the swap source/destination pickers or the already-migrated search path (#4954) or the unverified toggle (#4960).

## Acceptance Criteria
- List/picker/discovery surfaces read candidates from `TokenCatalogRepository`; named-static readers unchanged.
- Regression: curated chains surface an identical coin set to before on the migrated surfaces; verification-to-surface + spam filtering honored; enable/disable round-trips unchanged.
- Unit/regression tests for at least the primary migrated surface (e.g. `CoinSelectionViewModel.groupAssets` output parity for a curated chain).
- SwiftLint + `make test` green (bar the known snapshot/flaky cases). PR body carries the site-by-site migrate/leave ledger.

gastonm5:
## Problem

Token search / add-token / pickers are anchored to the static, hand-maintained `TokensStore` as both the catalog's floor and ceiling — a token can't be searched or held until a release ships a new entry, and every ticker/logo/`priceProviderId` is hand-verified per platform and rots. The app already fetches, caches, merges and trust-filters dynamic token lists in **two disconnected places** — `TokenSearchService` + `SwapTokenListCache` (1inch/Jupiter) and the swap-destination `DestinationTokenProvider`/`DestinationTokenRegistry` (SwapKit / THOR·Maya pools / secured assets) — that don't share a contract and don't feed the wallet search.

Closes #4941. Closes #4955. Closes #4966.

> **Consolidated PR.** This was originally a 3-PR stack (#4954 engine → #4960 unverified surface → #4967 direct readers). All three branches now live here on one branch against `main`, and `main` has been merged in.

## Approach — generalize the registry that already exists

Promote the swap-only `DestinationTokenProvider`/`DestinationTokenRegistry` into an app-wide **`TokenCatalogProvider` / `TokenCatalogRepository`** (`Core/Services/TokenCatalog/`). The static `TokensStore` **and** every dynamic source now conform to one protocol, merged by one engine. Trust rides on a **`CatalogToken { meta, verification, sourceKind }`** wrapper + **`TokenVerification`** (`.curated` / `.verified(source:)` / `.unverified`) — **`CoinMeta.uniqueId`/`==`/`hash` are untouched** (the app-wide dedup identity is unchanged).

Key design decision (kept the delicate swap-picker hot path byte-identical): the bucket method `tokens(for:)` stays the provider primitive, so `DestinationTokenRegistry`/`Provider` remain thin typealiases and the swap picker compiles unchanged; `catalogTokens(for:)` + the new `catalog(for:)` (dedup by `uniqueId`, highest-precedence `CoinMeta` wins, strongest `verification` kept) are layered on top. Two configured shared instances — `TokenCatalogRepository.shared` (swap destinations, unchanged provider set) and `.appCatalog` (bundled + 1inch + Jupiter) — decouple the two surfaces.

## Part 1 — the catalog engine (#4941, steps 1–6)

1. **Trust model** — `TokenVerification` + `CatalogToken` wrapper. `CoinMeta` identity unaffected (unit-tested).
2. **Promote the contract** — `TokenCatalogProvider`/`TokenCatalogRepository` in `Core/Services/TokenCatalog/`, `precedence`, `catalog(for:)` merge, back-compat typealiases; conformers `providerKind`→`kind`.
3. **`BundledTokensProvider`** — wraps `TokensStore`, `precedence = .max`, `.curated`, no network → the offline floor; honors the `sepolia`/`thorchainChainnet`/QBTC-MLDSA visibility gates (vault-independent subset; the MLDSA-key gate stays in the view model).
4. **Network providers + offline snapshot** — `OneInchCatalogProvider` (1inch's curated `/tokens` whitelist → `.verified("1inch")`, with the tokens 1inch itself tags `RISK:malicious`/`RISK:suspicious`/`RISK:unverified` downgraded to `.unverified` — see *1inch verification, corrected* below), `JupiterCatalogProvider` (`?query=verified` list → `.verified("Jupiter")`), and a `TokenCatalogDiskCache` (Codable per-chain, off-MainActor I/O) for cold-offline last-good.
5. **Route search through the repository** — `TokenSearchService` assembles from `appCatalog.catalog(for:)` instead of the hardcoded preset+1inch/Jupiter merge; `loadTokens([CoinMeta])` signature unchanged, so `SwapTokenListCache` (TTL / coalescing / fail-open) and the swap pipeline are untouched.
6. **Verification-to-surface** — `TokenSearchService.surfaceableTokens` withholds `.unverified` candidates; `CoinService.isSpamToken` retained as the second gate on discovery.

## Part 2 — the add-token surface (#4955): one unified, local-first pool

Both browse and search read the **same** catalog. Curated/local tokens are always present synchronously; provider tokens append when they load. There is **no opt-in toggle** — searching is the reveal.

- **Local-first ordering everywhere.** Held coins, then curated `TokensStore` presets, then provider breadth — deduped by `CoinMeta.uniqueId` (curated/local meta wins a collision).
- **Browse (empty query):** local + **verified** provider tokens. `.unverified` is NOT shown in browse.
- **Search (non-empty query):** filters the full pool (local + verified + `.unverified`), so typing reveals the badged unverified long-tail.
- **⚠ badge** on `.unverified` rows (`triangle-warning`, 12pt, `Theme.colors.alertWarning`, non-interactive) + **add-confirm** listing each token's ticker and **full contract address**.
- **Spam hard-gate** (`CoinService.isLikelySpam`) applies to both lists — spam never appears, browse or search.
- **Local-token search regression fixed:** search previously filtered only the async catalog result and the load's `catch` never re-derived, so a curated token vanished from search while the fetch was pending or failed. The pool now always includes the presets and **fails open** to local.

## Part 3 — direct `TokensStore` readers (#4966)

Only **2** of the ~30 direct `TokensStore.TokenSelectionAssets` readers are genuine whole-preset browsable-list readers; both now route through `BundledTokensProvider.curatedTokens(for:defaults:)` (synchronous, curated), so the sepolia / thorchain-chainnet / QBTC-MLDSA visibility gates live in one place:

| Site | Surface |
|---|---|
| `CoinSelectionViewModel.groupAssets` | Chain/token selection grid (`VaultSelectChainScreen`, address book, send tab) — byte-identical grouped output; only the vault-specific MLDSA-key gate stays in the VM |
| `TokenSelectionLogic.preExistingTokens` | Per-chain curated add-token floor |

Everything else is **left** deliberately: named statics (`.rune`/`.cacao`/`.tcy`/`.sruji`/`.qbtc`/…), `findTokenMeta(chain:contract:)`, native-token lookups and vault-seeding paths want *one specific curated coin*, not a browsable catalog. Full migrate/leave ledger in the wiki plan (`dynamic-token-datasource/direct-readers-plan.md`).

The migrated surfaces route through the **bundled** provider, not the network-fanning async `catalog(for:)`: `groupAssets` drives a chain grid consumed synchronously by the address book, send tab and chain detail, so making it async + injecting hundreds of dynamic tokens would not be behavior-preserving. **Flagged for maintainers** — surfacing the full dynamic catalog on the chain grid is a follow-up with proper async plumbing + UX review.

## Security posture

- **Verification-to-surface** — only `.curated`/`.verified` auto-appear in browse; `.unverified` (tokens 1inch itself flags `RISK:*`, lookalike USDC/USDT on a different contract, airdrop dust) is reachable only by explicit search, badged, and behind an add-confirm.
- **`uniqueId`, never bare ticker** — held-coin enrichment, search exclusion and curated-preset exclusion are all `uniqueId`-keyed and disjoint, so an unverified lookalike sharing a held ticker (fake USDC on another contract) can't ride into the held/browse set.
- **1inch trust signal** — 1inch's own `RISK:*` tags are carried into `verification`, so the ~8% of its whitelist it flags as malicious/suspicious/unverified never auto-surfaces. The CoinGecko `providers` gate is *not* available on this endpoint; it remains live on the separate discovery path (`EvmCoinFinder`).
- **candidate ≠ enabled** — the repository output is the *candidate* set; the enabled set stays `vault.coins` and the tombstone `HiddenToken`. Nothing enters `vault.coins` without explicit user action.
- **Spam is a hard second gate** on both lists, and `isSpamToken` is retained on the discovery path.
- **Disk snapshot trust** — the Caches-sandbox snapshot caps loaded verification at `.verified` (never the dedup-winning `.curated`), and verification is never decoded from a third-party source payload (providers assign it in code from a live signal).

## On-device round (7 follow-up commits)

An on-device pass on the add-token screen surfaced six defects, all fixed here:

- **Search filtered one keystroke behind.** Typing `USDCC` listed USDC; deleting the trailing `C` made USDC vanish. The screen re-filtered from `.onReceive($searchText) { _ in … }`, discarded the emitted value, and had the view model read `searchText` back — but `@Published` fires its publisher in `willSet`, *before* the new value is stored, so every filter ran against the previous query. (The swap picker's equivalent sink uses the emitted value and was never affected.) Now filtered from `didSet` with the query passed explicitly, and the view's subscription is gone. `testSearchResultsTrackTheLatestKeystroke` drives the exact repro.
- **The unverified add-confirm is a bottom sheet, not a system alert** — `UnverifiedTokenBottomSheet`, modeled on `SecurityScannerBottomSheet` (warning icon, safe action primary, continue-anyway secondary). It no longer enumerates the tokens: a wall of tickers and full contract addresses in a confirm is not read, and the per-token detail already lives on the badged cell behind it. Copy reworded in place across all 8 locales (existing keys reused, not orphaned) and re-sorted. ⚠ **Reviewers:** this removes the only surface that showed the contract address at add-time — the ⚠ badge is now the sole per-token signal.

- **Held tokens were unfindable by search.** The search pool omitted `selectedTokens` *and* `filteredTokens` dropped every held `uniqueId`, so searching a token the vault owns returned nothing — e.g. `vult` on Ethereum found nothing despite VULT being a curated preset, because the vault holds it. Held tokens now lead the pool (local-first) and the filter excludes nothing; the pool is already deduped by `uniqueId`, and a same-ticker lookalike on a different contract is still a distinct, badged entry. A test that pinned the old behaviour was rewritten, and `testHeldCuratedTokenIsFoundBySearch` now guards the reported case with the real VULT preset.
- **Names overlapped between columns.** `AssetSelectionGridCell` sized its label with `.fixedSize(horizontal: true, …)`, so a name wider than the 74pt cell laid out at full ideal width and ran under its neighbours. Now clipped to the cell, truncating in the **middle** — these long tickers share prefixes, so tail truncation renders `aEthLidoUSDC` and `aEthLidoUSDS` identically.
- **Logos span forever.** `AsyncImageView` used `CachedAsyncImage`'s `placeholder:` initializer, which shows the placeholder for every non-success phase, so a 404 or non-image body (routine for long-tail logo URLs) never resolved. Now phase-based: `.failure` renders the existing ticker-letter fallback. *Not covered:* a host that connects then stalls still spins until the default request timeout — bounding it needs a raw `URLRequest`, which SwiftLint confines to `Core/Networking`.
- **Grid updates were unanimated, and churned identity.** Added an animated diff — and, from the Codex pass, fixed `ForEach(elements, id: \.self)`: `AssetSection` hashes its asset array, so the section's identity changed on every keystroke and SwiftUI rebuilt the whole section, restarting every cell's logo request (cells visibly re-entered their loading state while typing). Now keyed on `section.type`, unique at all six call sites.

Note the grid-cell and image changes are in **shared** components: `AssetSelectionGridCell` is used by 5 other screens (chain selection, defi chain selection, custom RPC, tx-history filter, key-import chains) — long names there now truncate instead of overlapping — and `AsyncImageView` is app-wide, where the only behaviour change is failed remote logos falling back instead of spinning.

## Maintainer review round — 1inch verification, corrected

@johnnyluo found that the per-token verification gate never fired. **Reproduced and confirmed.**

`OneInchCatalogProvider.catalogTokens` → `OneInchService.fetchTokens` → `/1inch/swap/v6.0/{chain}/tokens`, and that endpoint **never returns a `providers` field**. Pulled live through the app's own proxy across five chains:

| Chain | tokens | with `providers` |
|---|---:|---:|
| Ethereum | 2,225 | 0 |
| BSC | 1,596 | 0 |
| Arbitrum | 825 | 0 |
| Polygon | 194 | 0 |
| Base | 165 | 0 |

Keys are exactly `address, symbol, decimals, name, logoURI, eip2612, tags`. `providers` exists only on the *different* `/token/v1.2/{chain}/custom` endpoint (bulk metadata by address) — which is what `EvmCoinFinder` uses. So `isCoinGeckoVerified` was always `false` here, `.verified(source: "CoinGecko")` was dead code on the catalog path, and **every** 1inch-sourced token was tagged `.unverified` — including USDC/WETH on chains where they aren't hand-curated. The split this PR described did not happen.

**Fix**, per @johnnyluo's directive that 1inch-discovered tokens should not be marked unverified:

- `/tokens` is 1inch's **curated swap whitelist**, so membership in it is the trust signal → `.verified(source: "1inch")`. The label names the source that actually vouches for the token; it no longer claims a CoinGecko signal that isn't present.
- The one per-token trust signal the response *does* carry is the `RISK:*` tag family. Tokens 1inch itself tags `RISK:malicious` / `RISK:suspicious` / `RISK:unverified` are downgraded to `.unverified` — badged, search-only, never in browse. `RISK:availability` is a routing/liquidity property, not a legitimacy one, and does **not** downgrade. Conflicting tags (1inch does emit `RISK:norisk` + `RISK:unverified` together) fail closed.
- This is what thread #2 asked for — "use a different signal for the catalog's per-token verification" — and it needs no extra request: the tags are in the response we already fetch.
- **Why not blanket-verified:** 1inch tags 16 Ethereum / 8 BSC / 1 Arbitrum token `RISK:malicious`. Marking the whole list verified would auto-surface those in browse. Conversely they are *not* dropped outright — the tag is noisy (it also hits Wrapped XRP, STASIS EURS and some Aave aTokens), so `.unverified` (reachable by search, badged, behind the confirm sheet) is the proportionate tier rather than hiding them.

Resulting split on the sampled chains: **4,603 verified / 402 unverified (8.0%)**. USDC, WETH, USDT, DAI and WBTC all carry `RISK:norisk` → verified. This also keeps the ⚠ badge + `UnverifiedTokenBottomSheet` surface reachable — with `.verified` at the provider level and no per-token downgrade, no `appCatalog` provider would emit `.unverified` at all and that UI would have been dead.

**Kept, not deleted:** `OneInchToken.providers` and `isCoinGeckoVerified` stay — `EvmCoinFinder` gates auto-discovery on them via `/token/v1.2/{chain}/custom`, where the field genuinely is returned. Only the *catalog path's* use of them was dead, and that is what was removed. A test now pins that the two endpoints' signals are independent.

## Verification

- **Cross-model review loop**: every step gated by a read-only `codex exec` pass (findings triaged fixed / rejected-with-reason / deferred), plus whole-branch passes on each part. Ledgers in the wiki plans (`pages/projects/vultisig/dynamic-token-datasource/`). Notable fixes: whole-branch P1 (a transient outage could shrink the cached catalog) fixed by capping — rather than flooring — disk verification; spam gated on both lists; add-confirm shows the contract; held-coin match by `uniqueId`.
- **CodeRabbit** on the engine: stale-retry finding addressed — `TokenSelectionLogic.showRetry` gated on `TokenSearchServiceError.networkError`/`.rateLimitExceeded`, which the catalog path can no longer throw (it fails open to the bundled floor + disk last-good), and no view consumed it; the dead retry predicate and the unreachable error cases are removed.
- **Post-merge integration**: `main` merged in (logging facade #4945 + spokes, `TokenDiscovererRegistry` #4935, `CustomTokenViewModel` #4936, `ChainConfig` #4914 …) with no conflicts; the three catalog files added here were converted from raw `OSLog` `Logger`s to the newly-landed `Log.wallet.service` / `Log.wallet.store` facade so they honour the per-feature gate.
- **Gate**: full `make test` on the pinned sim (iPhone 17 Pro Max) after every round — latest (after the 1inch verification fix): `** TEST SUCCEEDED **`, **4025 tests executed, 1 skipped, 0 failures** (clean even of the usual local `CreateVaultSnapshotTests.testCreateVaultScreen_iPhone16Pro` artifact). SwiftLint clean on every changed file. CI on the consolidated branch: both jobs green.
- **Consolidation review**: a further read-only `codex exec` pass over the consolidation delta (the `main` merge + the two follow-up commits) found no defects — it independently confirmed the catalog path can only throw `.cancelled`, that no consumer of the removed error cases or the retry predicate remains, that the logging conversion preserves OSLog privacy/interpolation semantics, and that the merge produced no duplicated discovery/filter logic or stale-path edits.

## What's deferred

- Verification-gating the swap **destination** registry (SwapKit / pool / secured) — that breadth is swappable-by-definition and intentionally surfaced; gating it is a separate product decision.
- Discovery-as-a-provider (now unblocked by #4935 merging), phase-2 community token-lists (Uniswap / Trust allow·deny), Android/Windows/SDK parity, and the full destination-picker migration onto `catalog(for:)`.

## Risks / please review on-device

Touches the token catalog and the scam-token surface, and reroutes the wallet/swap token list. **Flagging for human review + on-device testing:** wallet add-token browse vs search (local-first order, ⚠ badge, add-confirm), swap source/destination pickers (curated vs dynamic, dedup, curated logo/price winning), airplane-mode cold launch (bundled floor), the chain-selection grid after the `groupAssets` migration, and enable/disable + hidden-tombstone round-trips.

## Summary by CodeRabbit

* **New Features**
  * Added an app-wide token catalog powering wallet search/token add and swap destinations, combining bundled + external sources with deduping and chain-visibility rules.
  * Introduced curated/verified/unverified trust tiers, including an “unverified token” confirmation flow and an unverified badge.
* **Bug Fixes**
  * Improved catalog reliability with last-good disk snapshots and safer handling of verification strength on load failures.
  * Enhanced spam filtering and logo loading behavior (better loading/failure states).
* **UI/UX Improvements**
  * Smoother token grid updates, plus improved long-name truncation.
* **Tests**
  * Added coverage for catalog routing, caching, trust-tier behavior, and search/badging logic.


VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
```diff
@@ -27,52 +30,39 @@ struct TokenSearchService {
...
+        guard !Task.isCancelled else { throw TokenSearchServiceError.cancelled }
 
-        case .Solana:
-            let jupTokens = try await SolanaService.shared.fetchSolanaJupiterTokenList()
-            return jupTokens
-        default:
-            return []
-        }
+        return Self.surfaceableTokens(from: candidates)
     }
```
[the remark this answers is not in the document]
gastonm5: Fixed in `64dd3b8a1`.

Verified the finding before acting on it: `loadCatalog` → `fetchUncached` can only ever throw `.cancelled`, because `TokenCatalogRepository.catalog(for:)` is non-throwing and both network providers swallow failures and fall back to the last-good disk snapshot (or `[]`). So `.networkError` / `.rateLimitExceeded` were indeed unreachable — and `showRetry` had no consumer at all: no view read it, so the retry affordance was dead in two layers, not just mis-gated.

Rather than restore a fabricated error path, the dead code is removed: `TokenSelectionLogic.showRetry` + the `TokenSelectionViewModel` passthrough, and the `TokenSearchServiceError` cases the service can no longer produce (`.noTokens`, `.networkError`, `.rateLimitExceeded`), so the error type stops advertising failure modes that do not exist. The fetch error is still recorded on the view model as fail-open evidence and is asserted by `TokenSelectionPoolTests`.

Note this PR has since been consolidated: #4960 and #4967 were folded onto this branch and `main` was merged in.

VultisigApp/VultisigApp/Features/Wallet/TokenSelection/View/UnverifiedTokenBottomSheet.swift
```diff
@@ -0,0 +1,59 @@
...
+
+            VStack(spacing: 8) {
+                PrimaryButton(title: "cancel".localized, action: onCancel)
+
+                Button("continueAnyway".localized, action: onContinue)
+                    .frame(height: 42, alignment: .center)
+                    .foregroundStyle(Theme.colors.textButtonDisabled)
+                    .font(Theme.fonts.caption10)
+                    .frame(maxWidth: .infinity)
+                    .buttonStyle(.plain)
```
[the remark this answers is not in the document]
gastonm5: Agreed — fixed in `301bd21c1`.

Replaced the raw `Button` with `PrimaryButton(title:type:.secondary)`. Keeping the hierarchy deliberate: `cancel` stays `.primary` (the safe action) and continue-anyway is the subdued `.secondary`, so the risky path is still visually quieter without hand-rolling a style.

For the record on provenance: the raw-Button pattern was copied from `SecurityScannerBottomSheet`, which does the same thing for its own continue-anyway action. That is pre-existing and out of scope here, but it is the same design-system violation if anyone wants to sweep it.

VultisigApp/VultisigApp/Features/Wallet/TokenSelection/View/TokenSelectionGridCell.swift
```diff
@@ -11,14 +11,23 @@ struct TokenSelectionGridCell: View {
     let coin: CoinMeta
     let name: String?
     let showChainIcon: Bool
+    let verification: TokenVerification
     var onSelection: (Bool) -> Void
 
     @State var isSelected: Bool
```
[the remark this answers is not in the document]
gastonm5: Not taking this one — the premise does not hold in this repo.

`private_swiftui_state` is **not** enabled in `VultisigApp/.swiftlint.yml` (it is not in the config at all), and a full `swiftlint lint --config VultisigApp/.swiftlint.yml` run over this branch reports 0 violations in `Features/Wallet/`. So nothing flags this today.

Two more reasons to leave it: the line is **pre-existing** — this PR only added the `verification` parameter and the badge overlay to this cell, it did not introduce `isSelected` — and non-private `@State var isSelected` is the established pattern for this family of cells (`ChainSelectionGridCell.swift:15`, `SwapChainCell.swift:17`), because the parent seeds the initial value through the memberwise-style init. Making this one file private while its siblings keep the pattern buys no enforcement and costs consistency.

Worth doing as a deliberate sweep across all the selection cells (and enabling the rule so it stays fixed), but not as a drive-by in this PR.

VultisigApp/VultisigApp/Blockchain/Swaps/OneInch/OneInchToken.swift
```diff
@@ -38,4 +38,17 @@ extension OneInchToken {
...
     }
+
+    /// Trust-carrying catalog candidate. Unlike `toCoinMeta` (which drops the
+    /// legitimacy signal), this preserves 1inch's CoinGecko-verified flag as the
+    /// token's `verification` — the same signal `EvmCoinFinder` uses as its scam
+    /// gate. Note: 1inch `/tokens` carries no CoinGecko id, so `priceProviderId`
+    /// stays empty; the curated bundled provider wins dedup and supplies it for
+    /// known tokens.
+    func toCatalogToken(chain: Chain, sourceKind: String) -> CatalogToken {
+        let verification: TokenVerification = isCoinGeckoVerified
```
johnnyluo: I pulled the live response from the endpoint `OneInchCatalogProvider` actually calls (`/1inch/swap/v6.0/{chain}/tokens`, via the Vultisig proxy) for Ethereum, Arbitrum, Base, and Polygon. None of the ~3,400 tokens returned across those four chains has a `providers` field at all — the payload only contains `address`, `symbol`, `decimals`, `name`, `logoURI`, `eip2612`, `tags`. Example (Ethereum USDC):

```json
{
  "address": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
  "symbol": "USDC",
  "decimals": 6,
  "name": "USD Coin",
  "logoURI": "https://tokens.1inch.io/...",
  "eip2612": true,
  "tags": ["connector", "crosschain", "defi", "GROUP:USDC", "PEG:USD", "RISK:norisk", "stablecoin", "tokens"]
}
```

The `providers` field does exist in 1inch's API, but only in the response of the *different* `/token/v1.2/{chain}/custom` endpoint (bulk metadata lookup by address) — that's what `EvmCoinFinder` actually uses for its CoinGecko gate. `fetchTokens` (called from `OneInchCatalogProvider.catalogTokens`, see the other comment) hits the plain `/tokens` whitelist endpoint instead, which never returns `providers`.

So `providers` decodes to `nil` for every token this provider emits, `isCoinGeckoVerified` is always `false`, and `.verified(source: "CoinGecko")` here is currently dead code for the catalog path — every 1inch-sourced token (including reputable ones not already in the curated bundle, e.g. USDC/WETH on a chain where they aren't hand-curated) gets tagged `.unverified` today, contradicting the PR description's stated fix.

VultisigApp/VultisigApp/Core/Services/TokenCatalog/OneInchCatalogProvider.swift
```diff
@@ -0,0 +1,69 @@
...
+        let catalog = await catalogTokens(for: chain, forceRefresh: forceRefresh)
+        let metas = catalog.map(\.meta)
+        return DestinationTokenBucket(chain: chain, tokens: metas, uniqueIds: Set(metas.map(\.uniqueId)))
+    }
+
+    func catalogTokens(for chain: Chain, forceRefresh _: Bool) async -> [CatalogToken] {
+        guard service.isChainSupported(chain: chain), let chainID = chain.chainID else { return [] }
+        let sourceKind = kind
+        do {
+            let tokens = try await service.fetchTokens(chain: chainID)
```
johnnyluo, line 58: This calls `OneInchService.fetchTokens`, which hits `/1inch/swap/v6.0/{chain}/tokens` — I verified live that this endpoint's response never includes a `providers` field (checked Ethereum/Arbitrum/Base/Polygon, ~3,400 tokens, 0 with `providers`). See the comment on `OneInchToken.toCatalogToken` for the full response sample.

That means every token from this provider currently gets `.unverified` regardless of actual CoinGecko listing status — the verified/unverified split described in the PR body doesn't take effect for 1inch tokens. To get real per-token trust signal here, this would need to route through (or additionally call) the `/token/v1.2/{chain}/custom` endpoint that `providers` is actually sourced from (same one `EvmCoinFinder` uses), or use a different signal for the catalog's per-token verification.

johnnyluo (commented):
One finding from testing the 1inch verification path against the live API — details in the two inline comments below.

VultisigApp/VultisigApp/Blockchain/Swaps/OneInch/OneInchToken.swift
johnnyluo: for those tokens we discovered by using 1inch api endpoint , we should NOT mark it as unverified

gastonm5: Reproduced — you're right, and thank you for digging this out. I pulled `/1inch/swap/v6.0/{chain}/tokens` through the same proxy across five chains and got the identical result:

| Chain | tokens | with `providers` |
|---|---:|---:|
| Ethereum | 2,225 | 0 |
| BSC | 1,596 | 0 |
| Arbitrum | 825 | 0 |
| Polygon | 194 | 0 |
| Base | 165 | 0 |

Keys are exactly `address, symbol, decimals, name, logoURI, eip2612, tags`. So `providers` was `nil` for every token on this path, `isCoinGeckoVerified` was always `false`, and `.verified(source: "CoinGecko")` here was dead code — every 1inch-sourced token landed on `.unverified`. The verified/unverified split the PR body described never happened.

Worth noting the blast radius was bigger than mis-badging: `TokenSearchService.surfaceableTokens` filters on `autoSurfaces`, so with everything `.unverified` the **entire 1inch breadth was also being dropped from the swap source picker** on EVM chains, leaving it with just the curated `TokensStore` presets. That's fixed by the same change.

**Fixed in `6c22a56`.** `/tokens` is 1inch's curated swap whitelist, so membership in it is the trust signal — tokens are now `.verified(source: "1inch")`, naming the source that actually vouches for them rather than claiming a CoinGecko signal that isn't in the response.

`providers` / `isCoinGeckoVerified` are **kept**, not deleted: `EvmCoinFinder` gates auto-discovery on them via the `/token/v1.2/{chain}/custom` endpoint, where the field genuinely is returned. Only the catalog path's use of them was dead, and that's what was removed. `EvmCoinFinderTests` now has a test pinning that the two endpoints' signals are independent, so this can't get re-conflated.

VultisigApp/VultisigApp/Core/Services/TokenCatalog/OneInchCatalogProvider.swift
gastonm5: Reproduced — confirmed live against the same proxy across Ethereum, BSC, Arbitrum, Polygon and Base: 5,005 tokens, **zero** with a `providers` field. Details and the per-chain table are in the reply on `OneInchToken.toCatalogToken`.

You offered two routes: additionally call `/token/v1.2/{chain}/custom`, **or** use a different signal for the catalog's per-token verification. I took the second — and it needs no extra request, because the signal is already in the response we fetch.

`/tokens` carries a `RISK:*` tag family. Live distribution on Ethereum: `RISK:norisk` 706, `RISK:availability` 353, `RISK:suspicious` 116, `RISK:unverified` 71, `RISK:malicious` 16 (the rest are unannotated).

**Fixed in `6c22a56`:**

- Whitelist membership is the baseline trust signal → `.verified(source: "1inch")`.
- Any `RISK:` tag other than `RISK:norisk` / `RISK:availability` downgrades to `.unverified` — badged, search-only, never in browse. `RISK:availability` is a routing/liquidity property, not a legitimacy one. The rule is written as "unknown `RISK:` value ⇒ downgrade", so a classification 1inch adds after this ships fails closed instead of auto-surfacing unreviewed.
- Conflicting tags fail closed (1inch does emit `RISK:norisk` + `RISK:unverified` on the same token — 53 in the sample).
- No `RISK:` tag at all ⇒ verified. Much of the whitelist is unannotated and absence of a flag isn't a reason to withhold.

Resulting split across the sampled chains: **4,603 verified / 402 unverified (8.0%)**. USDC, WETH, USDT, DAI, WBTC all carry `RISK:norisk` → verified.

Two judgement calls I'd like your read on:

1. **Not blanket-verified.** 1inch tags 16 Ethereum / 8 BSC / 1 Arbitrum token `RISK:malicious`; marking the whole list verified would auto-surface those in browse.
2. **But not dropped either.** That tag is noisy — it also hits Wrapped XRP, STASIS EURS and some Aave aTokens — so hiding them outright would lose legitimate assets. `.unverified` (reachable by search, badged, behind the confirm sheet) is the proportionate tier.

If you'd rather every 1inch token be verified with no per-token downgrade at all, it's a one-line change — but note that would also leave the ⚠ badge and the unverified confirm sheet with no reachable path, since no other `appCatalog` provider emits `.unverified`.

VultisigApp/VultisigApp/Blockchain/Swaps/OneInch/OneInchToken.swift
gastonm5: Agreed, and done in `6c22a56`.

1inch-discovered tokens are no longer marked unverified by default. `/swap/v6.0/{chain}/tokens` is 1inch's curated swap whitelist, so being on it is itself the trust signal — tokens now come back `.verified(source: "1inch")`. The label names 1inch rather than CoinGecko, because there is no CoinGecko signal on this endpoint (that's the `providers` field, which only exists on `/token/v1.2/{chain}/custom` — `EvmCoinFinder`'s path, untouched).

The one carve-out: tokens **1inch itself** flags with a `RISK:` tag other than `norisk`/`availability` (`malicious`, `suspicious`, `unverified`) stay `.unverified` — still reachable by search and still addable, just badged and kept out of the browse list. That's ~8% of the whitelist; USDC/WETH/USDT/DAI/WBTC and the other 92% are verified. Rationale and the option to make it blanket-verified instead are in my reply on the `OneInchCatalogProvider` thread — happy to go either way on your call.

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Blockchain/Swaps/OneInch/OneInchToken.swift
gastonm5: Following up on this now that it's merged — surfacing the whole 1inch whitelist as `.verified` had a second-order effect worth closing out.

Because browse renders the full verified provider breadth, marking the whitelist verified took the add-token browse list from a curated handful to **~2,000 tokens on Ethereum**. Correct trust-wise, unusable as a list.

Follow-up on `feat/token-browse-ranking` (branched off the merge commit) bounds it:

- **Browse = held + curated presets + the best 20 _per provider_.** Not a truncation: the providers now return their lists ranked best-first, so it's a shortlist.
  - **1inch** — `/swap/v6.0/{chain}/tokens` carries no volume/mcap/liquidity field, so there's no real ranking to be had. It does carry your own tag curation, so tokens are tiered `bluechip` → `PEG:*` → `connector` → rest (Ethereum: 41 / 178 / 806 / 1,200 of 2,225). The code calls this a quality *tier*, not a rank, so nobody mistakes it for one. Head of Ethereum goes from `$MBG, ZRX, 0x0, 0xBTC…` to `1INCH, AAVE, LINK, COMP, CRV, WBTC…`; Base from `$OLIVIA2.0…` to `USDT, USDC, WETH, DAI, EURC, SOL`.
  - **Jupiter** — ranked by `organicScore`, which the app was decoding and discarding. Picked over `mcap` on the live data: `mcap` covers 66% of the 4,319-token list and its head fills with paper valuations (tSAT — $13.4B cap against $57K liquidity; HONx — $1.76B cap, zero liquidity) outranking ETH and PYUSD. `organicScore` is on 100% of entries and heads with USDC, USDT, SOL, ETH, JitoSOL, JUP.
- **Search still reaches everything** — the full verified *and* unverified breadth. The cap is browse-only and can never be why a token isn't found; there's a test pinning that a token ranked #500 is absent from browse and still found by search.
- Tier and trust stay orthogonal: a `RISK:*`-flagged bluechip still ranks high **and** is still `.unverified`, so it stays out of browse and behind the confirm sheet. Your fix here is unchanged.

Also fixed a pre-existing bug this exposed: `filteredTokens` ended in `.prefix(20)` on `main` too, so token search has always rendered at most 20 matches — searching `US` on Ethereum returned 20 of several hundred with no second page, hiding exactly the badged lookalikes the surface exists to reveal. Removed rather than raised (the grid is a `LazyVGrid`).

Leaving this thread open for you.

PR #4976, opened 2026-07-28
perf(wallet): 114 → 4.4 ms/s hitch time on the wallet chain list

### LINKED ISSUE #4974: [Fix] Wallet chain list scrolls at 114 ms/s hitch time — every row is masked, and the content tree rebuilds every frame [Wallet / Main]
## Problem

Scrolling the chain list on the Wallet tab is visibly janky, and gets worse the more chains a vault has. Measured on an iPhone 16 Pro (iOS 26.5.2) over a 20 s sustained scroll with the Animation Hitches instrument:

| | |
|---|---|
| hitch time ratio | **114.2 ms/s** |
| hitches | 34, totalling 2271 ms |
| worst single hitch | 758 ms |
| hitches over 100 ms | 4 |
| offscreen render passes per hitching frame | 256 – 494 (median 426) |

Apple's guidance is **< 5 ms/s** good and **> 10 ms/s** user-visible, so this is roughly **12× the threshold at which users feel it**.

Every hitch narrative is render-side — `Potentially expensive render, N offscreen passes` / `Potentially expensive GPU work` — which is what pointed at masking rather than CPU work.

This is a regression of the ground covered by #4495 (fixed in #4500). The `ChainRowModel` value projection from that PR is intact and still doing its job; the cost that came back is elsewhere.

## Expected Behavior

Scrolling the chain list holds 120 fps with hitch time under the 5 ms/s "good" bar, on a vault with many chains.

## Steps to Reproduce

1. Open a vault with a large number of chains (the effect scales with row count).
2. Wallet tab, let balances settle.
3. Scroll the chain list up and down continuously.
4. Profile with Instruments → Animation Hitches on a real device.

## Solution

Five independent causes, in descending order of measured impact.

**1. Every row is clipped, including rows with no corners to round.** `CommonListItemContainer` applies `.clipShape(.rect(...))` to *every* row. On middle rows all four radii are `0` — a visually no-op clip that still installs a mask and can force an offscreen pass. An N-row list pays ~N pointless masks. **This is the dominant cost by a wide margin.**

**2. The scroll-offset throttle invalidates the whole content tree every frame.** `VaultMainScreenScrollView` keeps its throttle bookkeeping in `@State` and writes `pendingValue` unconditionally on every scroll frame. The documented 5 fps throttle rate-limits the `scrollOffset` binding but not the thing that costs: the `@State` write invalidates the scroll view's body, which re-invokes `content()` — the entire wallet content tree — at display rate. (Two lesser defects in the same function: the trailing `Timer` is scheduled in `.default` mode, which does not run during scroll tracking, so the trailing update is dropped for the whole duration of a drag; and `coordinateSpaceName = UUID()` plus its `.coordinateSpace` are dead code.) `VaultMainScreenScrollView` is also the DeFi tab's scroll view, so both tabs pay this.

**3. The vault total is recomputed inside that invalidated body.** `VaultMainScreen` calls `homeViewModel.balanceText(for: vault)` eagerly on every body evaluation, which walks every coin through `RateProvider`. Same defect at `HomeMainHeaderView`.

**4. `RateProvider.rate(for:)` is a linear scan.** `rates.first(where: { $0.id == identifier })` over a `Set<Rate>` is a sequence scan, not a hash lookup, plus a `CoinMeta` allocation and a lowercased string interpolation per call. This sits on the path of every fiat render in the app, not just this screen.

**5. Rows are neither lazy nor `Equatable`.** The list is a plain `VStack`, so all rows realize up front; and `VaultChainCellView` holds a `Vault` reference plus closures, so SwiftUI cannot skip an unchanged row when the parent body runs.

### Must Do
- [ ] Clip only the first and last rows in `CommonListItemContainer` (a single-row list is both, and must keep all four corners)
- [ ] Move the scroll throttle's bookkeeping out of `@State`; schedule the trailing timer in `.common` mode; delete the dead coordinate space
- [ ] Publish a preformatted vault total instead of computing it in `VaultMainScreen` / `HomeMainHeaderView` bodies
- [ ] Back the rate cache with a dictionary keyed by `Rate.id`, preserving the newest-wins merge semantics
- [ ] Make the chain list lazy and its rows `Equatable`

### Must NOT Do
- Don't change any rendered output — this is behaviour-preserving throughout
- Don't undo the `ChainRowModel` projection from #4500, or the deliberate "no stale-then-fresh double reorder" behaviour it protects
- Don't collapse `GroupedChainCellView` to a wallet-only shape; the DeFi tab shares it

## Acceptance Criteria
- [ ] Build succeeds
- [ ] Hitch time ratio under 10 ms/s on a many-chain vault, measured the same way
- [ ] Chain list, DeFi list, chain detail, receive picker and folder picker all render identically
- [ ] Existing `VaultDetailViewModelTests` invariants stay green

gastonm5:
## Description

Fixes #4974

Wallet chain list scroll performance. Measured on an iPhone 16 Pro (iOS 26.5.2), 20 s sustained scroll of a many-chain vault, Animation Hitches, `xctrace --attach`, identical instrument sets and the same vault on both sides:

| | before | after |
|---|---|---|
| **hitch time ratio** | **114.2 ms/s** | **4.4 ms/s** |
| hitches | 34 | 8 |
| total hitch time | 2271 ms | 88 ms |
| worst single hitch | 758 ms | 17 ms |
| hitches over 100 ms | 4 | 0 |
| offscreen passes / hitching frame | 256 – 494 (median 426) | negligible |

Apple's bar is < 5 ms/s good, > 10 ms/s user-visible — so this moves the screen from ~12× the user-visible threshold to inside the "good" band.

**Stacked on #4975** (the rate-cache change), which the measurement above includes. GitHub will retarget this to `main` when that merges.

## The commits, and what each one is for

Every hitch narrative in the baseline was render-side, so the masking work is the load-bearing part; the rest are real defects found along the way that the profile confirmed as secondary.

- **`perf(list): clip only the end rows of a common list`** — **the dominant fix.** `CommonListItemContainer` applied `.clipShape(.rect(...))` to *every* row, including middle rows where all four radii are `0` — a visually no-op clip that still installs a mask. An N-row list paid ~N pointless masks. Now only `isFirst || isLast` are clipped (a single-row list is both and keeps all four corners). This alone accounts for the great majority of the improvement.
- **`perf(wallet): stop the scroll-offset throttle invalidating the content tree`** — the throttle kept `pendingValue`/`lastUpdateTime`/`throttleTimer` in `@State` and wrote `pendingValue` on every scroll frame, so the documented 5 fps throttle rate-limited the `scrollOffset` binding but not the body invalidation: `content()` — the whole wallet content tree — was rebuilt at display rate. Moved into a reference box so the bookkeeping is invisible to SwiftUI. Also fixes the trailing `Timer` being scheduled in `.default` mode (it never fires during scroll tracking) and deletes a dead `UUID()` coordinate space. **This view is also the DeFi tab's scroll view, so both tabs benefit.**
- **`perf(wallet): publish the vault total instead of computing it in the body`** — `homeViewModel.balanceText(for: vault)` walked every coin through `RateProvider` on every body evaluation. Now a published, vault-tagged `TotalBalance`, with a fallback to the on-the-fly compute when it is absent or belongs to another vault (the first frame precedes `onLoad`, and a vault switch briefly leaves the previous vault's value).
- **`perf(wallet): read the published vault total in the home header`** — same defect at `HomeMainHeaderView`, same shape of fix. The DeFi branch of that property was deliberately left alone: `VaultDetailViewModel` has no DeFi-balance trigger, and `HomeViewModel.defiBalanceText` and `DefiBalanceService.totalBalanceInFiatString` are already two different formulas that can disagree — reconciling them is separate work, not a hoist.
- **`perf(wallet): restore laziness and equatable rows in the chain list`** — the list was a plain `VStack` (all rows realized up front) and `VaultChainCellView` held a `Vault` reference plus closures, so no row could ever be skipped. Now `LazyVStack` + `.equatable()`, with `vaultPubKeyECDSA` in the equality so an equality-skipped row cannot keep closures captured against a previous vault.

## What was tried and rejected

`.drawingGroup()` on the two `blur(radius: 36)` background gradients was implemented, measured, and **dropped**: 4.7 ms/s with it vs 4.4 ms/s without — no measurable benefit, for roughly 21 MB of persistent texture per live background instance. Recording it here so nobody re-derives it. If the background blurs are ever revisited, the zero-cost answer is not to blur a static gradient at runtime at all.

## Which feature is affected?

## Checklist

## Additional context

`CommonListItemContainer` is shared by five call sites — wallet chain list, DeFi chain list, chain detail, receive chain picker, folder picker. All five were checked: no row draws outside its own bounds, so the middle-row clip was genuinely a no-op and removing it is not observable. `.plainListItem()` was also dropped from the modifier; of the five call sites only `AddFolderScreen` is inside a real `List`, and there the rows sit in a `VStack` that is itself the single list row and carries its own `.plainListItem()`.

- **Agent**: Claude Code
- **Issue**: #4974
- **Confidence**: high on the measurements and the CPU-side changes; medium on visual identity of the clip change, which is argued structurally rather than screenshot-verified
- **Needs human review for**: the rendered output of the clip change across all five list call sites

### DIFF
--- a/VultisigApp/VultisigApp/Components/List/CommonListItemContainer.swift
+++ b/VultisigApp/VultisigApp/Components/List/CommonListItemContainer.swift
@@ -25,27 +25,34 @@ struct CommonListItemContainer: ViewModifier {
             Separator(color: Theme.colors.borderLight, opacity: 1)
                 .showIf(!isLast)
         }
+        // The end rows carry the surface's corner radius themselves, so the
+        // group reads as one rounded card without a mask spanning the whole
+        // list. A container-height `clipShape` is a render-side cost that grows
+        // with the row count and forces the list to be laid out as one unit.
         .clipShape(
             .rect(
-                topLeadingRadius: isFirst ? 12 : 0,
-                bottomLeadingRadius: isLast ? 12 : 0,
-                bottomTrailingRadius: isLast ? 12 : 0,
-                topTrailingRadius: isFirst ? 12 : 0
+                topLeadingRadius: isFirst ? CommonListContainer.cornerRadius : 0,
+                bottomLeadingRadius: isLast ? CommonListContainer.cornerRadius : 0,
+                bottomTrailingRadius: isLast ? CommonListContainer.cornerRadius : 0,
+                topTrailingRadius: isFirst ? CommonListContainer.cornerRadius : 0
             )
         )
-        .plainListItem()
     }
 }
 
 /// Wrapping container for a group of `commonListItemContainer` rows: a single
 /// rounded surface with a hairline border, matching the Figma list style.
 struct CommonListContainer: ViewModifier {
+    static let cornerRadius: CGFloat = 24
+
     func body(content: Content) -> some View {
         content
-            .background(Theme.colors.bgSurface1)
-            .clipShape(RoundedRectangle(cornerRadius: 24))
+            .background(
+                RoundedRectangle(cornerRadius: Self.cornerRadius)
+                    .fill(Theme.colors.bgSurface1)
+            )
             .overlay(
-                RoundedRectangle(cornerRadius: 24)
+                RoundedRectangle(cornerRadius: Self.cornerRadius)
                     .strokeBorder(Theme.colors.borderLight, lineWidth: 1)
                     .allowsHitTesting(false)
             )
--- a/VultisigApp/VultisigApp/Components/List/CommonListItemContainer.swift
+++ b/VultisigApp/VultisigApp/Components/List/CommonListItemContainer.swift
@@ -19,24 +19,42 @@ struct CommonListItemContainer: ViewModifier {
         index == itemsCount - 1
     }
 
+    /// Only the end rows have a non-zero radius, so only they need a mask. A
+    /// middle row's clip is a plain rectangle the size of the row — visually a
+    /// no-op that still installs a mask and can cost an offscreen pass, once
+    /// per row, on every frame the list composites.
+    private var needsCornerClip: Bool {
+        isFirst || isLast
+    }
+
     func body(content: Content) -> some View {
+        // The end rows carry the surface's corner radius themselves, so the
+        // group reads as one rounded card without a mask spanning the whole
+        // list. A container-height `clipShape` is a render-side cost that grows
+        // with the row count and forces the list to be laid out as one unit.
+        if needsCornerClip {
+            // A single-row list is both first and last, so it keeps all four
+            // corners rounded.
+            row(content)
+                .clipShape(
+                    .rect(
+                        topLeadingRadius: isFirst ? CommonListContainer.cornerRadius : 0,
+                        bottomLeadingRadius: isLast ? CommonListContainer.cornerRadius : 0,
+                        bottomTrailingRadius: isLast ? CommonListContainer.cornerRadius : 0,
+                        topTrailingRadius: isFirst ? CommonListContainer.cornerRadius : 0
+                    )
+                )
+        } else {
+            row(content)
+        }
+    }
+
+    private func row(_ content: Content) -> some View {
         VStack(spacing: 0) {
             content
             Separator(color: Theme.colors.borderLight, opacity: 1)
                 .showIf(!isLast)
         }
-        // The end rows carry the surface's corner radius themselves, so the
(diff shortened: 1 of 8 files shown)


johnnyluo (approved):
LGTM

Issue #4977, opened 2026-07-28
[Fix] UTXO set is never re-validated between signing and broadcast [Send / Keysign]


gastonm5:
## Problem

The UTXO set for a send is selected when the keysign payload is built and never re-validated before broadcast. On MPC that gap is minutes wide while devices pair and sign, and Blockchair's dashboard response is itself cached for 60s — so even a complete, correctly-filtered snapshot can be stale by the time the transaction goes out.

The result is the same user-visible failure as #4970: the transaction is signed, broadcast, and rejected with `bad-txns-inputs-missingorspent` because an input was spent in the interval.

Split out of #4970, which fixes the *completeness and filtering* of the snapshot (pagination to `unspent_output_count`, plus confirmation/spendability filtering). That reduces the failure rate substantially but cannot eliminate this window.

## Expected Behavior

A transaction is not broadcast against a UTXO set that is known to be stale, and the user gets a clear, recoverable error rather than a network rejection.

## Solution

Re-fetch the UTXO set immediately before broadcast and compare it against the inputs the transaction actually spends.

The design question this needs answered before implementation — and the reason it is not folded into #4970:

**What happens when the set has changed after signing?** The transaction is already signed against specific inputs; it cannot be silently amended. Realistic options:

- Abort before broadcast with a clear error and return the user to the send screen to rebuild and re-sign. Safe, but costs a full MPC ceremony.
- Abort only when a *spent* input is detected, and broadcast otherwise (new UTXOs appearing is harmless — the tx does not reference them).
- Warn and let the user choose.

The second is probably the right default: the only fatal change is an input that has disappeared. But it needs a decision, and the check must not itself become a broadcast-blocking failure mode when the provider is briefly unavailable — failing closed on a provider hiccup would be worse than the bug.

### Must Do
- [ ] Re-fetch the UTXO set immediately before broadcast on the UTXO chains
- [ ] Compare against the inputs the signed transaction spends; detect any that are gone
- [ ] Decide and implement the on-change behaviour (see above)
- [ ] Make provider unavailability at this point non-fatal — do not turn a transient fetch failure into a failed send

### Must NOT Do
- Don't re-sign or amend a signed transaction automatically
- Don't duplicate #4970's pagination/filtering work — build on it

## Acceptance Criteria
- [ ] Build succeeds
- [ ] A send whose input is spent between signing and broadcast fails with a clear, actionable error instead of a network rejection
[...]

PR #4978, opened 2026-07-28
fix(utxo): fetch the complete UTXO set, spend only what is spendable, and show that as the balance

### LINKED ISSUE #4970: BTC UTXO fetch capped at 100 and unfiltered — parity with SDK/Windows (mirrors vultisig-android#5433)
## Context

A user sending 0.2 BTC and 0.4 BTC on Android had the transaction signed, broadcast, and rejected by the network with `bad-txns-inputs-missingorspent` — at least one selected input was no longer in the UTXO set. Root cause is an incomplete/stale UTXO snapshot. iOS carries the same defect, so filing here for parity. Android tracked in vultisig/vultisig-android#5433.

## Cause

`VultisigApp/VultisigApp/Blockchain/UTXO/Service/UTXOAPI.swift:50` fetches UTXOs with:

```
GET /blockchair/{chain}/dashboards/address/{address}
```

No `limit`, no `offset`, no pagination — and no `state` parameter either. Blockchair's default caps the `utxo` array at **100 entries**, returned **newest-first**. Probed live against an address with 696 unspent outputs:

```
real chain tip       : 959953
blockchair state     : 959953        # provider is AT the tip — not indexer lag
utxo returned/total  : 100 / 696     # 596 UTXOs invisible to the app
utxo block_id range  : 959942 -> 959953   # newest ~11 blocks only
context.cache        : duration 60s
?limit=200           : returns 200   # the cap is purely a missing parameter
```

Two further gaps:

1. `VultisigApp/VultisigApp/Blockchain/UTXO/Models/Blockchair/Blockchair.swift:29-41` — `BlockchairAddress` does not decode `unspent_output_count` at all, so iOS has no way to detect that the list was truncated.
2. `VultisigApp/VultisigApp/Features/Keysign/Services/KeysignPayloadFactory.swift:179-201` — UTXOs are mapped and then filtered only by dust threshold. No confirmation (`block_id > 0`) or spendability (`is_spendable`) filter.

The UTXO set is also selected at build time and never re-validated before broadcast. On MPC that gap is minutes wide while devices pair and sign.

## Why it tracks with amount

A small send is satisfied by one or two visible UTXOs. 0.2–0.4 BTC needs dozens of inputs drawn from a recency-biased 100-entry slice, on a snapshot up to 60s old. Every extra input is another chance one of them has been spent since.

## Reference implementation

Windows / desktop / extension already do this correctly via the SDK:

- `packages/core/chain/chains/utxo/client/getUtxoAddressInfo.ts:29-65` — `?limit=1000&offset=N`, loops until `utxo.length >= unspent_output_count`, and **throws** if it cannot retrieve them all.
- `packages/core/chain/chains/utxo/tx/getUtxos.ts:21` — filters `value > minUtxo[chain] && is_spendable !== false && block_id > 0`.

| Platform | Pagination | Truncation detectable | Filter |
|---|---|---|---|
| Windows / extension (SDK) | yes, loops to `unspent_output_count`, throws if short | yes, enforced | dust + `is_spendable` + `block_id > 0` |
| **iOS** | **no — capped at 100** | **model does not decode `unspent_output_count`** | dust threshold only |
| Android | no — capped at 100 | parses it, never checks it | none on the BTC path |

## Fix

1. Decode `unspent_output_count`, then paginate with `limit`/`offset` until the full set is retrieved; fail loudly if incomplete rather than silently building from a partial view.
2. Add the SDK's confirmation/spendability filter alongside the existing dust filter.
3. Re-fetch and validate the UTXO set immediately before broadcast, not only at build time.

## Verification

- Confirm behaviour matches Windows/extension on the same vault: same UTXO count retrieved, same set after filtering.
- Test with a vault holding >100 UTXOs — that is the case that currently breaks and cannot be caught by a small-balance test vault.
- Fixes should land consistently across iOS, Android (vultisig/vultisig-android#5433) and the SDK reference behaviour.

gastonm5:
## Description

Fixes #4970

UTXOs were fetched with no pagination, so Blockchair's default capped the list at **100 entries, newest-first**, and they were filtered only by dust. A large send then drew dozens of inputs from a recency-biased partial snapshot and could be broadcast spending an already-spent output → `bad-txns-inputs-missingorspent`. Reported on Android with real funds; iOS had the identical defect.

Three things land here: **complete the snapshot**, **filter it correctly**, and **make the displayed balance equal what is actually spendable**.

### 1. Pagination (`c6e8a6a29`, `3d9da2081`)

`unspent_output_count` and `is_spendable` are now decoded, and the dashboard is walked with `limit`/`offset` until the set is provably complete. Fails loudly rather than silently building from a partial view. Page cap is 20 × 1000 — a standard transaction tops out near ~1400 P2WPKH inputs at the 100 kvB standardness limit, so ~20 000 candidates is far beyond anything spendable, and the cap is really bounding worst-case latency.

Two findings worth calling out, because the obvious implementation is wrong in both:

- **Blockchair applies `limit`/`offset` to `transactions,utxo`, not just `utxo`.** `?limit=5` returns `context.limit: "5,5"` and five *transactions* alongside five UTXOs. A naive `?limit=1000` therefore pulls 1000 transaction hashes per page as a side effect, ×20 pages, on the send path. We pin the transactions side to zero: `limit=0,1000`. Verified accepted on bitcoin, litecoin, dogecoin, bitcoin-cash and zcash.
- **"Loop until `utxo.count >= unspent_output_count`" is unsafe** — and it is what the SDK does. The count is read once off page 0; if it is stale or under-reports, `>=` terminates the walk on a *full* page and hands back exactly the truncated set this issue is about. The walk now terminates only on a **short page**, with the count as an arrival check plus a stability check, since a mid-walk removal shifts every subsequent offset.

Verified end-to-end against a real 1313-UTXO Dogecoin address: 1000 + 313 = 1313 unique, zero overlap.

### 2. Spendability filter (`6e1b30d64`, `1bfc9042c`)

Confirmation and spendability filters now sit alongside the dust threshold — but **not** as a blanket "confirmed only".

A blanket exclusion has a worse bug in it: your change output is unconfirmed for ~10 minutes after every send, so spending 0.1 BTC from a 1.0 BTC UTXO would leave you unable to touch the 0.9 change, on the common path. Instead:

> **spendable = confirmed + our own unconfirmed change; a stranger's zero-conf excluded.**

That is Bitcoin Core's actual policy. We can implement it because the app already persists every transaction it broadcasts (`StoredPendingTransaction`), so a mempool UTXO whose `transaction_hash` is one of ours is provably our own change — the SDK cannot distinguish these and so excludes both.

Ownership is scoped by chain **and** vault. Two residual costs are documented at the call site: a parent evicted between selection and broadcast takes its child with it (a rejected broadcast, not lost money — unlike a stranger's zero-conf, where the counterparty can double-spend deliberately), and a long chain of unconfirmed sends eventually hits the network's 25-ancestor mempool limit.

### 3. Balance == spendable (`7f3d4bf25`, `1aa697ebb`, `6e41b0a5a`)

`BalanceService` read `address.balance`, which **includes** unconfirmed — so the displayed balance could exceed what was spendable, and a send could pass `validateBalanceWithFee` and then fail at `notEnoughUTXOError`. It now sums exactly the filtered set, through a **single shared predicate** (`SpendableUtxos`) that both the balance path and selection use, so the two cannot drift apart again. A test pins `balance == sum(spendable)`.

### 4. What that costs on the receiving side, and the pending line (`e94b806d1`)

The predicate in §2 is asymmetric on purpose — our own zero-conf in, a stranger's out — and the cost lands on the **receiving** side: an inbound payment still in the mempool no longer moves the balance, where `address.balance` used to include it. That is a real UX change and was not called out in the first version of this description.

It is worse than "the balance doesn't move for ten minutes": **nothing** happened. The local history records only transactions this device broadcast, so an inbound receive has never had a row there — the balance changing *was* the entire notification that a payment had arrived.

So the arriving amount is now a separately-labelled line under the balance on the coin screen, and deliberately **not** folded into it. Folding it in would put back the exact gap §3 closes — a send would clear the balance check and then die at input selection, because a stranger's zero-conf is not something this wallet can build an input from.

`SpendableUtxos.pendingInboundBalance` is `select`'s complement on **one** clause: same rows, same dust threshold, same `is_spendable` gate, confirmation flipped. The two therefore **partition** the address's outputs — the same money cannot appear twice, our own unconfirmed change stays spendable rather than "arriving", and an inbound output that is sub-dust or explicitly unspendable is not promised (it will not be spendable after confirmation either). Both numbers come from one snapshot in one call (`BalanceSnapshot`), so they cannot drift. Copy is localized across all 8 locales.

## Which feature is affected?
- [x] Sending  - Please attach a tx link here

## Parity

- **Android:** `linked: vultisig-android#5433` — same pagination/filter defect, already tracked.
- **Windows + extension:** `first: iOS is first to align displayed balance with spendable.` **Not parity — deliberately ahead, because the SDK behaviour is itself buggy.** `getUtxoCoinBalance` is `bigIntSum(getUtxos(...))` and `getUtxos` filters `block_id > 0 && is_spendable !== false && value > minUtxo`, so Windows/extension (a) display 0.00 after spending a whole UTXO until the change confirms, and (b) omit sub-dust holdings from the balance entirely. Copying it would have imported both. **Unfiled** — flagging here so it is on the record rather than rediscovered.
- **SDK:** `first:` as above. The pagination termination condition (`utxo.length >= unspent_output_count`) is also unsafe for the reason given in §1, and the `?limit=1000` two-collection issue applies there too.

## Checklist

## Additional context

**Wider blast radius than the issue title suggests.** `BlockchairService` builds its path from `coin.chain.name.lowercased()`, so this path serves **BTC, BCH, LTC, DOGE and ZEC** — not just Bitcoin. It also backs `BlockchairService+QBTCClaim` (claimable-UTXO enumeration, likewise capped at 100) and **BTC swaps**, which reach `selectUTXOs` via `SwapPayloadBuilder → buildTransfer` with no fetch of their own — so swaps were building from whatever truncated snapshot the last balance poll happened to leave.

**Dash is deliberately untouched.** Its UTXOs come from `getaddressutxos`, an address index built from connected blocks and blind to the mempool in both directions. Summing it for balance would make the wallet keep showing an input its own pending send had already consumed. Separately noted while here: that same blindness means a *chained* Dash send can double-spend and be rejected at broadcast — pre-existing, unrelated to this change, unfiled.

**Fixed in passing:** `savePendingTransaction()` never persisted `pubKeyECDSA` (vault-scoping the ownership check was impossible until it did), and `BlockchairAddress.scriptHex` had no `CodingKeys` so it looked for a key Blockchair never sends — silently `nil` since forever (unused, so no behaviour change).

**Known and deferred, documented in code:** `cleanupOld` drops terminal rows after 24 h, so a transaction still unconfirmed by then loses its exemption; legacy rows with `pubKeyECDSA == nil` can never acquire an owner (degrades to the previous confirmed-only behaviour, resolves in one confirmation). Pre-broadcast revalidation is split to #4977 — this reduces the failure rate substantially but Blockchair's own 60 s cache means it cannot eliminate the window.

**Review round 1 (NeOMakinG).** Four of the five threads produced code. Dash no longer paginates a UTXO array nothing on its path reads (`2fe591a16`) — it was paying up to 20 sequential round trips *and* three new completeness failure modes on the balance refresh, `validateUtxosIfNeeded` and the fee estimate, for a number that comes from `address.balance`; it now issues one `limit=0,0` request, verified live against the proxy. Shape-regressed UTXO rows are counted and logged instead of silently understating the balance (`88faaefff`) — every field on `BlockchairUtxo` is optional, so such a row decodes cleanly and is then dropped in silence; the count of rows that cannot identify an output has no legitimate non-zero value, unlike the gap between `address.balance` and the spendable sum. A failed pending-transaction read now propagates rather than reading as `0` (`d2be3eb0f`), which in the post-send state was the one answer guaranteed to be wrong. Plus §4 above.

The fifth — falling back to the retrieved set on the balance path — is **rejected**: a fetch failure already keeps the last known balance rather than blanking the wallet (`fetchBalanceUpdate` yields `rawBalance == nil`, `hasUpdates` is false, `applyBalanceUpdates` skips the coin), and a fallback would show a number the send path still refuses to fund.

**Telemetry, honestly.** Every completeness throw logs its numbers, but `Log.*` resolves to `Logger(.disabled)` in Release — the `LOGGING` compilation condition is Debug-only — and the app ships no remote analytics. These are developer-build diagnostics, not field signal. Flagging rather than letting "we log it" stand in for "we would see it".

**Two more deferrals worth stating.** A constant `unspent_output_count` does not *prove* the offset walk saw one coherent set: a simultaneous add and remove between pages keeps the count unchanged. That is the same window `#4977` exists to close, and Blockchair's own 60 s cache bounds what any fetch-side fix can achieve. Separately, `DefaultSendInteractor.validateUtxosIfNeeded` rewrites every UTXO fetch failure as "check your internet connection", which is now misleading for a deterministic provider-data failure — pre-existing copy, and fixing it means new user-facing strings in 8 locales, so it is left alone here.

- **Agent**: Claude Code
- **Issue**: #4970
- **Confidence**: high on the pagination and the balance/spendable equality, both of which are directly tested; medium on the unconfirmed-change ownership policy, which is a deliberate divergence from the SDK and the judgement call most worth a second opinion
- **Needs human review for**: the ownership policy in §2, and **on-device verification against a vault with >100 UTXOs** — the case that currently breaks and that a small test vault cannot exercise

## Summary by CodeRabbit

* **New Features**
  * Added pending inbound balance tracking and a new “pending incoming balance” row in coin details.
  * Improved UTXO balance accuracy using spendable outputs plus wallet-owned unconfirmed change.
  * Added paginated UTXO retrieval with duplicate prevention and completeness checks.
* **Bug Fixes**
  * Prevented unverifiable/partial provider UTXO data from producing misleading balances.
  * Improved Blockchair decoding and spendability handling; fixed Ripple native vs issued balance routing.
* **Tests**
  * Added/expanded coverage for pagination, decoding, spendable selection, pending attribution, and balance consistency.
* **Documentation**
  * Added localized “pending incoming balance” strings and clarified UI display behavior.


VultisigApp/VultisigApp/Blockchain/UTXO/Service/BlockchairService.swift
```diff
@@ -40,12 +63,136 @@ actor BlockchairService {
...
+            }
+
+            // Only the end of the list ends the walk. The count is read once,
+            // off the first page, so it can be stale or under-report what the
+            // provider is actually holding; stopping on it while a page came
+            // back full is how a recency-biased partial set would slip through
+            // again. It is the check applied on arrival, not the terminator.
+            guard isLastPage else { continue }
+
+            if utxos.count < expected {
```
NeOMakinG, line 180: `should-fix:` Fail-loud completeness contract rests entirely on Blockchair's unspent_output_count exactly matching the paged utxo array length — `if utxos.count < expected { throw .incompleteUtxoSet }` (pinned by testThrowsWhenRetrievedCountFallsShortOfReportedCount) is correct as written, but if Blockchair ever over-reports the count relative to what it actually serves for an address (e.g. it counts a mempool-spent output the utxo array already omits, or count and array are computed from snapshots microseconds apart under churn), that address permanently fails every balance refresh and every send. This is fail-safe (last-known balance kept, no bad tx) but a hard availability brick for the affected user until the provider's numbers reconcile. Verified against one real 1313-UTXO DOGE address at one instant — not proof this never happens under provider churn.

fix: Emit telemetry on every incompleteUtxoSet/utxoSetChangedWhilePaging throw so a systematic provider divergence is visible before it becomes a support flood. Consider whether the balance-display path specifically could fall back to the retrieved set rather than throwing (a slightly-short balance is a softer failure than a blank wallet), while keeping the send path fail-loud.

VultisigApp/VultisigApp/Core/Services/BalanceService.swift
```diff
@@ -469,9 +483,44 @@ class BalanceService {
...
+                chain: coin.chain,
+                vaultPubKeyECDSA: vaultPubKeyECDSA
+            )
+            return try Self.utxoSpendableBalance(
+                from: blockChairData,
+                dustThreshold: coin.coinType.getFixedDustThreshold(),
+                ownUnconfirmedTxHashes: ownUnconfirmed
+            )
+
+        case .dash:
```
NeOMakinG, line 598: `should-fix:` Dash balance now routes through full UTXO-array pagination it never reads — Dash's balance only consumes blockChairData.address?.balance?.description (line 525) — it never reads the utxo array. But it still routes through fetchBlockchairData → fetchBlockchairResponse, which now fully paginates the utxo array and can throw incompleteUtxoSet/unverifiableUtxoSet/utxoSetChangedWhilePaging. A large or churning Dash address gains a brand-new balance-refresh failure surface (plus up to 20 sequential round trips of latency) for a value that doesn't depend on the utxo set at all.

fix: Fetch Dash's balance with a minimal/unpaged request, or catch the pagination-completeness errors specifically on the Dash branch and still return address.balance.

VultisigApp/VultisigApp/Core/Services/BalanceService.swift
```diff
@@ -534,6 +583,55 @@ class BalanceService {
...
+    /// outputs are all present and all filtered out is a different case
+    /// entirely and does legitimately read zero.
+    static func utxoSpendableBalance(
+        from blockchair: Blockchair,
+        dustThreshold: Int64,
+        ownUnconfirmedTxHashes: Set<String>
+    ) throws -> String {
+        let rows = blockchair.utxo ?? []
+
+        if rows.isEmpty, let reportedBalance = blockchair.address?.balance, reportedBalance > 0 {
```
NeOMakinG, line 717: `should-fix:` Malformed utxo rows silently understate balance without tripping the completeness guard — The self-contradiction guard only fires on rows.isEmpty && reportedBalance > 0. A non-empty utxo array where some rows are shape-regressed (missing transaction_hash/index/value) slips past it — SpendableUtxos.select silently drops those rows and the displayed balance understates real funds without failing loud. Doesn't break the balance==spendable invariant (a malformed row can't be selected as a send input either, so both sides drop it equally) and can't cause a send to over-draw, which is why this is should-fix not blocking.

fix: Add telemetry when sum(spendable) diverges sharply from reportedBalance on a non-empty utxo array, to surface a decode regression that's currently invisible. A hard throw on any divergence would false-positive constantly since reportedBalance legitimately includes stranger zero-conf and sub-dust.

VultisigApp/VultisigApp/Blockchain/UTXO/SpendableUtxos.swift
```diff
@@ -0,0 +1,139 @@
...
+                  let value = row.value, let amount = Int64(exactly: value), amount >= dustThreshold,
+                  let index = row.index, let vout = UInt32(exactly: index)
+            else {
+                return nil
+            }
+
+            // A row whose depth we could not establish is treated as
+            // unconfirmed: absent evidence of a block is not evidence of one.
+            let isConfirmed = (row.blockId ?? 0) > 0
+            guard isConfirmed || ownHashes.contains(txHash.lowercased()) else {
```
NeOMakinG: `q:` Inbound unconfirmed receives no longer show up in the displayed balance at all — product-signed-off? — Correctly excludes a stranger's inbound zero-conf from the spendable set — but because the displayed balance is now sum(spendable), a user who just received BTC/DOGE and watches their balance not move for ~10 min until confirmation is a plausible support ticket. The PR body covers the outbound change-confirmation case well; the inbound-receive UX tradeoff isn't called out.

fix: Confirm with product whether hiding unconfirmed inbound balance entirely (vs. showing it as a separate 'pending' line) is the intended UX.

VultisigApp/VultisigApp/Core/Services/TransactionStatus/PendingTransactionStorage.swift
```diff
@@ -91,6 +108,65 @@ final class StoredPendingTransactionStorage {
...
+        }
+
+        do {
+            let unconfirmed = try modelContext.fetch(FetchDescriptor(predicate: predicate))
+            return Set(
+                unconfirmed
+                    .filter { $0.chain == chain && $0.pubKeyECDSA == vaultPubKeyECDSA }
+                    .map(\.txHash)
+            )
+        } catch {
```
NeOMakinG, line 174: `q:` SwiftData read failure collapses to confirmed-only right when own-unconfirmed-change is the wallet's only balance — unconfirmedTransactionHashes catches a SwiftData read failure and returns [], the documented safe degradation to confirmed-only. But in the specific post-send state where the wallet's only funds are its own unconfirmed change, a transient storage read failure would momentarily read the balance as 0 and block the follow-up send — the exact failure this lookup exists to prevent, just triggered by a storage hiccup instead of a provider one.

fix: Judgment call, not a defect: would letting the read error propagate up the balance path (preserving the prior known balance) be strictly safer than collapsing to confirmed-only here?

NeOMakinG (dismissed):
gm — multi-lane pass: claude-correctness + codex-adversarial. conceptual-only (no runtime this pass). requesting changes.

## should-fix
### 1. Fail-loud completeness contract rests entirely on Blockchair's unspent_output_count exactly matching the paged utxo array length — `VultisigApp/VultisigApp/Blockchain/UTXO/Service/BlockchairService.swift:164`
`if utxos.count < expected { throw .incompleteUtxoSet }` (pinned by testThrowsWhenRetrievedCountFallsShortOfReportedCount) is correct as written, but if Blockchair ever over-reports the count relative to what it actually serves for an address (e.g. it counts a mempool-spent output the utxo array already omits, or count and array are computed from snapshots microseconds apart under churn), that address permanently fails every balance refresh and every send. This is fail-safe (last-known balance kept, no bad tx) but a hard availability brick for the affected user until the provider's numbers reconcile. Verified against one real 1313-UTXO DOGE address at one instant — not proof this never happens under provider churn.
action: Emit telemetry on every incompleteUtxoSet/utxoSetChangedWhilePaging throw so a systematic provider divergence is visible before it becomes a support flood. Consider whether the balance-display path specifically could fall back to the retrieved set rather than throwing (a slightly-short balance is a softer failure than a blank wallet), while keeping the send path fail-loud.

### 2. Dash balance now routes through full UTXO-array pagination it never reads — `VultisigApp/VultisigApp/Core/Services/BalanceService.swift:514`
Dash's balance only consumes blockChairData.address?.balance?.description (line 525) — it never reads the utxo array. But it still routes through fetchBlockchairData → fetchBlockchairResponse, which now fully paginates the utxo array and can throw incompleteUtxoSet/unverifiableUtxoSet/utxoSetChangedWhilePaging. A large or churning Dash address gains a brand-new balance-refresh failure surface (plus up to 20 sequential round trips of latency) for a value that doesn't depend on the utxo set at all.
action: Fetch Dash's balance with a minimal/unpaged request, or catch the pagination-completeness errors specifically on the Dash branch and still return address.balance.

### 3. Malformed utxo rows silently understate balance without tripping the completeness guard — `VultisigApp/VultisigApp/Core/Services/BalanceService.swift:620`
The self-contradiction guard only fires on rows.isEmpty && reportedBalance > 0. A non-empty utxo array where some rows are shape-regressed (missing transaction_hash/index/value) slips past it — SpendableUtxos.select silently drops those rows and the displayed balance understates real funds without failing loud. Doesn't break the balance==spendable invariant (a malformed row can't be selected as a send input either, so both sides drop it equally) and can't cause a send to over-draw, which is why this is should-fix not blocking.
action: Add telemetry when sum(spendable) diverges sharply from reportedBalance on a non-empty utxo array, to surface a decode regression that's currently invisible. A hard throw on any divergence would false-positive constantly since reportedBalance legitimately includes stranger zero-conf and sub-dust.

## suggestion
### 1. Completeness check can't fire if Blockchair self-contradicts on page 0 (count:0, non-empty utxo array) — `VultisigApp/VultisigApp/Blockchain/UTXO/Service/BlockchairService.swift`
The completeness check (utxos.count < expected) never fires when expected reads as 0, so a page-0 response reporting unspent_output_count: 0 while its own utxo array for that page is non-empty would be accepted. Distinct from the already-handled 'count moves between pages' defect — this would need the provider to lie about zero on the very first read. Purely theoretical; not something to hold the PR for.
action: Optional: guard against expected == 0 with a non-empty first page as an additional unverifiableUtxoSet trigger, if worth the complexity.

## q
- Inbound unconfirmed receives no longer show up in the displayed balance at all — product-signed-off?
- SwiftData read failure collapses to confirmed-only right when own-unconfirmed-change is the wallet's only balance

## risk
Low. Change is strictly tighter than the pre-fix send path, so fund-safety can only improve. All new error paths are fail-safe: balance keeps last-known value on throw, swap sign-time check fails closed, ownership lookup degrades to confirmed-only. Residual risk is availability, not funds — the should-fix items describe conditions under which a balance refresh or send could fail-safely brick for an address whose Blockchair unspent_output_count diverges from its served rows. Blast radius spans BTC/BCH/LTC/DOGE/ZEC balance+send+swap and the QBTC-claim enumeration; Dash balance is newly gated on utxo-array pagination it doesn't need.

## verdict
REQUEST CHANGES

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` conceptual pass.

VultisigApp/VultisigApp/Blockchain/UTXO/Service/BlockchairService.swift
gastonm5: Good challenge — I dug into both halves and only one of them survives.

**Telemetry: already there, but I won't oversell it.** Every throw site logs the numbers before throwing — `utxoPageMissingAddress`, `utxoSetChangedWhilePaging`, `unverifiableUtxoSet`, `incompleteUtxoSet` and `utxoPageLimitExceeded` in `BlockchairService`, plus `utxoBalanceWithoutOutputs` in `BalanceService` — and `fetchBalanceUpdate` logs the failure again per coin. What I can't claim is that this gives us early warning: `Log.*` resolves to `Logger(.disabled)` in Release, because the `LOGGING` compilation condition is Debug-only (`project.yml`), and the app ships no remote analytics of any kind. So these are developer-build diagnostics, not a signal that reaches us before a support flood. Wiring a telemetry sink for a wallet's address-level fetch failures is a privacy + infrastructure decision well outside this PR — I'd rather flag that honestly than let "we log it" stand in for "we'd see it".

**Balance-path fallback: rejecting, on two counts.**

1. *The premise doesn't hold — this does not blank the wallet.* A throw on the balance path is caught in `BalanceService.fetchBalanceUpdate`, which yields a `CoinBalanceUpdate` with `rawBalance == nil`; `hasUpdates` is then false and `applyBalanceUpdates` skips the coin entirely. **The last known balance stays on screen.** The affected user sees a balance that has stopped updating and sends that fail — bad, but not a blank wallet. That's pinned by `testUpdateWithoutRawBalanceReportsNothingToWrite`. The only case that reads `0` is an address that has never had one successful fetch (a fresh import), and there a partial sum would be a wrong number of unknown size rather than a soft one.
2. *It would reintroduce the bug §3 removes.* The send path still throws, so the screen would show a number the send flow refuses to fund — the "passes `validateBalanceWithFee`, dies at `notEnoughUTXOError`" failure, back again by construction.

I also checked whether the guard could be relaxed narrowly — accept a short set when the walk ended on a **short page**, since with only one page there was no offset window for entries to slide through. It can't: a provider that silently caps the array below the requested `limit` also produces a short page, and accepting that is exactly the original bug. So there's no principled middle ground here that isn't just "trust a truncated list".

Where this thread *did* land something real is the next one down: your Dash observation was right, and it's fixed in `2fe591a16`.

VultisigApp/VultisigApp/Core/Services/BalanceService.swift
gastonm5: Valid, and worse than described — fixed in `2fe591a16`.

It isn't only the balance refresh. `DefaultSendInteractor` calls `fetchBlockchairData` for **every** chain with `chainType == .UTXO`, twice on the Dash send screen: `validateUtxosIfNeeded` and the fee estimate. So a large or churning Dash address was paying up to 20 sequential round trips *and* carrying three new failure modes (`incompleteUtxoSet`, `unverifiableUtxoSet`, `utxoSetChangedWhilePaging`) in three places, all for an array nothing on the Dash path reads.

Took your first option rather than the catch, since it removes the latency too. `BlockchairService.fetchBlockchairResponse` now short-circuits for chains whose spendable set it doesn't serve and issues a single `limit=0,0` request. Probed live against the proxy on a real Dash address before writing it:

```
GET /blockchair/dash/dashboards/address/XnT33…?limit=0,0
→ context.limit "0,0"; address.balance 73832101404000, unspent_output_count 8; utxo [] ; transactions []
```

The aggregate is all there; only the arrays are empty. Everything else is unchanged — `fetchBlockchairData` still validates the address key and still caches, so error semantics for Dash are what they were before this PR.

The `.dash` comment in `BalanceService` now says why the fetch is unpaged as well as why the balance comes from Blockchair.

Tests: `testDashRequestsTheAddressAggregateWithoutPagingAnyUtxoRows` (exactly one request, `(limit: 0, offset: 0)`), `testDashBalanceSurvivesAResponseThatTripsTheCompletenessGuards` (an address claiming 5000 unspent outputs and serving 3 — Dash returns its balance, the same response on Bitcoin still throws `incompleteUtxoSet`), and `testOnlyDashSkipsTheUtxoWalk` so the exemption can't quietly widen.

VultisigApp/VultisigApp/Core/Services/BalanceService.swift
gastonm5: Agreed and implemented in `88faaefff` — though not quite as suggested, because checking where the drop actually happens changed the answer.

Every field on `BlockchairUtxo` is optional, so a shape-regressed row **decodes cleanly**; it isn't dropped by the decoder, it reaches `SpendableUtxos.select` and is dropped there. That means we can detect it exactly rather than infer it. So instead of a threshold on `sum(spendable)` vs `reportedBalance`, the shape check is now split out from the policy filters and the count of rows that cannot identify an output at all — no `transaction_hash`, no `index`, no `value`, or an index that doesn't fit a `vout` — is logged.

That's strictly better than the divergence heuristic for the reason you already named, taken one step further: the gap between `address.balance` and the spendable sum is legitimately non-zero on **any** address holding a stranger's zero-conf or a sub-dust output, so any threshold on it is either noisy or too loose to catch anything. The unusable-row count has no legitimate non-zero value at all, so `> 0` is the whole condition.

Logged, not thrown, as you said — a malformed row can't be selected as an input either, so it can never cause a send to over-draw.

Two tests pin the classification rather than the log line: `testCountsRowsThatCannotIdentifyAnOutputAtAll` and `testAPolicyExclusionIsNotCountedAsMalformed` (dust / `is_spendable: false` / a stranger's zero-conf are all well-formed and must not fire it).

One caveat I'd rather state than leave implied: like every log in this app it compiles out in Release (`LOGGING` is Debug-only, and there's no analytics sink) — see my reply on the completeness thread. This makes the regression *findable* in a developer or QA build instead of invisible; it does not make it reportable from the field.

VultisigApp/VultisigApp/Blockchain/UTXO/SpendableUtxos.swift
gastonm5: You were right that the tradeoff was uncalled-out, and right that it needed a product answer rather than mine. I took it to Gaston; his call was **add a separate pending line**, so it's built rather than deferred — `e94b806d1`.

First, what a user saw before this commit, since it's worse than "the balance doesn't move": **nothing at all happened.** The local transaction history records only transactions this device broadcast (`TransactionHistoryRecorder.recordSend`), so an inbound receive has never had a row there; the balance changing *was* the entire notification of a payment arriving. With the balance now strictly `sum(spendable)`, an inbound zero-conf produced no visible change anywhere in the app until the next block.

What's there now:

- The headline balance is **unchanged** — still strictly `sum(spendable)`. The filter is untouched and the `balance == spendable` invariant is intact; that was the non-negotiable part.
- The arriving amount is a separate, separately-labelled line under it (`+0.05 BTC pending`), absent entirely when nothing is arriving, and hidden along with everything else when balances are hidden.
- It is computed from the **same rows in the same call** (`BalanceSnapshot`), never a second fetch. `SpendableUtxos.pendingInboundBalance` is `select`'s complement on exactly one clause — same dust threshold, same `is_spendable` gate, same shape requirements, confirmation flipped — so the two **partition** the address's outputs. The same money can't appear in both, and our own unconfirmed change (already spendable, already in the balance) is explicitly not counted as pending.
- An inbound output that's sub-dust or explicitly unspendable isn't reported as arriving: it won't be spendable after confirmation either, and promising it would be worse than staying silent.
- Copy is localized across all 8 locales, `ko` included.

Tests: `testAnInboundZeroConfIsReportedAsPendingAndNotAsBalance`, `testOurOwnUnconfirmedChangeIsSpendableAndNeverPending`, `testSpendableAndPendingNeverOverlapAndNeverDoubleCount`, `testAnAddressWithNothingArrivingReportsNoPendingAmount` (which is also what makes the line disappear when the payment confirms), plus the partition/gate cases in `SpendableUtxosTests`.

Scoped to the coin screen. It's new UI in an otherwise-correctness PR, so if you'd rather it shipped separately it's a clean single commit to split out.

VultisigApp/VultisigApp/Core/Services/TransactionStatus/PendingTransactionStorage.swift
gastonm5: Traced it properly, and you're right — propagating is strictly safer. Changed in `d2be3eb0f`.

What each option actually does:

- **Balance path.** Returning `[]` computes a *real* balance from a false premise, and in the post-send state that premise is exactly wrong: the input is gone from the UTXO set and the change is unconfirmed, so the answer is `0` and it gets persisted. Throwing instead means `fetchBalanceUpdate` captures the error, produces an update with `rawBalance == nil`, `hasUpdates` is false, and `applyBalanceUpdates` skips the coin — the **last known balance stands**. Strictly better: a stale number instead of a wrong one.
- **Send path.** Confirmed-only is safe in the sense that it can never over-draw, but it fails as `notEnoughUTXOError` — "you don't have the funds", when the funds are there and the lookup broke. Thrown, the failure surfaces as itself.
- **`refreshSpendableBalanceOrThrow`** (the swap sign-time funds check) already rethrows by design and its caller catches, blocks navigation and shows an alert. Failing closed there is what that function is for, so no unrelated caller breaks.

It also removes an inconsistency: every other read on `StoredPendingTransactionStorage` (`get`, `getAllPending`, `save`, `cleanupOld`) already throws. This was the only one that answered a failed question with a confident "nothing".

One honest limitation on the test. I couldn't synthesize a SwiftData fetch failure — I probed it directly, and fetching a model absent from the container's schema returns 0 rows rather than throwing, so there's no deterministic way to force the error path from a unit test. What is pinned is the mechanism that makes throwing the safe choice: `testUpdateWithoutRawBalanceReportsNothingToWrite` now asserts both halves — an update with no balance is skipped (the previous balance survives) and a genuine `0` still applies (an emptied wallet still renders as empty).

Residual, and I think acceptable: if the read failed *persistently* the balance would freeze rather than degrade to confirmed-only. A SwiftData store that can't be read is failing `Storage.shared.save()` too, and a frozen balance plus a hard send error is far more likely to be reported than a silently-wrong number.

johnnyluo (approved):
LGTM

PR #4980, opened 2026-07-29
fix(tokens): rank the provider catalogs, bound browse to the top 20 each, unbound search

### LINKED ISSUE #4979: [Fix] Token picker browse dumps the whole provider catalog, and search silently renders only 20 matches [Wallet / Tokens]
## Problem

Two separate defects on the add-token picker, both made materially worse by the dynamic token catalog (#4941, shipped in #4954).

**1. Browse is unbounded.** `TokenSelectionViewModel.recompute` puts the *entire* verified provider breadth into `browseProviderTokens`, and `TokenSelectionScreen.elements` renders all of it when the search field is empty. On Ethereum that is ~2,200 tokens from 1inch alone; on Solana, 4,319 from Jupiter. Nobody scans that. And because the providers return no meaningful order, the head of the list is arbitrary — Ethereum currently opens on `$MBG, ZRX, 0x0, 0xBTC`, and Base on `$OLIVIA2.0`.

**2. Search renders at most 20 matches, with no second page.** `TokenSelectionLogic.filteredTokens` ends in `.prefix(20)`. This predates the catalog work and was survivable when the picker held a few dozen curated presets. With thousands of tokens it means searching a common substring like `US` returns 20 arbitrary hits out of several hundred — and the ones it drops include exactly the badged lookalike tokens the unverified surface exists to make visible.

The two interact badly: bounding browse is only safe if search genuinely reaches everything, and today it doesn't.

## Expected Behavior

Browse opens on a short, scannable, *good* list. Search reaches every token in the catalog, unbounded.

## Solution

**Rank in the provider, bound in the browse layer.** Each provider returns its list already ordered best-first; the browse layer takes the top N per `sourceKind`. No score field on `CatalogToken` — ordering carries the ranking, and provider-specific ranking logic stays next to the provider-specific data that justifies it.

- **Jupiter** — rank by `organicScore`. The `/jup/tokens/v2/tag?query=verified` response already carries `mcap`, `liquidity`, `organicScore`, `holderCount`, `fdv`; iOS currently decodes none of them. `organicScore` is present on 100% of the 4,319 tokens where `mcap` is present on 65.8%, and `mcap` ranks badly: tSAT (#2 by cap, $13.4B, **$57K liquidity**) and HONx (#5, $1.76B, **zero liquidity**) outrank ETH and WBTC. `organicScore`'s head is USDC, USDT, SOL, ETH, JitoSOL, JUP, cbBTC, WBTC.
- **1inch** — no ranking field exists on `/tokens`. Rank by 1inch's own tag curation in tiers: `bluechip` → `PEG:*` → `connector` → rest. This is a *quality tier*, not a ranking, and should be commented as such.
- **Search** — remove the cap outright rather than raising it. Any bound reintroduces the same unreachability at a different threshold. The grid is a `LazyVGrid`, so cost is bounded by what's on screen, not by match count.

### Must Do
- [ ] Providers return ranked lists; browse takes top N per `sourceKind` (one named constant, not a literal)
- [ ] Cap **after** the local/hidden exclusion, so a curated or already-held token can't eat a slot
- [ ] Remove `.prefix(20)` from `filteredTokens`
- [ ] Verify the disk cache preserves provider order — if it round-trips through a set, ranking is silently lost on the offline path
- [ ] Test that a token ranked far down is absent from browse but findable by search

### Must NOT Do
- Don't bound search
- Don't change chains that have no remote provider (bundled-only pickers must be untouched)
- Don't reorder the swap picker, which is deliberately balance-ordered

## Acceptance Criteria
- [ ] Build succeeds
- [ ] Ethereum and Base browse open on recognisable majors, not arbitrary tickers
- [ ] A token outside the top N is reachable by search
- [ ] A bundled-only chain's picker is unchanged

gastonm5:
## Description

Fixes #4979

Follow-up to #4954. Two defects on the add-token picker, the second of which predates that PR.

**Browse was unbounded** — the entire verified provider breadth rendered on open: ~2,200 tokens from 1inch on Ethereum, 4,319 from Jupiter on Solana, in whatever order the provider happened to return. Ethereum opened on `$MBG, ZRX, 0x0, 0xBTC`; Base on `$OLIVIA2.0`.

**Search rendered at most 20 matches** — `filteredTokens` ended in `.prefix(20)`, with no second page. That is on `main` today and predates the catalog work; it was survivable with a few dozen curated presets and is not survivable with thousands. Searching a common substring like `US` returned 20 arbitrary hits out of several hundred, dropping exactly the badged lookalikes the unverified surface exists to reveal.

They interact: bounding browse is only safe if search genuinely reaches everything.

Now: **browse = held + curated presets + top 20 per provider; search is unbounded.**

## Ranking — measured, not assumed

Providers return their list already ranked; the browse layer takes the first 20 per `sourceKind`. No score field was added to `CatalogToken` — ordering carries the ranking, so provider-specific logic stays next to the provider-specific data that justifies it.

**Jupiter → `organicScore`, not `mcap`.** The `/jup/tokens/v2/tag?query=verified` response already carried `mcap`, `liquidity`, `organicScore`, `holderCount` and `fdv`; iOS decoded none of them. Across the live 4,319-token response:

| | coverage | top-20 quality |
|---|---|---|
| `mcap` | 65.8% | polluted — **tSAT** #2 ($13.4B cap, **$57K** liquidity), **HONx** #5 ($1.76B, **zero** liquidity), DHRx, PYPLx — all scoring 0, all outranking ETH, WBTC and PYUSD |
| `organicScore` | **100%** | USDC, USDT, SOL, ETH, JitoSOL, JUP, cbBTC, WBTC, PYUSD, mSOL |

92% of the list scores exactly 0, so the sort is stable — the tail does not reshuffle between fetches.

**1inch → tag tiers.** `/tokens` exposes no ranking field at all, so this ranks by 1inch's own curation: `bluechip` → `PEG:*` → `connector` → rest. Commented in the code as a *quality tier*, not a ranking, because we do not have one.

Post-tier distribution (mutually exclusive, best tier wins):

| chain | total | bluechip | PEG | connector | other |
|---|---:|---:|---:|---:|---:|
| Ethereum | 2,225 | 41 | 178 | 806 | 1,200 |
| BSC | 1,596 | 9 | 33 | 461 | 1,093 |
| Arbitrum | 825 | 8 | 63 | 178 | 576 |
| Polygon | 194 | 7 | 30 | 150 | 7 |
| Base | 165 | 9 | 39 | 77 | 40 |

Resulting browse heads: Ethereum `1INCH, AAVE, LINK, COMP, CRV, WBTC`; Base `USDT, USDC, WETH, DAI, EURC, SOL`.

## Correctness

- The cap applies to `browseProviderTokens` **only**. `searchableTokens` keeps the full breadth, with a test pinning that a token outside the top N is absent from browse and findable by search — capping must never be *why* something can't be found.
- Cap is applied **after** the local/hidden exclusion, so a curated or already-held token cannot consume one of the 20 slots.
- Per provider, not global — grouped by `sourceKind`.
- Chains with only `BundledTokensProvider` are unaffected.
- Disk-cache order is preserved, so the ranking survives the offline path rather than silently degrading to arbitrary.

## Which feature is affected?

## Checklist

## Additional context

**Search-cap removal is a behaviour change beyond the original ask**, called out for review: the cap was raised nowhere — it was removed. Any bound reintroduces the same unreachability at a different threshold, and the grid is a `LazyVGrid` so cost tracks what's on screen rather than the match count. Disagreement welcome; this is the one judgement call here.

Two comment corrections fell out along the way: `JupiterCatalogProvider`'s header claimed `coingeckoId` is preserved into `priceProviderId`, but the v2 endpoints return no `extensions` object at all, so it is always empty and `SolanaJupiterTokenExtensions` is effectively dead (comment fixed, model left alone); and both provider headers credited `SwapTokenListCache` for wallet-path freshness, which it does not provide.

Known and deliberately not addressed: `SwapCoinSelectionViewModel` re-sorts by fiat balance and discards provider rank. That is pre-existing, intentional (the swap picker is balance-ordered), and not made worse here.

- **Agent**: Claude Code
- **Issue**: #4979
- **Confidence**: high on the ranking evidence and the browse/search split, both measured against live provider responses and covered by tests
- **Needs human review for**: the unbounded-search decision above, and whether 20 per provider is the right number once it's on screen

## Summary by CodeRabbit

- **New Features**
  - Enhanced token catalog ordering using provider quality signals (including Jupiter organic ranking and 1inch tier curation).
  - Search now returns results across the full shortlist (no arbitrary cap) while keeping empty queries empty.
  - Browse results are more fairly capped per provider category, while preserving curated local tokens.
  - Token source attribution is retained to support clearer catalog organization.
  - Catalog rank order is preserved across disk caching.
- **Bug Fixes**
  - Prevented catalog data from one blockchain appearing on another after failed loads.
- **Tests**
  - Added/expanded regression coverage for ranking, ordering, search reach, browse capping, and cache round-trips.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
@@ -139,19 +139,13 @@ struct TokenSearchResult {
     /// so browse can cap each provider's breadth independently. Threaded as a
     /// side map for the same reason as `verificationByUniqueId`: it keeps the
     /// swap pickers' `[CoinMeta]` list types unchanged.
+    ///
+    /// Deliberately has no default: an omitted map would compile clean while
+    /// collapsing every provider into `cappedPerProvider`'s single unattributed
+    /// bucket, silently turning "the best N per provider" into "N in total".
+    /// Constructed through the synthesized memberwise init so that stays true
+    /// for any field added later.
     let sourceKindByUniqueId: [String: String]
-
-    init(
-        surfaceable: [CoinMeta],
-        unverified: [CoinMeta],
-        verificationByUniqueId: [String: TokenVerification],
-        sourceKindByUniqueId: [String: String] = [:]
-    ) {
-        self.surfaceable = surfaceable
-        self.unverified = unverified
-        self.verificationByUniqueId = verificationByUniqueId
-        self.sourceKindByUniqueId = sourceKindByUniqueId
-    }
 }
 
 /// Cancellation is the only failure this service can report: the catalog's
--- a/VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
@@ -106,20 +106,27 @@ struct TokenSearchService {
         let unverified = safe.filter { !$0.autoSurfaces }.map { $0.meta }
 
         var verificationByUniqueId: [String: TokenVerification] = [:]
+        var sourceKindByUniqueId: [String: String] = [:]
         for candidate in safe {
             verificationByUniqueId[candidate.uniqueId] = candidate.verification
+            sourceKindByUniqueId[candidate.uniqueId] = candidate.sourceKind
         }
 
         return TokenSearchResult(
             surfaceable: surfaceable,
             unverified: unverified,
-            verificationByUniqueId: verificationByUniqueId
+            verificationByUniqueId: verificationByUniqueId,
+            sourceKindByUniqueId: sourceKindByUniqueId
         )
     }
 }
 
 /// The wallet add-token picker's verification-aware view of the catalog. Kept a
 /// value type so it crosses the actor boundary cleanly.
+///
+/// `surfaceable` and `unverified` arrive in provider order, which is the
+/// providers' ranking (best-first) — see `OneInchToken.rankedForCatalog` /
+/// `SolanaJupiterToken.rankedForCatalog`. Nothing downstream may re-sort them.
 struct TokenSearchResult {
     /// Curated / verified, non-native — the tokens that auto-surface (default).
     let surfaceable: [CoinMeta]
@@ -128,6 +135,23 @@ struct TokenSearchResult {
     let unverified: [CoinMeta]
     /// `CoinMeta.uniqueId → verification` for badging. Only non-native entries.
     let verificationByUniqueId: [String: TokenVerification]
+    /// `CoinMeta.uniqueId → the `TokenCatalogProvider.kind` that won the token`,
+    /// so browse can cap each provider's breadth independently. Threaded as a
+    /// side map for the same reason as `verificationByUniqueId`: it keeps the
+    /// swap pickers' `[CoinMeta]` list types unchanged.
+    let sourceKindByUniqueId: [String: String]
+
+    init(
+        surfaceable: [CoinMeta],
+        unverified: [CoinMeta],
+        verificationByUniqueId: [String: TokenVerification],
+        sourceKindByUniqueId: [String: String] = [:]
+    ) {
+        self.surfaceable = surfaceable
+        self.unverified = unverified
+        self.verificationByUniqueId = verificationByUniqueId
+        self.sourceKindByUniqueId = sourceKindByUniqueId
+    }
 }
 
 /// Cancellation is the only failure this service can report: the catalog's
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Models/SolanaJupiterTokenInfo.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Models/SolanaJupiterTokenInfo.swift
@@ -8,6 +8,7 @@ struct SolanaJupiterToken: Codable {
         case decimals
         case logoURI = "icon"
         case extensions
+        case organicScore
     }
     let address: String?
     let name: String?
@@ -15,6 +16,11 @@ struct SolanaJupiterToken: Codable {
     let decimals: Int?
     let logoURI: String?
     let extensions: SolanaJupiterTokenExtensions?
+    /// Jupiter's own 0–100 token-quality metric, blending organic (non-wash)
+    /// volume, holder distribution and liquidity depth. Present on every entry of
+    /// the verified tag list — see `rankedForCatalog` for why it, and not `mcap`,
+    /// is what the catalog ranks by.
+    let organicScore: Double?
 
     // Custom init to handle missing fields gracefully
     init(from decoder: Decoder) throws {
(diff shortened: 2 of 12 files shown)


VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
```diff
@@ -128,6 +135,23 @@ struct TokenSearchResult {
     let unverified: [CoinMeta]
     /// `CoinMeta.uniqueId → verification` for badging. Only non-native entries.
     let verificationByUniqueId: [String: TokenVerification]
+    /// `CoinMeta.uniqueId → the `TokenCatalogProvider.kind` that won the token`,
+    /// so browse can cap each provider's breadth independently. Threaded as a
+    /// side map for the same reason as `verificationByUniqueId`: it keeps the
+    /// swap pickers' `[CoinMeta]` list types unchanged.
+    let sourceKindByUniqueId: [String: String]
+
+    init(
```
NeOMakinG: `suggestion:` TokenSearchResult.init defaults sourceKindByUniqueId to [:] — No live bug — the one production call site passes the map — but a future 3-arg construction would compile clean while collapsing every provider token into the single .empty bucket in cappedPerProvider, silently turning "20 per provider" into "20 total".

fix: Drop the default and make sourceKindByUniqueId required, so an omission fails at compile time instead of degrading browse silently.

NeOMakinG (commented):
gm — multi-lane pass: claude-correctness + codex-adversarial. conceptual-only (no runtime this pass). comment-only.

## suggestion
### 1. loadExternalTokens catch records the error but never logs it — `VultisigApp/VultisigApp/Features/Wallet/ViewModels/TokenSelectionViewModel.swift:151`
After discardRetainedCatalog() on a chain change, a thrown load degrades browse/search to empty with no signal — indistinguishable from a genuinely token-less chain. In production this path is only reachable on TokenSearchServiceError.cancelled (TokenCatalogRepository.catalog(for:) is non-throwing and fails open to the bundled floor + disk snapshot), and cancellation skips recompute anyway, so it is not a live user-facing failure today. But the injected loadCatalog in tests throws arbitrary errors, and if the catalog wiring ever becomes throwing this becomes a silent blank picker.
action: Add a Logger.warning in the catch naming the chain and error, so a cleared-then-failed state is observable if the fail-open architecture ever changes.

### 2. TokenSearchResult.init defaults sourceKindByUniqueId to [:] — `VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift:144`
No live bug — the one production call site passes the map — but a future 3-arg construction would compile clean while collapsing every provider token into the single .empty bucket in cappedPerProvider, silently turning "20 per provider" into "20 total".
action: Drop the default and make sourceKindByUniqueId required, so an omission fails at compile time instead of degrading browse silently.

## q
- Possible one-scheduling-hop stale frame if the view model is reused across a chain change

## risk
Low. UX-surface change confined to the wallet add-token picker; no signing, tx-build, balance, or fund path is touched. The one deliberate behavior change — removing the .prefix(20) search cap so search is now unbounded — is mitigated by LazyVGrid rendering only on-screen rows and a sub-millisecond filter over a few thousand short tickers per keystroke; the author explicitly flagged this as the judgment call and requested feedback. Chain-scoping and cap logic are covered by targeted tests including the exact cross-chain-leak and cap-after-exclusion invariants, and were independently re-verified outside the author's own fixtures via a standalone Linux Swift re-execution of the algorithm.

## verdict
COMMENTED

🤖 reviewed via [Claude Code](https://claude.com/claude-code) with codex `gpt-5.4` conceptual pass.

VultisigApp/VultisigApp/Core/Services/TokenSearchService.swift
gastonm5: Agreed, and done — the failure mode you describe is exactly the kind that never shows up as a crash or a test failure, just a quietly shorter browse list.

Rather than only dropping the default, I removed the hand-written memberwise `init` entirely. Once the `= [:]` was gone it was byte-for-byte the init Swift synthesizes, so it was a duplicate that could only drift. Leaning on the synthesized one also means the property is required at every call site *by default* — and so is any field added to `TokenSearchResult` later, unless someone deliberately writes a default. That gets the compile-time guarantee you asked for and keeps it, instead of re-arming the same trap on the next field.

Kept a note on the property itself so the "why" survives:

```swift
/// Deliberately has no default: an omitted map would compile clean while
/// collapsing every provider into `cappedPerProvider`'s single unattributed
/// bucket, silently turning "the best N per provider" into "N in total".
/// Constructed through the synthesized memberwise init so that stays true
/// for any field added later.
let sourceKindByUniqueId: [String: String]
```

**On the four test fixtures that used the 3-argument form** — I checked each rather than blanket-filling `[:]`, since an explicit `[:]` on a breadth test would have preserved a misleading assertion. None of them was meaningfully relying on the default:

| Test | Catalog | Was it relying on the default? |
|---|---|---|
| `testHeldCuratedTokenIsFoundBySearch` | empty | No — no token to attribute; `[:]` is the honest value |
| `testChainWithNoRemoteProviderIsUnaffectedByTheCap` | empty | No — same, and it asserts `browseProviderTokens.isEmpty` |
| `testBrowseShowsVerifiedButNotUnverifiedWhileSearchRevealsUnverified` | 1 verified + 1 unverified | No — 1 token vs a cap of 20 |
| `testSearchResultsTrackTheLatestKeystroke` | 1 token | No — same |

So no assertion changes meaning. The two empty-catalog fixtures pass `[:]` with a comment saying why; the two non-empty ones now attribute their tokens to `"oneinch"`, so every token in every fixture in that file has a real source kind — matching the invariant `TokenSearchService.searchResult` actually upholds (it populates a source kind for every safe candidate) instead of exercising the unattributed fallback bucket. `"oneinch"` is the literal already used by `rankedBreadth` throughout this file; provider `kind` is a plain `String` per provider with no shared constant to reference.

The one deliberate `[:]` that remains is `testCapKeepsTokensWithAnUnknownSourceBoundedRatherThanUncapped`, which calls `cappedPerProvider` directly and is specifically about unattributed tokens staying bounded.

Leaving this thread for you to resolve.

johnnyluo (approved):
LGTM

PR #4984, opened 2026-07-29
feat(review): ask for an App Store review after a confirmed transaction

### LINKED ISSUE #4961: Add in-app review prompt after successful transaction (ratings velocity is our store-conversion bottleneck)
## Problem

Vultisig iOS has **62 ratings** (US App Store). Trust Wallet has ~1.7M; Exodus and Zengo each have hundreds of thousands. Ratings count + velocity gate both search rank and product-page conversion on the App Store — before any keyword or metadata work pays off, this is the ceiling. It has been flat at 62 for 14+ days (marketing ASO monitoring).

We rank **#1 for "seedless wallet"** (US App Store) today; with two-digit ratings that position converts a fraction of what it should, and it's fragile.

## Proposal

Trigger the native review prompt (`SKStoreReviewController.requestReview(in:)`, or `AppStore.requestReview(in:)` on iOS 16+) at the moment of highest satisfaction:

- **After a successful send/swap confirmation** — transaction confirmed on-chain, user on the success screen. Not at app open, never after an error.
- App-side throttling before requesting: ≥3 successful transactions AND ≥7 days since install AND not prompted in the last 120 days. (Apple also caps the system prompt at ~3/year and may no-op silently — that's expected.)
- Native sheet only — no custom "rate us" chrome, no incentives (App Store policy-safe).

## Why now

ASO keyword/metadata iteration is underway on the marketing side; its conversion impact is capped by ratings volume. This is the single highest-leverage product change for organic App Store growth right now.

## Acceptance

- Review prompt fires on the success screen under the throttle rules.
- Analytics event logged when the prompt is *shown* (so marketing can correlate ratings velocity to the trigger; note the system won't tell us if the user actually rated).

gastonm5:
## Description

Fixes #4961

Fires the native App Store review sheet at the moment of highest satisfaction — a **confirmed** transaction — behind a throttle. iOS has 62 US ratings against Trust Wallet's ~1.7M, and ratings volume caps what the ASO work can convert.

Prompts when **all** hold: **≥3 confirmed transactions** · **≥7 days since install** · **not prompted in the last 120 days**. Uses SwiftUI's `@Environment(\.requestReview)` (deployment target is iOS 17, so no `SKStoreReviewController` and no window-scene plumbing). Native sheet only — no custom chrome, no incentives.

## 🔴 One acceptance criterion was deliberately not met

The issue asks for an **analytics event when the prompt is shown**. **This app has no analytics of any kind** — no Firebase, Mixpanel, Amplitude, Segment or PostHog (a repo-wide grep returns 48 hits for "Segment", all `SegmentedControl` UI). And `PrivacyInfo.xcprivacy` declares:

```
NSPrivacyTracking:     None
collected data types:  NONE
tracking domains:      NONE
```

Meeting the criterion would mean introducing telemetry to a self-custody wallet that has deliberately never collected anything, updating the privacy manifest, and changing the App Store data-collection disclosure. That's a product and privacy decision, not an implementation detail — so it's **out of scope here and left to its own issue and its own privacy review** (decision: Gaston, 2026-07-29).

It also buys less than it appears: **Apple never reports whether the user actually rated.** The event would record only *"we asked"* — which marketing can already derive from the release date plus these throttle rules, correlated against ratings velocity in App Store Connect.

## Design

`DoneScreen` is the single shared completion component behind every flow — send, swap, co-signer, QBTC claim **and custom-message signing** — so this is one hook rather than five. It observes `statusService.status` reaching `.confirmed`.

- `AppReviewPolicy` — a pure decision function over `(state, now)`. No I/O, fully unit-tested at every boundary (exactly 3, exactly 7 days, exactly 120 days).
- `AppReviewService` — persistence (`UserDefaults`; preference-shaped, and must survive independently of any vault), install-date seeding, and an injected clock so tests never sleep.

## The bug worth reading about

The first cut of the hook counted the transaction and asked the throttle as two **independent** questions, discarding `recordConfirmedTransaction`'s `Bool`. Caught in review, and the exploit is reachable in production:

`DoneStatusServiceFactory.signedMessage()` returns `NoPoller(initialStatus: .confirmed)`, so `SigningKeysignScreen.signedMessageBranch` renders a `DoneScreen` that is **`.confirmed` from the first frame, with an empty hash, and is not a transaction at all**. The empty hash correctly counted nothing — and then it prompted anyway. **Signing a custom message would have burned the 120-day window on a non-transaction.** Same shape for re-entering an already-counted hash after a cooldown lapsed.

Fixed by composing the sequence in the service rather than guarding at the call site, so the ordering is under unit test instead of living in a view:

```swift
func claimReviewPrompt(forConfirmedTransaction id: String) -> Bool {
    guard recordConfirmedTransaction(id: id) else { return false }
    guard shouldRequestReview() else { return false }
    recordPromptShown()
    return true
}
```

An ask can now only be spent off a **newly counted** transaction. Note the service already had idempotency tests — the gap was one layer up, in the view's *use* of them, which service-level tests could never have seen.

Second review finding, also real: `withAnimation` only *schedules* the spring, so the reveal flag flipped ~0.55 s before the hero arrived and the sheet could land on the very animation the guard existed to avoid. Replaced with `hasSettled`, set after awaiting the spring; a `.confirmed` arriving mid-reveal is turned away and retried at the tail of the reveal task, so nothing is dropped.

## Decisions worth a second opinion

- **Install date is seeded honestly, not backdated.** Upgrading users are stamped at upgrade and serve the full 7 days. Backdating would fire at the entire existing base within days of release — the users most likely to have formed opinions, spent on a rushed ask against a 120-day window.
- **Seeding placement is load-bearing.** It sits in the iOS-only `extension VultisigApp` `.onAppear`, before the `-UITesting` early return. ⚠️ `.onAppear` is **not** guaranteed once per launch (the shared `content` backs both `WindowGroup` and `DocumentGroup`) — the **persistence guard**, not the placement, is what makes it a single stamp.
- **Counted-id ring is bounded at 100.** Done screens are reachable only from post-signing flows, never from history, so a 100-transactions-ago hash can't be re-observed; and past a count of 3, an extra increment changes nothing.
- **Co-signer completions count** toward the tally even though the transaction isn't the user's own — a heavy co-signer can reach the threshold without ever sending.
- **The ask is recorded when `requestReview` is called**, even though Apple may silently no-op. There's no feedback from the API, so this can burn the window without a sheet appearing. The issue explicitly accepts that.

## Which feature is affected?

## Checklist

## Additional context

**Gate:** iOS `** TEST SUCCEEDED **` — 4065 tests, 1 skipped, 0 failures (`AppReviewPolicyTests` 18/18, `AppReviewServiceTests` 22/22). macOS `** BUILD SUCCEEDED **`, required because `DoneScreen` is cross-platform and gained `#if os(iOS)`. SwiftLint 0 in touched files.

**No new user-facing strings** — the sheet is entirely system-provided, so there's no localisation work.

Pre-existing and untouched: `DoneScreen.swift:72-75` carries a `#4777` issue reference in a comment, against the repo's no-issue-numbers-in-code rule. It's on `main`, not from this branch.

- **Agent**: Claude Code
- **Issue**: #4961
- **Confidence**: high on the policy, persistence and the claim/spend ordering — all directly unit-tested, including the empty-hash case
- **Needs human review for**: **the sheet itself was never observed.** The system throttles `requestReview` and often suppresses it silently with no API feedback, so its presentation is not meaningfully verifiable in a simulator and no claim is made about it. Also worth a second opinion: the honest-vs-backdated install date, and whether co-signer completions should count toward the tally.

## Summary by CodeRabbit

- **New Features**
  - Added App Store review prompts after eligible confirmed transactions.
  - Review requests are limited by transaction count, installation age, and cooldown periods.
  - Duplicate transactions and repeated prompts are automatically prevented.

- **Bug Fixes**
  - Review prompts now wait until transaction completion animations have settled.
  - Install and prompt history persists across app launches.

- **Tests**
  - Added coverage for review eligibility, throttling, persistence, and duplicate transactions.

### DIFF
--- a/VultisigApp/VultisigApp/App/VultisigApp.swift
+++ b/VultisigApp/VultisigApp/App/VultisigApp.swift
@@ -216,6 +216,13 @@ extension VultisigApp {
                 // funnel can read them without touching @Model.
                 CustomRPCStore.shared.reloadFromStore()
 
+                // Stamp the install date the App Store review throttle waits
+                // on. Idempotent, and deliberately placed at launch rather
+                // than at the first confirmed transaction: the 7-day rule has
+                // to be measured from when the user got the app, not from
+                // when they first happened to transact.
+                AppReviewService.shared.seedInstallDateIfNeeded()
+
                 if ProcessInfo.processInfo.isiOSAppOnMac {
                     continueLogin()
                 }
--- a/VultisigApp/VultisigApp/Components/TransactionDone/DoneScreen.swift
+++ b/VultisigApp/VultisigApp/Components/TransactionDone/DoneScreen.swift
@@ -24,6 +24,9 @@
 //
 
 import SwiftUI
+#if os(iOS)
+import StoreKit
+#endif
 
 struct DoneScreen<
     TokenContent: View,
@@ -33,6 +36,10 @@ struct DoneScreen<
     let input: TransactionDonePayload
 
     @StateObject private var statusService: DoneStatusService
+
+    #if os(iOS)
+    @Environment(\.requestReview) private var requestReview
+    #endif
 
     /// Nav-bar title for the screen. Defaults to "Done"; the initiator
     /// Send/Swap flows pass "Overview" so the in-place keysign→overview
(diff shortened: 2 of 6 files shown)


johnnyluo (approved):
LGTM

Issue #4932, opened 2026-07-27
Expand golden signing-vector coverage: TCY, SEI, EVM chainId matrix, Dash/Zcash, missing Cosmos chains, Bittensor/QBTC


realpaaao:
## Summary

A 2026-07-27 sweep of the golden signing-vector corpus (23 files, ~59 cases, mirrored across vultisig-ios, vultisig-android, and the TS core in vultisig-sdk) found that the two cross-platform keysign failures reported this month — vultisig/vultisig-windows#4464 (THORChain **TCY → RUNE**, Android↔extension) and vultisig/vultisig-windows#4369 (**SEI**, mobile↔extension/desktop) — both involve transaction shapes with **zero vector coverage**. This issue tracks expanding this repo's corpus; each implementation repo carries the same additions so the three corpora stay identical.

## Proposed new vectors (priority order)

1. **THORChain TCY** — TCY send + TCY → RUNE intra-chain swap (`MsgDeposit` + swap memo). Regression for vultisig/vultisig-windows#4464. `thorchainswap.json` currently only covers cross-chain L1 swaps; no non-RUNE-denom deposit case exists.
2. **SEI** — native send + ERC20 send (chainId 1329). Regression for vultisig/vultisig-windows#4369.
3. **EVM chainId matrix** — one native-send vector per currently uncovered EVM chain: Avalanche, Optimism, Base, Blast, CronosChain, Zksync, Mantle, Hyperliquid. Only 4 of ~13 EVM chains have a vector today; chainId is part of the signed pre-image, and SEI shows per-chain divergence is real.
4. **UTXO: Dash + Zcash** — no vectors today. Zcash especially: it's hand-rolled on Android (SwapKit signer), and hand-rolled paths are where implementations diverge most.
5. **Cosmos: Osmosis, Dydx, Noble, Akash** — supported chains with no vectors; add sends + one IBC transfer.
6. **Bittensor + QBTC** — hand-rolled in the TS core (`getPreSigningHashes`), no shared vectors.
7. **THORChain limit-order Place flow** — shipped 2026-07-15 (vultisig-ios#4319), currently unpinned.

## Acceptance / ground rules

- New `expected_image_hash` values must be generated by running **at least two of the three real signers** (Kotlin / Swift / TS) on the payload and asserting agreement — never authored by hand or captured from a single implementation. A vector seeded from one signer just fossilizes that signer's bugs.
[...]

PR #4996, opened 2026-07-30
fix(tests): un-skip the ChainHelper golden signing-vector suite

### LINKED ISSUE #4931: ChainHelperTests golden suite passes vacuously since #4319 — filename guard skips all 23 TestData fixtures (zero signing-conformance coverage)
## Summary

`ChainHelperTests.testChainHelpers` has been passing **vacuously since 2026-07-15**: a filename guard added in #4319 skips every fixture in `VultisigAppTests/TestData/`, so the signing-conformance suite currently decodes and runs **zero** test cases. iOS presently has no golden-vector coverage of pre-image hashes.

## The bug

`VultisigApp/VultisigAppTests/Chains/ChainHelperTests.swift`:

```swift
// Skip files that aren't ChainHelper fixtures — other suites
// may bundle their own JSON (e.g. Fixtures/LimitSwapMemos.json)
// that flatten to the same resourcePath at test time. Match by
// filename prefix and fail fast if a *named* ChainHelper file
// doesn't decode (so genuine schema regressions still surface).
let isChainHelperFixture = jsonFile.lastPathComponent.hasPrefix("ChainHelper")
guard isChainHelperFixture else { continue }
```

The guard's intent (skip other suites' bundled JSON) is fine, but **no file in the repo matches the prefix**: the 23 fixtures are named `thorchain.json`, `evm.json`, `utxo.json`, etc. Verified against the full repo tree — the only path containing "ChainHelper" is the test file itself. Every iteration hits `continue`; the loop body never executes; the test goes green.

Introduced in commit `9076a4dc` (#4319, merged 2026-07-15).

## Evidence it's been dead since then

- `TestData/mayaswap.json` is **malformed JSON** (trailing comma before the closing `]`). A strict `JSONDecoder` rejects it — if the runner were decoding fixtures, the suite would be red today.
- The iOS fixture set has silently fallen behind the equivalent corpora in vultisig-android (`app/src/androidTest/assets/`) and the TS core (`vultisig-sdk` `packages/core/mpc/keysign/tests/fixtures/mobile/`): missing `cosmos.json` cases "SignAmino - ATOM transaction with multiple messages" / "SignDirect - ATOM transaction (multi-send)", `solana.json` "Send Solana USDC", and `thorchain.json` "SignDirect - THORChain Secure asset: LTC transaction".

## Why it matters

These vectors are the per-repo mechanism that pins iOS's pre-image hashes for keysign. With the runner dead, a change that moves a hash on iOS merges green and only surfaces as a field failure — a cross-platform keysign that 404s waiting for the setup message, because the joiner polls a `message_id` (hash of the locally rebuilt payload) the initiator never uploaded. See vultisig/vultisig-windows#4464 for a live example of that failure class this week.

## Suggested fix

1. Replace the prefix allow-list with the inverse — a skip-list of the known non-fixture JSONs (or point the runner at the `TestData` directory explicitly instead of the flattened bundle root), so all 23 fixtures load again and `XCTFail` fires on any that don't decode.
2. Fix the trailing comma in `mayaswap.json`.
3. Backfill the 4 cases listed above so the iOS corpus matches Android and the TS core again.
4. Sanity check: after the fix, the suite should report a non-trivial executed-case count (Android/TS run ~59 cases over 23 files) — consider asserting a minimum case count so an empty run can never pass again.

gastonm5:
## Description

`ChainHelperTests.testChainHelpers` — the golden signing-vector conformance suite — has been passing **vacuously since 2026-07-15**. It walked the flattened test-bundle resource root and then did:

```swift
let isChainHelperFixture = jsonFile.lastPathComponent.hasPrefix("ChainHelper")
guard isChainHelperFixture else { continue }
```

No fixture is named that way — they are `thorchain.json`, `evm.json`, `utxo.json`, … — so every iteration hit `continue`, the loop body never executed, and the test went green having decoded **zero** cases. Baseline on `main`: `testChainHelpers` **passed in 0.005 s**.

Fixes #4931. Unblocks #4932.

### What changed

1. **`TestData/` is now bundled as a folder reference** (`project.yml`), so the corpus keeps its own directory inside the test bundle instead of flattening into the resource root next to `LimitSwapMemos.json` and `SigningGoldenVectors.json`. The filename heuristic existed *because* of that flat namespace; fixing the namespace removes the need to guess which JSON belongs to which suite. Same pattern already used for `Swap/SwapKit/__fixtures__` and `CosmosStakingFixtures`.
2. **Explicit registration replaces the wildcard walk.** A `ChainHelperFixture` enum names all 23 fixtures, and `testAllRegisteredFixturesExecute` runs every case in the registry in one test — so coverage does not depend on anyone remembering to add a per-fixture method.
3. **Per-fixture test methods for attribution.** Each fixture also gets its own `test…Fixture()`, so a failure names the fixture it came from, and a vector that genuinely needs quarantining can carry its own annotated `XCTSkip` instead of a blanket one. Cases therefore run twice; the whole class takes ~0.1 s.
4. **`testEveryBundledFixtureIsRegistered`** pins the bundled `TestData/*.json` set against the registry *in both directions*, and checks each fixture still has its attributing method. A fixture can no longer stop being executed without something turning red — that is the actual bug class here, not the one bad predicate.
5. **Each fixture pins how many vectors it must hold**, and the aggregate test asserts the corpus total (23 fixtures / 67 vectors). Filenames alone only prove a file still exists, not that it still asserts anything — a fixture emptied down to one vector would otherwise stay green. Adding or removing a vector is a deliberate act, so it now needs a deliberate edit, and the failure message names which fixture changed size and by how much. (#4932 will bump these when it adds vectors; that friction is the point.)
6. **`mayaswap.json` trailing comma fixed.** That file has been invalid JSON since it landed; `JSONDecoder` rejects it outright. It is the standalone proof the runner was inert rather than merely quiet — a live runner would have been red every day since.
7. Each case in a fixture now runs even if an earlier one throws, so one broken vector cannot hide the state of the rest of its file.

### Executed-test count (the evidence)

| | Whole `VultisigAppTests` suite | `ChainHelperTests` | Golden cases actually asserted |
|---|---|---|---|
| Before (`main` @ `a36c03509`) | 4351 executed, 0 failures | 1 (`testChainHelpers`, 0.005 s) | **0** |
| After | 4375 executed, 0 failures | 25 | **67** |

### Proof the suite is not vacuous again

Restoring coverage is easy to *claim* and easy to get wrong, so it was verified by perturbation rather than by trusting the green result:

- Every `expected_image_hash` in all 23 fixtures was flipped in its first hex character, and the suite re-run.
- Result: `Executed 25 tests, with 134 failures` — **134 = 67 × 2**, i.e. every corpus case failed once under its own per-fixture test and once under the aggregate coverage test. 67 distinct case names appeared in the failure messages, exactly matching the corpus.
- The aggregate `testAllRegisteredFixturesExecute` was among the failures, confirming it really is carrying the coverage rather than merely existing.
- The fixtures were then restored (`git checkout`) and the suite is green again.

So every one of the 67 cases genuinely executes and genuinely compares its computed pre-image hash against the committed vector. The perturbation was a local verification step only — no perturbed data is in this branch.

## Per-fixture triage

All 23 fixtures / 67 cases **pass** against the current Swift signers. No fixture needed quarantining; **there are no skips in this PR**.

| Fixture | Cases | Result | Verdict |
|---|---|---|---|
| `arb.json` | 1 | pass | — |
| `bsc.json` | 2 | pass | — |
| `cardano.json` | 2 | pass | — |
| `cosmos.json` | 4 | pass | — |
| `cosmos-sdk-sign-amino.json` | 2 | pass | — |
| `cosmos-sdk-sign-direct.json` | 2 | pass | — |
| `dot.json` | 1 | pass | — |
| `evm.json` | 2 | pass | — |
| `kujira.json` | 3 | pass | — |
| `lifiswap.json` | 2 | pass | — |
| `maya.json` | 3 | pass | — |
| `mayaswap.json` | 2 | pass, after the JSON-syntax repair | file was **undecodable**, not wrong — no hash touched |
| `pol.json` | 1 | pass | — |
| `solana.json` | 4 | pass | — |
| `solana-sign-data.json` | 1 | pass | — |
| `sui.json` | 1 | pass | — |
| `terra.json` | 5 | pass | — |
| `thorchain.json` | 9 | pass | — |
| `thorchainswap.json` | 5 | pass | — |
| `ton.json` | 3 | pass | — |
| `tron.json` | 4 | pass | — |
| `utxo.json` | 5 | pass | — |
| `xrp.json` | 3 | pass | — |

**No signing regression was found.** Every committed pre-image hash still matches what the current Swift signers produce. Given that this corpus has had no pressure keeping it correct since 2026-07-15 and `main` has moved a long way, that is a genuinely good outcome and not a foregone one — and it is worth stating plainly, because a red vector here would have been the same failure class as vultisig/vultisig-windows#4464 (joiner polls a `message_id` derived from a locally rebuilt payload the initiator never uploaded).

## What this PR does NOT verify

- **Only the Swift signer was run.** A green vector proves iOS is self-consistent with the committed hash; it does **not** prove the hash itself is right. If a vector had originally been captured from a buggy signer, iOS agreeing with it would look identical to iOS being correct. Cross-signer corroboration (Kotlin / TS) is out of reach from this repo.
- Consequently **no `expected_image_hash` was added, changed, or re-baselined here** — #4932's ground rule (a hash must be corroborated by at least two of the three real signers) cannot be satisfied from inside vultisig-ios. The only fixture edit in this branch is `mayaswap.json`'s trailing comma; `git diff a36c03509..HEAD -- VultisigApp/VultisigAppTests/TestData/` shows no changed hash string.
- **The 4 cases iOS is missing versus Android and the TS core are deliberately not backfilled** (issue suggestion 3): `cosmos.json` "SignAmino - ATOM transaction with multiple messages" and "SignDirect - ATOM transaction (multi-send)", `solana.json` "Send Solana USDC", `thorchain.json` "SignDirect - THORChain Secure asset: LTC transaction". Each needs a new hash, so they belong to #4932 with two-signer corroboration.
- Chains still unpinned by this corpus are unchanged by this PR — everything listed in #4932: TCY, SEI, the wider EVM chainId matrix, Dash/Zcash, Osmosis/Dydx/Noble/Akash, Bittensor, QBTC, and the THORChain limit-order Place flow.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

Test-only change. No production code and no app behaviour is modified — the diff is one test file, one XcodeGen resource declaration, and one JSON syntax repair.

## Parity

Test-infrastructure fix, not a signing change. The skipped-corpus bug is specific to the iOS runner's filename guard; Android and the TS core carry the same fixture corpus but no equivalent guard.

- Android: n/a — the iOS-only filename guard does not exist there (corpus *expansion* is tracked separately in vultisig/vultisig-android#5421, the sibling of #4932 rather than of this fix)
- Windows + extension: n/a — same, no equivalent guard
- SDK: n/a — same

## Checklist

## Additional context

Gate: clean build (DerivedData removed first, so the resource-copy change is exercised the way CI would) + full `make test` on a dedicated simulator → `** TEST SUCCEEDED **`, `Executed 4375 tests, with 1 test skipped and 0 failures`. SwiftLint clean on the changed file.

The corpus guards are the part worth reviewing hardest. The original bug was not "someone wrote a wrong predicate"; it was "a wildcard walk can quietly match nothing and nothing notices". Two cross-model review rounds each caught a second-order version of exactly that in earlier drafts of this fix:

1. the guard first compared only the *number* of test methods to the number of fixtures — satisfied by deleting one fixture's test and adding any unrelated one. Hence coverage now lives in a single test that iterates the registry, not in per-method bookkeeping;
2. the aggregate test then asserted only that it had run *more than zero* vectors — satisfied by deleting vectors out of a fixture down to the last one. Hence the pinned per-fixture counts.

Both are the same shape as the original bug, one level up, which is a decent argument for reviewing this file adversarially rather than by reading it.

- **Agent**: Claude Code
- **Issue**: #4931
- **Confidence**: high for the runner fix and the executed-case evidence; the *correctness of the vectors themselves* is explicitly unverified — see "What this PR does NOT verify"
- **Needs human review for**: whether the un-skipped corpus should now be treated as authoritative, given that only the Swift signer has ever been run against it

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the MayaSwap test fixture format so it loads successfully.

* **Tests**
  * Improved ChainHelper fixture coverage with explicit fixture registration and validation.
  * Added checks to ensure all bundled fixtures are registered and executed.
  * Enhanced test reporting so individual fixture failures do not stop remaining cases.

* **Chores**
  * Updated test resource packaging to preserve the TestData folder structure.

### DIFF
--- a/VultisigApp/VultisigAppTests/Chains/ChainHelperTests.swift
+++ b/VultisigApp/VultisigAppTests/Chains/ChainHelperTests.swift
@@ -24,17 +24,18 @@ struct ChainHelperTestCase: Codable {
 
 /// The golden signing-vector fixtures bundled under `TestData/`.
 ///
-/// Every fixture is named here and gets its own `test…` method rather than
-/// being discovered by a wildcard walk of the bundle. A wildcard walk is what
-/// let the entire corpus stop executing without anything turning red: the
-/// walk's "is this one of ours?" filter silently matched nothing, so the loop
-/// body never ran and the suite stayed green with zero conformance coverage.
+/// This list — not a wildcard walk of the bundle — is what the suite executes.
+/// A wildcard walk is what let the entire corpus stop running without anything
+/// turning red: the walk's "is this one of ours?" filter silently matched
+/// nothing, so the loop body never ran and the suite stayed green with zero
+/// conformance coverage.
 ///
-/// Explicit registration makes that failure mode impossible to reach silently:
-/// `testEveryBundledFixtureIsRegistered` pins this list against the JSON files
-/// actually present in the bundle and against the number of test methods on
-/// this class, so a fixture can neither be added without a test nor listed
-/// without one.
+/// Two tests keep that from happening again.
+/// `testAllRegisteredFixturesExecute` runs every case in this list in a single
+/// test, so coverage never depends on per-method bookkeeping, and
+/// `testEveryBundledFixtureIsRegistered` pins the list against the JSON files
+/// actually present in the bundle — in both directions — so a fixture can
+/// neither be added without being run nor listed without existing.
 enum ChainHelperFixture: String, CaseIterable {
     case arb
     case bsc
@@ -59,6 +60,18 @@ enum ChainHelperFixture: String, CaseIterable {
     case tron
     case utxo
     case xrp
+
+    /// The per-fixture test method that must exist so a failure is attributable
+    /// to one fixture instead of to the corpus as a whole. Derived from the
+    /// file name, so the guard checks the same name the method has to declare.
+    var testMethodName: String {
+        let camelCased = rawValue
+            .split(separator: "-")
+            .enumerated()
+            .map { $0.offset == 0 ? String($0.element) : $0.element.capitalized }
+            .joined()
+        return "test\(camelCased.prefix(1).uppercased())\(camelCased.dropFirst())Fixture"
+    }
 }
 
 final class ChainHelperTests: XCTestCase {
@@ -96,10 +109,27 @@ final class ChainHelperTests: XCTestCase {
     func testUtxoFixture() throws { try runFixture(.utxo) }
     func testXrpFixture() throws { try runFixture(.xrp) }
 
-    // MARK: - Corpus guard
+    // MARK: - Corpus guards
+
+    /// Authoritative coverage: executes every registered fixture in one test.
+    ///
+    /// The per-fixture methods above exist for attribution, but coverage must
+    /// not depend on someone remembering to add one — that bookkeeping is
+    /// exactly what failed before. This test runs the whole registered corpus
+    /// regardless of how many per-fixture methods survive, so the vectors
+    /// cannot go dark again. Cases therefore run twice; the corpus is small
+    /// enough that this costs a fraction of a second.
+    func testAllRegisteredFixturesExecute() throws {
+        var executedCases = 0
+        for fixture in ChainHelperFixture.allCases {
+            executedCases += try runFixture(fixture)
+        }
+        XCTAssertGreaterThan(executedCases, 0,
+                             "The golden corpus executed zero signing vectors — the suite is vacuous")
+    }
 
     /// Fails if the bundled corpus and the registered fixtures diverge in
-    /// either direction, so no fixture can silently stop being executed.
+    /// either direction, or if a fixture has lost its attributing test method.
     func testEveryBundledFixtureIsRegistered() throws {
         let bundle = Bundle(for: type(of: self))
         guard let urls = bundle.urls(forResourcesWithExtension: "json",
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5003, opened 2026-07-30
chore(swap): remove the limitSwapEnabled feature flag

### LINKED ISSUE #5000: Remove the limitSwapEnabled feature flag — ship limit swaps unconditionally
## Summary

Limit swaps still ship behind `limitSwapEnabled`, an **Advanced-settings toggle that defaults to `false`**. The feature is complete and has been landing fixes for weeks (limit-order Place flow shipped 2026-07-15 in #4319), but every user has to find Settings → Advanced → "Limit Swap" to see the Market/Limit tab row at all.

This issue removes the flag and makes limit swaps unconditionally available.

## What the flag is

`SettingsViewModel.swift:34`

```swift
@AppStorage("limitSwapEnabled") var limitSwapEnabled: Bool = false
```

It is a **local** `@AppStorage` toggle, not a server-driven `FeatureFlagService` flag, so there is no remote kill-switch to coordinate — removal is purely a client change.

## Call sites — 3 reads, 1 write

| File | Line | Role |
|---|---|---|
| `Features/Settings/ViewModels/SettingsViewModel.swift` | 34 | the `@AppStorage` property |
| `Features/Settings/Views/SettingsAdvancedView.swift` | 58-62 | the `SettingToggleCell` that writes it |
| `Features/Swap/Views/Screens/SwapDetailsScreen.swift` | 43 | `.showIf(...)` gating the Market/Limit tab row |
| `Features/Swap/Views/Screens/SwapDetailsScreen.swift` | 308 | `refreshCounter` gate |

## The non-obvious part: `refreshCounter` becomes dead code

`SwapDetailsScreen.swift:304-311` is not just a condition to flip — the existing comment says exactly why:

```swift
// When the Market/Limit tab row is visible (limit swap enabled) the
// countdown lives in that row via `tabRowCountdown`; keep the toolbar
// counter only for the flag-off market layout, which has no tab row.
if !settingsViewModel.limitSwapEnabled, detailsViewModel.showRefreshCounter {
    SwapRefreshQuoteCounter(timer: detailsViewModel.timer)
}
```

With the flag gone, `!limitSwapEnabled` is permanently false, so this view can never render. It should be **deleted** along with its call site at `:80`, leaving `tabRowCountdown` (`:335`) as the sole quote countdown on this screen — not left behind as an unreachable branch.

Note `SwapVerifyScreen.swift:421` has an identically-named `refreshCounter`. It is unrelated and must stay.

## Scope

- [ ] Delete the `@AppStorage` property
- [ ] Delete the Advanced-settings toggle cell
- [ ] Remove the `"limitSwapToggle"` key from **all 8** locale files (en, de, es, hr, it, ko, pt, zh-Hans) and re-run `sort_localizable.py`
- [ ] Drop the `.showIf(...)` so the tab row always renders
- [ ] Delete the now-unreachable `refreshCounter` and its call site
- [ ] Confirm no remaining reference to `limitSwapEnabled` anywhere

## Acceptance

- [ ] A fresh install shows the Market/Limit tab row with no settings change
- [ ] An existing user who never enabled the toggle also sees it (the default was `false`, so this is the majority case)
- [ ] Exactly one quote countdown renders on the swap details screen, in the tab row
- [ ] `Settings → Advanced` no longer shows a "Limit Swap" row, and no locale leaks a raw `limitSwapToggle` key
- [ ] Full test suite green

## Out of scope

- The orphaned `limitSwapEnabled` value left in `UserDefaults` on existing installs. It is inert once nothing reads it; a cleanup migration is not worth a keyshare-touching code path.
- Any behavioural change to limit swaps themselves. This is flag removal only.

gastonm5:
## Description

Removes the `limitSwapEnabled` feature flag so THORChain limit swaps ship unconditionally.

The feature has been complete and landing fixes for weeks, but it sat behind an Advanced-settings toggle that defaulted to `false` — so unless a user went digging in Settings → Advanced, the Market/Limit tab row on the swap details screen never appeared at all. The flag was a **local `@AppStorage` value, not a server-driven `FeatureFlagService` flag**, so there was no remote kill-switch to coordinate; removing the property is the whole of the change.

Fixes #5000

### What was deleted

**The flag itself**
- `SettingsViewModel.swift` — the `@AppStorage("limitSwapEnabled")` property.
- `SettingsAdvancedView.swift` — the `SettingToggleCell` that wrote it, **and** the `Row.limitSwap` enum case that placed it in the list. `Row` is `CaseIterable` and drives the screen's `ForEach`, so the case had to go with the cell; the issue's call-site map missed this one.
- `Localizable.strings` — the `"limitSwapToggle"` key, removed from **all 8** shipping locales (en, de, es, hr, it, **ko**, pt, zh-Hans) and re-sorted with `sort_localizable.py`.

**The gate**
- `SwapDetailsScreen.swift` — `.showIf(settingsViewModel.limitSwapEnabled)` on the Market/Limit tab row.

### What became dead as a consequence, and was removed

- **`SwapDetailsScreen.refreshCounter` and its toolbar item.** Its own comment said it existed *only* for the flag-off market layout, "which has no tab row". With the row always present, `!limitSwapEnabled` can never be true — the view was unreachable, not merely redundant. `tabRowCountdown` is now the single quote countdown on this screen.
- **The toolbar-ordering comment** explaining how the glass advanced-settings button and the countdown pill sequenced. Only one toolbar item remains, so there is no ordering left to explain.
- **`SwapDetailsScreen`'s `@EnvironmentObject var settingsViewModel`.** The two flag reads were its only uses anywhere on the screen. Safe to drop: `@EnvironmentObject` *reads* from the environment, it does not inject into it, and nothing here fed a descendant via `.environmentObject(settingsViewModel)`.
- **The `selectedSwapMode` doc comment**, which described "flag-off renders the existing Market layout pixel-identical to pre-feature" — and, separately, named two symbols that no longer exist in the codebase at all (`LimitSwapModeBody`, `SwapRoute.limitPair`). Rewritten against the real types.

### Deliberately kept — each still has a live consumer

| Symbol | Why it stays |
|---|---|
| `SwapDetailsViewModel.showRefreshCounter` | Still read by `tabRowCountdown`, and by `updateTimer` as the timer's park guard |
| `SwapRefreshQuoteCounter` | Still the countdown on `SwapVerifyScreen` — only the *details*-screen usage went |
| `SwapQuoteCountdownBadge` | It is what `tabRowCountdown` renders |
| `CustomToolbarItem(hideSharedBackground:)` | Still used by `SwapVerifyScreen` and `VaultSetupScreen` |
| `"swap.tab.market"` / `"swap.tab.limit"` | Label the tab row, which is now always visible |
| `SwapVerifyScreen.refreshCounter` | **Identically named but unrelated** — it gates on `currentTransaction.isLimit`. Untouched. |

### Behaviour worth a reviewer's eye

The tab row now renders for **every** vault, including one with zero THORChain-routable chains. That is not new behaviour introduced here — `canCurrentPairUseLimitSwap` already *disabled* (rather than hid) the Limit segment in that case, which is what flag-on users have been seeing all along. Flagging it because removing `.showIf` is what makes that path reachable for everyone.

## Which feature is affected?

> No swap tx link: this PR changes only which UI is visible. It touches no keysign, broadcast, quote, memo or fee-computation code — the limit-swap pipeline itself is byte-for-byte unchanged.

## Parity

- Android: `n/a` — already unconditional. `SwapScreen.kt` renders `SwapModeTabs` with no flag gate; its Limit tab shows a `swap_limit_coming_soon` placeholder because the feature isn't implemented there yet, but the row itself was never hidden.
- Windows + extension: `n/a` — already unconditional. `core/ui/vault/swap/form/SwapForm.tsx` registers the `market` / `limit` tab set with no feature-flag check.
- SDK: `n/a` — no UI-level flag exists. `packages/core/chain/swap/native/limitSwapMemo.ts` is memo construction only.

iOS was the only platform hiding the tab row, so this closes a divergence rather than opening one.

## Checklist

> No tests added, and none removed. Grepped: **no test in `VultisigAppTests` or `VultisigAppUITests` referenced `limitSwapEnabled` or `limitSwapToggle`**, `VultisigAppUITests` has no limit-swap coverage at all, and no snapshot test covers `SwapDetailsScreen` — so nothing asserted flag-off behaviour and there was nothing to rewrite. The change is a deletion of a visibility gate; there is no new logic to assert.

_None._ See "Not verified" below — the reviewer's device pass is the right place for this.

## Additional context

**Not verified here, and worth a manual pass:** that an existing install whose stored `limitSwapEnabled` is `false` renders the tab row correctly on device. The reasoning is sound — nothing reads the key any more, so the stored value is inert — but that is reasoning, not an observation, and this was not run on a device that previously had the toggle off.

Per the issue, the orphaned `limitSwapEnabled` value left in `UserDefaults` on existing installs is **out of scope**: it is inert once unread, and a cleanup migration is not worth adding a keyshare-adjacent code path.

- **Agent**: Claude Code
- **Issue**: #5000
- **Confidence**: high
- **Needs human review for**: on-device confirmation that the tab row appears for a user who previously had the toggle off; the "tab row now shows for vaults with no THORChain-routable chains" note above

## Summary by CodeRabbit

* **Changes**
  * Removed the Limit Swap toggle from Advanced Settings.
  * Limit Swap availability in Swap Details is now determined by supported trading pairs rather than the removed setting.
  * Simplified the Swap Details toolbar; refresh and countdown information now appears within the tab interface.
* **Localization**
  * Removed obsolete Limit Swap toggle translations across supported languages.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -751,7 +751,6 @@
 "limitSwap.verify.title" = "Du platzierst eine Limit-Order";
 "limitSwap.warning.priceAtOrBelowMarket" = "Kurs bereits verfügbar, erwäge einen Markt-Swap";
 "limitSwap.warning.priceFarAboveMarket" = "Wird möglicherweise nicht vor Ablauf ausgeführt";
-"limitSwapToggle" = "Limit-Swap";
 "liquidityPools" = "Liquiditätspools";
 "loading" = "Loading...";
 "loadingDetails" = "Lade Details";
--- a/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
@@ -749,7 +749,6 @@
 "limitSwap.verify.title" = "You're placing a limit order";
 "limitSwap.warning.priceAtOrBelowMarket" = "Price already available, consider market swap";
 "limitSwap.warning.priceFarAboveMarket" = "May not fill within expiry";
-"limitSwapToggle" = "Limit Swap";
 "liquidityPools" = "Liquidity Pools";
 "loading" = "Loading...";
 "loadingDetails" = "Loading details";
--- a/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
@@ -751,7 +751,6 @@
 "limitSwap.verify.title" = "Estás colocando una orden límite";
 "limitSwap.warning.priceAtOrBelowMarket" = "Precio ya disponible, considera un swap de mercado";
 "limitSwap.warning.priceFarAboveMarket" = "Puede no ejecutarse antes del vencimiento";
-"limitSwapToggle" = "Swap límite";
 "liquidityPools" = "Pools de liquidez";
 "loading" = "Loading...";
 "loadingDetails" = "Cargando detalles";
--- a/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
@@ -751,7 +751,6 @@
 "limitSwap.verify.title" = "Postavljaš limit nalog";
 "limitSwap.warning.priceAtOrBelowMarket" = "Cijena već dostupna, razmotri tržišnu zamjenu";
 "limitSwap.warning.priceFarAboveMarket" = "Možda se neće izvršiti prije isteka";
-"limitSwapToggle" = "Limit zamjena";
 "liquidityPools" = "Bazeni likvidnosti";
 "loading" = "Učitavanje...";
 "loadingDetails" = "Učitavanje detalja";
--- a/VultisigApp/VultisigApp/Core/Localizables/it.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/it.lproj/Localizable.strings
@@ -751,7 +751,6 @@
 "limitSwap.verify.title" = "Stai inserendo un ordine limite";
 "limitSwap.warning.priceAtOrBelowMarket" = "Prezzo già disponibile, valuta uno swap di mercato";
 "limitSwap.warning.priceFarAboveMarket" = "Potrebbe non eseguirsi entro la scadenza";
-"limitSwapToggle" = "Swap limite";
 "liquidityPools" = "Pool di liquidità";
 "loading" = "Caricamento...";
 "loadingDetails" = "Caricamento dettagli";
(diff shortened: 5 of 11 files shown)


johnnyluo (approved):
LGTM

PR #5006, opened 2026-07-30
fix(ripple): record an XRPL trust-line activation as an activation, not a quadrillion-token transfer

### LINKED ISSUE #5005: XRPL trust-line activation records in history as a 1,000,000,000,000,000 token transfer
## Summary

An XRPL trust-line activation (TrustSet) is persisted to local transaction history as a **token transfer of the trust-line limit**. A USDC activation records as a send of **1,000,000,000,000,000 USDC**.

The transaction is real and the user did pay for it — but it moved no tokens. It opened a trust line and locked an owner reserve.

## Mechanism

`RippleTrustLineLimit.defaultLimit` (`RippleTrustLine.swift:30`) is `10^(issuedCurrencyDecimals + 15)` = `10^30` base units, i.e. **1e15** display units. A TrustSet signs that as its `LimitAmount`, so the payload's amount field carries it.

**The codebase already knows a TrustSet's amount is not a transfer.** Two paths special-case it:

- `SendCryptoVerifyLogic.swift:88` — *"An XRPL TrustSet's amount is the trust-line LIMIT, not a transfer"*, branching on `tx.transactionType == .rippleTrustSet` at `:96`
- `SendDoneScreen.swift:57` — *"A TrustSet's amount IS the trust-line limit, so it has to claim the hero"*, routing to `RippleTrustSetPresentation.hero(for:)`

**The recording path does not.** `git grep -n "TrustSet" -- Components/TransactionDone/ Core/Services/TransactionStatus/` returns nothing. So `TransactionHistoryRecording.record()` falls through to:

```swift
TransactionHistoryRecorder.shared.recordSend(
    coin: payload.coin,                 // the token, e.g. USDC
    amountCrypto: payload.amountCrypto, // the trust-line LIMIT
    ...
)
```

Net effect: verify and the Done screen render the activation correctly, and then history contradicts both.

## Proposed fix — a typed row, not suppression

`TransactionHistoryData` already carries a `type` discriminator with a non-transfer case: `recordApprove` persists `type: .approve`. An ERC-20 approve and an XRPL TrustSet are the same category — a permissioning transaction that costs a fee and grants an allowance rather than moving funds.

So:

1. Add a trust-line-activation case to the history `type` enum.
2. Add a `recordTrustLineActivation` entry point on `TransactionHistoryRecorder`, modelled on `recordApprove` — no `amountCrypto` masquerading as a transfer.
3. Branch in `TransactionHistoryRecording.record()` on `transactionType == .rippleTrustSet` before the `recordSend` fallback.
4. Render the row using the copy `RippleTrustSetPresentation` already provides, so history agrees with the Done screen.

**Why not suppress the row** the way `isLimitOrderCancel` does: a cancel is suppressed because the order's own row already narrates its lifecycle and links the hash, so nothing is lost. A TrustSet has no parent row — suppressing it means the fee the user paid and the reserve they locked disappear from history entirely. It should be visible and correctly labelled, not hidden.

## Out of scope

Cleaning up rows already written by 1.43.69. The app is not in production, so there is no user data worth migrating.

## Acceptance

- [ ] A trust-line activation records as an activation, not a send
- [ ] No history row displays the trust-line limit as a transferred amount
- [ ] The row still shows the fee and links to the explorer
- [ ] History agrees with what the Done screen showed
- [ ] Full test suite green

gastonm5:
## Description

Fixes #5005

An XRPL trust-line activation (`TrustSet`) was persisted to local transaction history as a **token transfer of the trust-line limit**. Activating USDC recorded as a send of **1,000,000,000,000,000 USDC**. The transaction is real and the user paid for it — but it moved nothing. It opened a line and locked an owner reserve.

The codebase already knew a TrustSet's amount is not a transfer. `SendCryptoVerifyLogic.validateBalanceWithFee` skips the balance check for one, `SendCryptoVerifySummaryView` renders labelled issuer / currency / limit rows instead of Payment rows, and both done screens route it through `RippleTrustSetPresentation.hero`. Only the *recording* path did not — it fell through to `recordSend` — so history contradicted the receipt the user had just been shown.

### What the row shows now

A typed `.trustLineActivation` row, modelled on `recordApprove`: an ERC-20 approve and a TrustSet are the same category, a permissioning transaction that grants an allowance rather than moving funds.

| | Before | After |
|---|---|---|
| Type pill | `Send` | `Trust line` |
| Body | `1000000000000000 USDC` + a fiat price for it | `Activating the USDC trust line`, over the issuer |
| Card layout | expanded amount → recipient diagram | collapsed; it has neither |
| Detail sheet header | `1000000000000000` in the amount slot + fiat | the same sentence + issuer, matching the done screen |
| Detail sheet address row | `To  rMxCK…m5De` | `Issuer  rMxCK…m5De` |
| Fee row | fee | fee (unchanged) |
| Explorer link | works | works (unchanged) |
| Amount fields on the persisted row | the limit | **empty** |

The copy is not new wording invented for history — `RippleTrustSetPresentation.activationTitle` is extracted from the existing done-screen hero and used by both, so the receipt and the row cannot drift apart again.

**Recorded, not suppressed.** A limit-order cancel gets no row because the *order's* row narrates its lifecycle and keeps the hash linked, so nothing is lost. A TrustSet has no parent row — suppressing it would erase the fee the user paid and the ~0.2 XRP owner reserve they locked. It stays visible and is labelled honestly.

### Paths verified

- **Initiator** (`SendDoneScreen` → `DoneScreen.onAppear` → `TransactionHistoryRecording.record`) — **was affected, fixed.** `amountCrypto` was `"\(tx.amount) \(ticker)"`, i.e. the limit.
- **Co-signer** (`JoinKeysignDoneView.sendBranch` → the same `DoneScreen` → the same `record`) — **was affected, fixed by the same branch.** It builds a different `amountCrypto` (`keysignPayload.toAmountWithTickerString`) but hits the identical fallback.
- **`recordFromKeysignPayload`** — **checked, not affected.** A TrustSet never reaches it: that dispatch requires `swapPayload != nil` or a limit-swap memo, and a TrustSet has neither. Both devices therefore shared *one* defect at *one* line. A defensive gate was still added at the top of that method, because its `else` arm is an unconditional `recordSend` and it is the choke point every payload-driven row funnels through (same rationale `TransactionStatusPoller` gives for duplicating its tracker gate).
- **Downstream `type` consumers** — every switch and filter over `TransactionHistoryType` was checked: type pill (both switches), card expansion + collapsed content + address pill, detail sheet header / from-to cards / address label, the Swaps / Limit Orders / Send tab filters, `SwapTrackingRegistry` (metadata-driven, not type-driven), `TransactionStatusPoller` (chain-driven — the row still polls and still reaches successful/error), the history reset service. No exhaustive switch was left with a silent fall-through.
- **Empty amount fields** — the card and the detail sheet both branch away from the amount/fiat slots for this type, so nothing formats them; no service parses `amountCrypto`/`amountFiat`.

### What I could not verify

**I did not broadcast a real TrustSet.** That needs a funded XRP account and a live signing ceremony, neither of which is available here. The fix is verified by unit tests over the pure routing and row-construction seams and by reading every render path end to end — **not** by observing a row on a device. The rendering in particular (pill, card, sheet) is verified by code inspection and by the routing tests, not by a screenshot.

### Out of scope

- **Rows already written by 1.43.69 keep their `.send` type. No migration** — the app is not in production, so there is no data worth migrating (confirmed on the issue).
- **This bug class is not unique to XRPL.** The Codex review pass turned up several non-transfer operations still recorded as `.send`, the closest siblings being Cosmos **withdraw-rewards** (`.send` with an empty amount and the first validator as the destination), Solana **deactivate/unstake** (`.send 0 SOL` to a stake-account pubkey), THORChain **unbond/leave** (`.send 0`, real quantity in the memo), **governance vote** (`.send 0`), and the **dApp raw-sign paths** (`signRipple` / `signSui` / `signSolana`, whose rows are built from generic `toAmount`/`toAddress` fields that need not describe the opaque signed bytes). Root cause: `TransactionBuilder` turns every function transaction into a `SendTransaction` and the done screen persists its generic fields verbatim. Deliberately not folded in here — it wants its own issue, and this PR establishes the shape of the fix.

## Which feature is affected?
- [x] New Chain / Chain related feature  -  Please attach tx link here

No tx link: this changes how an already-broadcast transaction is *recorded locally*, and broadcasting a real TrustSet is not possible from this environment. Nothing on the signing, broadcast or fee path is touched.

## Parity

- Android: `first: XRPL trust-line activation` — no other platform can originate a TrustSet yet (iOS is the first to build XRPL token payloads at all), so the defect cannot exist there. Raise the sibling ticket when XRPL token support lands.
- Windows + extension: `first: XRPL trust-line activation` — same reason.
- SDK: `n/a: local transaction-history presentation, no SDK surface`

## Checklist

- [x] My changes generate no new warnings — SwiftLint 36 violations before and after (0 new)

## Additional context

`TrustLineActivationRecordingTests` covers two seams, because they fail in different ways:

- **Routing.** The precedence chain in `TransactionHistoryRecording` was extracted into a pure `route(for:) -> Route`, so the *order* of the dispatch is testable, not only the individual predicates. That matters: the bug was never a missing predicate, it was a branch in the wrong place — and a suite that only asserts "this payload IS a TrustSet" stays green through exactly that mistake. Tests assert a TrustSet routes `.trustLineActivation` and is never `.send` (readable, plain and *unreadable* token ids), on both the initiator's and the co-signer's payload shape, and that the new branch displaced none of the pre-existing routes (empty hash, preview payload, limit-order cancel, swap dispatch, plain send, `.claim`).
- **Row content.** Asserted field by field, then swept: no field of the persisted row may contain the limit string, written as a sweep so a field added later cannot quietly become the next place it leaks. A companion test proves the fixture really carries that string *and* that it is the value `RippleTrustLineLimit` actually signs — without it the sweep could pass vacuously.

Two further payloads pin the branch's *position*, not just its existence — a TrustSet carrying a limit-order memo (which genuinely qualifies for the keysign dispatch) and one with the `.claim` verb. Both are impossible on-chain; their only job is to make a demotion past either guard observable.

**Verified by mutation, not by assertion alone.** Deleting `if isTrustLineActivation(payload) { return .trustLineActivation }` and re-running the suite turns **9 of 14 tests red**, including all four route tests. Restored afterwards; the committed tree is the unmutated one.

One new user-facing string (`rippleTrustLineTypePill`), added to all **8** locales including `ko`, sorted with `sort_localizable.py`. Everything else reuses keys the trust-line flow already ships.

Full suite: **4390 tests, 1 skipped, 0 failures** (`** TEST SUCCEEDED **`), verified by the log marker.

- **Agent**: Claude Code (Opus 5)
- **Issue**: #5005
- **Confidence**: high on the recording path (unit-tested end to end through the dispatch); medium on the rendering, which is verified by code inspection rather than by running the screen
- **Needs human review for**: the visual treatment of the new row — the pill icon (`chain-link-3`), the card's two-line activation column and the detail-sheet header are design choices made without a Figma reference

## Summary by CodeRabbit

* **New Features**
  * Added dedicated XRPL trust-line activation entries to transaction history.
  * Trust-line records display the activation title, issuer address, chain-link icon, and localized type label.
  * Trust-line transactions are included in the Send history tab and support localized presentation in multiple languages.

* **Bug Fixes**
  * Improved transaction routing so trust-line activations are not incorrectly recorded as sends.
  * Added fallback handling when ticker or issuer details are unavailable.

* **Tests**
  * Added comprehensive coverage for recording, routing, persistence, and display of trust-line activations.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ripple/Models/RippleTrustSetPresentation.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ripple/Models/RippleTrustSetPresentation.swift
@@ -148,9 +148,17 @@ enum RippleTrustSetPresentation {
     }
 
     private static func heroContent(for display: Display) -> HeroContent {
-        .title(
-            text: String(format: "rippleTrustLineHeroTitle".localized, display.ticker),
-            caption: display.issuer
-        )
+        .title(text: activationTitle(ticker: display.ticker), caption: display.issuer)
+    }
+
+    /// The one sentence that names what a TrustSet did, given the trusted
+    /// currency's ticker.
+    ///
+    /// Shared by the done-screen hero and the transaction-history row on
+    /// purpose. History used to record an activation as a token transfer of the
+    /// limit, contradicting the receipt the user had just been shown; the two
+    /// surfaces now read from the same string so they cannot drift apart again.
+    static func activationTitle(ticker: String) -> String {
+        String(format: "rippleTrustLineHeroTitle".localized, ticker)
     }
 }
--- a/VultisigApp/VultisigApp/Components/TransactionDone/TransactionHistoryRecording.swift
+++ b/VultisigApp/VultisigApp/Components/TransactionDone/TransactionHistoryRecording.swift
@@ -21,23 +21,38 @@ private let logger = Log.wallet.other
 
 enum TransactionHistoryRecording {
 
-    /// Records the send-style transaction reflected in `payload` to tx
-    /// history. No-op when the payload doesn't carry a vault
-    /// `pubKeyECDSA` (e.g. previews), when the hash is empty, or when
-    /// the verb is `.claim` (QBTC has no tx-history schema yet).
+    /// What `record` will do with a payload.
     ///
-    /// Swap-side flows pre-record via `recordSwap` + `recordApprove`
-    /// from their screen layer (they carry from/to coin pairs we don't
-    /// surface on the unified payload) — those payloads ship with
-    /// `isSend == false` and no `keysignPayload`, so this helper skips
-    /// them to avoid double-counting. Cosigner paths with a
-    /// `KeysignPayload` are dispatched here via
-    /// `recordFromKeysignPayload` so the cosigner's swap-branch gets
-    /// captured too.
-    @MainActor
-    static func record(payload: TransactionDonePayload) {
-        guard payload.pubKeyECDSA.isNotEmpty else { return }
-        guard payload.hash.isNotEmpty else { return }
+    /// Split out from `record` so the DISPATCH ORDER is testable, not just the
+    /// individual predicates. The predicates were already pure; the precedence
+    /// between them was not, and precedence is where this dispatch has actually
+    /// gone wrong — a TrustSet reaching `recordSend` is a branch that exists but
+    /// sits in the wrong place, and a suite that only asserts "the payload IS a
+    /// TrustSet" stays green through exactly that mistake.
+    enum Route: Equatable {
+        /// No row. Previews and un-broadcast payloads, QBTC claims, limit-order
+        /// cancels, and the swap initiator (which pre-records in its own screen
+        /// layer, so recording here would double-count).
+        case skip
+        /// An XRPL trust-line activation.
+        case trustLineActivation
+        /// Delegated to `recordFromKeysignPayload`, which reads the swap-or-order
+        /// discriminator off the payload itself.
+        case keysignPayload
+        /// The plain send row.
+        case send
+    }
+
+    /// Which row `payload` earns, and — the part that matters — in which order
+    /// the question is asked.
+    ///
+    /// Pure and `static` so the whole precedence chain can be pinned by tests:
+    /// the recorder it drives is a `private init()` singleton that writes to
+    /// SwiftData, so `record` itself isn't reachable from a unit test.
+    static func route(for payload: TransactionDonePayload) -> Route {
+        guard payload.pubKeyECDSA.isNotEmpty else { return .skip }
+        guard payload.hash.isNotEmpty else { return .skip }
+
         // A limit-order CANCEL gets no row of its own. It is a step in an
         // order's life, not a transfer the user made, and recording it produces
         // a standalone "Send 0 RUNE" — or, on the L1 route, a send of dust —
@@ -49,9 +64,9 @@ enum TransactionHistoryRecording {
         // auditable: its hash is persisted on the `LimitOrder` and the order's
         // detail sheet links it to the block explorer, so the fee and the
         // transaction itself remain one tap away.
-        guard !isLimitOrderCancel(payload) else { return }
+        guard !isLimitOrderCancel(payload) else { return .skip }
 
-        // An XRPL trust-line activation is not a transfer. Branched BEFORE
+        // An XRPL trust-line activation is not a transfer. Asked BEFORE
         // everything below it, because every route beneath this line ends in a
         // `recordSend` that would persist the trust-line LIMIT as an amount sent
         // — 1,000,000,000,000,000 USDC for a USDC activation.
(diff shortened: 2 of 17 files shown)


johnnyluo (approved):
LGTM

PR #5010, opened 2026-07-31
fix: heal push notification registration

### LINKED ISSUE #4971: Push registration silently no-ops when the APNs token isn't in memory yet — and never heals
`PushNotificationManager.registerVault` guards on `deviceToken` being set — but `deviceToken` is never hydrated from keychain at init, and every opt-in entry point (intro sheet, settings screen, per-vault toggle) requests permission and registers **in the same task**, before the APNs callback can have delivered a token. The opt-in lands in SwiftData as ON; the server never hears about it.

Nothing heals afterwards: registration only re-fires when the token string *changes* (`setDeviceToken` early-returns otherwise), the `isVaultRegistered` reconcile call was removed as dead code (073d11fe9), and `reRegisterOptedInVaults(_:)` has zero call sites. Result: toggle ON, zero pushes, forever — the top candidate for users whose co-signing device never receives anything.

**Fix:** hydrate the token from keychain at init; on token arrival re-register opted-in vaults unconditionally; restore an `isVaultRegistered` reconcile pass on launch.

Receipts: `Core/Notifications/PushNotificationManager.swift` (guard in `registerVault`, early-return in `setDeviceToken`, `init` without hydration).

johnnyluo:
## Description

Hydrates the in-memory APNs token from keychain when `PushNotificationManager` is initialized and reconciles every opted-in vault whenever APNs supplies a token, even when that token is unchanged. This closes the race where local notification opt-in was saved before the APNs callback, while the server registration silently no-op'd and was never retried.

Closes #4971

## Which feature is affected?
- [ ] Swap - Please attach tx link for the swap here

## Checklist

## Test Plan

- [x] Focused `PushNotificationManagerTests`: 3 passed
- [x] Full unit suite: 4,575 passed, 0 failed, 1 skipped
- [x] `make build-check`: build succeeded
- [x] SwiftLint: changed files have zero violations; repository retains 39 pre-existing warnings

Not applicable; no UI changes.

## Additional context

The existing vault-level `isVaultRegistered` endpoint was not restored because it cannot verify whether this specific party/token is registered. Re-registering opted-in vaults on APNs token delivery provides device-specific reconciliation using the existing registration request.

- **Agent**: Codex
- **Issue**: #4971
- **Confidence**: high
- **Needs human review for**: none

Co-Authored-By: Codex <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved push notification registration by restoring the saved device token when the app starts.
  * Ensured opted-in vaults are re-registered even when the device token has not changed.
  * Continued excluding opted-out vaults from notification registration.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Notifications/PushNotificationManager.swift
+++ b/VultisigApp/VultisigApp/Core/Notifications/PushNotificationManager.swift
@@ -35,6 +35,7 @@ class PushNotificationManager: ObservableObject {
     ) {
         self.notificationService = notificationService
         self.keychainService = keychainService
+        deviceToken = keychainService.getDeviceToken()
     }
 
     // MARK: - Notification Delegate
@@ -83,12 +84,12 @@ class PushNotificationManager: ObservableObject {
         deviceToken = tokenString
         keychainService.setDeviceToken(tokenString)
 
-        guard tokenString != previousToken else {
-            logger.info("Device token unchanged, skipping re-registration")
-            return
+        if tokenString == previousToken {
+            logger.info("Device token unchanged, reconciling vault registrations")
+        } else {
+            logger.info("Device token changed, reconciling vault registrations")
         }
 
-        logger.info("Device token changed, re-registering vaults")
         Task {
             await reRegisterOptedInVaults()
         }
--- a/VultisigApp/VultisigApp/Core/Notifications/PushNotificationManager.swift
+++ b/VultisigApp/VultisigApp/Core/Notifications/PushNotificationManager.swift
@@ -29,6 +29,12 @@ class PushNotificationManager: ObservableObject {
 
     private let notificationDelegate = NotificationDelegate()
 
+    /// Set when `registerVault` is called before a device token is available
+    /// (the opt-in race from #4971: permission just granted, APNs hasn't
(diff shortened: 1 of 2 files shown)


VultisigApp/VultisigApp/Core/Notifications/PushNotificationManager.swift
```diff
@@ -83,12 +84,12 @@ class PushNotificationManager: ObservableObject {
         deviceToken = tokenString
         keychainService.setDeviceToken(tokenString)
 
-        guard tokenString != previousToken else {
-            logger.info("Device token unchanged, skipping re-registration")
-            return
+        if tokenString == previousToken {
```
johnnyluo: can you please not remove this check ,

johnnyluo: Resolved in 32ffce1c2. The unchanged-token guard is restored: ordinary unchanged APNs tokens now skip re-registration. A pending flag is set only when registration was attempted before a token existed, allowing the next token callback to reconcile that deferred opt-in. Added tests cover both the skip and pending catch-up paths.

johnnyluo:
The fix is not required

PR #5033, opened 2026-08-03
fix(limit-swap): derive the cancel dust ceiling from a pinned threshold, and make the coverage test structural

### LINKED ISSUE #5004: Limit-order cancel: dust ceiling has no XRP/TRON/SOL entry, so a cancel from any of them would be refused (latent; live on Android in vultisig-android#5464)
`limitOrderCancelDustCeiling` has no entry for XRP, TRON or SOL, and its `default` of `0.001` natural units is far below the dust THORChain actually requires on all three. Any cancel from one of those chains would throw `dustAmountExceedsCeiling` and could never be built.

This is **latent today, not a live bug** — see "Why it does not bite yet" below. Filing it because the trap is invisible at the point where someone would spring it.

## The numbers

`dust_threshold` is published in THORChain's 1e8 fixed point on every chain. Live values from https://thornode.ninerealms.com/thorchain/inbound_addresses (fetched 2026-07-30):

| chain | `dust_threshold` | native decimals | rescaled | attach (x2) | ceiling from `default: 0.001` | over by |
|---|---|---|---|---|---|---|
| XRP  | `100000000` | 6 | 1 XRP | **2 XRP** | 0.001 XRP (1,000 drops) | 2000x |
| TRON | `10000000`  | 6 | 0.1 TRX | **0.2 TRX** | 0.001 TRX (1,000 sun) | 200x |
| SOL  | `100000`    | 9 | 0.001 SOL | **0.002 SOL** | 0.001 SOL | 2x |

`limitOrderCancelDustAmount` computes `max(walletCoreDustFloor, rescaled) * dustSafetyMultiple` and then enforces `amount <= ceiling`, so every one of these throws.

The table's own sizing rule is stated in its doc comment — "roughly an order of magnitude above each chain's live `dust_threshold` doubled" — and the captured list it was sized against (DOGE, LTC, AVAX, GAIA, BCH, BSC, BTC, ETH) simply does not contain XRP, TRON or SOL. They fall through to `default`, which was written for ETH/BSC-scale thresholds.

## Why it does not bite yet

`thorchainChainPrefix(for:)` in `VultisigApp/VultisigApp/Features/LimitSwap/Logic/THORChainAssetSymbol.swift:150` returns a prefix only for THOR, ETH, AVAX, BSC, BTC, BCH, LTC, DOGE and GAIA, and `nil` for everything else. `chainFromThorchainSymbol` (same file, :223) is its strict inverse, so `computeSupportedChains(from:)` drops the XRP, TRON and SOL inbound rows and those chains can never be selected as a limit-order source.

So the ceiling is unreachable for them right now.

## Why it is worth fixing anyway

THORChain routes XRP, TRON and SOL today and all three are unhalted. Adding one to `thorchainChainPrefix` is a two-line change that looks complete and self-contained — placement would work, and **cancel would silently stop working on that chain**, surfacing only as "the amount a cancellation must attach could not be determined safely, try again shortly", a message that never resolves. Nothing in `thorchainChainPrefix` points at the dust table.

Android is the same table with a wider source set, so there it is live: `thorchainMemoAssetChainPrefix` already includes Ripple, Solana and Tron, and cancel is dead on all three. Raised on https://github.com/vultisig/vultisig-android/pull/5464.

## Suggested fix

Either:

1. Give the three chains explicit cases sized by the table's own rule (roughly an order of magnitude above the doubled attach) — e.g. `.ripple: 10`, `.tron: 2`, `.solana: 0.02`; or
2. Better, stop hardcoding a natural-units ceiling per chain and derive it from the `dust_threshold` the app already fetched, e.g. `ceiling = rescaled_threshold * K`. That keeps the "a bad remote value cannot quietly donate a meaningful sum" guarantee the ceiling exists for, without a table that silently goes stale every time a chain is added.

Either way, extend `LimitOrderCancelDustTests.swift:251` to assert, for **every** chain in the supported set, that the computed attach is `<= ceiling` against the published thresholds. That test is what makes the next chain addition fail loudly instead of silently.

## Files

- `VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitOrderCancelDust.swift:216` — the ceiling table
- `VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitOrderCancelPayloadAssembler.swift:98` — the only caller
- `VultisigApp/VultisigApp/Features/LimitSwap/Logic/THORChainAssetSymbol.swift:150` — the gate that keeps it latent

gastonm5:
## The trap

`limitOrderCancelDustCeiling(for:)` was a hand-computed per-chain table in **natural** units with a `default: 0.001`. XRP, TRON and SOL had no entry, and that default sits far below the dust THORChain actually requires on all three — so a cancel from any of them would throw `dustAmountExceedsCeiling` and could never be built. The user would see only *"the amount a cancellation must attach could not be determined safely, try again shortly"*, a message that never resolves.

`dust_threshold` is published in THORChain's 1e8 fixed point on **every** chain, whatever precision that chain uses. Re-fetched from `/thorchain/inbound_addresses` on 2026-08-03:

| chain | `dust_threshold` | decimals | attach (×2) | old ceiling | |
|---|---|---|---|---|---|
| XRP | `100000000` | 6 | **2 XRP** | 0.001 XRP | refused, 2000× short |
| TRON | `10000000` | 6 | **0.2 TRX** | 0.001 TRX | refused, 200× short |
| SOL | `100000` | 9 | **0.002 SOL** | 0.001 SOL | refused, 2× short |
| BASE | `1000` | 18 | 0.00002 ETH | 0.001 ETH | passed *by luck* |

**Latent today, not a live bug.** `thorchainChainPrefix(for:)` returns `nil` for those chains and `chainFromThorchainSymbol` is its strict inverse, so `computeSupportedChains(from:)` drops their inbound rows and they can never be selected as a limit-order source. It is worth fixing anyway because adding one to `thorchainChainPrefix` is a two-line change that looks complete and self-contained — placement would work, and **cancel would silently stop working on that chain**. Nothing at that gate points at the dust table. Android is the same table with a wider source set, so there it is live (vultisig-android#5464).

Two things the issue does not state, found while verifying: **BASE is a 12th published chain**, and it survives the same `default` purely because its threshold happens to equal Ethereum's. And **THORChain publishes no inbound row for DASH, ZCASH or NOBLE**, though all three carried ceiling entries.

## Why the ceiling could not simply be derived

The issue's option 2 is `ceiling = rescaled_threshold × K`. **A purely derived ceiling is vacuous and does not ship here.**

The ceiling exists for one reason, stated in `dustAmountExceedsCeiling`'s own doc comment: `dust_threshold` is a **remote** value that decides how much of the user's money is irreversibly `donateToPool`'d, with no refund path. It is the only upper bound in the file; every other floor is a lower bound. The attach is `threshold × 2`, so any ceiling of the form `threshold × K` with `K ≥ 2` is satisfied by *every* possible remote value, however hostile. The guard would still be there and would never fire again.

`min(derived, absoluteCap)` does not rescue it either — the derived term never binds, so the expression collapses to the cap. `max(...)` is worse: it lets the remote value raise its own cap.

**The binding bound has to be local.** What a formula can remove is the hand-arithmetic, and that is what this does.

## What changed

Pin, per chain, the `dust_threshold` THORChain publishes — **in THORChain's own 1e8 fixed point, the same unit system the live value arrives in** — and derive the ceiling from it:

```
ceiling = chainSmallestUnits(pinned × limitOrderCancelDustSafetyMultiple × limitOrderCancelDustCeilingHeadroom)
```

- **The pinned numbers are verbatim copies of the endpoint.** Verifying an entry is a `curl` and a string compare, not a 1e8 → native rescale done in someone's head. That rescale is exactly what went wrong before: the file records that the table two revisions ago was sized against the EVM `ConvertAmount` floor instead and so sat *below* the real attach on both LTC and AVAX, refusing cancels users were entitled to make.
- **The rescale is now done by `chainSmallestUnits`** — the same function the amount goes through, so the ceiling and the amount cannot end up in different unit systems.
- **The ceiling is composed from `limitOrderCancelDustSafetyMultiple`**, so raising the safety multiple widens the ceiling with it instead of silently eating the margin.
- **`headroom = 10`** is the rule the old table documented for itself ("roughly an order of magnitude above each chain's live `dust_threshold` doubled") and the centre of the spread it actually implemented, which ranged 5× (DOGE, BCH, BSC) to 50× (BTC, ETH) with no rationale on any row.
- **No `default:` branch.** An unpinned chain throws `dustCeilingUnpinned` naming itself. It routes through the existing `.dust(...)` mapping, so no new user-facing string.

All 12 published chains are pinned, including BASE/XRP/TRON/SOL which the prefix gate does not yet admit — pinning ahead of the gate is what stops the next chain addition from breaking cancellation. DASH/ZCASH/NOBLE are dropped: no published threshold means nothing honest to pin, and a cancel from them is impossible upstream anyway (`resolveThorchainInboundVault` refuses them at `isThorchainRoutable`). Inventing numbers for them is the defect this removes.

## What the ceiling still guarantees

**Unchanged:** the most a wrong or hostile `dust_threshold` can cause the app to donate is a locally-known, per-chain constant, independent of anything the remote says. After this it is uniformly **10× the normal attach** on every chain.

| | BTC | ETH | GAIA | LTC | AVAX | BCH | BSC | DOGE |
|---|---|---|---|---|---|---|---|---|
| old | 0.001 | 0.001 | 0.5 | 0.02 | 0.02 | 0.001 | 0.001 | 10 |
| new | 0.0002 | 0.0002 | 0.2 | 0.02 | 0.02 | 0.002 | 0.002 | 20 |

BTC and ETH get **5× tighter**, GAIA 2.5× tighter, LTC/AVAX unchanged; BCH/BSC/DOGE 2× looser, which is the price of a uniform rule and still ~$1–4 of exposure.

**Never guaranteed, before or after:** a threshold move *within* the headroom is honoured verbatim. That is the point of headroom — a legitimate protocol change should not brick cancelling. Beyond 10× it fails loudly with both numbers, and the fix is a one-number edit copied from the endpoint.

**Considered and rejected:** a fiat-denominated cap. It is the conceptually right bound — the hazard is "a meaningful sum of money", not "a number of drops" — but rates are themselves remote, the lookup is async, and it would drag `RateProvider` into a function that is currently pure and unit-testable without any of it. Noted as a deliberate choice, not an oversight.

## The structural test

**This is the more important half.** `testEveryChainsPublishedThresholdSitsUnderItsCeiling` already existed — but it looped a hardcoded 8-entry array, *the same captured list the ceiling table itself was sized against*. It could only ever confirm what the table already said, and was blind to a chain nobody remembered to add. That is structurally why XRP/TRON/SOL were missed.

The set is now composed from the production predicates and never listed:

```swift
Chain.allCases.filter { isThorchainRoutable(chain: $0)
                        && !limitOrderCancelIsThorchainSourced(sourceChainRawValue: $0.rawValue) }
```

Add a chain to `thorchainChainPrefix` without pinning its threshold and `testEveryChainACancelCanBeSentFromHasAPinnedThreshold` fails in CI, telling you to read the value off `/thorchain/inbound_addresses`. Native precision comes from the app's own coin catalog rather than a literal, so a chain's decimals cannot be mis-stated here either.

Also asserted: the pinned table is verbatim what the endpoint publishes **and pins nothing it does not**; twelve hand-computed ceilings as literals in both smallest and natural units (the one place the formula is checked against independent arithmetic); the headroom boundary from both sides; that WalletCore's local signer floor — the *other* floor, which the ceiling is not derived from — also fits; and the XRP/TRON/SOL/BASE regression in its own numbers, each row proving the attach really did exceed the removed default before proving it fits now.

## Out of scope

Adding XRP/TRON/SOL/BASE to `thorchainChainPrefix`. That would make them user-selectable limit-swap sources and needs its own memo/routing verification. This only removes the trap that would spring when someone does.

## Verification

- `make test` on a dedicated simulator: **`** TEST SUCCEEDED **`** — 4622 tests executed, 1 skipped, **0 failures**. `LimitOrderCancelDustTests` 40 tests, 0 failures (30 before).
- SwiftLint clean: 39 warnings repo-wide, **none in either touched file** (unchanged from `main`).
- Cross-model review: OpenAI Codex (gpt-5.5, read-only) gated each commit. Step 1 → 1 MINOR (fixed: `decimals >= 0` now guarded in `limitOrderCancelDustCeiling`, since Swift evaluates it as an *argument* to the function whose guard used to cover it — `Int.min` would have trapped on `8 - decimals`). Step 2 → 1 MAJOR (fixed: the headroom boundary derived both sides from the same constant and was tautological). Whole-branch pass: **No findings**. Nothing rejected, nothing deferred.

Closes #5004

### DIFF
--- a/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitOrderCancelDust.swift
+++ b/VultisigApp/VultisigApp/Features/LimitSwap/Logic/LimitOrderCancelDust.swift
@@ -34,6 +34,17 @@ enum LimitOrderCancelDustError: Error, Equatable {
     /// be honoured verbatim and then doubled. Every other floor in this file is
     /// a lower bound; this is the only upper one.
     case dustAmountExceedsCeiling(chain: String, computed: String, ceiling: String)
+    /// No `dust_threshold` is pinned for this chain, so there is no locally-known
+    /// bound on what a remote value could ask us to donate.
+    ///
+    /// Deliberately fatal, and deliberately not a default. A default is a number
+    /// nobody checked against the chain it is being applied to: the one this
+    /// replaced was 0.001 natural units, which sat three orders of magnitude
+    /// under XRP's real requirement and would have refused every cancel from it.
+    /// Refusing HERE says which chain is unconfigured; a wrong default says only
+    /// that some amount exceeded some ceiling, on a chain where cancelling then
+    /// never works.
+    case dustCeilingUnpinned(chain: String)
     /// The computed dust is too small for THORChain to observe at all.
     ///
     /// ⚠️ The loud failure that the 2026-07-22 ETH rehearsal needed and did not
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5046, opened 2026-08-03
refactor(design-system): standardise container corner radii on 24 via Theme.radius

gastonm5:
Part of #5024. Phase 2 of four.

> ⚠️ **Stacked on #5040, and must not merge before it.** The base branch is `feat/radius-tokens-5024`, not `main` — every line here calls `Theme.radius`, which only exists in that PR. Merging this first would not compile.

## What this does

Phase 1 (#5040) added the token scale and migrated 15 sites to prove the API, with zero visual change. This phase is the one that actually moves pixels: it standardises **container** surfaces on **24**.

> "The new corner radius is 24 standardized like we did for settings, main screen and swaps"

That instruction is what settles this, and it had to, because **the design file does not**. A second sampling round found Figma holding `Input fields/Default` at 16 in one place and 12 in another, and "card" at 24 (banner), 16 (transaction history) and 24-top/20-bottom (Vult tier) — three answers for the same component class. So this is not "sync the code to Figma"; there is no consistent source to sync to. The evidence is written up in the design-system notes attached to #5024.

## The rule: by role, from the call site

| | |
|---|---|
| **Container** — the outer bounding surface of a card, banner, panel, picker or list group; the thing sitting on the page background that holds other content | → `Theme.radius.xl` (24) |
| **Not a container** — anything nested inside one: icon tiles, avatar wells, rows within a group, badges, chips, progress tracks, skeleton blocks, inner detail panels | → an explicitly **smaller** step |
| **Fully round** — radius ≥ half the smaller dimension of an explicitly framed box | → `Theme.radius.pill`, pixel-identical |

**16 (84 sites) and 12 (67 sites) were the two largest populations in the codebase, and most of them are not containers.** Blanket-converting either would have been visually destructive. Of the 84 sites at 16, **49** were promoted; of the 67 at 12, **20**.

### Nesting: the two levels must differ

A first pass held `containerStyle` at 12 because promoting it produced 24 inside 24 on the referral screen. That diagnosis was right and the remedy was wrong: the problem is not the value, it is that **both levels were reading the same one**. Resolved by ruling that the **outer** level is the container:

- `containerStyle` defaults to the container step — of its eleven call sites, seven are the outer surface on their page.
- The four that are genuinely nested pass an explicit `Theme.radius.md`.

Deriving that list **from the call sites rather than the declarations** is the whole discipline here, and two of the four are not where the name suggests:

- **`ReferralCodeBoxView`'s value box** is nested in *both* of its usages — `referredCodeSection` wraps it in a container, and `yourReferralCodeView` renders it inside the referral-code section. It is never the outer surface anywhere.
- **`BannerView`** is nested in one usage and outer in the other: the "save on swaps" banner sits on the page, the "collected rewards" banner sits inside the referral section. One default cannot serve both, so it takes a `radius` parameter defaulting to the container step, and the nested call site passes the smaller one.

The remaining seven inherit the new default: both referral sections, both backup screens, and `ContainerView`, whose thirteen call sites are all top-level cards or rows — several of them DeFi position cards sitting at 12 while the staking cards beside them had already moved to 24.

### Ruling: form sections are containers

> "For phase 2, check VaultServerBackupScreen. Sections should have new corner radius/padding like we did for send/swaps etc"

Settles the last open "container or field group?" question. `FormExpandableSection` (all four initialisers) and `SendFormExpandableSection` — which also backs `FormPickerSection` — move their defaults from `md` to `xl`.

**All 19 call sites traced, and they are unanimous**: every one is a direct child of `FormScreen`'s content builder, and `FormScreen` draws no surface of its own, so every section is the outer container on its screen. Nothing needed an override. The inputs *inside* them — `CommonTextField`, `AmountTextField`, `FunctionAddressField` — stay at 12, so the nesting reads correctly in the direction that matters.

**Padding reference: the 2026 Send-details cards** (`SendDetailsAssetTab` / `AddressTab` / `AmountTab`) — because they are *this same view* with the new design already applied. Adopting their values as the defaults makes "like send" structural rather than a comment, and the three tabs stop restating what is now the default. Picking a number off a neighbouring card would have been a guess: the nearby cards do not agree with each other (send verify summary pads 24, swap verify summary pads 16, `BannerView` pads 24).

| | Before | After |
|---|---|---|
| Section corner radius | 12 | **24** |
| Section vertical padding | 16 | **20** |
| Section horizontal padding | 16 | 16 (unchanged) |

### Ruling: all referral banners take 24 — reversing an earlier call on this branch

> "For referral main screen, make sure all banners have the new corner radius also (collected rewards also)"

The collected-rewards banner drops the explicit `md` an earlier commit here gave it. **Flagging this as a reversal rather than letting it read as a tidy-up:** collected-rewards is nested inside `referralCodeSection`, which is also 24, so this is exactly the equal-radii-nested shape commit 11 argued against and stepped down to avoid.

What makes it hold up on screen is that `BannerView` applies **24pt of its own padding**, so the inner and outer edges are separated by a real gap rather than sitting a hairline apart — but this is the one change in the PR a reviewer should judge on a device rather than from the diff.

The `radius` parameter stays as the seam; there is simply no call site using it today.

**The nesting rule is not repealed.** `expiresOnView` is a plain nested content box in the same section and stays at 12, because the ruling named banners. Whether nested content boxes follow is the open question at the top of the list below.

### These two rulings cost zero literal edits

Both are **default changes**, so ~20 rendered surfaces (15 `FormExpandableSection`, 4 `FormPickerSection`, 1 banner) moved without touching a single radius literal. Before phase 1 this ruling would have been a 20-site sweep — that is the token layer paying for itself.

## Counts

| | |
|---|---|
| Raw radius literals at the base commit | **355** |
| Raw radius literals now | **181** |
| **Literal sites migrated to a token** | **174** |
| — of which **change a value** | **77** (16→24 ×49, 12→24 ×20, 10→24 ×4, 20→24 ×4) |
| — of which are **pixel-identical** | **97** (already-24 ×46, pill ×40, md ×6, lg ×2, + 3 multi-line rewrites) |
| Named `CGFloat` constants / parameters retyped to `CornerRadius` | **8** |

Only **8** raw `24` literals remain, and both groups are deliberate: the macOS one-time-code field and its paste button (an input row, out of scope) and `CrossPlatformSheet` ×4 (sheets keep the native radius).

## Commits — grouped by surface area

| # | What | Visual? |
|---|---|---|
| 1 | `pill` — the four "fully round" magic numbers collapse onto one token | no |
| 2 | Send / swap verify + done summary cards | **yes** |
| 3 | Keysign / keygen panels | **yes** |
| 4 | DeFi cards and banners — the largest block | **yes** |
| 5 | Transaction-history cards | **yes** |
| 6 | List containers + empty states | **yes** |
| 7 | Remaining option cards and promo banners | **yes** |
| 8 | The 46 containers already at 24 → read from the token; form-section `cornerRadius` retyped | no |
| 9 | Two container helpers that hid their radius behind a constant | **yes** (Tron) |
| 10 | Three more named radius constants → token | no |
| 11 | Fix: three nested-container cases put back to their inner step | corrective |
| 12 | Fix: three corner badges made to follow the surface they are drawn into | corrective |
| 13 | `containerStyle` defaults to the container step; the nested cases say so | **yes** |
| 14 | Address-book cells → container step; the circle idiom → `pill` | **yes** |
| 15 | `Core/Platform` sweep (see below) | **yes** (macOS) |
| 16 | Fix: two form-control containers kept matched to the fields beside them; scanner masks carry the token style | corrective |
| 17 | Form sections take the container step and the send-details padding | **yes** |
| 18 | Every referral banner takes the container step | **yes** |

## `Core/Platform` — a tree the earlier passes filtered out

Reported plainly, because a sweep that silently skips a tree is worse than one that says so.

**It was a deliberate exclusion, not an enumeration bug.** Every earlier commit filtered `Core/Platform/macOS` out on the grounds that the gate builds the iOS destination, so edits there would ship unverified. That was the wrong trade — this is one multiplatform target, macOS ships, and "unverified" is an argument for building macOS, not for leaving a tree alone. Commit 15 sweeps it, and **the macOS target is now built explicitly** (see Verification).

**Did the same enumeration miss anything else?** Checked directly. No directory — but yes, one *form*:

- Every directory in the app target does appear in the literal census; `App/`, `Blockchain/`, `Model/` and `Resources/` simply contain no radius literals.
- **`cornerSize: CGSize(width: 24, height: 24)` was invisible to the census**, because the number sits inside a `CGSize` rather than after a `cornerRadius:` label. Three sites, all scanner viewports — and this *had already caused a bug*: all three draw the frame twice, once as a stroked shape and once as the rounded punch-out in the dimming mask, and on iOS an earlier commit tokenised the stroke while leaving the mask a literal. That would have detached the dimming from the frame the next time the token moved. All three now read the same token.
- Two further forms turned up in the same check and are correctly out of scope: `textField.layer?.cornerRadius = 12` in an AppKit-bridged text field (an input), and `RectangleCornerRadii` on the address-field accessory buttons (input-adjacent).

macOS specifics: both scanner viewports now match their iOS twin; the "use a different email" control is a text chip, so it keeps 12 and reads it from the token. The one-time-code digit fields are input fields and are left alone, and the paste button beside them is sized and rounded to match that row, so it moves when the row does and not before.

## What changed, and why each one is a container

<details>
<summary><b>Value-changing sites — click to expand</b></summary>

| File | Old | New | Why it is a container |
|---|---|---|---|
| `Components/TransactionDone/DoneScreen.swift` | 16 ×4 | `xl` | Done-screen hero card + transaction-details card |
| `Components/TransactionDone/SwapDoneSummaryCard.swift` | 16 ×4 | `xl` | Swap done summary card + hero card |
| `Components/TransactionDone/SignedMessageDoneTokenContent.swift` | 12 | `xl` | Done-screen body card for the dApp-signing flow |
| `Features/Send/.../SendCryptoVerifySummaryView.swift` | 16 ×2 | `xl` | The card holding the transaction being signed |
| `Features/FunctionCall/Views/CosmosStakingVerifySummaryView.swift` | 16 ×2 | `xl` | Same card, staking flow |
| `Features/Swap/Views/Screens/SwapVerifyScreen.swift` | 10 | `xl` | Swap verify summary card |
| `Features/Swap/Views/Screens/SwapDoneScreen.swift` | 16 ×2 | `xl` | Limit-order info banner, anchored on the page |
| `Features/Keysign/Views/ErrorView.swift` | 16 ×2 | `xl` | Expandable error panel |
| `Features/Keysign/Views/KeysignSwapConfirmView.swift` | 10 | `xl` | Swap confirm summary card |
| `Features/Keysign/Views/KeysignCustomMessageConfirmView.swift` | 10 | `xl` | Custom-message confirm card (inner scroll panels stay at 16) |
| `Features/Keysign/Views/KeysignDiscoverServiceView.swift` | 10 | `xl` | Mediator-discovery panel |
| `Features/Keygen/Views/JoinKeygenView.swift` | 12 | `xl` | Join-keygen status panel |
| `Features/Defi/DefiMain/View/DefiMainBalanceView.swift` | 16 ×2 | `xl` | DeFi balance banner |
| `Features/Defi/DefiChain/View/Main/DefiChainBalanceView.swift` | 16 ×2 | `xl` | Per-chain balance banner |
| `Features/Defi/Common/Yield/View/YieldTopBanner.swift` | 16 ×2 | `xl` | Yield top banner |
| `Features/Defi/DefiChain/View/Governance/QBTCGovernanceView.swift` | 16 ×2 | `xl` | Governance empty/error state panel |
| `Features/Defi/DefiChain/View/Governance/GovernanceProposalRow.swift` | 16 ×2 | `xl` | "One proposal card in the governance list" |
| `Features/Defi/DefiChain/View/Governance/GovernanceProposalDetailScreen.swift` | 16 ×2 | `xl` | The screen's `card(title:content:)` helper |
| `Features/Defi/DefiChain/View/Staked/CosmosStakeDefiView.swift` | 16 ×4 | `xl` | Total-staked + active-delegations cards |
| `Features/Defi/DefiChain/View/Staked/SolanaStakeDefiView.swift` | 16 ×2 | `xl` | `cardBackground` / `cardBorder` |
| `Features/Defi/DefiChain/View/SelectPosition/DefiChainSelectPositionsScreen.swift` | 12 | `xl` | "No positions" empty-state panel |
| `Features/Defi/Protocols/Tron/TronConstants.swift` | 16 | `xl` | Drives 8 Tron dashboard / resources / actions cards |
| `Features/TransactionHistory/Views/TransactionHistoryCardView.swift` | 16 ×4 | `xl` | The history card, plus the "via" badge corner that traces it |
| `Features/TransactionHistory/Views/TransactionHistoryDetailSheet.swift` | 16 ×4 | `xl` | Hero card + details card |
| `Features/AddressBook/Views/AddressBookChainSelectionScreen.swift` | 12 | `xl` | List container clipping the scrolling group |
| `Features/FunctionTransaction/Ton/PoolSelection/TonPoolSelectionScreen.swift` | 12 | `xl` | List container |
| `Features/FunctionTransaction/Maya/AssetSelectionList/AssetSelectionListScreen.swift` | 12 | `xl` | List container |
| `Features/FunctionTransaction/SharedFunctions/.../StakingValidatorPickerScreen.swift` | 12 | `xl` | List container |
| `Features/Referral/Views/PreferredAssetSelectionView.swift` | 12 | `xl` | List container |
| `Features/Referral/Views/ReferralVaultSelectionScreen.swift` | 12 | `xl` | Vault list group |
| `Features/Wallet/TokenSelection/Screens/CustomTokenScreen.swift` | 12 ×2 | `xl` | Found-token list group |
| `Features/Wallet/ChainDetail/View/ChainDetailListView.swift` | 12 | `xl` | Empty state for a list already at 24 |
| `Features/Common/AssetSelection/ChainNotFoundEmptyStateView.swift` | 12 | `xl` | Empty-state panel |
| `Components/ActionBanner/ActionBannerView.swift` | 12 ×3 | `xl` | Bottom-docked empty-state panel (top corners only) |
| `Features/Onboarding/.../ImportVaultSelectionSheet.swift` | 16 | `xl` | Import-option card |
| `Features/Onboarding/.../DerivationPathSelectionSheet.swift` | 16 | `xl` | Derivation-path option card |
| `Features/Reshare/Views/BeforeYouReshareBottomSheet.swift` | 20 ×2 | `xl` | Reshare intro card |
| `Features/Reshare/Views/ReshareScreen.swift` | 20 ×2 | `xl` | Reshare option card |
| `Features/Referral/Views/ReferralInitialScreen.swift` | 16 | `xl` | Referral entry card |
| `Features/Referral/Views/ReferralSendOverviewView.swift` | 16 | `xl` | Send-overview summary card |
| `Features/QBTCClaim/Views/QBTCClaimSelectionView.swift` | 16 ×2 | `xl` | Claim top banner |
| `Features/QBTCClaim/Components/ClaimQbtcPromoBanner.swift` | 12 ×2 | `xl` | Full-width promo banner |
| `Features/FunctionCall/Models/FunctionCallAddThorLP.swift` | 12 | `xl` | Enable-Thorchain CTA panel |
| `Features/FunctionTransaction/Cosmos/.../CosmosWithdrawRewardsTransactionScreen.swift` | 16 | `xl` | Withdraw-rewards form panel (rows stay at 12) |
| `Components/Layout/View+Container.swift` | 12 | `xl` | The app's shared container chrome — 7 outer call sites, incl. `ContainerView` ×13 |
| `Components/Cells/AddressBookCell.swift` | 12 | `xl` | Each cell draws its own border on the page background; the list adds no group |
| `Components/Cells/SendCryptoAddressBookCell.swift` | 12 | `xl` | Same component class in the send flow |
| `Core/Platform/macOS/.../MacScannerView.swift` | 24 lit. | `xl` | Scanner viewport, matching its iOS twin |
| `Core/Platform/macOS/.../MacAddressScannerView+macOS.swift` | 24 lit. | `xl` | Scanner viewport |

</details>

Three explicit `style: .continuous` arguments went with the literals they qualified — the token carries the style, and a call site naming it again is deciding geometry the design system already decided. Verified equal: the SDK's implicit default is `.continuous`, which is what the token holds.

## Resolved rulings

- ✅ **Nesting** — the outer level is the container; nested surfaces take an explicit smaller step. The two levels must *differ*.
- ✅ **Circle idiom** — radius ≥ half the smaller dimension of an explicitly framed box means `pill`. Swept mechanically; 7 sites converted. The same sweep correctly left the 64×64 icon well at 16 and the 180×120 loader at 10.
- ✅ **Address-book cells** — containers.
- ✅ **`Core/Platform`** — in scope, both platforms.
- ✅ **Form accordions are containers** — defaults `xl` / 16 / 20, matching the 2026 Send-details cards.
- ✅ **Referral banners uniform at 24**, including the nested collected-rewards banner.

## Still needs a human call

**1. Do nested *content boxes* follow the banners to 24, or only banners?** ⭐ *This one likely settles a whole class — worth answering before the others.*
`ReferralMainScreen.expiresOnView` is `.containerStyle(padding: 14, radius: Theme.radius.md)`: a plain nested box inside the 24pt referral section, sitting directly beside the collected-rewards banner just ruled up to 24. The banner ruling named banners, so it was **not** extended here. If content boxes follow, the same answer also covers `ReferralCodeBoxView`'s value box and the two `containerStyle` cases on the backup screen — and the nesting rule effectively collapses into "everything on a page is 24".

2. **Inline alert / note strips** — `InformationNote` (12), `WarningView` (12), `InfoBannerView` (12), the Cosmos redelegate cooldown notice (12). Full-width surfaces on the page background, which makes them containers by the letter of the rule, but they read as callouts, not cards. *(`SwapDoneScreen`'s limit banner **was** promoted because it uses the card idiom — `bgSurface1` + hairline border — rather than the tinted-alert idiom these four use.)*
3. **`PeerDiscoveryScreen` QR frame (33, 8pt border)** — a container, but 33 looks deliberately matched to the QR module rounding and the thick border.
4. **`VultDiscountTierView` footer (20) / `LockedFeatureSheet` card (14)** — 20 is a *real* Figma value; the tier card is 24-top / 20-bottom on purpose.
5. **Selectable rows that are also standalone bordered surfaces** — `StakingValidatorCard`, `TonPoolCard`, `QBTCClaimSelectionView`'s UTXO row (all 16). Card or row?
6. **Dropdown containers (10)** — `GenericSelectorDropDown`, `FunctionCallSelectorDropdown`, `FunctionCallContractSelectorDropDown`.
7. **Drop zones (10)** — `ImportWalletUploadSection`, `FileQRCodeImporterMac`.
8. **Round-looking sites with no explicit frame** — the two swap-arrow badges, `SendDetailsAmountTextField`'s unit toggle, `SwapFromToChain`'s chevron (a probable no-op `.cornerRadius` on a bare glyph). The pill rule likely applies, but the box is intrinsic so the inequality cannot be *proved*, and the rule's value is that it never needs judgement.

## Input fields — untouched, listed for one ruling

Explicitly out of scope. Figma has them at both 12 and 16, the code has 12, and whether an input is a "container" is exactly the ambiguity above:

`CommonTextField` · `MemoTextField` · `DestinationTagTextField` · `StyledTextField` · `StyledNumberField` · `StyledFloatingPointField` · `HiddenTextField` · `CommonTextEditor` · `SearchTextField` · `GasLimitSettingsView`'s number field · `CustomRPCDetailScreen`'s endpoint editor · `SendDetailsAddressFields`' from-field · `AddressBookTextField` · `SendDetailsAmountTextField` · **`ServerBackupVerificationView`'s macOS one-time-code fields + paste button** · **`BackspaceDetectingTextField`'s AppKit `layer?.cornerRadius = 12`** · the `AddressFieldAccessoryStack` buttons

## Also left alone

- **Sheets.** `CrossPlatformSheet` (24) and `BottomSheetContainer` keep the native platform radius; the scale deliberately has no sheet step and none was added.
- **Nested panels inside a promoted card** — the eight `Sign*DisplayView` `bgSurface2` panels, `DisclosureSection`, the two scroll panels in `KeysignCustomMessageConfirmView`, `SettingsCustomMessageView`, `RippleTrustLineActivationSheet`.
- **Rows, cells, chips, badges, icon tiles, skeletons** — including the Figma-confirmed inner icon tile at 16.
- **The token scale itself, the lint rule (phase 3), and corner *style***.

## Review

Cross-model review loop, four rounds of `codex exec --sandbox read-only`:

| Round | Scope | Result |
|---|---|---|
| 1 | pill commit | **NO FINDINGS** |
| 2 | the six container commits | 3 findings — all nesting mistakes, all valid, all fixed |
| 3 | whole branch | 1 finding — valid, fixed |
| 4 | the ruling commits, asked specifically to hunt newly-flattened nesting | 4 findings — 3 valid and fixed, 1 rejected |
| 5 | the two default flips above, asked again about newly-flattened nesting | **NO FINDINGS** |

**Every finding across all four rounds was the same bug**, which is worth stating because it is exactly what phase 3's lint rule will *not* be able to catch:

> A radius literal means one of two things — *this surface's own geometry*, or *a restatement of a neighbour's*. Only the second kind breaks when a container moves, and neither the declaration nor a lint rule can tell them apart. You have to read the call site.

Round 2 found three views that read as containers from their declaration but render *inside* another container. Round 3 found the transaction-history "via" badge, whose `bottomTrailingRadius: 16` existed only to trace the card's corner. Round 4 found two `ContainerView`s that are **form controls** — the address-book chain selector and the referral payout selector, each sitting in a stack of text fields at 12 and widened to 24 by the default flip — plus the scanner masks carrying the token's `points` but not its `style`. The `cornerSize` gap above is the same bug reached by a different route.

Commit 16 fixes those three. The fourth finding — that `AddressBookCell`'s `99 → pill` does not meet the framed-box proof — is **rejected**: 99 is a member of the explicitly-ruled "fully round" family (50/60/77/88/99/100), and the arithmetic proof exists for numbers *outside* that family. The cell is also bounded well under the 198pt it would need to stop clamping: both text rows are `lineLimit(1)` beside a fixed 32pt logo.

## Verification

- **macOS target built explicitly** (`-destination platform=macOS`): `** BUILD SUCCEEDED **`. Required, because `Core/Platform/macOS` is behind `#if os(macOS)` and an iOS build never compiles it.
- `xcodebuild test`, **full suite**, iPhone 16 Pro / iOS 26.5: `** TEST SUCCEEDED **` — 4621 tests, 1 skipped, 0 failures.
- SwiftLint on first-party sources: **39 violations — exactly the repo baseline, none added.**
- **No before/after screenshots.** These surfaces are behind an onboarded, funded vault — send verify, swap done, DeFi positions, transaction history, keysign — and I cannot drive the app that far. A human should eyeball, in rough order of blast radius: the **DeFi tab** (balance banners, staking cards, governance list, and the position cards that moved via `ContainerView`), the **referral screen** (the nesting rule's hardest case), the **send/swap verify + done** screens, **transaction history** (card + detail sheet, specifically the "via" badge corner), the **address book** (both the settings list and the send-flow picker), the **backup flow**, and on **macOS** the server-backup verification screen and the QR scanners.


johnnyluo (approved):
LGTM

PR #5048, opened 2026-08-03
build(lint): keep corner radii on the token scale (phase 4 of 4)

gastonm5:
The last piece of #5024 — the guard that keeps corner radii on the scale.

The other three parts have merged (#5040 tokens, then #5046, which carried the container sweep and the tokenisation of the rest with it). This is now a **standalone PR against `main`**, not the tail of a stack, and it was rebuilt on current `main` to say so: one commit, `.swiftlint.yml` plus the files that carry a directive.

## What this adds

One SwiftLint `custom_rule`, `no_raw_corner_radius`, following the shape and `severity: warning` of the two rules already in `.swiftlint.yml`.

The regex is three alternatives, one per binder family the phase-3 census found — because a raw number binds to a corner in more shapes than `cornerRadius:`:

1. **Direct binders** — `.cornerRadius(` / `cornerRadius:` / `layer?.cornerRadius =`, the per-corner `topLeadingRadius:` family, `RectangleCornerRadii`'s `topLeading:` family (which names corners and never says "radius"), plus the AppKit spellings `addRoundedRect(in:cornerWidth:cornerHeight:)` and `NSBezierPath(roundedRect:xRadius:yRadius:)` — this is a macOS target too.
2. **Named constants** one hop from use (`let footerRadius: CGFloat = …`) — the form that hid six of the survivors.
3. **`cornerSize:` / `cornerRadii:` taking a `CGSize`**, where the number sits inside the `CGSize` and usually on a later line.

## ⚠️ What it does not do

**It enforces "use a token". It cannot enforce "use the *right* token."**

Whether a surface takes `xl` or a smaller step depends on whether it is a container or sits inside one — and that is only visible at the **call site**, never at the declaration. Every finding across the earlier phases was that same bug: a badge tracing its parent's corner, a `containerStyle` nested inside another, skeletons not matching the cards they stand in for. A regex sees none of it.

So this PR prevents *new literals*. It does not prevent *drift*. The rule's own message says so, and #5024's "drift can't re-accumulate" should be read as half-delivered.

## The 19 sites it fires on today

Turned on against current `main`, the rule fires **19** times.

**Sixteen are the deliberate survivors** the sweep left behind. Each is silenced with an inline `// swiftlint:disable:this no_raw_corner_radius` placed next to the comment that already justified it — not by broadening the regex or excluding files, so the exception stays attached to its reason. Three of those sixteen had no comment; the 16×1 step connectors and the 3pt Tron progress bars now say why they are off the scale.

| Site | Value | Why it stays |
|---|---|---|
| `ForegroundNotificationBannerView` ×2 | 40 | banner clipped to the screen's top edge |
| `NotchedRectangle.notchRadius` | 27 | a circular cavity, not a corner |
| `TooltipShape.arrowCornerRadius` | 2 | arrow tip fillet, sized against the arrow |
| `TooltipShape.arrowJunctionRadius` | 4 | arrow-to-bubble fillet, on the diagonal |
| `TronResourcesCardView` ×3 | 2 | end caps of a 3pt bar, already past the clamp |
| `FastVaultPasswordScreen` | 1 | 16×1 hairline step connector |
| `VaultSetupScreen` | 1 | 16×1 hairline step connector |
| `PeerDiscoveryScreen` ×2 | 33 | the QR frame the pairing flow is drawn around |
| `KeyImportActiveChainsView` | 20 | equidistant from 16 and 24 — a design call |
| `LockedFeatureSheet.cardCornerRadius` | 20 | mirrors the tier card's Figma bottom |
| `VultDiscountTierView.bottomCornerRadius` | 20 | the 24-top / 20-bottom asymmetry is a Figma value |
| `ModalBackgroundView` | 34 | matches a platform sheet default the scale deliberately doesn't own |

**The other three are new, and are a finding rather than an exception.** `AssetSelectionGridCellSkeleton`'s `24` and `4` and `AssetSelectionContainerView`'s failure-row `12` landed on `main` in #5030 *after* the census was taken, while the earlier phases were still in review — exactly the regression this rule exists to catch, arriving before the rule did. All three are plain on-scale values with no justification and no reason to be literals, so they are **tokenised** (`xl`, `xs`, `md`) rather than silenced. The skeleton now uses the same `xl` tile the real `AssetSelectionGridCell` draws and the same `xs` bar its DeFi sibling skeletons already use.

`VultisigAppTests` / `VultisigAppUITests` are excluded, matching the existing custom rules.

## Verification

**Baseline unchanged.** `main` reports **39** SwiftLint violations. With the rule active and the 16 directives + 3 tokenisations in place, this branch reports **39** — and the two violation lists are byte-identical after sorting, not merely the same count. The rule is usable in CI on day one.

**The rule fires.** A probe file added to the tree with `.cornerRadius(17)`, `RoundedRectangle(cornerRadius: 17)` and `UnevenRoundedRectangle(topLeadingRadius: 17, bottomTrailingRadius: 17)` took the count **39 → 43**, all four flagged. Probe removed; count back to 39, list byte-identical to baseline.

**All three families fire.** An isolated probe covering every alternative produced 8 hits — the 4 above, plus `let someCornerRadius: CGFloat = 31` and `var footerRadius = 21` (family 2), plus `cornerSize: CGSize(width: 19, …)` and a line-wrapped `cornerRadii: CGSize(\n width: 23, …)` (family 3).

**No false positives.** In that same probe, **zero** hits from any of: `RadialGradient(startRadius: 4, endRadius: 88)`, `.blur(radius: 12)`, `.shadow(radius: 8)`, `UnevenRoundedRectangle(topLeadingRadius: 0, bottomTrailingRadius: 0.0)`, `RoundedRectangle(cornerRadius: Theme.radius.xl.points)`, `Theme.radius.xl.shape`, `.cornerRadius(Theme.radius.md)`, `CGSize(width: Theme.radius.xl.points, height: 4)`, and the named constants `blurRadius` / `shadowRadius` / `startRadius`. `cornerRadius: 12` written inside a comment or a string literal is also silent — those match kinds are excluded.

The zero-guard covers every spelling (`0`, `00`, `0_0`, `0.0`, `-0.0`, `0e0`, `0x0`, `0b0`, `0o0`) because a zero corner means "square this corner" and there is no token for it — while `0.5`, `0_1`, `0x10` and `1e1` are real radii and do fire.

## Known blind spots, stated rather than hidden

All need the number to be legible where it is bound:
- a bare `let radius = N` — `radius` alone is the same word blur, shadow and gradients use, so matching it would cry wolf
- a wrapped `cornerRadius: CGFloat(N)`
- **a literal returned from a function** — `radius(for:) -> CGFloat { case .small: 99 }`

That last one is how `PrimaryButtonStyle.cornerRadius(for:)` — the app's most-rendered radius — stayed hidden through three phases. A regex can see the `99`; nothing on that line says it is a corner. Only the call site does.

Family 2 matches on the *name*, an approximation in both directions. It is deliberately **not** narrowed to names containing "corner": this tree uses `topRadius`, `footerRadius` and `arrowJunctionRadius` for real corners. The cost is that a rounded *shape* radius that is not a corner (`NotchedRectangle.notchRadius`, a circular cavity) also fires and needs a directive. One directive on a documented site beats three silent gaps.

## Gates

Sixteen of the nineteen Swift edits are trailing comments, so they cannot change behaviour. The three `AssetSelection` tokenisations are real code, so this ran the full gate rather than resting on that argument:

- iOS unit tests (iPhone 17 Pro, iOS 26.5) — **TEST SUCCEEDED**, 4683 tests, 0 failures, 1 skipped
- macOS build (`-destination platform=macOS`) — **BUILD SUCCEEDED**
- SwiftLint — **39**, byte-identical to `main`'s baseline

## Not closing #5024

Human-call items remain open: the QR frame's `33`, the `1`/`2` sites that are already capsules, `VultDiscountTierView.footerInnerShadow` drawing at `xl` while the footer clips to 20, `SwapPercentageButtons` rendering differently on iOS and macOS, and whether nested content boxes follow banners to 24.

Part of #5024

## Summary by CodeRabbit

* **Style**
  * Added a lint rule encouraging consistent use of the app’s theme-based corner-radius scale.
  * Updated asset selection screens to use themed rounded shapes for improved visual consistency.
  * Preserved intentional custom corner radii for banners, tooltips, progress indicators, onboarding, and other specialized UI elements.


johnnyluo (approved):
LGTM

PR #5054, opened 2026-08-04
fix(keychain): distinguish an absent item from an unreadable Keychain

gastonm5:
## The bug, which predates the passcode work

`Keychain.get` returned `nil` for **every** `SecItemCopyMatching` failure, not only for
`errSecItemNotFound`. A caller could therefore not tell these apart:

- **there is no item for this key** — the Keychain answered, and the answer was "nothing"
- **the item could not be read** — a locked device (`errSecInteractionNotAllowed`), a missing
  entitlement, an IO error. The Keychain did not answer at all

They license opposite decisions. Absence says a value may safely be written. An unreadable
Keychain says one may already be there. Collapsed into `nil`, a transient read failure looks
exactly like permission to overwrite whatever is stored — and where the stored item is key
material, that is the difference between saving a secret and destroying one.

This is not hypothetical or new: it already sits under the fast-vault password path on `main`
today. It is also the foundation the passcode/at-rest-encryption work builds on, where the same
collapse would let a failed read mint a fresh data key over an existing wrapper and orphan every
encrypted key share on disk. Fixing it separately, on its own merits, keeps that review honest.

## What changed

- **`KeychainReadResult<Value>`** — `absent` / `present(Value)` / `unavailable(OSStatus)`,
  returned by `Keychain.getString` / `getInt` and by every `KeychainService` read.
  `errSecItemNotFound` is the only status that maps to `absent`.
- An item that exists but cannot be decoded — bytes that are not UTF-8, a version string that is
  not an integer, a success with no data — reports `unavailable(errSecDecode)`. It is present, so
  calling it absent would invite a caller to write over it.
- **`KeychainItemStore`** puts the three `SecItem` calls behind a seam. A unit-test bundle has no
  keychain entitlement, so every real `SecItem` call answers `errSecMissingEntitlement` — one of
  the very statuses that must not be mistaken for an absent item, and one that cannot be *shown*
  to be told apart if it is the only status a test can produce.
- Every existing consumer now **states** its choice instead of inheriting it from a collapsed
  `nil`: the display-only and re-obtainable reads (fast-vault hint and password, device token)
  call `valueTreatingUnavailableAsAbsent` with a comment saying why that is safe there, and
  `AppMigrationService` folds `unavailable` into "nothing migrated" in a named method that
  records the reasoning — the migrations are idempotent, so repeating them is cheaper than
  skipping them, and nothing on that path writes a secret.

## No behaviour changes

This is an API-shape change plus explicitness, nothing more. Every call site takes the same
branch it took before, for every combination of present / absent / read-failed / undecodable.
Nothing about passcodes, key shares, encryption or migration is touched.

## Tests

- `KeychainTriStateReadTests` — fault-injects eight `OSStatus` values a real device hands back
  (`errSecInteractionNotAllowed`, `errSecNotAvailable`, `errSecAuthFailed`,
  `errSecMissingEntitlement`, `errSecUserCanceled`, `errSecDecode`, `errSecParam`, `errSecIO`)
  and asserts that **not one of them reaches a caller as `absent`**, through `Keychain` and
  through all four `KeychainService` reads. Plus the decode-failure and success-without-data paths.
- `KeychainWriteSeamTests` — pins the write path across its move behind the seam: delete-then-add
  order, the generated query, `kSecClass`, the value bytes, the accessibility attribute, the
  dropped `kSecReturnData` flag, `nil` meaning delete, and a failing add still being ignored.
- `AppMigrationServiceKeychainReadTests` — pins the one shipping consumer that acts on absence:
  `unavailable` runs the migrations rather than skipping them, and a recorded latest version still
  skips them.

The seven remaining consumers are SwiftUI views and view models that reach
`DefaultKeychainService.shared` directly with no injection point; giving them one is a refactor
this PR deliberately does not do. Their collapse is expressed through the named
`valueTreatingUnavailableAsAbsent` affordance, which is itself covered.

## Verification

- `make test` — **`** TEST SUCCEEDED **`**, 4704 tests, 1 skipped, 0 failures
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 39 warnings, unchanged from
  the base commit, none in any touched file
- Reviewed by `codex exec --sandbox read-only` over two rounds: no CRITICAL or MAJOR findings;
  three MINOR findings (test-only global-state leak, unpinned write path, a missing `kSecClass`
  assertion) all fixed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unavailable or unreadable secure storage.
  * Prevented inaccessible saved passwords and hints from appearing or triggering autofill.
  * Ensured device registration and migrations retry when stored values cannot be read.
  * Improved backup export behavior when saved credentials are unavailable.

* **Tests**
  * Added coverage for missing, available, invalid, and inaccessible secure-storage values.
  * Added tests for secure-storage writes, deletions, and migration behavior.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Migration/AppMigrationService.swift
+++ b/VultisigApp/VultisigApp/Core/Migration/AppMigrationService.swift
@@ -33,11 +33,7 @@ struct AppMigrationService {
             return
         }
 
-        // Get the last migrated version from Keychain
-        // there is no way to figure out whether it is a new installation , or an update from a version without AppMigrationService
-        // So when the last migrated version is nil , we need to do the migration
-        // thus we need to make sure the migration is idempotent
-         let lastVersion = keychainService.getLastMigratedVersion() ?? -1
+        let lastVersion = lastMigratedVersion()
 
         // If already migrated to the latest version, skip
         if lastVersion >= latestMigrationVersion {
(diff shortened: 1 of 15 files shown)


johnnyluo (approved):
LGTM

Issue #5017, opened 2026-08-03
[ADD] Kamino Earn vaults — Solana DeFi positions (deposit / withdraw / position view)


realpaaao:
## Summary

Integrate **Kamino Earn vaults (kVaults)** as a new position type that can be enabled under the **Solana DeFi positions**. Users deposit USDC or SOL into curated lending vaults and manage the position (balance, APY, PnL, deposit/withdraw) from the DeFi tab — same enablement pattern as the existing per-chain DeFi positions.

Kamino is the largest Solana DeFi protocol (~$3.2B TVL; Earn vaults ~$516M across 164 vaults). The integration is REST-only against `https://api.kamino.finance` (no API key) and reuses the existing Solana ed25519 keysign + broadcast path — no new signing infrastructure.

## Launch vault set (3)

| Tier | Vault | Asset | Curator | Vault address |
|---|---|---|---|---|
| Conservative | Steakhouse USDC | USDC | Steakhouse Financial | `HDsayqAsDWy3QvANGqh2yNraqcD8Fnjgh73Mhb3WRS5E` |
| RWA yield | RWA USDC | USDC | RockawayX | `DWSXb18xZApz29vnQpgR2m6MynCT7PznaXt7Ut7M7KaP` |
| SOL yield | Allez SOL | SOL | Allez Labs | `A1so1bPD3W1TfeFwboDh8yfAAVaVtcdAYBYCjhg2mJQ` |

Keep the vault list in a config/registry that can be extended without a code change where the platform allows it.

Risk labeling: Steakhouse USDC = conservative USDC lending; **RWA USDC lends against tokenized private credit (OnRe reinsurance ~51%, Huma receivables ~30%, Obligate bonds ~19%)** — show it one risk tier above the conservative vault, not as "T-bill backed"; Allez SOL = SOL lending yield.

## API

Base: `https://api.kamino.finance` — docs: https://kamino.com/docs/build/api-reference/introduction

Actions (return a base64 Solana transaction ready to keysign + broadcast):
- `POST /ktx/kvault/deposit` — body `{ wallet, kvault, amount }` (amount in decimal, not lamports)
- `POST /ktx/kvault/withdraw` — same shape
- `POST /ktx/kvault/deposit-instructions` / `withdraw-instructions` — instruction-level variants if we need to assemble the tx ourselves

Data:
- `GET /kvaults/vaults/{pubkey}` — vault state (shares mint, fees, min deposit)
[...]

PR #5055, opened 2026-08-04
feat(defi): show Kamino Earn vault positions on the Solana DeFi tab

gastonm5:
## Description

Layer 2 of a 4-layer stacked change adding **Kamino Earn vaults (kVaults)** to the Solana DeFi surface. This layer makes positions *visible*: a `@Model` to persist the snapshot, an Earn segment under Solana, and the totals wiring. **Read-only** — no deposit, no withdraw, no transaction building. Those are layer 3.

Part of #5017. Deliberately not `Fixes` — the issue closes with the final layer.

> **Stacked PR.** Base is `feat/kamino-plumbing-5017` (#5053), so this diff is only layer 2. Review #5053 first.

### What's here

**`KaminoPosition` `@Model`**, keyed `vaultAddress + pubKeyECDSA`, holding `isEnabled`, shares, token amount, APY, PnL and `lastUpdated`, with amounts stored as exact base-unit strings plus their scale.

Enablement lives on this model rather than in `DefiPositions`, and that placement is load-bearing: `DefiPositions.init(from:)` uses required `decode` for every bucket and vault backups encode those rows, so adding a fourth bucket would break decoding of **existing** backups. Putting the flag on a model we're adding anyway costs nothing and removes the hazard rather than mitigating it.

**`DefiChainPositionType.earn`** with `selectionIndex = nil`, following the `.governance` precedent; Solana's position types become `[.stake, .earn]`. The picker element becomes `DefiSelectableAsset`, an enum over `CoinMeta` and `KaminoVaultDescriptor` — `AssetSelectionContainerSheet` and `AssetSection` were already generic over `Asset: Hashable`, so neither needed changing. A Kamino vault never flows into `DefiChainSelectPositionsScreen.onSave`'s `CoinService.addToChain`, so it can't become a real `Coin` row with a market-data lookup and an SPL balance fetch that would return zero.

**Value derivation.** Shares → token amount via `metrics.tokensPerShare`; fiat via rate lookup. Note `metrics.sharePrice` is **USD-denominated** while `tokensPerShare` is the token rate — they differ by ~74× on the SOL vault, so using `sharePrice` as a token rate would overstate the position enormously. `apy30d` is a fraction, not a percentage.

**`DefiBalanceService`** folds Kamino into the Solana total and position count. This is the reason the snapshot is persisted at all: that service computes synchronously from persisted state, which a view-model-only value can never feed.

8 new keys across all 8 locales, including `ko`.

### Correctness issues caught in review

Three were real defects rather than style, and all three are the kind that silently show a wrong number:

- **An unparseable `totalShares` zeroed a live position.** A `?? "0"` fallback conflated "vault absent from `/positions`", which genuinely means zero, with "value present but unparseable", which means the read failed. The second case now keeps the cached row.
- **A snapshot could be written into the wrong vault.** The screen's `StateObject` outlives a vault switch, so an in-flight refresh would read the new vault's owner address and persist it into the old vault's rows. Fixed with `update(vault:)` plus a generation stamp, which also discards out-of-order refreshes.
- **Nothing fetched a rate for the underlying token.** The rate cache is populated from the vault's own coins, so a user whose USDC lives *inside* the Kamino vault and nowhere in their wallet saw a real position valued at $0 — in the card and in the synchronous DeFi total. The SOL vault was incidentally safe because SOL is always present.

### Known gaps, deliberately not fixed here

- **The DeFi chain list sorts by wallet-coin balance**, so a Solana entry whose value is mostly Kamino sorts as though it were zero. Pre-existing — TON and Cosmos staking already diverge the same way — and changing the sort key reorders every chain, so it doesn't belong in this PR.
- **Refresh errors are not surfaced.** `DefiChainMainScreen` deliberately surfaces only the bond error; stake and LP also fail silently onto persisted rows. Matching that is intentional; staleness UI is layer 4.

### A wider problem this surfaced — worth a decision beyond this feature

While verifying the backup path, it turned out that **no DeFi preference survives a `.vult` restore today**. `Vault(proto:)` and `mapToProtobuff()` carry key material and identity only, so `defiChains`, `defiPositions` and `enabledDefiProviders` are all lost on restore — this is not specific to Kamino and is not introduced here.

This PR adds `Vault.enabledKaminoVaults` (Codable, `decodeIfPresent`, materialised back into rows by an idempotent hydration), which covers the **JSON import path only**. Carrying DeFi state in the proto is a `commondata` schema change across four platforms, and doing it for Kamino alone would make it the only DeFi opt-in that survives a restore — a strange asymmetry to ship by accident. Flagging it rather than deciding it here.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

No tx link: this layer is read-only and broadcasts nothing. Transaction links arrive with layer 3.

## Parity

Kept even though this layer touches no keysign, broadcast or fee path, because the feature as a whole is cross-platform.

## Checklist

Gate: 4871 tests, 1 skipped, 0 failures. SwiftLint 39 warnings / 0 serious — identical to the pre-existing baseline, none in the touched files.

_To add: the Earn segment under the Solana DeFi tab._

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
@@ -99,7 +99,8 @@ final class KaminoEarnViewModel: ObservableObject {
     ///   must not make a position look like it vanished);
     /// - a failed per-vault state/metrics read keeps that row's cached values,
     ///   because without `tokensPerShare` the shares cannot be valued at all;
-    /// - a failed `/pnl` read only drops the profit-and-loss line.
+    /// - a failed `/pnl` read keeps the last figure that was read, and only
+    ///   leaves the profit-and-loss line blank when there has never been one.
     ///
     /// An empty `/positions` response is a real "holds nothing" answer and is
     /// allowed to zero a row — the vault stays listed, since the user enabled it.
@@ -206,7 +207,14 @@ final class KaminoEarnViewModel: ObservableObject {
                 continue
             }
 
+            // A `/pnl` failure is a "could not confirm", exactly like the reads
+            // above, so it keeps what the last successful one established. The
+            // figure is persisted as well as displayed: without the fallback a
+            // single flaky read would write `nil` over a real value, and the
+            // next launch would seed a row whose profit-and-loss line has
+            // quietly gone — a display gap outliving the outage that caused it.
             let pnlToken = await pnlToken(owner: owner, descriptor: descriptor)
+                ?? cachedPnlToken(for: descriptor)
 
             freshRows.append(
                 KaminoEarnRow(
@@ -282,6 +290,10 @@ final class KaminoEarnViewModel: ObservableObject {
 
     /// Profit and loss in the underlying token. Quietly `nil` on failure — it is
     /// one display line, and dropping it must never cost the user their position.
+    ///
+    /// `nil` means "could not read", never "the position has no profit and
+    /// loss": a real zero arrives as a parseable `0`. The caller is what turns
+    /// that into the cached figure rather than into an erasure.
     private func pnlToken(owner: String, descriptor: KaminoVaultDescriptor) async -> Decimal? {
         do {
             let pnl = try await service.fetchPnl(owner: owner, vault: descriptor.address)
@@ -292,6 +304,13 @@ final class KaminoEarnViewModel: ObservableObject {
             )
             return nil
         }
+    }
+
+    /// The profit and loss a previous pass established for this vault, read from
+    /// the rows on display — which are the persisted ones until this refresh
+    /// publishes, exactly the source `keepCached` reads.
+    private func cachedPnlToken(for descriptor: KaminoVaultDescriptor) -> Decimal? {
+        rows.first(where: { $0.id == descriptor.address })?.pnlToken
     }
 
     private func persist(_ snapshots: [KaminoPositionSnapshot]) {
--- a/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
@@ -32,11 +32,18 @@ final class KaminoEarnViewModel: ObservableObject {
     private let storage: KaminoPositionStorageService
     private let priceService: CryptoPriceServiceProtocol
     private let logger = Log.defi.viewModel
-    /// Bumped by every `refresh` and by `update(vault:)`. A refresh publishes
-    /// only if it is still the current one, so a slower earlier call cannot
-    /// overwrite a newer result — or resurrect a vault the user just disabled,
-    /// since the enabled set is read when the refresh starts.
-    private var refreshGeneration = 0
+    /// The refresh currently in flight, if any.
+    ///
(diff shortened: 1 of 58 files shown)


VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
```diff
@@ -0,0 +1,328 @@
...
+                )
+            )
+            snapshots.append(
+                KaminoPositionSnapshot(
+                    vaultAddress: descriptor.address,
+                    shares: shares,
+                    tokenAmount: tokenAmount,
+                    apy30d: info.apy30d,
+                    pnlToken: pnlToken
+                )
```
johnnyluo, line 235: **Correctness:** A failed `/pnl` read isn't protected by the "keep cached value on failure" rule the rest of this refresh follows. Every other read here treats failure as "could not confirm, keep what we had" — an unparseable share value or a failed `vaultInfo` fetch calls `keepCached(descriptor)` and `continue`s, leaving the row *and* the persisted snapshot untouched. `pnlToken(owner:descriptor:)` doesn't: on a failed `/pnl` call it returns `nil`, and that `nil` flows into both `freshRows` (display) and `snapshots` (line 220-227), which then gets written via `KaminoPositionStorageService.upsert` → `KaminoPosition.apply(_:)`, **overwriting whatever PnL value was previously persisted**.

Concretely: if a user has a real, previously-fetched PnL of $5 persisted from a prior successful refresh, and the next refresh hits a transient `/pnl` failure while `/positions` and `vaultInfo` both succeed, the persisted `pnlToken` silently resets to `nil`. On the next app launch, `seedFromPersistedSnapshot()` reads that `nil` and the PnL row disappears — not because the user's position changed, but because one flaky read wiped a cached, correct value. The doc comment above ("a failed `/pnl` read only drops the profit-and-loss line") reads as "hide it this render," but the implementation persists that drop rather than scoping it to display.

Suggest mirroring `keepCached`: on a `/pnl` failure, fall back to the existing row's `pnlToken` (from `rows.first(where: { $0.id == descriptor.address })`) for both the displayed row and the persisted snapshot, instead of defaulting to `nil`. Worth a test alongside the existing `testFailedVaultHydrationKeepsThatRowsCachedValues`-style coverage: a PnL failure after a prior successful fetch must not erase the cached PnL.

johnnyluo (commented):
Inline finding from review — see comment below.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoEarnViewModel.swift
gastonm5: Good catch — fixed in `f7c8bd5`.

`/pnl` was the one read in that loop not following the rule the others do. The `nil` went into `snapshots` as well as `freshRows`, and `KaminoPosition.apply(_:)` writes `pnlToken` unconditionally, so a transient failure overwrote a figure a previous refresh had established — and because `seedFromPersistedSnapshot()` reads the store, the gap outlived the outage into the next launch.

It now falls back to the cached row's `pnlToken`, the same source `keepCached` reads. `nil` from that read is unambiguously "could not read" rather than "no profit and loss" — a real zero arrives as a parseable `0` — so the fallback cannot mask a genuine transition to nothing.

Plus the regression test you asked for, `testAFailedPnlReadKeepsTheCachedFigureRatherThanErasingIt`: a successful read, then `/pnl` down, asserting the displayed row, the persisted snapshot, and a freshly seeded view model all still carry the earlier figure.

johnnyluo (approved):
LGTM

PR #5056, opened 2026-08-04
feat(defi): deposit into and withdraw from a Kamino Earn vault

gastonm5:
## Description

Layer 3 of a 4-layer stacked change adding **Kamino Earn vaults (kVaults)** to the Solana DeFi surface. This is the layer that moves funds: asset-denominated **deposit** and **withdraw** flows, each ending in a `signSolana` payload.

Part of #5017. Deliberately not `Fixes` — the issue closes with the final layer.

> **Stacked PR.** Base is `feat/kamino-earn-view-5017` (#5055), so this diff is only layer 3. Review #5053 → #5055 first.
>
> ⚠️ **One path is deliberately refused.** A *farm-staked* withdraw — the path every real user takes — is blocked on an on-chain capture that has not happened, and refuses rather than guessing. See "The farm-staked withdraw" below.

### The pipeline

At verify time: build via the Kamino API → validate → simulate → inject ComputeBudget → **re-simulate requiring `err == nil`** → attach as `signSolana`.

Validation runs in two phases, and they must not be collapsed. Before injection, *any* ComputeBudget instruction is a refusal. After injection, both are required and pinned to the locally chosen values. That asymmetry is what stops an attacker-supplied unit price from draining SOL through priority fees — the network charges `limit × price`, bounded only by the payer's balance.

### The safety-critical part: `refreshSolanaBlockhash`

The general path of `BlockChainService.refreshSolanaBlockhash(for:)` returns `signData: nil`. Without a Kamino branch, the device would rebuild and sign a **plain transfer** from `coin` / `toAddress` / `toAmount` — real funds, wrong destination — instead of the Kamino transaction. This is not wiring; it is the difference between a deposit and a loss.

It turned out to need **two separable properties**, not one:

- **Preserving the raw bytes must be unconditional.** Dropping them is what manufactures the transfer, whoever built them. The `if let signSolana = payload.signSolana` branch returns *before* the rebuild that sets `signData: nil`.
- **Splicing the blockhash must be conditional** on the `kaminoPayload` marker — those same 32 bytes hold the nonce in a durable-nonce transaction, where overwriting them would be wrong.

The argument that it can't fall through doesn't rest on reading the branch. `KeysignDiscoveryViewModel.setData` assigns `self.keysignPayload = finalPayload` after the refresh, and *both* consumers — the returned `KeysignInput` and the `ProtoSerializer.serialize` that relays to a co-signer — read that same stored property, so no pre-refresh copy survives. `FastVaultKeysignBootstrap` does the same with a local. Both are initiator-only; a joiner builds from proto and never calls the refresh. All three `KeysignPayload` copy helpers now carry `kaminoPayload`, checked specifically because `withChainSpecific` sits beside the staking branch and silently dropping the marker there would have degraded a deposit to the no-splice path with no error.

Two tests pin the behaviour rather than the implementation: one asserts `getPreSignedImageHash(keysignPayload:)` equals `getPreSignedImageHashForRaw` of the payload's own bytes — only true when the raw branch was taken — and another asserts the spliced blockhash *is* the fresh one, with wire size, every static account, every instruction's program index / accounts / data, and every lookup table byte-identical across the splice. A third re-runs the full validator on the spliced bytes.

### The SOL maximum

Derived from the actual payer lamport delta of the validated, built transaction rather than a fixed "fee + 2 ATA rents" — the ATA count varies with which accounts exist, farm initialisation may fund more, and the priority fee is `limit × µlamports-per-CU`.

Two things fell out of doing it properly:

- **The pre-balance cancels out**: `max = payerLamportsAfter + probe − rentExemptMinimum(0)`. No `getBalance`, and no race between two reads.
- **There is a rent-exempt floor the plan missed.** Leaving the payer at exactly zero is fine — the account is deleted — but leaving them *non-zero and below* rent exemption is rejected on chain with `InsufficientFundsForRent`. That's a cliff roughly 890,880 lamports wide sitting directly beneath the naive maximum. The floor is fetched from RPC, not hardcoded.

One assumption is now named in the code: the runtime debits the fee during transaction loading, so the post-simulation balance is net of it. If a node ever reported pre-fee, the maximum would be too large by the fee and the *final* simulation refuses — a refusal, never a signed transaction that fails.

### Correctness issues caught in review

All at one boundary the earlier layers never reached: where a human types a number and a machine reads it.

- **`1,23` parsed as `123` on an `en_US` device.** Grouping separators were stripped, never validated. Groups are now validated against the locale's own `groupingSize` / `secondaryGroupingSize`.
- **The canonical amount was rendered POSIX**, but downstream reads `SendTransaction.amount` through `parseInput`, which tries the device locale first — so `10.123456` read as **ten million** on a comma-decimal device. Now rendered through the device's own formatter, with the test asserting the round-trip rather than a literal.
- **The amount was read twice** — once for the payload, once after the network round trips for the summary. A keystroke during preparation would have the user approve a different amount from the one in the bytes. A single `makeDeposit()` now returns both from one captured value; the regression test perturbs the field from inside the suspended preparation.
- Percentage buttons truncated a 9-decimal maximum to 4 dp.
- An empty prioritization-fee sample inherited `SolanaHelper`'s 1,000,000 default, which is also this feature's clamp *ceiling* — a 50× over-tip on a quiet network.
- ASCII-only digit guards refused Eastern Arabic / Devanagari numerals that the device's own formatter writes.

### Withdraw, and why a partial can never become a full exit

The withdraw form is asset-denominated and converts to shares before the call: `shares = assets / tokensPerShare`, rounded **down**, in exact `BigInt` arithmetic.

The hazard is the `u64::MAX` sentinel. The API validates nothing, and a withdraw above the user's share balance is silently rewritten to `u64::MAX` — *withdraw everything*. So an over-request by a single base unit turns a partial withdraw into a full exit, and round-tripping a maximum through assets-and-back is exactly how you would introduce that base unit.

Four independent layers each refuse it, and each is tested:

1. **Arithmetic** — `KaminoWithdrawMath.shares` refuses above the maximum, returns the exact read balance *only* at the maximum, and truncates below. Since `maximumTokens` truncates too, `tokens < maximumTokens` implies the result is `< held`. Tested at +1, +78 and +1e9 over, and swept across the top 400 base units below.
2. **Form** — the ceiling is re-enforced in `makeWithdraw` rather than trusted from the screen.
3. **Request** — `KaminoService.buildWithdrawTransaction` refuses `u64::MAX` outright, so the sentinel cannot be produced by arithmetic at all.
4. **Bytes** — the validator pins the withdraw instruction's `u64` to exactly the shares requested. This is the layer that survives a balance changing between read and send: the API rewrites the request to the sentinel, and the transaction is refused before simulation and before any signature.

**A blocker found in review, which is not Kamino-specific.** Any amount at or above the maximum was sent as the whole held balance — and `FormScreen` renders `PrimaryButton(...).disabled(isContinueDisabled)` and **never reads `validForm`**. So a mistyped extra digit was a validator-approved full exit. Verifying it showed the premise is broader than the finding: `validForm` is decorative for *every* caller built on `FormScreen`, the Kamino deposit form included. Fixed here in four places, including an `isContinueDisabled` passthrough that the withdraw screen uses — which is also the hook for the wider fix in layer 4.

### The farm-staked withdraw

The validator's withdraw template only knows the *unstaked* instruction sequence, the only shape ever observed on chain. Every fresh deposit auto-stakes into the vault's farm, so a farm-staked withdraw — the path every real user takes — is **explicitly refused** rather than built on a guess. A guessed sequence that happens to validate would be the worst available outcome.

Three tests pin that refusal so it cannot be lost silently: `testRefusesAWithdrawCarryingAFarmInstructionAfterIt`, `…BeforeIt` (both orderings), and `testAFarmStakedPositionRefusesAndBuildsNothing`.

One decoded mainnet withdraw of a farm-staked position unlocks exactly two edits:

1. `KaminoTransactionValidator.expectedSequence(for:)`, at the block marked `EXTENSION POINT — the farm-staked withdraw` in the `.withdraw` case. It needs a `Step.Kind` for the unstake (program `farms`, discriminator `sha256("global:<name>")[0..<8]`, added to `KaminoInstructionDiscriminator` and `Context.matches`), its account layout checked the way `validateFarmsStake` checks the deposit's, and the step inserted at its observed position.
2. `KaminoWithdrawEligibility.resolve` — drop the `.farmStaked` refusal so a staked balance resolves to `.withdrawable`.

**RESOLVED — the capture exists (2026-08-05), and the path now works.** It landed in layer 4 (#5057) rather than here, because the shared `KaminoInstructionSequence` it extends only exists in that layer and splitting it back down would mean rewriting all four PRs.

Read #5057 for the detail, but two of its findings correct this PR's description directly:

- **The sequence turned out to be conditional on the amount, not on the vault having a farm.** At or below the unstaked balance it is still the 2-instruction shape described above; above it, the farms pair appears and the unstake is exactly the shortfall.
- **The `u64::MAX` reasoning in "why a partial can never become a full exit" above needs one correction.** The sentinel fires at `requested ≥ balance`, so sending the exact held balance is what *produces* it — the four guards would therefore have refused every legitimate full exit rather than allowing one. Layer 4 resolves it by making the maximum strictly below the balance (≤ 1 base unit of dust). The four layers still hold; what was wrong was the claim that a 100 % withdraw could never reach the sentinel.

### Deferred to layer 4

`SendCryptoVerifyViewModel` recomputes its fee row from the display-only transaction, so the displayed fee ignores the injected compute budget. Pre-existing for Circle and Yield too; verify-screen semantics are layer 4's scope, and the real fee is inside the signed bytes.

**Corrected in layer 4:** an earlier revision of this description said the fee was bounded at 0.00035 SOL. That is only the compute-budget portion — `SolanaHelper.defaultFeeInLamports` adds a flat, app-wide 0.001 SOL base on top of which it sits.

## Which feature is affected?
- [x] Sending — Kamino deposit builds and signs a Solana transaction
- [ ] Swap

**tx links — broadcast on mainnet, all three curated vaults, all succeeded.**

| Vault | Transaction | On-chain instruction sequence |
|---|---|---|
| Steakhouse USDC | [`5HMLRdK8…EAYWcSN`](https://solscan.io/tx/5HMLRdK8aWnusHYkn3fEs1azL5LNNVGpa6Z2iMMVEwJCwGFoZ8NpQf7XZq89widXMBZkigrJYgDH9UMMbEAYWcSN) | `Deposit` → `RefreshReservesBatch` → `InitializeUser` → `Stake` |
| RWA USDC | [`4jQirmaf…XRDohLb`](https://solscan.io/tx/4jQirmafbkYpighhsJ6dh275o9mHjwo6JK8EM9RRanfKLphnAJGtgexoLCSG7u3rXLjoVn1MrnapHLiaJXRDohLb) | `Deposit` → `RefreshReservesBatch` → `InitializeUser` → `Stake` |
| Allez SOL | [`25vcmZxz…E6uRAM`](https://solscan.io/tx/25vcmZxzc6HSpniUr7HEGoTapuWpZEx2uAj4Pgc6bfGKTubD5uVFWbw8ESdnxiiWmz1Jw1rN5oC6XTDxNwE6uRAM) | `Deposit` → `RefreshReservesBatch` → `InitializeUser` → `Stake`, preceded by System + Token for the wrapped-SOL account |

Each one confirms the parts of this layer that simulation alone could not:

- **The injected compute budget survives to chain.** Every transaction carries exactly two leading `ComputeBudget` instructions — the pair this layer injects and the post-injection validator pins. Fees landed at 11,400–13,000 lamports.
- **The blockhash splice held.** These were signed after `refreshSolanaBlockhash` replaced the 32-byte field in the already-built bytes; a dropped-bytes regression would have broadcast a plain transfer instead.
- **The SOL path exercises its own branch.** The Allez SOL deposit carries the extra System + Token instructions for the wrapped-SOL account, which the two USDC deposits do not.
- `InitializeUser` + `Stake` on all three confirms deposits auto-stake into the vault's farm, which is why shares never land in the user's associated token account.

The matching **withdraw** — which needs the farm-staked path — is in layer 4 (#5057), since this layer deliberately refuses it.

## Parity

Required — this touches keysign, fee computation and the pre-sign refresh.

## Checklist

Gate: 4976 tests, 1 skipped, 0 failures. SwiftLint 39 violations / 0 serious — identical to the pre-existing baseline, none in any touched file.

_To add: the deposit form, the withdraw form, and the verify screen._


VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoWithdraw.swift
```diff
@@ -0,0 +1,266 @@
...
+    static func shares(
+        forTokens tokens: KaminoTokenAmount,
+        held: KaminoShareAmount,
+        maximumTokens: KaminoTokenAmount,
+        tokensPerShare rate: KaminoRate,
+        shareDecimals: Int
+    ) -> KaminoShareAmount? {
+        guard tokens.baseUnits > 0, held.baseUnits > 0 else { return nil }
+        guard tokens.baseUnits <= maximumTokens.baseUnits else { return nil }
+        guard tokens.baseUnits < maximumTokens.baseUnits else { return held }
```
johnnyluo, line 221: [P2] This exact-maximum branch still returns the full held share balance. The PR description notes Kamino rewrites requests at `requested >= balance` to the `u64::MAX` withdraw-all sentinel, so a legitimate 100% unstaked withdraw can come back widened and fail validation instead of working. The maximum path should avoid sending exactly the held balance, or otherwise align with the layer 4 sentinel handling.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoDepositViewModel.swift
```diff
@@ -0,0 +1,285 @@
...
+    struct DepositHandoff {
+        let payload: KeysignPayload
+        let transaction: SendTransaction
+    }
+
+    /// Builds the payload the verify screen signs and the summary it shows, or
+    /// `nil` with `error` set.
+    func makeDeposit() async -> DepositHandoff? {
+        guard let vaultInfo, let depositCoin else { return nil }
+        guard let amount = enteredAmount(), amount.isValidRequestAmount else {
```
johnnyluo, line 206: [P2] `makeDeposit` only checks that the entered amount parses and fits the request range. Since `FormScreen` still ignores `validForm`, below-minimum or over-balance deposits can reach the build/simulation pipeline, and missing-coin cases can silently no-op. Please enforce the same minimum/balance/missing-coin gates here or pass an explicit disabled state from the deposit screen.

VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoWithdraw.swift
gastonm5: I don't think this one bites — though the argument name invites the reading, since `held` is never the held balance.

The exact-maximum branch is fed `held.maximum`, which is `KaminoSharePosition.spendable`, and `spendable` is the total with one base unit taken off when the reported figure is exactly representable (and the truncated figure otherwise) — see the doc comment at `KaminoWithdraw.swift#L58-L95`. So the request is strictly below the balance by construction and the `>=` rewrite cannot fire. `KaminoWithdrawEligibility.resolve` hands that value out as `maximum`, and `KaminoWithdrawViewModel` passes it as `held`, so a 100% withdraw sends the asset value of `spendable` rather than of the position.

The one-base-unit step-back *is* the sentinel handling, paid up front rather than detected afterwards: on the launch vaults it is 10^-6 of a share, roughly a ten-thousandth of a cent. The measurements behind it are in that same comment — requests of `3137.14326`, `71.999441` and `1683.002283` against balances of exactly those figures each came back as `u64::MAX`, and each passed through as an ordinary share count one base unit below.

Happy to rename `held` to `spendable` through `KaminoWithdrawMath` if the argument name is what made this hard to read — that is the part I'd agree is misleading.

VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/KaminoDepositViewModel.swift
gastonm5: Fixed in `6837867`.

You're right, and the reason is worse than an oversight: the gate existed, but one layer up in #5057 — which is not something this PR should depend on to be correct. Both halves you asked for are now in this PR:

- `makeDeposit()` re-enforces the vault's minimum and the available ceiling before any network call, mirroring what `KaminoWithdrawViewModel` already did. Neither bound can move funds wrongly (the API accepts a below-minimum deposit and the chain rejects it afterwards; an over-balance one fails simulation), but both used to cost the user several round trips to be told what the form already knew.
- `isDepositUnavailable` is wired to `isContinueDisabled`, which covers the missing-coin case you mention along with an unhydrated vault and a balance that cannot reach the minimum however it is spent. A below-minimum *amount* is deliberately not in there — the user can fix that one, and disabling the button would remove the tap that reveals the field error.

Three tests: `testABelowMinimumAmountIsRefusedBeforeAnyNetworkCall`, `testAnAmountAboveTheMeasuredMaximumIsRefusedBeforeAnyNetworkCall`, `testContinueIsDisabledOnlyWhenNoAmountCouldWork`.

The ceiling is compared in exact base units rather than against `availableAmount`: that `Decimal` comes back through `Coin.balanceDecimal`, which is formatter-backed, so at large balances it can round either way — refusing an exact full-balance deposit, or admitting one a base unit over. `availableBaseUnits` is published from the same measurement that produces the rendered figure, so the number enforced and the number shown cannot diverge.

johnnyluo (approved):
LGTM

VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoWithdraw.swift
johnnyluo: Checked this against the actual code on `feat/kamino-flows-5017`: `KaminoSharePosition` here has no `spendable` property, and `KaminoWithdrawEligibility.resolve` hands out `position.unstaked` directly as the maximum — not a value stepped back by one base unit. So in this PR's diff, `held`/`maximum` genuinely is the full read balance, and a 100% withdraw request will still equal it and risk the `u64::MAX` rewrite.

I found the `spendable` mechanism you're describing in #5057 (`feat/kamino-verify-5017`) instead — it's not in this branch. That matches this PR's own description, which lists the sentinel-on-exact-balance issue under "Deferred to layer 4" / "RESOLVED... landed in layer 4 (#5057)".

Resolving this thread on that basis: the fix is real, just one PR up the stack rather than in this one. Worth double-checking #5056 and #5057 land together (or that #5056 isn't merged standalone) so a 100% withdraw is never reachable in a shipped state without the fix.

PR #5057, opened 2026-08-04
feat(defi): decode Kamino transactions on the verify screen, and withdraw farm-staked shares

gastonm5:
## Description

Layer 4 — the final layer — of the stacked change adding **Kamino Earn vaults (kVaults)** to the Solana DeFi surface. This layer makes the transaction *readable* before it is signed, on both the initiating device and any co-signer.

Part of #5017. The issue closes when the whole stack lands.

> **Stacked PR.** Base is `feat/kamino-flows-5017` (#5056), so this diff is only layer 4. Review #5053 → #5055 → #5056 first.

### The problem this closes

`SignSolanaDisplayView` showed only program id, account count and data length. A co-signer approving a Kamino deposit or withdraw was approving **blind** — the bytes could have said anything.

`KaminoTransactionDecoder` now reads a deposit or withdraw back out of the wire message **offline**. The vault has to be identified backwards: lookup tables cannot be resolved without the network and the vault account lives in one, so each curated vault's share account is *derived* for the transaction's signer and matched against the instruction's share slot — a static key in all three captured vectors, verified independently by recomputing the ATA PDAs against the fixtures.

`KaminoVerifyPresentation` cross-checks that decode against the marker on the initiator, or `toAddress` / `toAmount` on a co-signer. **A disagreement, or a Kamino transaction that cannot be decoded at all, blocks signing on both devices** — at the button and at the confirm seam.

### What review found, and the pattern in it

Round 1 was the heaviest of the feature: **6 BLOCKER + 3 MAJOR + 1 MINOR**. Eight of the ten were the same mistake — *a check that is sound on the initiator, assumed to hold on a peer*:

- The decoder read only the kVault instruction, so a transfer could ride alongside a plausible deposit and still display "Deposit 10 USDC". Fixed by moving the instruction template into a shared `KaminoInstructionSequence` that both the validator and the decoder walk.
- Detection trusted the relayed `coin.chain` — an attacker-supplied off switch.
- A `.mismatch` verdict left Sign and Join enabled.
- A co-signer's withdraw was labelled "verified" while its headline amount was unchecked; that state is now `.amountUnverifiable`.

Round 2 added **2 BLOCKER + 1 MAJOR**: a **legacy** transaction invoking kVaults fell through the v0 parser into silence, bypassing both signing guards; and matching program and discriminator proved nothing about the *arguments* — an inflated compute-unit price is a direct SOL drain and is plainly readable offline.

The whole-branch pass over all 7 commits found **2 MAJOR + 1 MINOR**, both MAJORs at seams no per-step review could see: stripping the injected ComputeBudget selected the *pre-injection* template, which still matched, so bytes the layer-1 validator refuses read as **verified** while the fee row quoted a budget they didn't carry; and `farms::stake` could carry a specific amount instead of the `u64::MAX` sentinel.

The lesson is recorded in the review ledger: the decoder was first written as *"what can a co-signer establish?"*, when the right question is *"what would the initiating device have refused, minus only what genuinely needs the network?"* Almost nothing turned out to need the network.

### `FormScreen.validForm` — removed, not enforced

This is the judgement call most worth a second opinion.

`FormScreen` rendered `PrimaryButton(...).disabled(isContinueDisabled)` and never read `validForm`, so Continue was tappable regardless of validation on **every** screen built on it. The obvious fix — enforce `validForm` — turns out not to be available: Continue advances focus on the bond screens, opens the validator or pool picker on three others, and everywhere triggers the `validateErrors()` that reveals errors on an untouched form. Disabling it would break all of that.

So `validForm` is **removed** rather than enforced. Validity *is* enforced, in each caller's `transactionBuilder`. **12 callers changed, none behaviourally.** The Kamino deposit form separately adopts `isContinueDisabled` for the three states no amount can reach, and `makeDeposit` re-enforces the minimum and ceiling in exact base units.

### The fee row

`PrebuiltPayloadFee` now reads the payload instead of re-estimating from the display-only transaction, so the number shown is the number being signed — on both devices.

**Correction to what layer 3's description said:** the bound is *not* 0.00035 SOL total. `SolanaHelper.defaultFeeInLamports` is a flat, app-wide 0.001 SOL, and the compute budget adds up to 0.00035 SOL on top of it. The flat base is deliberately left alone — changing it for one flow is a repo-wide decision — and that is now stated in the code.

### Staleness

Implemented narrowly: a row that a refresh did not confirm shows when it was last true. The refresh *error* stays out of the chain banner, matching stake and LP — layer 2's reasoning for the banner still holds. What earns the per-row line is that a Kamino figure has **no other source in the app**, so a silently stale one is indistinguishable from a current one.

### Polish

Risk tiers and the wSOL rent disclosure (a SOL deposit creates a wSOL ATA and never closes it, stranding ~0.00204 SOL) now also appear on the verify screen — where the transaction is still refusable, and where a co-signer sees them for the first time. 20 new keys × 8 locales including `ko`.

## Step 8 — the farm-staked withdraw now works

The refusal every earlier layer carried is **gone, and it was earned**: the transaction was captured (no funds spent, nothing signed — the API builds unsigned transactions for any address), both farms discriminators were confirmed by deriving them from their Anchor names rather than matching observed bytes, and a *mixed* staked/unstaked position was found and captured too.

Three things the capture overturned, each of which the plan had wrong:

1. **The sequence is conditional on the AMOUNT, not on the vault having a farm.** A request at or below the unstaked balance is still the old 2-instruction shape; above it, 5 (or 6 on the SOL vault). The unstake is exactly the **shortfall** — `requested − unstaked` — not the staked balance, and the validator pins it in `u128` arithmetic rather than bounding it, so an over-release is refused too.
2. **The "a full withdraw cannot tip into the `u64::MAX` sentinel" premise was exactly backwards.** The API fires the sentinel at `requested ≥ balance`, so sending the exact held balance is precisely what produces it — meaning the shipped validator would have refused **every legitimate full exit**. Resolved by making the maximum the largest amount *strictly* below the balance (≤ 1 base unit of dust) rather than by accepting the sentinel, whose bytes are identical to an over-request and which a co-signer has no intent to condition on. Truncation alone would not have been enough: **70 of 80 live positions have exactly-representable totals**, so this is the majority case, not an edge.
3. **`withdrawUnitLimit` of 220,000 was too small.** A staked withdraw costs 283k–309k CU, so every staked shape aborted in simulation on the compute meter. Raised to 400,000; all six vectors re-simulate `err: null`.

Also caught: a full **wrapped-SOL** withdraw carries **two** `closeAccount` instructions, which the shipped `repeatable: false` would have refused — and, underneath that, a latent bug from layer 1 where a `match` step that was both `required` and `repeatable` never enforced its requirement at all, silently voiding the wSOL close requirement.

Review found 3 BLOCKERs in this step, **all in the offline decoder, none in the online validator**. The most interesting was `matchesFarm` being inert because the farm is ALT-resolved: failing closed there would have refused every legitimate staked withdraw on a co-signer, so it is instead closed by deriving the farms `user_state`, a PDA over `(farm, owner)` and a static key in all six vectors.

## Steps 9–11 — a live-API conformance suite, and the two defects it found

Every other Kamino test runs against golden vectors captured on 2026-08-04/05. That is what makes
them deterministic and also what makes them blind: they freeze `api.kamino.finance` at a moment in
time, so a changed minimum, a reordered instruction or a renamed field passes all 384 of them and
breaks in a user's hands.

`KaminoLiveAPITests` asserts the same claims against the live API — vault identity against the pinned
registry, both transaction shapes through the validator and the offline decoder, the
amount-conditional withdraw sequence, the `u64::MAX` sentinel boundary, and the compute a staked
withdraw needs. **Opt-in behind `KAMINO_LIVE=1`, so `make test` and CI never run it** (without the
flag all 9 tests skip in 0.02 s); a network test in the gate is a flaky gate.

Two rules it is written to, both learned in its first three runs: it asserts **shape and invariants,
never balances**, because the sampled wallets move and a wallet that exited is a skip with its
address named; and **a dropped connection, timeout, 429 or 5xx skips rather than fails**, because a
live suite that cries drift when the network hiccups is ignored by its second false alarm.

### ⚠️ Both published minimums are refused by the program that publishes them

Neither could have been caught by a frozen vector — every captured amount sits well clear of both
floors. Both are measured against the live chain at one base unit of resolution.

**The withdraw minimum was in the wrong unit *and* too low.** `state.minWithdrawAmount` was read as
SHARE base units — which is what the field's name suggests, and what the withdraw endpoint's own
units suggest. It is a **token** figure, and the program refuses a withdraw whose token value does
not exceed **twice** it (`WithdrawAmountBelowMinimum`, custom 7004, `vault_operations.rs:325`):

| Vault | published | refuses | accepts |
|---|---|---|---|
| Steakhouse USDC | 1000 | 1898 share base units | **1899** |
| Allez SOL | 1000 | 1861 | **1862** |

Read as a share count the published figure is 1000 — roughly half of what executes — so the form
advertised a minimum, accepted it, and then failed at the simulation gate with an error about an
amount the user had just been told was allowed. Both withdraw shapes share the boundary, so it
belongs to the vault withdraw rather than the farm release. On the SOL vault the two readings differ
by a factor of ~930; on the dollar vaults the rate is near 1, which is why the wrong reading looked
right for so long.

**The deposit minimum is in the right unit but still refused at the published figure**
(`DepositAmountBelowMinimum`, custom 7026, `vault_operations.rs:113`): Steakhouse accepts 100,008
against a published 100,000, Allez 10,000,010 against 10,000,000. A handful of base units, tracking
the share rate, so it reads as a rounding loss in the shares minted rather than a second threshold.

⚠️ **That one was not cosmetic — it broke the SOL vault's Max button.**
`maxNativeDepositLamports` measures the reserve by preparing a probe deposit of exactly `minDeposit`,
and the preparer requires the probe to simulate cleanly, so a probe below the program's floor made
the whole measurement throw.

Both are now derived rather than parsed: the withdraw minimum from the token figure ×3 converted to
shares **rounded up**, the deposit minimum with a margin of one part in a thousand (16-base-unit
floor). Both multiples carry margin above what was measured **on purpose** — the rules are readings
of observed behaviour, not documented contracts, since the kVault program source is not published —
and both are invisible in money: a tenth of a percent of a ten-cent minimum. Re-verified after the
change, and the boundaries were still exactly where they were hours later.

Review also caught two display consequences, both fixed: the withdraw form's minimum rounds **up**
while its maximum rounds **down**, so a position sitting on the minimum could render a minimum above
its own balance while Max still worked; and the deposit form stayed enabled for a wallet holding
**less than** the minimum, where no typeable amount satisfies both validators.

**The generalisable lesson, recorded in the wiki: a published minimum is a claim, not a contract.**
Any integration that surfaces a third party's minimum should simulate at exactly that figure before
believing it.

## Remaining known gaps
- **The offline decode cannot verify account identities *inside* a matched instruction.** That genuinely requires the lookup table. Documented precisely rather than papered over; the initiating device closes it.
- **Offline, *how many* shares a farm release moves can only be bounded, not pinned** — the exact figure needs a balance that is not on the wire. The initiating device pins it exactly.
- Out of scope and unchanged: no DeFi state survives a `.vult` restore (pre-existing, affects every provider), the DeFi chain-list sort key, and Fast Vault compatibility for a non-staking `signSolana`.

## Rebased onto current main

The whole stack was rebased onto `origin/main@f5e206120` on 2026-08-05, so all four PRs merge cleanly. Since vultisig-ios CI builds each PR *merged with main*, this also means the local green now reflects what CI will build.

There was exactly one textual conflict, in `SendCryptoVerifyViewModel`. Ours added `isKaminoDecodeRefused`, main added `hasLoadError`, both as terms on the same predicate. They are independent fail-closed holds on the same button — the figures were never resolved, versus the bytes contradict the figures — so they **compose** rather than replace each other:

```swift
!isValidForm || isLoading || hasBalanceError || hasLoadError || isKaminoDecodeRefused
```

Taking either side alone would have silently dropped a guard. `BlockChainService` and the eight locale files auto-merged and were verified rather than trusted — the Kamino branch in `refreshSolanaBlockhash` is byte-identical to its pre-rebase self, and the locales are a clean union with all 8 carrying identical key sets.

`git range-diff` confirms **7 of the 8 commits are patch-identical** across the rebase; the only one that differs is this predicate line and its comment. Post-rebase gate: **5183 tests, 1 skipped, 0 failures** — the rise from 5067 is exactly main's new tests, with the Kamino contribution unchanged at 384.

## Which feature is affected?
- [x] Sending — changes what the verify screen displays and when signing is blocked
- [ ] Swap

**tx link — the farm-staked withdraw, broadcast on mainnet and succeeded.**

[`2GrF4kcQ…5L7MvG`](https://solscan.io/tx/2GrF4kcQ2vD7mcTx3kDnE3mJqJbvici1pdu1uVYk6CroEcZ6saWi9Ee9iBDDht72ATjSXp3dqiqDaoDsSd5L7MvG) — Steakhouse USDC, fee 13,000 lamports, `err: null`.

On-chain instruction sequence: `Unstake` → `WithdrawUnstakedDeposits` → `Withdraw` → `RefreshReservesBatch`.

This is the specific thing this layer adds. Layer 3 refuses this path outright, because a deposit auto-stakes and every real position is therefore farm-staked; the two `farms` instructions ahead of the vault withdraw are what release the shares before they can be redeemed. The transaction also carries the injected `ComputeBudget` pair, so the fee it actually paid is the one the validator pinned.

The three **deposits** covering all curated vaults are on layer 3 (#5056).

## Parity

Required — this changes keysign display semantics and the fee row on both devices.

## Checklist

Gate: **5197 tests, 10 skipped, 0 failures** (the 10 skips are 1 pre-existing plus the 9 opt-in live tests). SwiftLint 0 violations across the Defi sources and tests.

_To add: the verify screen on initiator and co-signer._


VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoTransactionDecoder.swift
```diff
@@ -0,0 +1,601 @@
...
+        return false
+    }
+
+    /// Decodes the single kvault deposit or withdraw these bytes perform, or
+    /// `nil` when they do not perform exactly one that this app recognises.
+    ///
+    /// `nil` is a refusal to describe, never a "probably fine". Every guard
+    /// below is a case where the honest answer is that the bytes are not the
+    /// shape this decoder knows how to read.
+    static func decode(_ transaction: SolanaV0Transaction) -> KaminoDecodedTransaction? {
```
johnnyluo, line 200: decode derives the signer from account 0 but never applies `SolanaV0Transaction.validateUnsignedSingleSigner`. The raw signing path signs the message verbatim and only replaces signature slot 0, so a relayed Kamino transaction with multiple required signatures or prefilled extra signatures can still be shown as verified even though the initiator-side validator rejects that shape. Require the same single unsigned fee-payer precondition before returning a `KaminoDecodedTransaction`.

VultisigApp/VultisigApp/Blockchain/Solana/Models/SolanaProgramDerivedAddress.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// the ed25519 curve — an on-curve hash is an ordinary public key that
+    /// somebody could hold the private key for, so it is skipped rather than
+    /// returned. Searching downward from 255 is what makes the answer unique.
+    static func find(seeds: [Data], programId: String) -> String? {
+        guard let program = Base58.decodeNoCheck(string: programId), program.count == 32 else { return nil }
+        // Solana allows 16 seeds in total and the bump occupies the last slot,
+        // so a caller may supply at most 15.
+        guard seeds.allSatisfy({ $0.count <= 32 }), seeds.count < 16 else { return nil }
+
+        // 255 down to 1, matching the runtime. Bump 0 is not a value
```
johnnyluo, line 40: Solana `findProgramAddress` searches bumps 255 down through 0, but this loop stops at 1 and the comment says bump 0 is impossible. For the rare seed/program pair whose canonical PDA only succeeds at bump 0, this returns nil and makes the offline farm-position check reject a valid staked Kamino withdraw. Include 0 and add a regression vector.

VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoTransactionDecoder.swift
gastonm5: Fixed in `4dcece2`.

Confirmed the gap: `decode` derived the signer from slot 0, but nothing on the joining path asserted the precondition `KaminoTransactionValidator` applies before building, and `KaminoVerifyPresentation.state` went from `decode` straight to `.verified`. The raw signing path then hashes the message verbatim and replaces slot 0 only, so a relayed transaction requiring a second signature — or carrying one somebody else had already contributed — was summarised as an ordinary deposit with the other signer's authority nowhere on the screen.

`decode` now requires `validateUnsignedSingleSigner(feePayer: signer)` before returning, so those bytes surface as `.unreadable` rather than verified. Two tests: a transaction with `numRequiredSignatures = 2`, asserting both `decode == nil` and `state == .unreadable`, and one whose signature slot is already filled.

VultisigApp/VultisigApp/Blockchain/Solana/Models/SolanaProgramDerivedAddress.swift
gastonm5: I checked this against the reference implementations before touching it, and I don't think bump 0 should be included — `findProgramAddress` stops at 1 in both of them:

- web3.js, `PublicKey.findProgramAddressSync`: `let nonce = 255; while (nonce != 0) { … nonce--; }` — the loop exits before 0 is ever tried. ([publickey.ts](https://github.com/solana-foundation/solana-web3.js/blob/master/src/publickey.ts))
- Rust, `Pubkey::try_find_program_address`: `let mut bump_seed = [u8::MAX]; for _ in 0..u8::MAX { … bump_seed[0] -= 1; }` — 255 iterations starting at 255, so the last value handed to `create_program_address` is 1; `bump_seed` only reaches 0 as the loop falls out and returns `None`. ([pubkey.rs](https://docs.rs/solana-program/1.6.7/src/solana_program/pubkey.rs.html))

So "the canonical PDA only succeeds at bump 0" is not a case that exists on chain: the program that derives the farm position uses the same algorithm, so it could never produce that address either. `stride(from: 255, through: 1, by: -1)` matches the runtime, and a valid staked withdraw cannot be rejected by it — including 0 would instead make us accept an address the chain does not agree is a PDA, which is what the comment there is guarding against.

Left as is, but happy to be shown a counter-example.

johnnyluo (approved):
LGTM

PR #5064, opened 2026-08-05
fix(send): read a balance exactly past Int64 so a send can't validate against funds that aren't there

### LINKED ISSUE #5037: Balance parse rounds up past Int64: `toBigInt(decimals:)` goes through Double, so large balances validate against more funds than exist
## What
`String.toBigInt(decimals:)` — the idiom used to turn a coin's `rawBalance` into base units — parses through `NumberFormatter` → `Double`. Past `Int64` that loses precision and **rounds upward**, so any balance above roughly 9.2 (for an 18-decimal asset, ~9.2 ETH) is reported as *larger than it actually is*.

## Why it matters
Every guard that compares an amount against the balance is then comparing against an inflated number. The concrete failure is a send that passes validation locally and is rejected at broadcast — after a completed MPC ceremony, so the signing round is wasted.

This was found while fixing #5019 (max send on EVM L2s). That PR fixes the parse **only inside its own max-send clamp** (`fix(send): read the raw balance exactly when clamping an EVM max send`), because handing the clamp an inflated balance would have reintroduced the very bug it was fixing. The lossy idiom is still live everywhere else.

## Where (non-exhaustive — needs a sweep)
- `SendCryptoVerifyLogic.validateBalanceWithFee` — the `amount + fee > balance` guard
- `isAmountExceeded`
- other `rawBalance.toBigInt(decimals:)` call sites across the send path

## Ask
Sweep every `toBigInt(decimals:)` use that feeds a balance comparison or an amount derivation and move it to the exact-parse path introduced in #5019. Add a test pinning a >`Int64` balance so the rounding direction can't regress.

## Notes
Rounding *up* is the dangerous direction — it lets the app believe the user can afford something they can't. A downward-rounding bug would merely strand dust.

gastonm5:
## Problem

Every send affordability guard turns a coin's `rawBalance` into base units with `String.toBigInt(decimals:)`. Past `Int64` — about **9.223 on an 18-decimal asset** — that parse rounds the balance, and on the common inputs it rounds **UP**. The guards then compare the requested amount against funds the vault does not hold.

The concrete failure: a send clears local validation and is rejected at broadcast — **after a completed MPC ceremony**, so the signing round is wasted. Rounding *up* is the dangerous direction; a downward error would merely strand dust.

## Root cause — the exact `Double` hop

`toBigInt(decimals:)` → `toDecimal()` → `parseInput()` → **`NumberFormatter.number(from:)`**.

`NumberFormatter.number(from:)` returns an `Int64`-backed `NSNumber` (`objCType == "q"`) while the value fits in `Int64`, and a **`Double`-backed** one (`objCType == "d"`) once it doesn't. `.decimalValue` then faithfully converts an already-destroyed value. Setting `generatesDecimalNumbers = true` does not help — ICU parses through a double either way; measured, both give the same wrong digits.

Measured (identical in `en_US`, `de_DE`, `pt_BR`):

| input (18-dec base units) | `objCType` | old `toBigInt(decimals: 18)` | delta |
|---|---|---|---|
| `9223372036854775807` (`Int64.max`) | `q` | `9223372036854775807` | exact |
| `9223372036854775808` | `d` | `9223372036854776000` | **+193** |
| `12345678901234567890` (~12.35 ETH) | `d` | `12345678901234570000` | **+2110** |
| `99999999999999999999` (~100 ETH) | `d` | `100000000000000000000` | **+1** |

## Fix strategy — leaf fix **plus** a shared accessor, not a call-site-only sweep

Two options were weighed, per the issue.

**(A) Sweep the call sites onto a shared helper, leave `toBigInt(decimals:)` lossy.** Rejected as the only action: it leaves a loaded gun in a `String` extension with an inviting name, and the next `rawBalance.toBigInt(decimals:)` re-fires the bug. #5034 already demonstrates that instinct — it reached for the idiom and had to work around it locally.

**(B) Make `toBigInt(decimals:)` itself exact.** The issue framed this as the broader blast radius. **On this tree it isn't**: the complete set of `toBigInt(decimals:)` call sites is 11, all in Send plus one Ripple trust-line view model. The widely-used overload is the zero-argument `toBigInt()`, which is already exact (`BigInt(String)`, no `Double`).

Two premise corrections that drove the choice, both verified against the source:

1. **`toBigInt(decimals:)` does not scale a human decimal.** It truncates to at most `decimals` fraction digits and then integer-parses, yielding `0` when anything fractional survives — so `"1.5".toBigInt(decimals: 18) == 0`, not `1.5e18`.
2. **There is therefore no "human-decimal amount" caller to protect.** `SendCryptoLogic.amountInRaw` multiplies by `pow(10, decimals)` *first* and passes `Decimal.description`, an integer string; the `decimals:` argument there is a vestigial no-op guard. Making the leaf exact for integer strings cannot break amount entry.

**Chosen: B, then A on top.**

1. **Leaf** — an exact fast path for a plain ASCII-digit string. Removes the trap permanently, and fixes `amountInRaw` in place, which a call-site-only sweep would have left broken for any typed amount above ~9.22 on an 18-decimal asset.
2. **Accessor** — `Coin.balanceRaw` in `Core/Extensions/CoinExtension.swift`, defined as `rawBalance.toBigInt(decimals: decimals)`. With the leaf fixed this *is* #5034's private `exactRawBalance(of:)`, so that helper has a home to collapse into. It also names the quantity instead of repeating an idiom whose `decimals:` argument does nothing. Placed in the extension file rather than `Model/Coin.swift`, which is SwiftData `@Model` territory.

The fast path triggers only on an all-ASCII-digit string, where the truncation is a provable no-op. **Signs, decimal points, grouping separators, locale forms, exponent notation, whitespace, garbage and empty all fall through to the byte-identical existing path** — the fallback property #5034 established: a non-integer degrades to the shared parse rather than collapsing to zero. Verified by a differential probe across `en_US`/`de_DE`/`pt_BR` over 33 input shapes: every non-digit-only input matches the old implementation exactly; every digit-only input past `Int64` differs by being right.

Deliberately **not** widened to fractional strings — that would also flip `"1.000000000000000001".toBigInt(decimals: 18)` from `1` to `0`, a semantic change no caller asked for. No production call site passes a fractional string; the residual is documented in the doc comment rather than fixed speculatively.

## Call-site sweep

Ten sites moved from `x.rawBalance.toBigInt(decimals: x.decimals)` to `x.balanceRaw`:

| File | Site |
|---|---|
| `SendCryptoLogic.swift` | `:52` `isAmountExceeded`, `:227` `verifyMaxCandidateRaw` |
| `SendCryptoVerifyLogic.swift` | `:142` `validateBalanceWithFee`, `:207` Cardano min-ADA reserve, `:217` ERC-20 gas balance, `:354` XRP trust-line reserve |
| `SendDetailsViewModel.swift` | `:803` UTXO max probe amount, `:1186` `validateERC20GasBalance` |
| `RippleTrustLineActivationViewModel.swift` | `:100` `spendableDrops` |
| `VultisigAppTests/Send/Mocks/MockSendInteractor.swift` | `:90` max-plan stub — the mock modelled the bug, so a test against a large balance would have passed vacuously |

**Left alone — already exact** (plain `BigInt(String)`, no `Double`): `SendCryptoLogic:98` `canBeReaped`, `:196` `terraClassicMaxValue`, `Coin.getMaxValue:351`, `DefiChainScreenModel:53`, `SwapCryptoLogic:94`/`:108`, `LimitSwapFormViewModel:354`, `SolanaStakingVerifyResolver:38`. Migrating these for cosmetic consistency would widen the diff on a fund-safety PR with no behaviour to gain.

**Left alone — in scope but not this idiom**: `SendCryptoLogic:26` `amountInRaw` keeps its call. It is not a balance read, and the leaf fix corrects it in place.

After the sweep the only remaining `rawBalance.toBigInt(decimals:)` in the tree is inside `Coin.balanceRaw` itself.

## Tests

- `StringExtensionTests` — exactness at and past `Int64.max` (including a value past `Decimal`'s 3.4e38 ceiling); a sweep over `10^15…10^30 ± {1, 0, 1, 7}` asserting the parse can **never** report more than the string holds; and the fallback shapes (fractional, signed, exponent, garbage, empty, leading zeros, negative `decimals`) asserted unchanged.
- `CoinExtensionTests` — `balanceRaw` exactness, the one-directional property, and the fallback that keeps `"1e18"` at `1e18` instead of collapsing to zero (`BigInt("1e18")` is `nil`, so a naive exact-only helper would have).
- `SendValidationTests` / `SendCryptoVerifyViewModelTests` — the guards that actually spend funds: a vault holding **99.999999999999999999 ETH** must reject a **100 ETH** send. The old read returned a flat `1e20`, so both `isAmountExceeded` and `validateBalanceWithFee` waved that send through. These fail if the fix is reverted.

## ⚠️ Conflicts with #5034 — recommended merge order

**#5034 (`fix/evm-max-send-clamp-5019`) is open and unmerged.** This branch is cut from `origin/main`; nothing is cherry-picked from it. Both branches edit `SendCryptoLogic.swift` and `SendCryptoVerifyLogic.swift`, so they **will** conflict. That is expected and accepted.

**Merge #5034 first, then rebase this branch onto it.** On the rebase:

- #5034's private `exactRawBalance(of:)` should collapse into `Coin.balanceRaw` so there is exactly one helper, not two.
- #5034's `EvmMaxSendClampTests.testClampReadsBalancesBeyondInt64Exactly` carries an `XCTAssertNotEqual` asserting the *shared* parser is still lossy, with the comment "if this starts passing, the shared balance parser gained full precision and this guard is stale". This branch is exactly that event — invert that assertion on the rebase.

## Verification

- **Full unit-test suite green**: `** TEST SUCCEEDED **` — **4807 tests, 1 skipped, 0 failures**, iOS Simulator (dedicated `gate-5037`, iPhone 16 Pro, `en_US`). All 7 new tests pass.
- `swiftlint --config VultisigApp/.swiftlint.yml` on all 11 touched files: **0 violations**.
- Root cause and every measured figure above reproduced in a standalone Swift probe against `Foundation`, not inferred.
- Cross-model review: three per-step Codex passes plus a whole-branch pass. Two findings raised and both fixed — an over-claiming doc comment on `balanceRaw`, and locale-sensitive fraction literals in the general-path test (widened to five-digit fractions, which no locale groups).
- **Not verified on device.** No manual send was performed against a real >9.22 ETH balance.

## 🔒 Fund safety — please review as such

This changes how the send/signing path decides whether a transfer is affordable. It needs **human review and on-device verification**, not just a green suite. The behaviour to confirm by hand: a vault with more than ~9.22 of an 18-decimal native asset can no longer be talked into a send that exceeds its real balance, and a normal send below that threshold is unchanged.

## Known adjacent gap (not fixed here)

`Coin.balanceDecimal` is the same bug one layer up — `rawBalance.toDecimal() / pow(10, decimals)`, also `Double`-backed. It is mostly display-facing, but two guards outside Send weigh amounts against it:

- `SwapCryptoLogic.balanceError` (`:890-891`) — the **swap** affordability rule, for both the spend and the gas coin.
- `FunctionCallCustom.isFormValid(for:)` (`:110`) — the submit-time gate on custom function calls.

Out of scope for an issue scoped to `toBigInt(decimals:)`, and a different idiom needing a different fix, but the same rounded-up-balance failure is reachable through them. Worth a follow-up issue; `balanceRaw`'s doc comment says so explicitly rather than implying the sweep is universal.

Closes #5037

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved balance accuracy for assets with very large values and high decimal precision.
  * Prevented rounding and floating-point conversion errors during send validation, maximum-send calculations, fee checks, and reserve checks.
  * Ensured transactions exceeding the available balance by the smallest unit are correctly rejected.
  * Improved handling of balances represented in scientific notation and other non-standard formats.

* **Tests**
  * Added coverage for large balances, exact integer parsing, rounding prevention, invalid inputs, and send-validation edge cases.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
@@ -5,6 +5,7 @@
 //  Created by Johnny Luo on 15/3/2024.
 //
 
+import BigInt
 import Foundation
 import WalletCore
 
@@ -18,6 +19,17 @@ extension Coin {
 
     var isDefiOnly: Bool {
         Coin.defiOnlyTickers.contains(ticker.uppercased())
+    }
+
+    /// The balance in base units.
+    ///
+    /// `rawBalance` is a base-unit integer string, and this is the one reading
+    /// of it that every affordability guard shares — so no guard ever compares
+    /// against a balance that is merely close to the real one. A value that
+    /// isn't a plain integer still goes through the shared parse rather than
+    /// collapsing to a zero balance.
+    var balanceRaw: BigInt {
+        rawBalance.toBigInt(decimals: decimals)
     }
 }
 
--- a/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
@@ -21,13 +21,17 @@ extension Coin {
         Coin.defiOnlyTickers.contains(ticker.uppercased())
     }
 
-    /// The balance in base units.
+    /// The balance in base units, read exactly.
     ///
-    /// `rawBalance` is a base-unit integer string, and this is the one reading
-    /// of it that every affordability guard shares — so no guard ever compares
-    /// against a balance that is merely close to the real one. A value that
-    /// isn't a plain integer still goes through the shared parse rather than
-    /// collapsing to a zero balance.
+    /// `rawBalance` is a base-unit integer string, and this is the reading of it
+    /// that the send affordability guards share, so none of them can drift onto
+    /// a value that is merely close to the real one. A value that isn't a plain
+    /// integer still goes through the shared parse rather than collapsing to a
+    /// zero balance.
+    ///
+    /// Not yet universal: guards outside Send still weigh amounts against
+    /// `balanceDecimal`, which divides through the same `Double`-backed parser
+    /// this exists to avoid.
     var balanceRaw: BigInt {
         rawBalance.toBigInt(decimals: decimals)
     }
--- a/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
@@ -199,9 +199,27 @@ extension String {
         return valueBigInt
     }
 
-    // We must truncate before converting to bigInt.
+    /// Base-unit integer value of a numeric string, truncated to at most
+    /// `decimals` fraction digits. Anything still fractional after the
+    /// truncation yields zero — this reads an already-scaled value, it does not
+    /// scale one (`"1.5"` with 18 decimals is zero, not 1.5e18).
+    ///
+    /// A string of plain ASCII digits is read exactly. The general path goes
+    /// through `NumberFormatter`, which returns a `Double`-backed `NSNumber`
+    /// once the value no longer fits in `Int64` — about 9.223 on an 18-decimal
+    /// asset — and keeps only 17 significant digits. That commonly rounds the
+    /// value UP, and a balance reported larger than it really is lets an
+    /// affordability guard pass a send the chain then rejects at broadcast,
+    /// after the signing ceremony has already run.
+    ///
+    /// Non-integer input still takes the `NumberFormatter` path, imprecision
+    /// included, rather than collapsing to zero; no caller passes one, because
+    /// a surviving fraction is already the zero case above.
     func toBigInt(decimals: Int) -> BigInt {
-        self.toDecimal().truncated(toPlaces: decimals).description.toBigInt()
+        if decimals >= 0, isNotEmpty, allSatisfy({ $0.isASCII && $0.isNumber }), let exact = BigInt(self) {
+            return exact
+        }
+        return self.toDecimal().truncated(toPlaces: decimals).description.toBigInt()
     }
 
     func isValidDecimal() -> Bool {
--- a/VultisigApp/VultisigApp/Features/Send/Logic/SendCryptoLogic.swift
+++ b/VultisigApp/VultisigApp/Features/Send/Logic/SendCryptoLogic.swift
@@ -49,7 +49,7 @@ enum SendCryptoLogic {
         }
 
         let amountRaw = amountInRaw(coin: coin, amount: amount)
-        let balanceRaw = coin.rawBalance.toBigInt(decimals: coin.decimals)
+        let balanceRaw = coin.balanceRaw
 
         if (sendMaxAmount && (coin.chainType == .UTXO || coin.chainType == .Cardano || coin.chainType == .Ton))
             || !coin.isNativeToken {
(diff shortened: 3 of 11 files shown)


johnnyluo (approved):
LGTM

PR #5068, opened 2026-08-06
fix(import): stop a vault import from silently overwriting a stored vault

### LINKED ISSUE #5066: Fix vault import silently overwriting a different vault on a unique-attribute collision
## Summary

Importing a vault can **silently replace a different, unrelated vault** already on the device, destroying its key shares. The user sees a successful import and no error.

This is **pre-existing on `main`** — not introduced by any open PR. It was found while reviewing unrelated import work.

## Mechanism

`Vault` marks **three** fields as `@Attribute(.unique)`: `name`, `pubKeyECDSA` and `pubKeyEdDSA`.

`EncryptedBackupViewModel.isVaultUnique` rejects an incoming vault only when **both** public keys match an existing one:

```swift
if vault.pubKeyECDSA == backupVault.pubKeyECDSA &&
    vault.pubKeyEdDSA == backupVault.pubKeyEdDSA {
    return false
}
```

So a vault that collides on **one** key — or on **nothing but the name** — passes the guard and reaches `context.insert`. SwiftData treats a unique-attribute collision as an **upsert**: the incoming vault replaces the stored one, and the stored vault's key shares go with it.

Nothing throws. Nothing is logged. The import reports success.

## Steps to reproduce

1. Have a vault stored, e.g. named `Savings`.
2. Import a `.vult` for a genuinely different vault (different key material) that happens to be named `Savings`.
3. The import reports success.
4. `Savings` is now the imported vault. The original is gone, along with its key shares — with no backup, its funds are unreachable.

The same happens when only `pubKeyECDSA` matches and `pubKeyEdDSA` differs, or vice versa.

## Impact

Permanent, silent loss of a vault's key material. A name collision is not exotic — `Main`, `Savings` and `Test` are exactly the names people reuse across devices, and the vault name is user-chosen and freely editable.

## Note on the existing test

`ZipImportDuplicateTests` currently pins the same-ECDSA / different-EdDSA case as **acceptable**, so the suite endorses one of the dangerous paths. Fixing this means revisiting that expectation, and deciding deliberately whether such a vault is a duplicate to reject or a distinct vault to admit under a non-colliding identity.

## Suggested direction

Treat "is this a duplicate?" and "can this be inserted?" as two different questions.

- The uniqueness guard must cover **every** `@Attribute(.unique)` field, not just the two that identify a vault semantically.
- A name collision between genuinely different vaults should not be a rejection — it should be resolved (disambiguate the incoming name) rather than allowed to upsert.
- Worth auditing every other `context.insert` path against its model's full set of unique attributes; this class of bug is silent by construction.

## Environment

Found on `main` (also present on the passcode stack, which does not change this behaviour). Affects iOS and macOS.

gastonm5:
> [!WARNING]
> **Fund-safety change — needs human + on-device review.** This touches the vault import path, where the failure mode is *permanent, silent loss of a vault's key shares*. The unit tests exercise the decision function; the premise (SwiftData upsert-on-unique-collision) deserves a real device/simulator import run before merge.

## Problem

Importing a vault could **silently replace a different, unrelated vault** already on the device and destroy its key shares. The user saw a successful import and no error.

## Root cause

`Vault` marks **four** fields `@Attribute(.unique)` — `name`, `pubKeyECDSA`, `pubKeyEdDSA` and `publicKeyMLDSA44` (`VultisigApp/VultisigApp/Model/Vault.swift:12-15`; the issue says three, `publicKeyMLDSA44` landed later).

`EncryptedBackupViewModel.isVaultUnique` rejected an incoming backup only when **both** pubkeys matched:

```swift
if vault.pubKeyECDSA == backupVault.pubKeyECDSA &&
    vault.pubKeyEdDSA == backupVault.pubKeyEdDSA { return false }
```

So a backup colliding on **one** pubkey — or on **nothing but the name** — passed the guard and reached `modelContext.insert`. SwiftData treats a unique-attribute collision as an **upsert**: the incoming vault replaces the stored row, and the stored vault's key shares go with it. Nothing throws, nothing is logged, the import reports success.

Four call sites fed that guard and then inserted, all in `EncryptedBackupViewModel.swift`: `importVaults` (zip batch), `restoreVaultBack` (`.bak`), and both branches of `restoreVault` (new `BackupVault` format + legacy `Vault` JSON fallback).

## Fix

The guard was answering two questions with one predicate. They are now separate:

**"Is this a duplicate?"** — `isVaultUnique(backupVault:vaults:)`. Any **single** shared public key (`pubKeyECDSA`, `pubKeyEdDSA` *or* `publicKeyMLDSA44`) means the same key material, so any one match rejects. A vault cannot legitimately share one key with a *different* vault — @johnnyluo confirmed on the issue that both pubkeys "should be unique and have no collision at all", so a partial collision means something is wrong, and inserting it would upsert key shares away. Empty/`nil` keys carry no identity and never count as a match.

**"Can this be stored without clobbering something?"** — `importDecision(for:existing:)`, returning:

| | |
|---|---|
| `.duplicate` | same key material — skip, `vaultAlreadyExists` |
| `.insert(name:)` | safe; `name` is the backup's own name, disambiguated (`Savings` → `Savings (2)`) when a stored vault holds it |
| `.unsafeCollision` | still collides on a unique attribute after the name was resolved — refuse rather than upsert |

- **Name collisions are resolved, not rejected.** `availableVaultName(basedOn:taken:)` derives from the incoming vault's own name rather than reusing `Vault.getUniqueVaultName`, which generates `Vault #N` from scratch and would throw away the user's chosen name. It checks the stored vaults *and* everything already renamed earlier in the same zip batch, and terminates by pigeonhole (at most `taken.count` names are unavailable; the scan covers `taken.count + 1` candidates), so there is no arbitrary cap and no unbounded loop.
- **Empty pubkeys.** `pubKeyECDSA`/`pubKeyEdDSA` default to `""`, and `""` is a real value in a SQLite unique index (unlike `NULL`), so two key-less backups would upsert each other. They are *not* treated as duplicates — `""` says nothing about identity — but the final gate refuses them with a clear error rather than letting the insert through. `publicKeyMLDSA44` is normalized with `nilIfEmpty` on both sides, because the proto path already does that (`Vault+ProtoMappable.swift:37`) while the JSON path decodes `""` verbatim.
- The `.unsafeCollision` check deliberately **restates the raw index semantics** — one clause per `@Attribute(.unique)` field — rather than reusing the duplicate rule, whose empty/`nil` tolerance is exactly what would let a key-less backup through. It is the single place to extend if a fifth unique attribute is ever added.
- All four import paths now go through one `insertIfSafe(_:existing:modelContext:)` helper, so there is a single place where a vault becomes stored.

No schema change: `Model/` is untouched.

## Deliberate test-expectation inversion

`ZipImportDuplicateTests.testIsVaultUnique_withPartialKeyMatch_returnsTrue` pinned the same-ECDSA / different-EdDSA case as **acceptable** — i.e. the suite endorsed one of the dangerous paths. It is now `..._returnsFalse`, with a comment explaining why. **This is an intentional behaviour change**: a backup sharing one pubkey with a stored vault is now refused where it previously imported (and destroyed the stored vault).

## `context.insert` audit (the issue asks for it)

Every `@Model` with `@Attribute(.unique)` was checked against every insert site. **No other production insert path is in the same silent-key-material-loss class**, so nothing outside the import path was changed.

<details>
<summary>Full audit — 14 models with unique attributes</summary>

| Model | unique attr(s) | insert sites | verdict |
|---|---|---|---|
| `Vault` | `name`, `pubKeyECDSA`, `pubKeyEdDSA`, `publicKeyMLDSA44` | `EncryptedBackupViewModel` ×4 | **fixed here** |
| `Vault` | ” | `KeygenViewModel.swift:214, 486, 828, 901` | guarded — all behind `confirmDuplicateVaultIfNeeded` + user confirmation; the upsert is *intended* for the reshare/replace case |
| `Vault` | ” | `Core/Extensions/VaultExtension.swift:14-15` | preview/test fixture only (both fixtures share `pubKeyECDSA`, so the "two sample vaults" preview only ever shows one) |
| `CustomRPCOverride` | `chainRaw` | `CustomRPCStore.swift:68, 145` | guarded (fetch-then-update; migration checks membership) |
| `StoredPendingTransaction` | `txHash` | `PendingTransactionStorage.swift:84` | guarded (predicate fetch, update branch) |
| `DatabaseRate` | `id` (`fiat-crypto`) | `RateProvider.swift:185` | guarded (update-else-insert) |
| `CirclePosition` | `id` (`circle_<pubKeyECDSA>`) | `CirclePositionStorageService.swift:22` | guarded; regenerable cache |
| `YieldPosition` | `id` (`<provider>_<pubKeyECDSA>`) | `YieldPositionStorageService.swift:39, 61` | guarded; regenerable cache |
| `YieldRedemptionRecord` | `id` | via relationship, `YieldPositionStorageService.swift:74` | delete-then-replace; unreachable today (Circle emits no redemptions) |
| `LimitOrder` | `id` (`<txHash>_<pubKeyECDSA>`) | `LimitOrderStorageService.swift:64` | guarded (throws `.duplicate`; rejects empty `inboundTxHash`) |
| `StakePosition` | `id` | `DefiPositionsStorageService.swift:114, 161, 181` | guarded |
| `LPPosition` | `id` | `DefiPositionsStorageService.swift:42, 208` | guarded |
| `BondPosition` | `id` | `DefiPositionsStorageService.swift:77` | guarded |
| `ReferralCode` / `ReferredCode` | `id` | `ReferralViewModel.swift:125`, `ReferredViewModel.swift:114` | random `UUID()` keys |
| `ChainPublicKey` | `id` (`<chain>-<publicKeyHex>`) | relationship-driven only | see note below |

Three things worth a maintainer's eye, all filed as observations rather than changed here:

1. **`confirmDuplicateVaultIfNeeded` checks only `pubKeyECDSA`, never `name`** (`KeygenViewModel.swift:178-194`) — yet `name` is unique too. A name collision at any of the four keygen inserts would upsert a *different* vault away. Today that is only prevented at a distance, by `VaultNameValidator` on the initiating device and `isVaultNameAlreadyExist` on the joining device; nothing enforces it at the insert.
2. **`ChainPublicKey.id` is `"<chain.name>-<publicKeyHex>"` with no vault component.** Two vaults derived from the same seed (the same mnemonic key-imported twice, or an import alongside the original) produce identical ids, so the second vault's relationship-driven insert *steals* the row from the first. Not key material — a derived public key, regenerable — but it silently mutates the older vault's `chainPublicKeys`, which `VaultDefaultCoinService` reads to pick derived-vs-root pubkeys.
3. **`YieldRedemptionRecord.id` is provider-supplied and vault-agnostic** — harmless today, but the next windowed yield provider should namespace it by `pubKeyECDSA` the way `YieldPosition.makeID` does.

</details>

## Tests

`VultisigApp/VultisigAppTests/Utils/ZipImportDuplicateTests.swift`, one case per collision shape:

- duplicate detection: name-only, ECDSA-only, EdDSA-only, MLDSA-only, both-pubkeys, empty-pubkeys, empty-MLDSA, nil-MLDSA
- `importDecision`: keeps a free name, renames on a name-only collision, walks past already-taken suffixes, refuses an empty-pubkey collision, admits an empty-pubkey vault when nothing collides
- `availableVaultName` bound over a 50-long collision chain
- three end-to-end cases against a real in-memory `ModelContainer` proving the stored vault and its key shares **survive**: a name-only collision renames, a partial-pubkey collision is skipped, and two same-named vaults in one zip both land

## Localization

Two new keys — `vaultImportWouldOverwriteExisting`, `zipImportUnsafeVaults` — added to all **8** shipping locales (`en de es hr it ko pt zh-Hans`), sorted with `VultisigApp/scripts/sort_localizable.py`.

## Verification

- SwiftLint: no new warnings (22 pre-existing, none in the touched files).
- Build: `xcodebuild build -destination 'generic/platform=iOS Simulator'` — marker: `** BUILD SUCCEEDED **`
- Tests: full suite on a dedicated iPhone 17 Pro Max simulator — marker: `** TEST SUCCEEDED **`, **4893 tests, 1 skipped, 0 failures**

Closes #5066

## Summary by CodeRabbit

* **New Features**
  * Improved vault backup imports by preventing unsafe overwrites and key collisions.
  * Automatically renames safe imports when vault names conflict.
  * Reports imported, duplicate, skipped, and unsafe vaults during multi-vault imports.
  * Applies the same safety checks to individual vault restores.
* **Localization**
  * Added translated import-safety messages in English, German, Spanish, Croatian, Italian, Korean, Portuguese, and Simplified Chinese.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -1542,6 +1542,7 @@
 "vaultCreated" = "Tresor erstellt";
 "vaultDetailsDescription" = "Details zu deinem Tresor ansehen";
 "vaultDetailsTitle" = "Tresor-Details";
+"vaultImportWouldOverwriteExisting" = "Dieses Backup kann nicht importiert werden: Es würde einen bereits auf diesem Gerät vorhandenen Tresor überschreiben.";
 "vaultInfo" = "Vault -Info";
 "vaultLibType" = "Tresortyp";
 "vaultManagement" = "Tresorverwaltung";
@@ -1686,3 +1687,4 @@
 "yourVaultSetup" = "Ihre Tresor-Einrichtung";
 "zipImportAllDuplicates" = "Alle %d Tresor(e) existieren bereits und wurden übersprungen: %@.";
 "zipImportPartialSuccess" = "%d Tresor(e) importiert. %d Duplikat(e) übersprungen: %@.";
+"zipImportUnsafeVaults" = "%1$d Tresor(e) importiert. %2$d übersprungen. %3$@ konnte nicht importiert werden: Es würde einen bereits auf diesem Gerät vorhandenen Tresor überschreiben.";
--- a/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
@@ -1542,6 +1542,7 @@
 "vaultCreated" = "Vault created";
 "vaultDetailsDescription" = "View vault name, part and type";
 "vaultDetailsTitle" = "Details";
+"vaultImportWouldOverwriteExisting" = "This backup can't be imported: it would overwrite a vault already on this device.";
 "vaultInfo" = "Vault Info";
 "vaultLibType" = "Vault Type";
 "vaultManagement" = "Vault";
@@ -1686,3 +1687,4 @@
 "yourVaultSetup" = "Your vault setup";
 "zipImportAllDuplicates" = "All %d vault(s) already exist and were skipped: %@.";
 "zipImportPartialSuccess" = "Imported %d vault(s). Skipped %d duplicate(s): %@.";
+"zipImportUnsafeVaults" = "Imported %1$d vault(s). Skipped %2$d. %3$@ could not be imported: it would overwrite a vault already on this device.";
--- a/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
@@ -1542,6 +1542,7 @@
 "vaultCreated" = "Bóveda creada";
 "vaultDetailsDescription" = "Ver los detalles de tu caja fuerte";
 "vaultDetailsTitle" = "Detalles";
+"vaultImportWouldOverwriteExisting" = "Esta copia de seguridad no se puede importar: sobrescribiría una bóveda que ya está en este dispositivo.";
 "vaultInfo" = "Información de bóveda";
 "vaultLibType" = "tipo de bóveda";
 "vaultManagement" = "Gestión de bóveda";
@@ -1686,3 +1687,4 @@
 "yourVaultSetup" = "Tu configuración de bóveda";
 "zipImportAllDuplicates" = "Las %d bóveda(s) ya existen y fueron omitidas: %@.";
 "zipImportPartialSuccess" = "Se importaron %d bóveda(s). Se omitieron %d duplicado(s): %@.";
+"zipImportUnsafeVaults" = "Se importaron %1$d bóveda(s). Se omitieron %2$d. %3$@ no se pudo importar: sobrescribiría una bóveda que ya está en este dispositivo.";
--- a/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
@@ -1542,6 +1542,7 @@
 "vaultCreated" = "Trezor stvoren";
 "vaultDetailsDescription" = "Pogledajte detalje vašeg sefa";
 "vaultDetailsTitle" = "Detalji";
+"vaultImportWouldOverwriteExisting" = "Ova sigurnosna kopija ne može se uvesti: prebrisala bi trezor koji već postoji na ovom uređaju.";
 "vaultInfo" = "Informacije o trezorima";
 "vaultLibType" = "Tip trezora";
 "vaultManagement" = "Upravljanje trezorima";
@@ -1686,3 +1687,4 @@
 "yourVaultSetup" = "Vaša postavka trezora";
 "zipImportAllDuplicates" = "Svih %d trezor(a) već postoji i preskočeno je: %@.";
 "zipImportPartialSuccess" = "Uvezeno %d trezor(a). Preskočeno %d duplikat(a): %@.";
+"zipImportUnsafeVaults" = "Uvezeno %1$d trezor(a). Preskočeno %2$d. %3$@ nije moguće uvesti: prebrisalo bi trezor koji već postoji na ovom uređaju.";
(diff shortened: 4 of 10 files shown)


johnnyluo (approved):
LGTM

PR #5069, opened 2026-08-06
fix(swap): inset the advanced-swap settings card from the sheet's bottom edge

### LINKED ISSUE #5016: Advanced swap sheet: settings card sits flush against the bottom edge — needs bottom padding
In the **Advanced swap** sheet, the rounded card that holds the rows (Slippage Tolerance / Gas Limit / Select route / Use External Recipient) ends flush against the bottom of the sheet. Its bottom border and the sheet edge nearly touch, so the card reads as clipped rather than as a card. It needs the same breathing room below it that it already has on the sides.

Image

**Cause:** `VultisigApp/VultisigApp/Features/Swap/Views/AdvancedSwapSheet.swift:154` — the card gets `.padding(.horizontal, 16)` and nothing at the bottom, and the sheet is a fixed-height detent (`mainDetentHeight`, `AdvancedSwapSheet.swift:90`) sized to just header + rows, so there is no slack left over.

**Ask:**

- Add `.padding(.bottom, …)` to the card in `mainView` — 16 to mirror the horizontal inset, or 24 to mirror the row padding.
- Bump the `mainDetentHeight` base by the same amount. The detent is a fixed height, so padding on its own would squeeze/clip the card instead of opening a gap.
- Measure the gap below the bottom safe-area inset, so it doesn't visually collapse on devices with a home indicator.

**Done when:** the gap under the card is visible and equal to the side inset in every row combination — EVM (Gas Limit row present), non-EVM, provider selection on/off, and secured mint (no External Recipient row).

gastonm5:
## Problem

The rounded settings card in the Advanced Swap sheet was inset 16pt on the sides and nothing at the bottom, so its bottom border ran straight into the sheet edge and the card read as clipped rather than as a card.

## Root cause

Two separate things, one per presentation style:

- **The card had no bottom inset.** `mainView` applied `.padding(.horizontal, 16)` and nothing below. macOS presents this sheet **sized to its content** (`CrossPlatformSheet`, no detents) — which is the presentation in the issue screenshot — so a missing inset shows up directly as a flush edge.
- **The iOS detent is fixed and had to make room for it.** `mainDetentHeight` sizes the sheet's content area and the content is top-aligned, so an inset the height doesn't budget for is clipped by the sheet edge rather than opening a gap. The base is bumped by the same 16pt.

While in that arithmetic: it counted the Gas Limit and Select route rows but not the **External Recipient** row, which a secured mint hides — so the secured-mint sheet was one full row (70pt) taller than the rows it rendered. The height is now derived from the rows the card actually draws, which is also what makes the inset below the card identical in every combination.

## Fix

- `.padding(.bottom, MainLayout.cardInset)` on the card, with the horizontal inset reading the same constant so the two can't drift. **16pt, not 24** — the acceptance criterion is that the gap equals the *side* inset, and the side inset is 16.
- `mainDetentHeight` / `mainRowCount` are now `static` functions of `(isGasLimitSupported, canSelectProvider, isSecuredMint)`: `chrome + rows × rowHeight + cardInset`. Same shape as before, one row per row rendered, secured mint included.
- New `AdvancedSwapSheetLayoutTests` pins the height for every reachable row combination and measures the real header/row views to prove the detent still has room for the card *and* its inset.

Heights: 1 row 206 · 2 rows 276 · 3 rows 346 · 4 rows 416 (was 260 / 260 / 330 / 400, with secured mint mis-sized).

### Why the row and chrome allowances were not tightened

Measured on an iOS 26.5 simulator: header 76pt, one-line row 68pt, separator 1pt. That makes the fixed detent look 30–45pt too generous — but rows are not fixed-height. Measured at a 343pt card width (iPhone SE):

| row | value | height |
|---|---|---|
| Use External Recipient | `Off` | 68 |
| Use External Recipient | `0x1234…abcd` | 81.67 (2 lines) |
| Externen Empfänger verwenden | `0x1234…abcd` | 98.67 (3 lines) |

The spare allowance is what keeps a wrapped row from clipping the card, so it stays. The test asserts the fit with one row wrapped, not with every row on one line.

## Verification

**Rendered.** The sheet needs a live vault + quote to reach in the app, so it was driven directly instead — both presentations, on an iOS 26.5 simulator:

- **Content-sized (macOS-equivalent, the platform in the issue screenshot):** before/after captured via `ImageRenderer` at the sheet's ideal height. Before reproduces the screenshot exactly — the card's bottom corners cut off by the sheet's rounded edge. After shows a 16pt gap matching the side inset.
- **iOS detent:** the real sheet was presented in a `UIWindow` on the simulator and its geometry read back — sheet view height `310` = detent `276` + `safeAreaInsets.bottom` `34`, confirming the detent value sizes the *content* area and the system adds the home-indicator strip below it. So the gap is measured above the safe-area inset and can't collapse into the home indicator. Window screenshot captured.

**Reasoned, not rendered.** The 3- and 4-row combinations and the secured-mint case were verified against the measured header/row/separator heights rather than each being screenshotted; the card is the same `VStack` with one more row in each, and every combination is asserted in `AdvancedSwapSheetLayoutTests` against those measurements. Secured mint additionally requires a THORChain secured-asset pair, which was constructed in-process to confirm `isSecuredMint` gates the row (`BTC.BTC` → `btc-btc`).

Not verified on real macOS — no macOS run in this environment. The macOS path is `CrossPlatformSheet`'s content-sized sheet, which the content-sized render models.

**Gate** (own simulator, iPhone 17 Pro Max / iOS 26.5):
- `make build-check` → `** BUILD SUCCEEDED **`
- `make test` (full suite) → `** TEST SUCCEEDED **`, 0 failed cases
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` → 22 violations, unchanged from `main` (none in the touched files)

Reviewed read-only by `codex exec` before the gate: no production defect; three low-severity test findings, two fixed in the follow-up commit (side/bottom insets tied to one constant, the wrapping test now asserts its own premise), one rejected — `mainRowCount` mirroring `mainView`'s conditions can only be made drift-proof by making the row list data-driven, which is beyond a layout fix; the coupling is called out in the doc comment.

Closes #5016

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Advanced Swap sheet sizing across different row combinations.
  * Ensured layouts correctly accommodate secured mint external-recipient details and wrapped content.
  * Applied consistent bottom spacing within the main card.

* **Tests**
  * Added coverage for rendered rows, sheet heights, content fitting, and dynamic text wrapping.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Swap/Views/AdvancedSwapSheet.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/Views/AdvancedSwapSheet.swift
@@ -85,13 +85,62 @@ struct AdvancedSwapSheet: View {
         }
     }
 
-    /// Main-sheet height grows with the optional rows it renders: the Gas Limit
-    /// row (EVM only) and the Select route row (provider selection available).
     private var mainDetentHeight: CGFloat {
-        var height: CGFloat = 260
-        if isGasLimitSupported { height += 70 }
-        if vm.canSelectProvider { height += 70 }
-        return height
+        Self.mainDetentHeight(
+            isGasLimitSupported: isGasLimitSupported,
+            canSelectProvider: vm.canSelectProvider,
+            isSecuredMint: vm.isSecuredMint
+        )
+    }
+
+    /// Fixed heights the main state is built from. The detent sizes the content
+    /// area — the system adds the bottom safe area on top of it — so everything
+    /// the card needs has to be accounted for here, the inset below it included:
+    /// content is top-aligned, so anything that doesn't fit is clipped by the
+    /// sheet edge rather than compressed.
+    enum MainLayout {
+        /// The header, plus headroom. The header itself measures 76pt; the rest
+        /// absorbs a row wrapping onto a second line, which a long title does in
+        /// several locales (and in English once the External Recipient row shows
+        /// an address). Unused headroom only ever becomes extra space below the
+        /// card, so it's cheaper than clipping the card when a row grows.
+        static let chrome: CGFloat = 120
+        /// One `AdvancedSwapMainRow` — 68pt on one line — plus its 1pt separator.
+        static let rowHeight: CGFloat = 70
+        /// Gap below the card, mirroring its horizontal inset so the card reads
+        /// as inset on all four sides instead of clipped by the sheet edge.
+        static let cardInset: CGFloat = 16
+    }
+
+    /// Rows the card renders. Mirrors the conditions in `mainView` and has to
+    /// stay in step with them: Slippage is always shown, Gas Limit is EVM-only,
+    /// Select route needs more than one quote to pick from, and a secured mint
+    /// drops External Recipient.
+    static func mainRowCount(
+        isGasLimitSupported: Bool,
+        canSelectProvider: Bool,
+        isSecuredMint: Bool
+    ) -> Int {
+        var rows = 1
+        if isGasLimitSupported { rows += 1 }
+        if canSelectProvider { rows += 1 }
+        if !isSecuredMint { rows += 1 }
+        return rows
+    }
+
+    /// Main-sheet height: fixed chrome, one row per row the card actually
+    /// renders, and the inset that keeps the card off the sheet's bottom edge.
+    static func mainDetentHeight(
+        isGasLimitSupported: Bool,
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5071, opened 2026-08-06
fix(keysign): resolve KeyImport custom-message keys by derivation path

### LINKED ISSUE #5067: [Fix] KeyImport custom-message signing falls back to the root keyshare for EVM vaults without an Ethereum entry [Keysign]
# [Fix] KeyImport custom-message signing falls back to the root keyshare for EVM vaults without an Ethereum entry [Keysign]

## Problem

`KeysignViewModel.startKeysignDKLS(isImport:)` resolves a custom message's chain, defaulting to `.ethereum`, then looks the key up with an **exact** chain match:

```swift
// VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:449-458
var targetChain: Chain = .ethereum // Default to Ethereum
if let chain = Chain.allCases.first(where: { $0.name.caseInsensitiveCompare(customMessagePayload.chain) == .orderedSame }) {
    chainPath = chain.coinType.derivationPath()
    targetChain = chain
}
publicKey = self.vault.chainPublicKeys.first(where: { $0.chain == targetChain })?.publicKeyHex
```

If there is no exact `chainPublicKeys` entry for `targetChain`, `publicKey` is `nil` and line 481 falls back to `vault.pubKeyECDSA` — the vault's root share at BIP32 path `m`.

That fallback is reachable and wrong. Key import does not force Ethereum into the chain set, so a user can create a KeyImport vault with, say, only BSC, Base, or Arbitrum. Every EVM chain shares the derivation path `m/44'/60'/0'/0/0`, so the BSC entry **is** the correct key for an Ethereum-named custom message — but the exact match misses it and iOS signs with the path-`m` root share instead.

Since KeyImport signing uses `chainPath = ""` (no in-ceremony derivation), a signature under the root share verifies against no address the vault displays and fails `ecrecover` against the user's EVM address.

It also breaks co-signing with Android, which already handles this correctly: `Vault.getEcdsaSigningKey` tries an exact chain match first, then falls back to a **derivation-path** match, and only then to the root key (`data/src/main/kotlin/com/vultisig/wallet/data/models/Vault.kt:91-115`). When iOS picks root and Android picks the EVM-sibling share, the two devices load shares of different keys and the DKLS ceremony aborts on `SignError::InvalidFinalSessionID` (`dkls23-rs/src/sign/dsg.rs:147-152, 265-267`).

Note iOS itself hardcodes `chain: Chain.ethereum.name` on every custom message it initiates (`SettingsCustomMessageView.swift:221`), so an Ethereum-named payload reaching a non-Ethereum EVM vault is the normal case, not an exotic one.

## Expected Behavior

For a KeyImport vault, a custom message naming an EVM chain selects any imported keyshare that shares the same derivation path, and only falls back to the root key when the vault genuinely has no key on that path.

## Steps to Reproduce

1. Import a seed phrase into a new vault, selecting an EVM chain but **not** Ethereum (e.g. BSC only).
2. Vault Settings → Sign Custom Message. Enter method `personal_sign` and any message. (`customMessagePayload.chain` will be `"Ethereum"`.)
3. Observe that `chainPublicKeys` has no `.ethereum` entry, so `publicKey` is `nil` and the ceremony runs with `vault.pubKeyECDSA`.
4. The produced signature does not recover to the vault's EVM address. With an Android co-signer the ceremony aborts instead of completing.

## Solution

Mirror Android's three-step resolution in the KeyImport branch: exact chain match, then a derivation-path match across `chainPublicKeys`, then the existing root fallback. Because all EVM chains resolve to `m/44'/60'/0'/0/0`, a path match returns cryptographically the same key the exact match would have — it just also succeeds when the vault imported a sibling EVM chain instead of Ethereum. Keep the root fallback at line 481 as the last resort so behaviour for genuinely unimported chains is unchanged.

Apply the same fallback to the EdDSA path so `SchnorrKeysign` picks consistently.

### Must Do
- [ ] Add a derivation-path fallback to the `chainPublicKeys` lookup in the custom-message branch of `startKeysignDKLS(isImport:)` (`KeysignViewModel.swift:449-458`), after the exact match and before the root fallback.
- [ ] Apply the same fallback to the EdDSA/Schnorr key selection on the import path.
- [ ] Add a test: a KeyImport vault holding only a BSC `chainPublicKeys` entry resolves the BSC key — not `pubKeyECDSA` — for a custom message with `chain = "Ethereum"`.

### Must NOT Do
- Don't change unrelated code
- Don't remove the `publicKey ?? vault.pubKeyECDSA` fallback at line 481
- Don't change the transaction path (line 447) — for transactions the chain always comes from the vault's own enabled chains, so the exact match always hits

### Out of Scope
- Forcing Ethereum into the KeyImport chain selection UI — that is a product decision, not this fix.

## Acceptance Criteria
- [ ] Build succeeds
- [ ] A KeyImport vault with a non-Ethereum EVM chain signs an Ethereum-named custom message with the imported EVM keyshare
- [ ] A KeyImport vault with no key on that derivation path still falls back to the root key (unchanged behaviour)
- [ ] The resulting `personal_sign` signature recovers to the vault's EVM address
- [ ] Cross-platform: the same message co-signs successfully with an Android device

## References
- Android reference implementation: `data/src/main/kotlin/com/vultisig/wallet/data/models/Vault.kt:91-115` (`getEcdsaSigningKey` / `getEddsaSigningKey`)
- Related Android fix: vultisig/vultisig-android#5525 — resolves the custom-message chain for KeyImport share selection; merged, from external contributor @arpicato

gastonm5:
## Problem

A KeyImport vault holds one pre-derived TSS share **per imported chain** and signs with `chainPath = ""` — no in-ceremony BIP32 derivation — so the share handed to the ceremony *is* the signing key. `KeysignViewModel.startKeysignDKLS(isImport:)` picked that share for a custom message by **exact chain match** against `vault.chainPublicKeys`.

The app stamps every custom message it initiates with `chain: "Ethereum"` (`SettingsCustomMessageView`), but key import never forces Ethereum into the chain set. So a vault imported with BSC, Base or Arbitrum alone matched nothing, `publicKey` stayed `nil`, and `publicKeyECDSA: publicKey ?? vault.pubKeyECDSA` handed the ceremony the **root share at BIP32 path `m`**.

The resulting signature recovers to no address the vault displays and fails `ecrecover`. Cross-device it is worse: Android already resolves this correctly, so the two devices load shares of *different* keys and the DKLS ceremony aborts on `SignError::InvalidFinalSessionID`.

## Root cause

`chainPublicKeys.first(where: { $0.chain == targetChain })` is an identity match, but every EVM chain derives from the same path `m/44'/60'/0'/0/0`. The BSC entry **is** the correct key for an Ethereum-named message — the exact match just couldn't see it.

## Fix

Mirror Android's resolution (`Vault.getEcdsaSigningKey` / `getEddsaSigningKey`, ported in vultisig/vultisig-android#5525) in the custom-message branch of the import path only: **exact chain match → same-curve derivation-path match → the existing root fallback**, which stays as the last resort. The single resolved key feeds both the DKLS and Schnorr branches, so ECDSA and EdDSA pick consistently.

Deliberately unchanged: the transaction branch (the chain always comes from the vault's own enabled chains, so the exact match always hits), the `?? vault.pubKeyECDSA` / `?? vault.pubKeyEdDSA` last-resort fallbacks, the non-import path, and `VultisigApp/Blockchain/Tss/`.

### Worth a reviewer's eye #1: the fallback's curve check is doubled, on purpose

The two platforms persist `ChainPublicKey.isEddsa` **differently** for QBTC, the one MLDSA chain:

| | predicate | stored `isEddsa` for QBTC |
|---|---|---|
| iOS | `!chain.isECDSA` (`signingKeyType` returns `.MLDSA`) | `true` |
| Android | `chain.TssKeysignType == TssKeyType.EDDSA` (QBTC is `MLDSA`) | `false` |

`Vault+ProtoMappable` maps the flag straight through from the shared proto backup, and QBTC shares Cosmos' derivation path (both `coinType: .cosmos`). So the fallback matches the curve twice — the persisted flag *and* the candidate chain's own `isECDSA`, the latter computed from iOS's own `Chain` table on both sides and therefore immune to the divergence. Without it, a vault restored from an Android backup could hand an MLDSA share to a Cosmos-named message.

### Worth a reviewer's eye #2: the *exact* match is deliberately left unchecked

Android's resolvers curve-filter the exact match too. This PR does not, and the reasoning is worth stating because it looks like a parity gap:

1. **It's unreachable.** QBTC is the only chain whose `isEddsa` the platforms disagree on, and QBTC never reaches this helper — the `.MLDSA` branch signs with `vault.publicKeyMLDSA44`. For every chain that *does* reach it, both write paths produce identical metadata.
2. **The unfiltered failure mode is the safer one.** `chain` is the entry's primary identity (it forms the `@Attribute(.unique)` id and is written from the same import job as the key), so a share filed under a chain *is* that chain's key. If the flag were ever inconsistent, filtering would fall through to the root share at path `m` — a signature that verifies against nothing, i.e. exactly the bug this PR removes.

`testExactMatchWithInconsistentCurveMetadataStillWins` pins this so it isn't "fixed" later by someone chasing parity.

## Tests

`KeyImportCustomMessageKeyResolutionTests` — 13 cases pinning the reported bug (a BSC-only import vault resolves the BSC share, not `pubKeyECDSA`), every EVM sibling, exact-match precedence, the preserved root fallback, curve isolation in both directions, non-import behaviour, and both cross-platform cases above. Codex confirmed the BSC-only test fails under exact-only resolution, i.e. it would catch a revert.

`testEvmSiblingsShareEthereumsDerivationPath` asserts at runtime over `Chain.allCases where chainType == .EVM` that all 16 EVM chains really do derive from `m/44'/60'/0'/0/0` — the premise the whole fallback rests on is now a guard, not a comment, so a future chain whose coin type carries a different path fails loudly here.

## Verification

- `make build-check` → `** BUILD SUCCEEDED **`
- `make test` (full suite, dedicated simulator) → `** TEST SUCCEEDED **` — 4881 tests, 1 skipped, 0 failures
- SwiftLint: 0 violations on both changed files
- Cross-model review: a read-only Codex pass per commit plus a whole-branch pass. No blocker-level findings. The one should-fix that survived was resolved test-only, after the whole-branch pass correctly refuted my first rejection rationale for it.

## ⚠️ Needs human + on-device review

This is signing-path code in a HIGH-security-tier wallet, so please review accordingly.

**Two acceptance criteria from the issue were NOT verified and cannot be by automated tests** — both need a real seed import and a second device:

- [ ] the resulting `personal_sign` signature recovers to the vault's EVM address
- [ ] the same message co-signs successfully with an Android device

I am explicitly **not** claiming either. The unit tests prove the *selection* logic picks the imported EVM share instead of the root share; they do not prove the end-to-end ceremony or the cross-device handshake.

Closes #5067

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift
@@ -433,6 +433,39 @@ class KeysignViewModel: ObservableObject {
         self.txid = newTxid
     }
 
+    /// Resolves the public key a custom message must be signed with.
+    ///
+    /// A key-import vault holds one pre-derived share per imported chain and signs with an
+    /// empty chain path, so the share handed to the ceremony *is* the signing key — nothing
+    /// derives it further. An exact chain match is preferred; failing that, any share on the
+    /// same derivation path is the very same key (every EVM chain derives from
+    /// `m/44'/60'/0'/0/0`), which is what lets a vault imported with BSC alone sign the
+    /// Ethereum-named messages the app produces.
+    ///
+    /// A path sibling must also be on the same curve as the target, checked two ways: the
+    /// persisted flag, and the curve the candidate's own chain implies. The persisted flag
+    /// alone is not enough — iOS and Android write it differently for the one MLDSA chain
+    /// (QBTC), and QBTC shares Cosmos' derivation path, so a vault restored from an Android
+    /// backup could otherwise resolve a Cosmos message to the QBTC share.
+    ///
+    /// Returns `nil` when nothing matches, leaving the caller's root-key fallback intact.
+    /// Non-import vaults get the exact match only: their root key is the signing key.
+    static func customMessagePublicKey(for chain: Chain,
+                                       in chainPublicKeys: [ChainPublicKey],
+                                       isImport: Bool) -> String? {
+        if let exactMatch = chainPublicKeys.first(where: { $0.chain == chain }) {
+            return exactMatch.publicKeyHex
+        }
+        guard isImport else { return nil }
+        let derivationPath = chain.coinType.derivationPath()
+        let isEddsa = !chain.isECDSA
+        return chainPublicKeys.first {
+            $0.isEddsa == isEddsa
+                && $0.chain.isECDSA == chain.isECDSA
+                && $0.chain.coinType.derivationPath() == derivationPath
+        }?.publicKeyHex
+    }
+
     func startKeysignDKLS(isImport: Bool) async {
         do {
             // Check if we have either keysignPayload or customMessagePayload
@@ -455,7 +488,9 @@ class KeysignViewModel: ObservableObject {
                     // Fallback to Ethereum if chain name cannot be parsed
                     chainPath = TokensStore.Token.ethereum.coinType.derivationPath()
                 }
-                publicKey = self.vault.chainPublicKeys.first(where: { $0.chain == targetChain })?.publicKeyHex
+                publicKey = Self.customMessagePublicKey(for: targetChain,
+                                                        in: self.vault.chainPublicKeys,
+                                                        isImport: isImport)
             } else {
                 throw HelperError.runtimeError("keysign payload is nil")
             }
--- a/VultisigApp/VultisigAppTests/KeyImport/KeyImportCustomMessageKeyResolutionTests.swift
+++ b/VultisigApp/VultisigAppTests/KeyImport/KeyImportCustomMessageKeyResolutionTests.swift
@@ -106,6 +106,31 @@ final class KeyImportCustomMessageKeyResolutionTests: XCTestCase {
         XCTAssertNil(resolved)
     }
 
+    /// The **exact** match deliberately does not filter on the persisted curve flag, unlike
+    /// Android's resolvers. `chain` is the entry's primary identity — it forms the
+    /// `@Attribute(.unique)` id and is written from the same import job as the key itself —
+    /// so a share filed under a chain *is* that chain's key, whatever the flag says. Were
+    /// the flag ever inconsistent, filtering would drop through to the root share at path
+    /// `m`: a signature that verifies against nothing, which is the precise failure this
+    /// fix exists to remove. Trusting `chain` over the flag keeps the safer failure mode.
+    ///
+    /// Neither platform's write path can actually produce this state for a chain that
+    /// reaches this helper: QBTC is the only chain whose `isEddsa` the two platforms
+    /// disagree on, and the `.MLDSA` branch signs with `vault.publicKeyMLDSA44` instead of
+    /// this helper's result. The test pins the intent so the check is not "fixed" later.
+    func testExactMatchWithInconsistentCurveMetadataStillWins() {
+        let vault = makeKeyImportVault(keys: [(.ethereum, ethKey)], isEddsaOverride: true)
+
+        let resolved = KeysignViewModel.customMessagePublicKey(
+            for: .ethereum,
+            in: vault.chainPublicKeys,
+            isImport: true
+        )
+
+        XCTAssertEqual(resolved, ethKey, "The chain's own share must win over an inconsistent curve flag, never the root key")
+        XCTAssertNotEqual(resolved, vault.pubKeyECDSA)
+    }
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5079, opened 2026-08-07
test(import): state the protection state the import tests are written for (#5078)

### LINKED ISSUE #5078: CI: main's iOS job is red — an import test inherits the host Keychain's fail-closed state
`main`'s iOS job has been failing since #5070 merged, on a single test:

```
ZipImportDuplicateTests.testZipImportRenamesInsteadOfOverwritingAStoredVault
  XCTAssertEqual failed: ("1") is not equal to ("2") - A name-only collision must not replace the stored vault
  XCTAssertEqual failed: ("["Savings"]") is not equal to ("["Savings (2)", "Savings"]")
  XCTAssertTrue failed
```

Every PR that re-runs CI inherits it. macOS passes.

## Not a production bug

Worth stating up front, because the failure reads far more alarming than it is: **no shipped import path is broken.** The refusal it triggers is correct, intended behaviour reached in a configuration the test never meant to be in.

## What happens

1. #5070 routed every import through `ProtectedVaultImporter.commit`, which normalizes the **incoming** vault's key shares via `KeyshareProtector.shared`.
2. That protector derives its state from a real Keychain read (`KeyshareKeySession.currentState`): `.present` → `.locked`, `.absent` → `.disabled`, and `.unavailable` → `.locked`, **failing closed on purpose** — the comment there is explicit that `.disabled` is "a licence to write plaintext", and taking it wrongly would store an unprotected share behind a passcode that is still set.
3. **The unit-test bundle has no Keychain entitlement** (documented at `VultisigAppTests/Security/KeychainWriteTests.swift:11`), so the read is `.unavailable`, the state is `.locked`, `seal` throws, and the import is refused.
4. `restoreMultipleVaults` catches, reports `vaultRestoreFailed`, and stores nothing — so the test sees 1 vault instead of 2.

Verified by probe rather than inference: calling `commit` directly with a share-bearing vault throws `locked`.

On a real device with no passcode set the Keychain is readable, the state is `.absent` → `.disabled`, and the protector is a passthrough.

## Why only this one test

Normalization runs over the **incoming** shares. `testZipImportRenamesInsteadOfOverwritingAStoredVault` is the only zip test whose incoming vault carries key shares; its three siblings use share-less fixtures and pass straight through.

That property is the part worth remembering: **a suite full of share-less fixtures looks healthy while every real vault — all of which carry shares — would be refused.**

Confirmed by bisect: the test passes at `c861dae3d` (#5068, where it landed) and fails from `322ddd625` (#5070) onward.

## Fix

Test-side, following the helper already in `BackupImportProtectionTests`: inject `KeyshareProtector(state: { .disabled })` and a fresh `KeyshareWriteCoordinator()` in `setUp`, so the tests state "no passcode set" instead of inheriting it from the host Keychain.

Plus a regression test for `.locked`, which had **no coverage at all** — `.unlocked` and `.disabled` were both well covered, which is why this arrived as a mystery rather than a named assertion.

## Follow-up, not fixed here

The same latent coupling exists elsewhere and passes only because those fixtures are share-less:

- `ZipImportDuplicateTests`' BAK path → `restoreVaultBack` → `Vault(proto:)` constructs with `KeyshareProtector.shared` **before** the injected importer is reached (`EncryptedBackupViewModel.swift:744`).
- `KeygenVaultCommitTests` at three sites.

Adding key shares to any of those fixtures would produce the same mystery failure.

gastonm5:
Closes #5078.

Unblocks `main`, whose iOS job has been red since #5070 merged. Every PR that re-runs CI inherits the failure.

**No production code changes, and no production bug.** The refusal the test hit is correct behaviour reached in a configuration the test never meant to be in.

## What was happening

#5070 routed imports through `ProtectedVaultImporter.commit`, which normalizes the **incoming** vault's key shares via `KeyshareProtector.shared`. That protector reads the real Keychain: `.absent` → `.disabled`, `.present` → `.locked`, and `.unavailable` → `.locked`, failing closed on purpose. The test bundle has no Keychain entitlement (`KeychainWriteTests.swift:11`), so the state is `.locked`, `seal` throws, and any import carrying key shares is refused — the view model catches it, reports `vaultRestoreFailed`, and stores nothing.

Verified by probe, not inferred: calling `commit` directly with a share-bearing vault throws `locked`. Confirmed by bisect: green at `c861dae3d` (#5068, where the test landed), red from `322ddd625` (#5070).

**Only one test fell** because normalization runs over the *incoming* shares, and it is the only zip test whose incoming vault has any. Its three siblings use share-less fixtures. That is the part worth remembering: a suite full of share-less fixtures looks healthy while every real vault, all of which carry shares, would be refused.

## The change

- `ZipImportDuplicateTests.setUp` now states the protection state instead of inheriting it — `KeyshareProtector(state: { .disabled })`, which is what "no passcode set" means, plus a fresh `KeyshareWriteCoordinator()` rather than the shared one carrying another test's lease residue. Same shape as the existing `makeViewModel(state:)` helper in `BackupImportProtectionTests`.
- A regression test for `.locked`, which had **no coverage**. `.unlocked` and `.disabled` were both well covered, which is exactly why this surfaced as a mystery instead of a named assertion.

## Testing

- iOS full suite: `** TEST SUCCEEDED **`, 5307 tests, **0 failures** — first clean run since #5070.
- macOS: `** BUILD SUCCEEDED **`.
- SwiftLint: no new warnings.
- `codex exec --sandbox read-only` reviewed the diagnosis as well as the diff, and was asked specifically whether this papers over a production bug. Four categories clean; it independently located the entitlement evidence.

## Known, not fixed here

The same latent coupling survives where fixtures happen to be share-less, and would produce the same mystery if shares were ever added:

- `ZipImportDuplicateTests`' BAK path → `restoreVaultBack` → `Vault(proto:)` uses `KeyshareProtector.shared` **before** the injected importer is reached (`EncryptedBackupViewModel.swift:744`).
- `KeygenVaultCommitTests`, three sites.

Left out deliberately: this PR's job is to get `main` green, and that cleanup is a wider test refactor. Recorded in #5078.

## Summary by CodeRabbit

* **Bug Fixes**
  * Backup imports are now verified to fail safely when the required key is locked or unavailable, without creating incomplete vault or coin records.

* **Tests**
  * Expanded automated coverage for protected backup imports and duplicate archive handling to help prevent regressions in import security and reliability.

### DIFF
--- a/VultisigApp/VultisigAppTests/Security/BackupImportProtectionTests.swift
+++ b/VultisigApp/VultisigAppTests/Security/BackupImportProtectionTests.swift
@@ -105,6 +105,37 @@ final class BackupImportProtectionTests: XCTestCase {
         XCTAssertEqual(try context.fetchCount(FetchDescriptor<Coin>()), 0, "default coins must not outlive a refused import")
     }
 
+    /// The third protection state, which nothing pinned until an import test
+    /// started failing for a reason no assertion named.
+    ///
+    /// `.locked` is not only "a passcode is set and not yet entered". A Keychain
+    /// that cannot be read resolves here too, deliberately — `.disabled` is a
+    /// licence to write plaintext, and taking it wrongly stores an unprotected
+    /// share behind a passcode that is still set. So the refusal is the intended
+    /// behaviour and this states it: a backup carrying key shares does not
+    /// import, and nothing of it is left behind.
+    ///
+    /// It also records the shape of the surprise. Normalization runs over the
+    /// *incoming* shares, so a vault with none imports perfectly well in this
+    /// state — which is why a suite full of share-less fixtures can look healthy
+    /// while every real vault, all of which carry shares, is refused.
+    func testAnImportIsRefusedWhileTheKeyIsOutOfReach() throws {
+        let sut = makeViewModel(state: .locked)
+        sut.decryptedContent = try backupVaultJSONHex()
+
+        sut.restoreVault(modelContext: context, vaults: [])
+
+        XCTAssertFalse(sut.isVaultImported)
+        XCTAssertEqual(
+            try context.fetchCount(FetchDescriptor<Vault>()), 0,
+            "a share that cannot be sealed must not reach the store"
+        )
+        XCTAssertEqual(
+            try context.fetchCount(FetchDescriptor<Coin>()), 0,
+            "default coins must not outlive a refused import"
+        )
+    }
+
     // MARK: - What the restored vault opens on
 
     /// The reported regression, through the batch entry point and with no
--- a/VultisigApp/VultisigAppTests/Utils/ZipImportDuplicateTests.swift
+++ b/VultisigApp/VultisigAppTests/Utils/ZipImportDuplicateTests.swift
@@ -23,7 +23,21 @@ final class ZipImportDuplicateTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
-        viewModel = EncryptedBackupViewModel()
+        // "No passcode set" has to be stated, not inherited. The default
+        // protector reads the real Keychain, which the test host cannot read, and
+        // an unreadable Keychain deliberately fails closed to `.locked` — under
+        // which `seal` throws and *any* import carrying key shares is refused.
+        // Left to the default, these tests would be asserting collision handling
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5095, opened 2026-08-10
fix(keygen): make commitVault all-or-nothing when the save throws

### LINKED ISSUE #5073: commitVault is not atomic when context.save() throws
## Description

`KeygenViewModel.commitVault` is not atomic when `context.save()` throws: the in-memory mutations that preceded the save have already happened, so the store and the object graph can end up disagreeing about a vault that was just created.

Raised by CodeRabbit on [#5070](https://github.com/vultisig/vultisig-ios/pull/5070) (`VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift`). Filed separately rather than fixed there: it is data-integrity work in the keygen path with its own blast radius, it is not caused by the passcode feature, and folding it into an already-large security PR would make both harder to review.

## What to look at

- Which mutations happen before `save()`, and which of them are observable if the save fails.
- Whether a failed commit should roll back, retry, or surface — a vault that half-exists after keygen is worse than one that visibly failed, because the user believes they have a wallet.
- Interaction with the passcode work landing in #5070: `commitVault` runs under `KeyshareWriteCoordinator`'s lease and key shares may be sealed at that point. See `docs/passcode-keyshare-encryption/invariants.md` § *the write that did not take*, which documents an adjacent case.

## Why it matters

This is the moment a vault's key shares become durable. A partial commit here is the class of bug where the user's funds exist on-chain but the device cannot sign for them.

## Which feature is affected?
- [x] Create vault ( Secure / Fast)
- [ ] Sending
- [ ] Swap

gastonm5:
Fixes #5073

## What was actually wrong

`KeygenViewModel.commitVault` inserts a freshly generated vault and its default coins into the app-wide `ModelContext` and then saves. The issue's premise holds, but the mechanism is the reverse of the natural reading, and it is worse than "the store and the object graph disagree":

- **The store is never half-written.** A throwing `save()` writes nothing.
- **The disagreement is that the app already believes the vault exists.** Measured on an in-memory container: with the vault and two coins inserted and no save, `FetchDescriptor<Vault>` returns **1** and `FetchDescriptor<Coin>` returns **2**. SwiftData resolves pending inserts in every fetch, so the vault list, `VaultNameValidator` and every other read see the vault the moment the alert says it could not be saved.
- **A later save makes it durable.** The main context autosaves and every service in the app saves it, so the pending insert lands with no further user action — and *outside* the write lease, where a passcode transition can land between the normalization and the write, putting the shares on the wrong side of the sealed-iff-a-passcode-is-set invariant.
- **The abort guarantee is lost.** `deferVaultPersistence` exists so that abandoning the review leaves the vault's name reusable; a failed commit took the name anyway.

### The mutation inventory

| # | Statement | Observable if `save()` throws? |
|---|---|---|
| 1 | `normalizedKeyshares(of:protector:)` | no — pure, and it throws before anything else |
| 2 | `vault.keyshares = shares` | **yes**, on the caller's object — see the share-form trap |
| 3 | `setDefaultCoinsOnce(vault:)` — `context.insert(coin)`, `coin.vault = vault`, `vault.defiChains = …` | **yes** — pending inserts in the shared context |
| 4 | `context.insert(vault)` | **yes** — pending insert in the shared context |
| 5 | `try context.save()` | the store is untouched |
| 6 | `startTokenDiscovery()` | no — outside and after the lease block; already correct on `main` |

### The share-form trap

`vault.keyshares = shares` rewrites the caller's vault *before* a save that can fail, and `main` never puts it back. Plaintext keygen → the user sets a passcode during the review → the commit seals the shares onto the object → the save throws → the user disables the passcode and retries → the retry cannot open shares sealed under a key that no longer exists, and the vault is **permanently uncommittable**. So the undo has to reach the object, not only the context.

## Rollback, retry, or surface

**Surface**, and undo. The screen already renders the error and stays put, so "Looks Good" *is* the retry, at the right granularity for the failures a save actually produces here — the container unreachable, data protection unavailable, the disk full — none of which a second attempt one line later would fix. `ProtectedVaultImporter` sets the same precedent: it retries `prepare` and never the insert's save.

`context.rollback()` is **not** safe unconditionally, measured: it destroys a foreign pending insert outright and reverts a foreign pending edit. And `commitVault` does not always receive a new vault — a secure reshare on a device that was already a signer has its vault saved by `finalizeDKLSKeygen`'s `needsInsert == false` branch and *still* routes through the review screen. So:

| Shape | Undo |
|---|---|
| the store does not hold the vault yet | `rollback()`, and the commit is **refused up front** with `.busy` if the context was carrying other work |
| the store already holds it (secure reshare) | delete only the coins this call attached, plus the DeFi chains derived with them; never touch the row |

Which shape it is, is asked of the **store** through a `ModelContext` of its own, not of `vault.modelContext` — that answers "registered", which a pending insert also is.

### Two designs that were tried and measured away

- **`context.delete(vault)` as the universal undo.** The row does go back, and the `.cascade` on `Vault.coins` and `Vault.chainPublicKeys` then strips the caller's *object* — so the retry rebuilds a key-import vault that no longer knows which chains it can derive and stores it with none. Not deterministic run to run. On the reshare shape the same delete cascaded an existing wallet's coins away silently.
- **Flushing the context first**, so the rollback is always safe. Rejected: a reshare that has set `vault.signers` and is suspended before its completion check is exactly the half-finished flow `KeyshareSweeper` refuses to flush.

### Interaction with the sealing already done under the lease

None that can strand key material. The seal `commitVault` performs is purely in-memory — `KeyshareNormalizer` returns a new `[KeyShare]`, mints no data key and deletes none — so there is no durable half to leave behind. The undo has exactly two places to reach, the context and the object, and both run inside `withWriteLease`.

### `@Attribute(.unique)` upsert

Reachable at this call site and deliberately untouched. `Vault` carries four independent unique attributes and `keyshares` is a plain stored property, so an upsert replaces a stored vault's shares; the `pubKeyECDSA` case is documented and prompted for, a `name`-only or `pubKeyEdDSA`-only collision is not. But an upsert makes `save()` **succeed**, so it is not on this issue's path — and the one thing the fix must not do is remove a row it did not insert, which is why the shape is decided by a fetch against the store. Same call as the pre-existing import hazard in `invariants.md`; it deserves its own issue.

## Tests

Eight cases in `KeygenVaultCommitTests`. Five fail against `main`'s `commitVault`, verified by running them against it. Three pass against `main` and are kept deliberately because they guard the fix rather than the bug — each fails against a rejected variant, verified the same way.

## Gate

`** BUILD SUCCEEDED **` and `** TEST SUCCEEDED **` in the same run — 5315 tests, 1 skipped, 0 failures. SwiftLint: 22 violations, all pre-existing, none in the touched files.

## Needs a device pass

The `.busy` refusal is new user-visible behaviour on a path that is hard to reach deliberately. A secure keygen + a secure reshare on device would confirm it does not fire in practice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vault commit reliability when saving fails or conflicting changes are detected.
  * Prevented incomplete or rejected vaults from being retained after failed commits.
  * Restored original key shares when processing cannot be completed safely.
  * Preserved or removed associated coins and DeFi chains correctly during rollback.
  * Added clearer localized messaging for temporarily unavailable commit actions.
  * Enabled safer retries after transient commit failures.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/Keygen/ViewModels/KeygenViewModel.swift
@@ -33,6 +33,22 @@ enum KeygenCommitError: Error, Equatable, LocalizedError {
     /// pointing both at the first would make the second one false.
     case unsealableKeyshare(pubkey: String)
 
+    /// The store is carrying uncommitted work from somewhere else, and this
+    /// commit would have to insert a vault it could only take back by
+    /// discarding that work with it.
+    ///
+    /// Refused rather than resolved, and the three ways of resolving it were
+    /// each tried and rejected: flushing the other work commits a half-finished
+    /// flow because the user tapped a button on this one; rolling it back
+    /// discards it; and deleting only this vault's own row strips the caller's
+    /// object of its coins and chain public keys through the cascade, so the
+    /// retry stores a vault that no longer knows which chains it can derive.
+    ///
+    /// Transient: any `save()` from anywhere settles the context — the main
+    /// context autosaves — and the same button then works. It is the same
+    /// answer `KeyshareSweeper` gives to a dirty context, for the same reason.
+    case busy
+
     init(_ failure: KeyshareNormalizationFailure) {
         switch failure {
         case .unreadable(let pubkey):
@@ -56,6 +72,8 @@ enum KeygenCommitError: Error, Equatable, LocalizedError {
             return "keysharesUnreadableVaultNotSaved".localized
         case .unsealableKeyshare:
             return "keysharesUnsealableVaultNotSaved".localized
+        case .busy:
+            return "somethingWentWrongTryAgain".localized
         }
     }
 }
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5104, opened 2026-08-11
feat(defi): tag every Kamino transaction with the vs attribution memo

gastonm5:
## Description

Layer 5 of the stacked change adding **Kamino Earn vaults (kVaults)** to the Solana DeFi surface. This layer tags every Kamino transaction this app originates, so the deposits it drives can be measured.

Part of #5017 — the [scope addition filed as a comment](https://github.com/vultisig/vultisig-ios/issues/5017#issuecomment-5249982356). The issue closes when the whole stack lands.

> **Stacked PR.** Base is `feat/kamino-verify-5017` (#5057), so this diff is only layer 5. Review #5053 → #5055 → #5056 → #5057 first.

### Why the memo exists

Kamino's kvault API takes **no referrer or partner parameter**, so attribution cannot be a request field — it has to be something in the transaction. One SPL Memo instruction (`MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr`) carrying the literal `vs`, which is the filter Kamino's own attribution and the public Dune queries key on. It has to be **byte-identical on iOS, Android and Windows**, so the literal lives in one place (`KaminoAttribution`) that the injector, the validator and the verify-screen decoder all read from.

No extra signer, no accounts, ~37 bytes.

### Where it is injected, and why there

`KaminoTransactionPreparer.inject` already performs one byte edit — the ComputeBudget pair — and then validates and simulates the result. The memo is a second edit at the same point, so **what gets proven is what gets signed**:

```
API response → validate (as built) → simulate
             → injectingComputeBudget → injectingMemo("vs")
             → validate (injected) → simulate → keysign
```

It is **appended**, not prepended. The memo takes no part in what the transaction does, so it belongs behind the instructions that move money — and putting it last leaves every existing instruction at the index the shared template already expects, with only the two compute-budget instructions ahead of them.

`injectingComputeBudget` and the new `injectingMemo` now share one private byte editor (`injecting(programKey:leading:trailing:)`) rather than carrying two copies of the key-append and index-shift arithmetic. Duplicating safety-critical byte editing is worse than one refactor guarded by the golden vectors — which is exactly what guards it: the frozen `injected` fixtures are unchanged and still assert byte-identity through the refactor.

### What the guards say

`KaminoInstructionSequence` is shared by the validator (online, initiating device) and the decoder (offline, any device), so both admit **exactly one** memo carrying **exactly** `vs`:

| Case | Result |
|---|---|
| Memo `vs`, no accounts, after injection | accepted |
| Memo present *before* injection | `unexpectedMemo` — Kamino emits none, so it came from somewhere else |
| Memo with any other bytes (`v`, `vsx`, `VS`, `vs `) | refused; on the verify screen the transaction is `.unreadable` |
| Memo listing accounts | `malformedInstruction` — a memo naming accounts is the Memo program *attesting* they signed, which attribution is not |

The tag is matched **whole**, never as a prefix: a memo is free-form bytes, and `vs` followed by anything else is a different memo.

### What review caught

The tag was never asserted against its literal. Every test derived it from the constant — fixtures, injector, validator, decoder, live simulation — so a drift from `vs` to any other valid two-byte string would have left the entire suite agreeing with itself and green, while the attribution it exists for silently matched nothing. It is now pinned to `"vs"` / `[0x76, 0x73]` directly.

### Fixtures

`KaminoTransactionFixtures.<vector>.injected` is unchanged — it is the artifact produced by the reference implementation and simulated on mainnet. The tagged form is **derived** (`.tagged` applies `injectingMemo`) rather than pasted as a second frozen blob: a recorded blob would have no provenance of its own and would assert the same edit against itself. The edit's exactness is pinned separately in `SolanaV0TransactionTests`.

### Two claims no offline test can make

Both went into the opt-in live suite (`testTheTaggedDepositStillFitsAndStillExecutes`), which simulates the tagged transaction against mainnet and reports each number with its headroom:

1. **Size** — the 1,232-byte packet cap. The vectors this was developed against are August's; a vault whose transaction grows could push the tagged form over, and the injection would then throw on a path the user cannot avoid.
2. **Compute** — the pinned unit limits were measured on **untagged** transactions (252k/287k/309k against 320k/350k/400k). The Memo program costs compute of its own, so the tagged form has to still fit *with* the margin rather than land on the limit.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

## Parity

Required — this changes the bytes that get signed.

- SDK: `n/a` — the SDK has no Kamino provider yet (noted as an unfiled gap on #5057).

**The tag must be byte-identical on all three platforms** — `vs`, lower case, no whitespace. That is the whole point of it.

## Checklist

Gate: **5740 tests, 11 skipped, 0 failures** (the skips are 1 pre-existing plus the opt-in live tests). SwiftLint 0 violations across the touched sources and tests.

_None — no user-visible change. The memo is invisible on the verify screen by design; it is not a spend and not a destination._

- **Agent**: Claude Code
- **Issue**: #5017
- **Confidence**: high
- **Needs human review for**: the injection refactor in `SolanaV0Transaction` — it edits the bytes that get signed, and the ComputeBudget path now runs through shared code.


johnnyluo (approved):
LGTM

johnnyluo:
Superseded by #5117 — this PR was auto-closed by the merge queue when its base branch (`feat/kamino-flows-5017`, merged as #5056) was deleted; GitHub's automatic base-change didn't stick before the queue closed it out.

PR #5105, opened 2026-08-11
feat(defi): bring the Kamino Earn UI to the design

gastonm5:
## Description

Layer 6 of the stacked change adding **Kamino Earn vaults (kVaults)** to the Solana DeFi surface. This layer brings the Earn UI to the design — the segment itself, and the promo that points at it from the main screen.

Part of #5017 — the [UI design filed as a comment](https://github.com/vultisig/vultisig-ios/issues/5017#issuecomment-5249982356). Figma: [DeFi Page — Earn — Kamino Vaults](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=79469-243134). The issue closes when the whole stack lands.

> **Stacked PR.** Base is `feat/kamino-memo-5017` (#5104), so this diff is only layer 6. Review #5053 → #5055 → #5056 → #5057 → #5104 first.

### What the design asked for

**1. Earn leads — everywhere.** On the Solana segments, and in the select-positions picker too: two lists describing the same thing and disagreeing about which comes first is how a user hunts for a section they were just looking at. The picker's order turned out to be presentational only — a tapped cell is filed through `selectionIndex`, which each type carries itself, so nothing there is addressed by array position; the test pins the new order *and* those indices, since their independence is what makes the move safe.

**1a. Segment order.** The design's segmented control reads `Earn  Staked`; the app had it the other way round. This also makes Earn the segment the screen opens on — `onLoad` selects the first type. The order is presentational only: the position picker files a selection by `selectionIndex`, which Earn does not have, so nothing about persistence or routing depends on it.

**2. Each vault says what it has earned.** This was not on screen. The figure existed — the card carried a "Profit & Loss" row — but only in the underlying token and only as a label beside a number. It now reads `Earned: 200 USDC` with its fiat value opposite, matching the `Deposited: 1,000 USDC → $1,000.23` line above it, which is the pair the design draws rather than two stacked columns.

A vault holding nothing drops both figure rows and takes a full-width Deposit, as the design draws its empty cards: there is no position to describe, and rows of zeros describe nothing.

**3. The main screen promotes it.** The design puts *"New on Solana / Start earning with Kamino"* in the main screen's banner carousel, and it was the one piece of that Figma with nothing behind it. Everything it needed already existed — `VaultBannerType` carries copy, icon and dismissal rule, `CarouselBannerView` already renders title-in-tertiary above subtitle-in-primary, and `PromoBannerDismissalStore` already honours a per-banner TTL — so this is one enum case, not a new surface. It shows for any vault that holds Solana — deliberately not gated on already having a position, since someone earning in one vault is a good audience for the other two, and dismissal is what retires it. It leads the carousel, and tapping it lands on Earn directly, since that is now Solana's first segment.

### What review caught

**A zero is two different claims.** Enabling a vault writes a zero placeholder, and the card was reading it as "holds nothing" — which hid **Withdraw** from a user who had deposited on another device, or whose refresh failed right after depositing. A row now records whether its figures were read or seeded, and the way out of a position is never removed on the strength of a read that has not happened. (The figures stay off the card either way; the withdraw form reads the position itself and is the thing that can say "nothing to withdraw" truthfully.)

**The carousel could strand itself.** `BannersCarousel` never clamped its index when the banner array is replaced from outside, so dropping the page the user was on left `currentIndex` and `scrollPosition` past the end — and the auto-advance declines to move at one banner, so nothing corrected it. Reachable today (back up a vault while the backup banner is showing); fixed here with a `clamped(_:count:)` beside the existing removal math, and unit-tested.

**A loss is not something earned.** The source is `totalPnl`, which goes negative. `Earned: -3 USDC` in red still asserts the loss was earned and leaves the colour to correct it, so below zero the label reads `Lost: 3 USDC` with the figure unsigned beside it.

### Three design points deliberately NOT followed

Flagged rather than changed, because each one would cost something the issue or the app already established:

| Design | This PR | Why |
|---|---|---|
| Subtitle `Kamino · Solana` | `Curated by <curator>` + risk tier | #5017 **requires** risk labeling — "show RWA one risk tier above the conservative vault, not as T-bill backed". A static string would undo an explicit acceptance criterion. |
| Third vault is `JLP Vault` | `RWA USDC` (RockawayX) | The issue's launch table names the three vaults; the registry is the source of truth. The design appears to predate it. |
| No total card | "Kamino Earn" total across vaults | Removing it loses the aggregate the DeFi total is built from. |

Happy to take any of these the other way if the design is the newer decision.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

## Parity

- SDK: `n/a` — presentation only, no SDK surface.

## Checklist

Gate: **5746 tests, 11 skipped, 0 failures**. SwiftLint 0 violations across the touched sources and tests. New strings (`kaminoEarnDeposited` as a format, `kaminoEarnEarned`, `kaminoEarnLost`, `kaminoBannerTitle`, `kaminoBannerSubtitle`) are in all **8** shipping locales, sorted, entry counts equal at 1,785.

_To add: the Earn segment with a funded vault and an empty one, plus the main-screen banner, against the Figma frames._

- **Agent**: Claude Code
- **Issue**: #5017
- **Confidence**: high
- **Needs human review for**: the three design deviations above — they are judgement calls between the Figma and the issue text.

## Summary by CodeRabbit

* **New Features**
  * Added a Kamino Earn promotional banner for eligible Solana vaults, with direct navigation to DeFi.
  * Improved Kamino Earn cards with clearer deposited, earned, and lost amounts, position-aware actions, and loading states.
  * Added attribution memos to supported Kamino Solana transactions.

* **Bug Fixes**
  * Prevented banner carousel positions from exceeding available banners.
  * Improved withdrawal availability handling for confirmed and unconfirmed positions.
  * Updated Kamino messaging across supported languages.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -634,8 +634,8 @@
 "kaminoEarnApy30d" = "APY (30 T.)";
 "kaminoEarnCuratedBy" = "Kuratiert von %@";
 "kaminoEarnDeposit" = "Einzahlen";
-"kaminoEarnDeposited" = "Eingezahlt";
-"kaminoEarnPnl" = "Gewinn & Verlust";
+"kaminoEarnDeposited" = "Eingezahlt: %@";
+"kaminoEarnEarned" = "Verdient: %@";
 "kaminoEarnRiskConservative" = "Konservativ";
 "kaminoEarnRiskPrivateCredit" = "Private Credit";
 "kaminoEarnTitle" = "Kamino Earn";
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -627,6 +627,8 @@
 "joinSend" = "Senden beitreten";
 "joinSwap" = "Tausch beitreten";
 "joinTransactionSigning" = "Verbinden Sie die Transaktionsunterzeichnung";
+"kaminoBannerSubtitle" = "Verdiene jetzt mit Kamino";
+"kaminoBannerTitle" = "Neu auf Solana";
 "kaminoDepositMinimum" = "Mindesteinzahlung ist %@ %@.";
 "kaminoDepositMissingCoin" = "Füge %@ auf Solana zu deinem Vault hinzu, bevor du einzahlst.";
 "kaminoDepositTitle" = "Einzahlen";
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -636,6 +636,7 @@
 "kaminoEarnDeposit" = "Einzahlen";
 "kaminoEarnDeposited" = "Eingezahlt: %@";
 "kaminoEarnEarned" = "Verdient: %@";
+"kaminoEarnLost" = "Verlust: %@";
 "kaminoEarnRiskConservative" = "Konservativ";
 "kaminoEarnRiskPrivateCredit" = "Private Credit";
 "kaminoEarnTitle" = "Kamino Earn";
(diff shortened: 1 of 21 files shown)


johnnyluo (approved):
LGTM

johnnyluo:
Closing/reopening to force GitHub to re-fire pull_request checks — the last two pushes never queued an Actions check-suite.

Issue #5086, opened 2026-08-10
Sealed shares can outlive their Keychain key across a restore, and launch reads it as 'no passcode'


realpaaao:
Replaces what #5083 covered. We're keeping at-rest encryption (vultisig/vultisig-sdk#1838), so this state has to be caught instead of designed away.

## What can happen

The encrypted shares live in the SwiftData store. The key that opens them is a Keychain item.

The store goes into every device backup — nothing sets `isExcludedFromBackup`. The Keychain item is `WhenUnlocked`, not `ThisDeviceOnly` (`KeychainService.swift:80-90`), so an unencrypted computer backup doesn't carry it to a new device.

Restore that way and you have the boxes without the key.

## Why the app doesn't notice

`alignLockModeWithTheWrappedKey` (`KeyshareInstallReconciler.swift:285-298`) sees the key is missing, logs a warning, sets `mode = .deviceAuth`, moves on. No lock screen, no message.

There's a repair path for the opposite case — Keychain survives, container is new — at `:131-191`. Nothing for this one.

`hasSealedShare()` already exists and would catch it. It's called from one place: `setPasscode` (`PasscodeService.swift:221`). Never at launch.

So the app opens normally. Vault names, addresses and balances all render, because `Vault.getKeyshare` swallows the error with `try?` and returns nil. Signing fails with "fail to get local keyshare" (`DKLSKeysign.swift:65`).

## Repro

Passcode on, vaults sealed. Back up the iPhone to a computer without ticking "Encrypt local backup". Restore onto a different iPhone. App looks healthy, can't sign.

## Fix

At launch, if `hasSealedShare()` is true and the wrapped key is confirmed absent, stop. Show a clear screen: these vaults are encrypted, the key isn't on this device, restore from your `.vult`.

The check exists. It needs a caller in `alignLockModeWithTheWrappedKey`.

## While we're here

iOS export is already correct — `mapToProtobuff` opens every share and throws if it can't. Add a regression test so it stays that way. It's the recovery path the whole design leans on.

## Done when

- [ ] Sealed shares with no key show an explicit restore-from-backup screen at launch, never a silent open
- [ ] The screen names the `.vult` as the way back
- [ ] Test pinning that export unseals or throws

realpaaao:
Design for this screen is tracked in vultisig/vultisig-sdk#1845 — the "your vaults are encrypted and the key is gone" recovery state, shared across all three clients. Worth waiting for it rather than each platform inventing its own copy.

PR #5107, opened 2026-08-11
fix(import): let a .vult replace the vault this device can no longer open

gastonm5:
> Stacked on #5106. Review that one first — this PR's diff is the single commit on top of it.

## What

#5106 puts a recovery screen in front of a device whose key shares are sealed with a key it no longer holds, and tells the user to restore from their `.vult`. That advice was a dead end: `importDecision` sends the same vault to `.duplicate` via `isVaultUnique` — it *is* the same vault, by key material — and the import path skips duplicates. So a user following the screen imported nothing, and the screen returned on every launch, forever.

This PR makes the route work.

Part of #5086.

## How

**A fourth decision, returned only when the stored row is provably unopenable.** `.replaceOrphaned` requires the colliding stored vault to have **every** share sealed and the wrapper to be **confirmed** absent. `ProtectedVaultImporter.commit` takes the row it displaces, deletes it inside the lease it already holds, and saves it in the same transaction as the insert.

**Remove-then-import was the other shape, and it was rejected.** It puts a destructive step ahead of the recovery, so a user with the wrong file or the wrong password loses the row and gains nothing — and "delete this to fix it" is the exact register vultisig/vultisig-sdk#1845 rules out.

### The predicate, clause by clause

Each one is the difference between a recovery and a loss, so none of them is a convenience:

- **Exactly one colliding stored vault.** A backup matching two is a question this cannot answer, and answering it by picking one destroys the other's key material.
- **A *confirmed* absent wrapper**, never `.unavailable`. A Keychain that merely went quiet may still hold the key, and the shares would open again on the next launch.
- **Every stored share sealed, and at least one.** A partly-sealed vault is not a state the sweep can produce, and an empty one has nothing to be orphaned about.
- **The backup *is* that vault.** Every non-empty signing identity the stored row carries — `pubKeyECDSA`, `pubKeyEdDSA`, `publicKeyMLDSA44` — has to be declared identically by the backup. Matching on one public key is what makes something a duplicate; it is not enough to make it a replacement. A file naming a different EdDSA key would swap a vault that cannot sign for one that still cannot. An identity the backup *adds* is fine; one it drops is not.
- **The backup carries a share for every key** — its own declared ones and the stored row's — with no empty share values. `ProtectedVaultImporter` proves that whatever the backup carries opens here, and it proves it before the deletion; but a file with an empty `keyshares` array normalizes perfectly well and would buy the deletion with nothing.
- **The raw unique-index check still runs**, against everything the replacement is *not* displacing. `pubKeyECDSA`/`pubKeyEdDSA` default to `""`, and `""` is a real value in the index.

### The two orderings

- **Normalization runs ahead of the deletion**, so a backup this device cannot open is refused while the row it would have replaced is still there.
- **A replacement over a context carrying somebody else's unsaved work is refused outright.** A pending deletion cannot be withdrawn row by row, and `rollback()` is only this method's to call when the context was clean on the way in — so the per-row failure path would leave the orphan on its way out with nothing arriving in its place. `.busy`, and the user retries.

### Never the upsert

The deletion is explicit. SwiftData resolves a duplicate `@Attribute(.unique)` insert by overwriting the stored row with the incoming values, with nothing thrown and nothing logged — which would make "the replacement was stored" and "the original was destroyed and nothing replaced it" the same code path. The success test captures the orphan's `persistentModelID` before the import and asserts the persisted row has a different one, so it cannot pass against an implementation that leans on the upsert.

## Known limits, for a reviewer to weigh

- **Nothing parses a share to confirm its bytes really derive the public key it is labelled with.** That is TSS work behind a critical boundary, and the ordinary import path extends a backup the same trust. The checks above are what keep this path from extending it any *further*.
- **The recovery screen is not re-raised when the user backs out of the import.** Inherited from #5106 and documented there; it is derived again at the next launch. Cross-model review flagged it twice and it is left as a deliberate product call rather than being resolved inside a fix PR — worth a decision.

## Testing

`OrphanedVaultReplacementTests`, 18 cases, every fixture carrying real key shares — a suite of share-less vaults would pass while refusing every vault a user actually has:

- the orphan replaced end to end through `EncryptedBackupViewModel.restoreVault`, asserted on a fresh `ModelContext` and on the row identity;
- the replacement keeping the backup's own name;
- an ordinary duplicate still skipped; a present wrapper, an unreadable wrapper, a partly-sealed vault, a share-less vault, and a backup colliding with two stored vaults all refused;
- a backup declaring a different EdDSA key, one declaring a key it has no share for, one carrying no shares, one carrying an empty share, and one missing a share the stored row holds, all refused;
- an incomplete backup that would otherwise have upserted over an unrelated healthy vault — both stored vaults survive;
- two copies of the same vault in one ZIP replacing the orphan exactly once;
- a failed save leaving the orphan in place with its own shares, a replacement refused over a busy context, and a backup this device cannot open refused without touching the orphan.

Gate, on a dedicated `iPhone 17 Pro Max` / iOS 26.5 simulator:

```
Executed 5503 tests, with 1 test skipped and 0 failures (0 unexpected)
** TEST SUCCEEDED **
```

macOS:

```
** BUILD SUCCEEDED **
```

`swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 22 violations, unchanged from `main`. 0 new.

Two rounds of read-only `codex exec` review on the commit plus two whole-branch rounds; ledger in the wiki at `pages/projects/vultisig/passcode-app-lock/codex-review-ledger-5086.md`.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Security/Keyshare/ProtectedVaultImporter.swift
+++ b/VultisigApp/VultisigApp/Core/Security/Keyshare/ProtectedVaultImporter.swift
@@ -10,9 +10,10 @@ import SwiftData
 private let logger = Log.app.store
 
 enum ProtectedVaultImportError: Error, Equatable {
-    /// A passcode transition is in progress. The import is refused rather than
-    /// queued: the transition is about to rewrite every stored share, and a
-    /// vault inserted underneath it would never be seen.
+    /// A passcode transition is in progress, or a *replacement* was asked for
+    /// over a store carrying somebody else's unsaved work. The import is refused
+    /// rather than queued: the transition is about to rewrite every stored
+    /// share, and a vault inserted underneath it would never be seen.
     case busy
     /// An incoming share is sealed and does not open here — the wrong key, or a
     /// value that is not a share at all.
@@ -111,9 +112,15 @@ struct ProtectedVaultImporter {
     ///   that exists, and it is aimed at rows this method can still take back —
     ///   so the caller starts it once `commit` has returned, never from inside
     ///   `prepare`.
+    ///
+    /// - Parameter orphaned: stored vaults the incoming ones replace, deleted in
+    ///   the same save that stores their replacements. See below for why this is
+    ///   an explicit delete and why it refuses a context that is carrying
+    ///   anything else.
     @MainActor
     func commit(
         _ vaults: [Vault],
+        replacing orphaned: [Vault] = [],
         into context: ModelContext,
         prepare: (Vault) -> Bool = { _ in true }
     ) throws {
@@ -127,6 +134,9 @@ struct ProtectedVaultImporter {
         }
         defer { coordinator.end(episode) }
 
+        // Ahead of the deletion, and that ordering is the whole safety of a
+        // replacement: a backup whose shares this device cannot open must be
+        // refused while the row it would have replaced is still there.
         let normalized = try statedAsAnImportFailure {
             try vaults.map { try normalizer.normalizedShares(of: $0) }
         }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

Issue #1845, opened 2025-02-17
[BUG] Pairing devices crashes on new onboarding keygen


realpaaao:
**Describe the bug**
The pairing device (iOS) crashes on completed keygen when finished successfully

**To Reproduce**
1. Initiate Keygen
2. Join with an iOS device
3. finish keygen

**Expected behavior**
Able to create a vault with an iOS device as pairing device

System:
 - Version `main branch
`

johnnyluo:
hmmm, I having been using this everyday , but it doesn't crash for me

realpaaao:
@johnnyluo @Rockindash this still happens for me on testflight with my iPhone 11

johnnyluo:
Can't reproduce it

PR #5111, opened 2026-08-11
feat(passcode): ask for a backup before the passcode goes on

gastonm5:
## Description

Setting a passcode encrypts every stored key share under a key this device holds. If that key is later lost while the shares remain — a restore that did not carry the Keychain across is the ordinary way — the `.vult` backup is the only route back, and nothing substitutes for it.

Until now **Set passcode** navigated straight to the entry screen and said none of that. The only mention was a line of caption text further down the settings list (`passcodeManageExplanation`), which is a warning you read after deciding, if at all.

Tapping it now raises a sheet first, offering the backup flow or a way past on the user's word that a backup already exists.

No issue — filed straight from a conversation about the recovery side of the same problem (#5086).

### Three decisions worth reviewing

**It is a prompt, not a gate.** The app cannot tell whether a backup exists or is current. `Vault.isBackedUp` is a plain `Bool` set by any export *or import* and never cleared when the vault changes afterwards, so treating it as proof would wave a stale backup through while nagging someone whose backup is fine. It is therefore neither read to skip the prompt nor written when the user answers — recording the answer would put a claim in the store that later code might trust.

**"Back up now" goes to the backup *selection* screen**, not an export of the current vault. A passcode seals the key shares of *every* stored vault, and `VaultBackupSelectionScreen` is the only entry point that can export more than one.

**The prompt is an item-based sheet rather than a boolean one.** Both answers end in a navigation push, and only `crossPlatformSheet(item:onDismiss:)` can run one after the dismissal has settled — pushing while the sheet is still on screen races the animation and the destination can be dropped. That hazard is documented on the overload itself (`CrossPlatformSheet.swift:15-19`); `TransactionHistoryScreen` uses it for the same reason.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

None of the above — Settings only, no keysign/broadcast/fee surface touched.

## Checklist

**No tests.** The change is a presentation decision inside a SwiftUI view — which sheet is raised and where each answer navigates. The unit-test target has no view-hosting harness, and the one piece of logic worth pinning (the answer surviving the dismissal) lives in `@State` on the view. Verified by running it instead; see below.

None attached — verified on-device by @gastonm5, who also made the final sheet sizing and detent adjustments.

## Additional context

**Gate**
- iOS: `Executed 5465 tests, with 1 test skipped and 0 failures (0 unexpected)` → `** TEST SUCCEEDED **`
- macOS: `** BUILD SUCCEEDED **` (the iOS target never compiles `#if os(macOS)`, and this sheet has a macOS-only path below macOS 26 where `crossPlatformSheet` renders as a `ZStack` sibling rather than a real presentation)
- SwiftLint: 22 violations, identical to `main` — 0 new
- Localisation: 4 new keys across all 8 shipping locales, each file at 1745 entries

**Relationship to the shared design work.** vultisig/vultisig-sdk#1845 specifies two passcode screens for all three clients. This is a lighter take on its *screen 1*: that issue asks for a **blocking** gate requiring a current backup, whereas the "I already have a backup" button here makes this a speed bump. If that design lands, this is the surface it would replace.

Two things it deliberately does not attempt, both of which #1845 wants and neither of which is currently expressible: a **"your backup is old"** state — `isBackedUp` is a bool with no export timestamp or vault-version marker behind it, so staleness cannot be detected — and any blocking behaviour.

The recovery side of the same problem is #5086 → #5106 / #5107.

- **Agent**: Claude Code
- **Issue**: none
- **Confidence**: medium — gate green and confirmed working on device, but no second-model review pass and no automated test coverage
- **Needs human review for**: the copy (it says "a key held on this device" and deliberately **not** that the key never leaves it — the wrapper is `kSecAttrAccessibleWhenUnlocked`, not `ThisDeviceOnly`, so it does travel in encrypted backups and device migration), and the item-based sheet plumbing in `ManagePasscodeScreen`

## Summary by CodeRabbit

* **New Features**
  * Added a backup prompt before passcode setup, allowing users to back up immediately or confirm an existing backup.
  * Added localized prompt content in English, German, Spanish, Croatian, Italian, Korean, Portuguese, and Simplified Chinese.
  * Improved navigation based on the selected vault and backup choice.
* **Accessibility**
  * Added accessibility identifiers for backup prompt actions.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -943,6 +943,10 @@
 "partOf" = "Teil %d von %d";
 "passcodeAlreadySet" = "Auf diesem Gerät ist bereits ein Passcode gesetzt. Gehen Sie zurück, um ihn zu ändern oder zu entfernen.";
 "passcodeAutoLockTitle" = "Automatische Sperre";
+"passcodeBackupPromptBackUpNow" = "Jetzt sichern";
+"passcodeBackupPromptHasBackup" = "Ich habe bereits ein Backup";
+"passcodeBackupPromptSubtitle" = "Ein Passcode verschlüsselt Ihre Key-Shares mit einem Schlüssel, der auf diesem Gerät liegt. Geht dieser Schlüssel verloren, ist Ihr .vult-Backup der einzige Weg zurück – der Passcode selbst lässt sich nicht wiederherstellen.";
+"passcodeBackupPromptTitle" = "Erstellen Sie ein Backup, bevor Sie einen Passcode festlegen";
 "passcodeBiometricDisableFailed" = "Die biometrische Entsperrung konnte nicht deaktiviert werden. Versuchen Sie es erneut.";
 "passcodeBiometricEnableFailed" = "Die biometrische Entsperrung konnte nicht aktiviert werden. Ihr Passcode funktioniert weiterhin.";
 "passcodeBiometricNotAvailable" = "Dieses Gerät kann die biometrische Entsperrung nicht verwenden.";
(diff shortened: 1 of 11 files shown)


johnnyluo (approved):
LGTM

PR #5113, opened 2026-08-11
fix(defi,rates): stop YieldPosition and DatabaseRate churning the store on no-op refreshes

gastonm5:
## Description

Two independent sites where SwiftData `@Model` state is written without comparing first, so every refresh re-invalidates SwiftUI readers even when nothing changed.

SwiftData's `@Model` setter calls `withMutation` **without comparing** old and new, so assigning an identical value still publishes a change. Measured against the app's own shipping types in an `NSWindow` + `NSHostingView` harness: an unguarded `apply(_:)` called with an identical DTO produced **344,063** body evaluations in 8s (~44,000/s); guarded, **1**. It is amplified app-wide by the root `@Query` at `ContentView.swift:17`, and the amplification is **cross-type** whenever a `save()` follows — which every storage service here does.

Guard shape follows #5109: an all-or-nothing early return covering every write in the method, including any timestamp.

### `YieldPositionStorageService`

- `:26-30` — three unguarded writes including `lastUpdated = .now`. Standard equality guard.
- `:69-81` — `syncRedemptions` **deleted and recreated every redemption row on every refresh**. No equality guard can fix that: the delete+insert is itself store churn that fires the amplifier unconditionally. Replaced with a diff/merge.

**Identity key: `YieldRedemption.id`.** It is the provider's own request identifier, it is already what `YieldRedemptionRecord` persists as `@Attribute(.unique)`, and it is already the key the read path (`YieldViewModel.seed`) maps back out on. Deliberately **not** derived from the mutable fields — a redemption keeps its identity precisely while it moves `pending → claimable → settled`, and that transition must read as an update, not a delete plus an insert.

The delete-recreate was **not** load-bearing: the existing `testUpsertReplacesStaleRedemptionRows` already asserts the same-id/changed-status semantics the merge preserves. New rows attach via the inverse (`record.position = position`) rather than by appending to `position.redemptions` — appending is a get-modify-set of the whole relationship, and rows deleted earlier in the same pass can still appear in the array the getter returns.

> ⚠️ Worth knowing when weighting this: **`syncRedemptions` is currently dead in production.** The only `DefiYieldProvider` is Circle, which passes `redemptions: []` unconditionally (`CircleYieldProvider.swift:100`). Today its only cost is the unconditional `position.redemptions = []` relationship write; the diff/merge is prophylactic for the first windowed provider.

### `RateProvider` — needed more than a guard

`:178-187` was the app's largest unguarded write burst. Its previous defence — "no view reads `DatabaseRate`" — is refuted by the cross-type amplification above.

But an equality guard alone **cannot** fix it. `NaN != NaN`, so `guard stored != incoming else { return }` always falls through for a non-finite rate. Worse, SwiftData **coerces NaN to `0` on persist** (measured; infinities survive), so the stored value never compares equal to the incoming NaN either. **A single NaN quote would have re-dirtied its row on every refresh, forever, silently defeating the guard.**

Non-finite rates are therefore dropped at the entry to `save(rates:)`, ahead of both the cache merge and the persistence loop. Flagging this as a real scope extension beyond "guard the writes".

## Which feature is affected?
- [ ] Sending
- [ ] Swap

None of the above — DeFi yield persistence and the rate cache. SwiftUI/SwiftData-specific, no analogue on the other platforms, so no parity section.

## Checklist

`YieldPositionWriteGuardTests` (12) and `RateProviderWriteGuardTests` (9). All assert via `withObservationTracking` rather than on values — identical values cannot distinguish guarded from unguarded — plus per-field sweeps.

The redemption diff additionally asserts on `persistentModelID`. `ObjectIdentifier` was tried first and is **unsound**: SwiftData re-materializes fresh instances for the same row after a save, so instance identity changes for reasons unrelated to deletion.

**Verified non-vacuous**, each guard reverted then restored: balance guard → 4 fail; redemption diff → 8 fail; rate equality guard → 2 fail; non-finite filter → 6 fail; warm-start filter → 1 fail.

Gate: `Executed 5486 tests, with 1 test skipped and 0 failures (0 unexpected)`, `** TEST SUCCEEDED **`. SwiftLint 22 — identical to `main`'s baseline.

## Review notes

One finding was **rejected with reason** and documented in code: a suggestion to preserve redemption ordering. A SwiftData to-many is unordered, so the wholesale assignment this replaces did not preserve order either; imposing one needs a sort index and would restore the churn this PR removes.

A later round raised deleting rows at warm-start load; the **filtering** half was taken and the **deleting** half rejected — a launch-time store write is precisely the mutation class this branch exists to remove.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate persistence by ignoring invalid values and avoiding unnecessary updates.
  * Improved yield position synchronization, preventing duplicate records and preserving unchanged data.
  * Ensured stale redemption records are removed during synchronization.

* **Tests**
  * Added coverage for rate caching, validation, persistence, and update behavior.
  * Added coverage for yield balance and redemption updates, duplicates, deletions, and no-op saves.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
+++ b/VultisigApp/VultisigApp/Core/Services/Rates/RateProvider.swift
@@ -97,18 +97,34 @@ final class RateProvider {
         // Defer fetch to avoid re-entrant SwiftData calls during init.
         // Must run on MainActor — ModelContext is MainActor-isolated.
         Task { @MainActor [weak self] in
-            guard let self else { return }
-            let descriptor = FetchDescriptor<DatabaseRate>()
-            do {
-                let objects = try Storage.shared.modelContext.fetch(descriptor)
-                self.replaceRates(with: objects.map { Rate(object: $0) })
-                // Persisted rates make fiat renderable immediately on a warm
-                // start; announce them so views that cached a pre-load fiat
-                // value recompute instead of waiting on a network refresh.
-                self.ratesDidChange.send()
-            } catch {
-                Log.chain.service.error("Failed to load rates: \(error.localizedDescription, privacy: .public)")
-            }
+            self?.loadPersistedRates()
+        }
+    }
+
+    /// Populates the cache from the persisted rates.
+    ///
+    /// Called once, from `init`. Not `private` because the initializer of a
+    /// process-wide singleton cannot be re-entered from a test, and the
+    /// non-finite filter below is worth verifying rather than asserting.
+    @MainActor
+    func loadPersistedRates() {
+        let descriptor = FetchDescriptor<DatabaseRate>()
+        do {
+            let objects = try Storage.shared.modelContext.fetch(descriptor)
+            // The same non-finite filter `save(rates:)` applies, on the way back
+            // in: a build predating that filter could have persisted an infinity,
+            // and this is the one path that would put it back into the cache
+            // every fiat render reads. Filtered rather than deleted — a
+            // launch-time store write is exactly the kind of gratuitous mutation
+            // these guards exist to stop making, and the row is overwritten by
+            // the next finite quote for that coin anyway.
+            replaceRates(with: objects.map { Rate(object: $0) }.filter { $0.value.isFinite })
+            // Persisted rates make fiat renderable immediately on a warm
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5114, opened 2026-08-11
test: stop TestStore vault fixtures collapsing into one row

gastonm5:
## Description

`TestStore.makeVault(pubKey:)` varied `name` and `pubKeyECDSA` but **hard-coded `pubKeyEdDSA: "test-pub-eddsa"`**. `Model/Vault.swift:12-15` marks `name`, `pubKeyECDSA`, `pubKeyEdDSA` and `publicKeyMLDSA44` **each independently** `@Attribute(.unique)`, and SwiftData resolves a unique collision by **upserting** — each insert silently replaces the previous row, with no error.

So every vault built by this helper collided. Any test creating more than one got one, per-case isolation was fiction, and assertions shaped `for vault in fetch() { … }` or checking ordering **passed vacuously** over the single survivor. 68 call sites across 27 files.

**Fixing it turned 98 test cases red.** Every one had been green for the wrong reason.

### The repair

| File | Cases | What it was silently doing |
|---|---|---|
| `Send/SendCryptoVerifyViewModelTests` | 72 | Never installed an in-memory container — inserted into whatever context the previous test class left behind, accumulating rows that all upserted into one. One test built a *payload* vault and a *transaction* vault that were the same row. |
| `Send/SendTransactionTests` | 23 | Same — **all 23** failed, so the store was already contaminated before the class started |
| `Keysign/SendPreviewOverrideTests` | 2 | Same |
| `Send/SendPhaseDRegressionTests` | 1 | Same (cross-class leakage from a vault left by an earlier class) |

All four now install a per-test container with one shared vault in `setUp` — one vault, many transactions, which is what these tests always meant. Every case was "relying on cross-case leakage"; **no product bugs surfaced**, and nothing outside `VultisigAppTests/` is touched.

> ⚠️ **Side effect worth knowing:** because those four classes had no container at all, the unit-test suite was writing `Vault` rows into the **app's real SwiftData store** in the simulator. Every `TestStore.makeVault` caller now installs one, so that's closed.

The repair is isolated in its own commit (`test: give the Send fixture classes their own in-memory store`) so it can be dropped and split if reviewers prefer.

### Two details worth review attention

**`publicKeyMLDSA44` is deliberately left `nil`, not varied.** A unique index treats NULLs as distinct, so nil fixtures never collapse (the repo already relied on this — `KeygenVaultCommitTests.swift:45`), and giving it a value would flip every product branch that reads the field as "this is a quantum vault": `CoinFactory+QBTC`, the MLDSA keysign key at `KeysignViewModel.swift:548,633`, `CoinSelectionViewModel.swift:66,83`, `VaultAdvancedSettingsScreen.swift:62,68`. That reasoning is in the doc comment and pinned by a test.

**Collision detection is *pre*-insert, not a post-insert row count.** SwiftData resolves the collision at **save**, so counting after the insert reads +1 even for a duplicate and would surface the problem at whatever unrelated place saved next. `assertNoUniqueCollision` fetches first and `XCTFail`s naming the exact colliding fields, with `file:`/`line:` defaulted to the call site.

### Still latent, reported not changed

`Send/SendFormFixture.swift:83-92` and `FunctionCall/FunctionCallFixture.swift:22-33` hard-code all three unique fields but never insert into a context — they collapse the moment a caller inserts two. `ReferralFormFixture` delegates to the former. Every other `@Model` with a unique attribute (`BondPosition`, `StakePosition`, `LPPosition`, `CirclePosition`, `YieldPosition`, `LimitOrder`, `DatabaseRate`, `CustomRPCOverride`, `ChainPublicKey`, `PendingTransaction`) was surveyed: all test constructions are ad-hoc with ids derived from real inputs, no shared helper hard-codes one.

One pre-existing vacuous assertion is **deferred, not fixed**: `SendPhaseDRegressionTests.testFeesInReadableReflectsTxFeeNotGas` — `feesInReadable` returns `.empty` when the vault has no native coin and the test's own escape hatch then skips the assertion. Verified pre-existing on `e20272b86`; making it real requires seeding the shared `RateProvider` singleton, which deserves its own change.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

None of the above — test infrastructure only. No product code changed.

## Checklist

`Helpers/TestStoreTests.swift` — three fixtures survive a `save()` as three rows; the guard fires on a duplicate (via `XCTExpectFailure`, which also proves the fetch sees still-unsaved fixtures); nil MLDSA keys don't collapse.

Gate, quoted from the log markers:

- Baseline, unmodified worktree — `** TEST SUCCEEDED **`, `Executed 5465 tests, with 1 test skipped and 0 failures (0 unexpected)`
- Helper fix only — `** TEST FAILED **`, `Executed 5468 tests, with 1 test skipped and 105 failures (0 unexpected)` (98 cases)
- Final — `** TEST SUCCEEDED **`, `Executed 5468 tests, with 1 test skipped and 0 failures (0 unexpected)`
- SwiftLint 22 → 22, no regression.

Staleness explicitly ruled out: disk stayed at 21–22 GiB free throughout (a full volume can make `xcodebuild` emit `** TEST SUCCEEDED **` from a stale run), the count moved 5465 → 5468, and the helper-fix-only run went red with failures originating in the edited helper — which a stale run cannot fake.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by using fresh in-memory storage for each test.
  * Added validation that test vault fixtures remain distinct and reject duplicate data.
  * Updated send and transaction tests to reuse consistent fixtures, improving reliability across fee, payload, balance, and transaction scenarios.
  * Added regression coverage for vault data integrity, including optional key fields.

### DIFF
--- a/VultisigApp/VultisigAppTests/Helpers/TestStore.swift
+++ b/VultisigApp/VultisigAppTests/Helpers/TestStore.swift
@@ -14,6 +14,7 @@
 
 import Foundation
 import SwiftData
+import XCTest
 @testable import VultisigApp
 
 /// Token returned by `TestStore.installInMemoryContainer()`. Holds the previous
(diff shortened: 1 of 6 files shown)


johnnyluo (approved):
LGTM

PR #5118, opened 2026-08-12
perf(defi): cut the Kamino deposit's exposure to the Solana proxy stall

gastonm5:
## Description

A SOL deposit into a Kamino Earn vault "often gets stuck on the loading screen for quite a while, then goes through". There is no local lock. The app is sitting on a request that the backend intermittently holds open for a minute, and the deposit flow made **12** of those requests before anything was signed.

This PR cuts the client's exposure. **It does not fix the backend half** — see below.

> **Stacked PR.** Base is `feat/kamino-earn-ui-5017` (#5105), so this diff is only the latency work. **#5105 must merge first.**

Part of #5017. No dedicated issue exists for the stall itself, so this PR does not close one.

### The measured root cause

`api.vultisig.com/solana/` parks a fraction of requests for **exactly ~60.5 s** and then answers a valid `200`. Measured from a standalone HTTP probe with no app involved, splitting each request into DNS / TCP+TLS / send / wait-for-response, interleaving control hosts so every target saw identical local conditions:

```
vultisig /solana/   STALL total=60551 ms (connect+tls=75, wait=60476) http=200
```

The handshake is normal. **The entire minute is wait-for-response** — the bytes leave the device and the connection then sits idle, which is exactly why it feels like a client-side lock and why a trace taken behind the stalling component shows no request arriving.

A controlled 150-round run, three targets interleaved:

| Target | n | p50 | p99 | max | stalls |
|---|---|---|---|---|---|
| `api.vultisig.com/solana/` | 150 | 485 ms | **60,551 ms** | 60,556 ms | **7 (4.7%)** |
| `api.vultisig.com/coingeicko/…` (**same host**) | 150 | 485 ms | 857 ms | 953 ms | 0 |
| `api.kamino.finance/kvaults/…` | 150 | 97 ms | 485 ms | 1,244 ms | 0 |

Across the whole investigation — two independent probes, **~700 requests — 24 stalls, all 24 on `api.vultisig.com/solana/`**, against **0** on the same host's other routes (150), on `api.kamino.finance` (150) and on a public Solana node, `solana-rpc.publicnode.com` (240).

Other measured properties: not method-specific (tiny `getBalance` and large `simulateTransaction` both stall); a warm keep-alive connection does not dodge it (5.8% vs 3.3% fresh); the duration is a constant, not a distribution (every stall landed in 60,474–60,561 ms, nothing at all between ~0.6 s and ~60.5 s); and it is bursty rather than steady-state. A latency that repeatable is a fixed 60-second timeout upstream, after which a retry succeeds.

**So this is route-specific and server-side.** Not the Cloudflare edge, not the local network, not Solana RPC in general — it is whatever `/solana/` proxies to.

> ⚠️ **The backend half is NOT fixed by this PR.** It needs to go to whoever owns `api.vultisig.com/solana/` with the numbers above. Everything here is the client reducing how many times it draws that ticket, and how much a drawn ticket costs. The 4.7% figure is one afternoon on one network — the *shape* (route-specific, wait-phase only, fixed ~60.5 s, bursty) is solid, the rate should not be read as an SLO.

### Why SOL, and why *often*

Each proxy round trip is an independent chance to draw the stall, and the wrapped-SOL vault made the most of them: `resolveAvailableAmount` short-circuits to the plain token balance for the USDC vaults, but the SOL vault runs `maxNativeDepositLamports`, which builds, validates and simulates a whole probe deposit just to measure what the transaction costs. On top of that the address lookup table was fetched **four times** per deposit, uncached — `KaminoTransactionPreparer.finish` validates as-built and `inject` validates again, and each validate re-fetched the tables, twice per prepare across two prepares.

### Before / after

Counted from the code, for a wrapped-SOL deposit from form open through to a built payload — the span the user waits on before the verify screen:

| | before | after (cold) | after (warm) |
|---|---|---|---|
| `api.vultisig.com/solana/` round trips | **12** | **7** | **4** |
| chance of drawing ≥1 sixty-second stall @ 4.7% | **44%** | 29% | **17.5%** |

"Warm" is a second form open inside the caches' 60 s, or any deposit once the lookup table, the fee sample and the 24 h rent minimum are resident. "Cold" is a fresh session.

And each remaining draw now costs a ~20 s timeout plus one retry instead of a 60 s freeze: `SolanaAPI`/`KaminoAPI` inherited the 60.0 s `TargetType` default, sitting exactly on the boundary of the stall.

### What changed

| | |
|---|---|
| Read the lookup table once per prepare | `finish`'s first validate threads its resolved tables into `inject`'s second. `injectingComputeBudget` provably does not touch `addressTableLookups` — it only appends static keys and instructions — so pass 2 was re-fetching byte-for-byte what pass 1 had just fetched. |
| TTL-cache the lookup table (~60 s) | In `SolanaService`, alongside the existing rent/epoch caches, keyed by **resolved endpoint + table address** so a custom-RPC override can never be served an entry read from another endpoint. |
| Cache the priority fee sample (~60 s) | `getRecentPrioritizationFees` is a network-wide median, not user-specific, already clamped with a fallback. Shared with the withdraw form. |
| Stop the SOL form fetching an unread balance | On the wSOL path the ceiling comes entirely from the probe and the screen reads only `availableAmount` — the price + `getBalance` pair was two round trips whose result was never read. |
| Reuse a hydrated vault | `KaminoEarnViewModel` already calls `fetchVaultInfo` for every enabled vault, and moments later the form the user tapped into refetched the same thing — `KaminoRoute.deposit` carries only the descriptor, so the hydration was discarded at the navigation boundary. A shared `KaminoVaultInfoCache` behind `KaminoService.fetchVaultInfo` keeps it. |
| Per-method timeouts + retry | 20 s for bounded reads, **60 s retained for broadcasts** (aborting one does not undo it, and the recovery path is a fresh signature over fresh bytes) and for genuinely unbounded bulk reads (`getVoteAccounts`, the stake-account `getProgramAccounts` walk, `getTokenAccountsByOwner(_, .programId)`). Any non-proxy target keeps 60 s — the measurement was taken on the proxy, so the mitigation applies to the proxy. Idempotent reads retry once on timeout. |
| The Kamino **exit** path keeps 60 s | `.userPositions` and the `.withdraw` build are the two calls used only to *leave* a position, and neither is on the deposit path the short limit exists to protect. Neither is retried into success by cutting the wait either: a server slow enough to blow 20 s will blow the next 20 s too, so a shorter limit converts "slow" into "cannot withdraw through Vultisig" rather than into a faster answer. The way *in* stays short — a refused deposit build costs an attempt on a form that remains usable, and it is on the path the measurement was taken from. |

### What is deliberately NOT cached

**`simulateTransaction` on the real signed bytes, and the real `POST /ktx/kvault/deposit` build.** These are the safety gate: the simulate proves the exact bytes execute and measures the payer's remaining lamports, and the build is amount-specific and carries a blockhash. Caching either voids the feature's whole safety argument. The four proxy calls left on the warm path are precisely those — two simulates on the reserve probe and two on the real build.

Caching is safe where it *is* applied because every path is fail-closed: a stale lookup table missing appended entries makes the transaction's indexes come up short and the validator refuses (`lookupIndexOutOfRange`); a stale `vaultLookupTable` mismatches the built transaction and refuses (`unexpectedLookupTable`); every money-deciding account is compared against a locally derived value. None can produce a wrong signature.

### Separate fix: the silently dead form

Independent of the proxy, and worth fixing regardless. `KaminoDepositViewModel.onLoad` set `self.error`, but `KaminoDepositScreen` only ever rendered its own `@State private var error`, which is assigned exclusively in `handleVerify()`. So a failed hydration was **never surfaced**: the user got a form whose Continue button is permanently disabled (`isDepositUnavailable` is true when `availableAmount <= 0` or `vaultInfo == nil`), with no message and no retry. `KaminoWithdrawScreen` had the identical shape.

Both screens now bind to the view model's load error and offer a retry. A reload that fails also tears down everything a previous pass derived, so a failed pass cannot leave a stale ceiling or position sitting under the error banner, and `onLoad` guards against overlapping tasks from rapid retry taps.

### Follow-ups this surfaced (not in this PR)

- Report the proxy stall to the backend owners — the actual fix.
- Deferring the SOL reserve probe to the percentage/Max buttons would take the common path lower still, but `availableAmount` also feeds `AmountBalanceValidator` and `isDepositUnavailable`, so the form needs an "unmeasured" state distinct from "zero". Product call.
- `SendCryptoVerifyViewModel.loadGasInfoForSending` unconditionally awaits `updateBalance(for:)`, so the verify screen pays a price + `getBalance` round trip even for a pre-built payload whose balance checks it then skips. Shared Send infrastructure, out of scope here.
- Nothing evicts the cached `KaminoVaultInfo` on an `unexpectedLookupTable` refusal, so a repointed vault fails deterministically for the rest of the TTL. Bounded to a minute by design.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

## Parity

- Android: `linked: vultisig-android#5485` — Kamino Earn is not implemented there yet; the mitigation applies when it is.
- Windows + extension: `linked: vultisig-windows#4541` — same.
- SDK: `n/a` — client caching and timeout policy in the iOS networking layer, no SDK surface.

> The **backend** half is platform-agnostic: every client that reads through `api.vultisig.com/solana/` draws the same stall. That belongs to the API owners, not to a sibling app ticket.

## Checklist

## Additional context

Gate on a dedicated en_US simulator: **`** TEST SUCCEEDED **` — 5975 tests, 11 skipped, 0 failures.** SwiftLint unchanged at 22 pre-existing violations, 0 serious, none in a touched file. The one new string (`kaminoLoadFailed`) is in all **8** shipping locales (`LOCALE_DIRS` is the source of truth and includes `ko`), sorted, entry counts equal at 1,803.

Reviewed commit-by-commit by a second model across the five implementation steps plus a whole-branch pass, then re-reviewed in full over two rounds after the rebase onto #5105 — since the combination of the caching work and the UI work had not been reviewed together before. Those rounds produced two MAJORs, both about the Kamino timeout classification and both fixed and pinned by tests: the position read and then the withdraw build should keep 60 s, which is what settled the rule into "calls used only to leave a position keep 60 s; everything on or shared with the deposit path stays at 20 s". The same pass confirmed the rebase is patch-identical, that the new main-screen banner consumes no cached Kamino data, and that a cached `KaminoVaultInfo` cannot turn zero shares into a position or hide a confirmed one.

- **Agent**: Claude Code
- **Issue**: #5017
- **Confidence**: high
- **Needs human review for**: the per-method timeout classification in `SolanaAPI` — "is this response bounded?" is the rule, and a wrong answer there aborts a legitimate slow read.

## Summary by CodeRabbit

* **New Features**
  * Added clearer Kamino vault loading errors with retry actions across deposit and withdrawal screens.
  * Improved form safeguards by disabling actions when required vault or position data is unavailable.
  * Added short-term caching for Kamino vault details and selected Solana data.

* **Bug Fixes**
  * Improved timeout handling and retry behavior for eligible network requests.
  * Prevented duplicate loading operations and stale form data after failed reloads.
  * Improved transaction validation and lookup-table reuse.

* **Localization**
  * Added Kamino loading-failure messages in supported languages.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaAPI.swift
@@ -109,6 +109,56 @@ struct SolanaAPI: TargetType {
 
     var method: HTTPMethod { .post }
 
+    /// The `TargetType` default, kept for everything the shorter one would not
+    /// be an improvement for.
+    static let defaultTimeout: TimeInterval = 60
+    /// Short enough to sit clear of the proxy's stall boundary. Forty times the
+    /// measured p99 of a healthy request through it. Repo precedent: `CetusAPI`
+    /// 10 s, `StakewizValidatorMetadataProvider` 15 s, `CosmosStakingAPI` 30 s.
+    static let proxyReadTimeout: TimeInterval = 20
+
+    /// Deliberately not one number for everything.
+    ///
+    /// The Vultisig RPC proxy intermittently accepts a request and then parks it
+    /// for a shade over 60 seconds before answering `200` — measured at
+    /// 60.47–60.56 s across two dozen occurrences, a constant rather than a
+    /// distribution, which is the signature of a fixed upstream timeout and a
+    /// retry. The 60 s default sits exactly on that boundary, so the app spent
+    /// the whole minute waiting and then usually got its answer.
+    ///
+    /// Three carve-outs, because the measurement is narrower than "Solana".
+    ///
+    /// **A custom endpoint** is not the proxy and was never measured. A user
+    /// running their own node may reasonably be slower, and shortening their
+    /// timeout would fix nothing they have.
+    ///
+    /// **A broadcast** is the one request whose outcome matters more than its
+    /// latency. Aborting it does not undo it: a node that accepted the
+    /// transaction is indistinguishable from one that never saw it, and the
+    /// recovery path is a fresh ceremony over a fresh blockhash — different
+    /// bytes, which is the one thing that could execute twice. Left long so the
+    /// answer arrives rather than the question being abandoned.
+    ///
+    /// **The bulk reads** legitimately take a long time, and the rule for them
+    /// is whether the response is bounded. `getVoteAccounts` returns the whole
+    /// validator set, `getProgramAccounts` walks the stake program, and
+    /// `getTokenAccountsByOwner` filtered by *program* returns every token
+    /// account a wallet holds — so a slow answer there is the work, not a stall.
+    /// The same call filtered by *mint* returns at most a couple of accounts and
+    /// stays on the short timeout.
+    var timeoutInterval: TimeInterval {
+        guard usesProxyPath else { return Self.defaultTimeout }
+        switch rpcMethod {
+        case .sendTransaction,
+             .getVoteAccounts,
+             .getStakeAccountsByOwner,
+             .getTokenAccountsByOwner(_, .programId):
+            return Self.defaultTimeout
+        default:
+            return Self.proxyReadTimeout
+        }
+    }
+
     var task: HTTPTask {
         // All Solana RPC methods share a JSON-RPC envelope. Params arrays are
         // heterogeneous (String, dict) which Swift's type system can't model
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift
@@ -88,12 +88,19 @@ class SolanaService: SolanaAddressLookupTableFetching, SolanaFinalizedBlockhashP
     /// the blockhash as not yet seen (propagation lag, not true expiry).
     private static let maxBroadcastAttempts = 3
 
+    /// How long a read of an Address Lookup Table stays usable. Injectable for
+    /// the same reason `broadcastRetryBackoff` is: a test has to drive expiry
+    /// without waiting out the real interval.
+    private let addressLookupTableTTL: TimeInterval
+
     init(resolver: RPCEndpointResolving = CustomRPCStore.shared,
          httpClient: HTTPClientProtocol = HTTPClient(),
-         broadcastRetryBackoff: Duration = .seconds(2)) {
+         broadcastRetryBackoff: Duration = .seconds(2),
+         addressLookupTableTTL: TimeInterval = SolanaService.defaultAddressLookupTableTTL) {
         self.resolver = resolver
         self.httpClient = httpClient
         self.broadcastRetryBackoff = broadcastRetryBackoff
+        self.addressLookupTableTTL = addressLookupTableTTL
     }
 
     /// Builds a pure `SolanaAPI` value with the resolved host and proxy-path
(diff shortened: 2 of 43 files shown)


johnnyluo (approved):
LGTM

PR #5126, opened 2026-08-12
fix(thorchain): price bRUNE at RUNE parity instead of a nonexistent pool

### LINKED ISSUE #5101: [Fix] bRUNE shows $0 — priced via nonexistent THOR.BRUNE pool instead of RUNE parity [Wallet]
## Problem
bRUNE displays $0 in the wallet.

## Cause
`ThorChainTokenPriceSource` strips `x/` from `x/brune` and queries pool `THOR.BRUNE` (`TokenPriceSource.swift:175-185`). That pool doesn't exist:

```
GET /thorchain/pool/THOR.BRUNE
→ {"code":2,"message":"asset: THOR.BRUNE is a derived asset"}
```

`getAssetPriceInUSD` returns `0.0` on failure (`ThorchainService.swift:378-380`).

The declared intent is RUNE parity — `TokensStore.swift:3145` sets `priceProviderId: "thorchain"` (RUNE's feed) — but `ThorChainTokenPriceSource` never reads `priceProviderId`.

## Fix
Special-case `x/brune` in `ThorChainTokenPriceSource` to reuse RUNE's price. Android already prices bRUNE at RUNE parity deliberately (`TokenPriceRepository.kt`).

Verify: bRUNE row shows ~RUNE price instead of $0.

gastonm5:
## Description

bRUNE renders **$0** in the wallet. `ThorChainTokenPriceSource` prices a THORChain token by stripping `x/` from its denom and asking `/thorchain/pool/THOR.<SYMBOL>` — but bRUNE is a THORChain *derived* asset and has no pool. Verified live against the node:

```
GET /thorchain/pool/THOR.BRUNE
→ {"code":2,"message":"asset: THOR.BRUNE is a derived asset","details":[]}
```

`getAssetPriceInUSD` reports `0.0` for that, so the rate cached under the denom is zero and the row renders $0.

Rather than hardcode bRUNE, this reads the intent `TokensStore` already declares: **when the pool has no usable price, fall back to the CoinGecko feed the curated entry names via `priceProviderId`.** bRUNE declares `thorchain` — RUNE's own feed — so it lands at RUNE parity with bRUNE named nowhere in the pricing code.

### The pool stays *first*, deliberately

Porting Android's feed-first ordering would have moved tokens that price correctly today. `TCY` declares `priceProviderId: "tcy"` but is **deliberately** priced from the Available `THOR.TCY` pool (commit `67a4f85`, "tcy price from the pool"), and RUJI/sRUJI from `THOR.RUJI`. Pool-first leaves every working price byte-identical and only fills the gap where the pool has nothing to say. Regression tests pin TCY, RUJI and sRUJI to their pool prices and assert the declared feed is never even consulted for them.

### ⚠️ How to reproduce — the issue's premise needs one correction

**A freshly-added bRUNE will not show $0, so a reviewer testing on a clean vault will not reproduce the bug.** `RateProvider.cryptoId(for:)` routes any THORChain coin that *carries* a `priceProviderId` to the provider-id fetch, so a curated bRUNE never reaches the pool lookup at all.

What is actually broken is a `Coin` **persisted before the curated entry landed** (#4799). bRUNE existed on Rujira first, so auto-discovery stored it with ticker `BRUNE` and an empty `priceProviderId` — and `CoinService.addToChain` returns early for an existing coin, so that empty id is never re-synced. Those coins fall to the contract path, hit `THOR.BRUNE`, and are pinned at $0 forever.

To reproduce: a vault that held bRUNE before upgrading past #4799 (or set an existing bRUNE coin's `priceProviderId` to `""`).

This is also why the fix **keys off the contract denom, not the ticker** — the broken coins stored `BRUNE`, not `bRUNE` (the lowercase-preserving guard shipped *with* the curated entry), so a ticker-keyed special case would have missed exactly the coins that are broken.

### Siblings, verified live against `/thorchain/pool/…`

| Denom | Pool asset | Result | Priced today? |
|---|---|---|---|
| `x/brune` | `THOR.BRUNE` | **derived asset** | **$0 — this PR** |
| `x/ruji`, `x/staking-x/ruji` | `THOR.RUJI` | `Available` | ✅ pool price — untouched |
| `x/staking-tcy`, `tcy` | `THOR.TCY` | `Available` | ✅ pool price — untouched |
| `x/nami-…-rcpt`, `x/staking-x/brune` | — | never reaches the strip | ✅ yield NAV path — untouched |
| `thor.lvn`, `thor.rkuji` | `THOR.LVN` / `THOR.RKUJI` | **derived asset** | ⚠️ repaired for free by the same rule |
| `thor.lqdy` | `THOR.LQDY` | **pool doesn't exist** | ❌ still $0 — no feed declared, unchanged |

None of the four assets the issue implied might share the cause actually do. **LVN and RKUJI *are* the identical root cause** (derived, no pool) under the same stale-coin precondition, and the general rule repairs them at **no extra code** — a consequence of reading the declared feed, not scope expansion.

### Deliberately out of scope

- **The real root cause: a stored `Coin.priceProviderId` is never re-synced from its curated `TokensStore` entry.** That is a cross-chain data migration and is why bRUNE, LVN and RKUJI can go stale at all. **Not filed** — flagging it here for a maintainer to raise.
- NAV-exact pricing for sRUJI/sTCY (both currently price at their underlying's value; Android does the liquid-bond NAV for `x/staking-tcy`).
- Mapping secured-asset denoms to their real pool asset (`btc-btc` → `btc.btc`), which Android does and iOS does not.
- Per-currency rates for contract-keyed sources. The fallback emits **USD only**, symmetric with the pool price it replaces: emitting the feed's other currencies would strand them once an asset returned to pool pricing, since the pool path has no per-currency rates to refresh them with. `SolanaTokenPriceSource` has the same USD-only shape — systemic, its own change.

### Implementation notes

- The declared feed is normally already cached (provider-id prices resolve before contract prices, and any vault holding bRUNE also holds native RUNE). A *single-coin* refresh — `BalanceService.updateBalance(for:)` behind a coin-detail screen — carries no such coin, so on a cold cache the feed is fetched once before it is read, mirroring the CACAO pre-fetch `MayaChainTokenPriceSource` already performs. Deduped per refresh so two contracts naming the same feed don't issue the same request twice.
- When there is no declared feed at all (LQDY), the pool's zero is still reported — behaviour unchanged.
- `ThorChainTokenPriceSource` gains injected yield/pool/feed closures defaulting to `ThorchainServiceFactory`, `RateProvider.shared` and `CoinGeckoRates`, mirroring how `SuiTokenPriceSource` is already structured; the `x/` + `STAKING-` sanitisation moves to a static `poolAssetName(for:)` so the (unchanged) mapping can be pinned directly.

Fixes #5101

## Which feature is affected?
- [x] New Chain / Chain related feature  -  Please attach tx link here

> No tx link: this is a THORChain **price-display** fix. It changes no transaction, memo, fee or keysign path — only which source a cached USD rate is read from.

## Parity

- Android: `n/a:` already correct — `TokenPriceRepositoryImpl.refresh` buckets every `priceProviderID`-bearing token to CoinGecko and never asks a pool, and `mapThorPoolAsset("x/brune")` matches no pool so nothing overwrites it. bRUNE prices at RUNE parity there by construction; iOS is the platform catching up, so there is no Android defect to link or file.
- Windows + extension: `n/a:` no separate implementation — both consume `@vultisig/core-chain`, whose token registry carries the same `priceProviderId: 'thorchain'` for `x/brune` (see the SDK verdict). Verified read-only from the SDK registry the app imports, **not** by running the desktop app; if a maintainer knows of a Windows-side pool lookup that bypasses the registry, that would be worth a second look.
- SDK: `n/a:` already correct — `packages/core/chain/coin/knownTokens/cosmos.ts` registers `'x/brune'` with `priceProviderId: 'thorchain'`, and vultisig-sdk#1317's changelog states bRUNE "is auto-discovered as a normal wallet token priced against RUNE (`priceProviderId: 'thorchain'`)". Nothing to fix there.

> The iOS-only defect is the **pool-first contract path**, which no sibling has: the others price straight from the declared feed. This PR brings iOS's fallback in line with that intent without giving up the pool prices iOS deliberately uses for TCY/RUJI.

## Checklist

None. Reproducing the $0 row requires a vault holding a bRUNE coin auto-discovered before #4799 (an empty stored `priceProviderId`), which I could not construct on a simulator without a funded mainnet THORChain vault. The behaviour is pinned by unit tests instead — see below.

## Additional context

**Gate:** `make test` on a dedicated en_US simulator, judged by the log marker:

```
** TEST SUCCEEDED **
Executed 5953 tests, with 11 tests skipped and 0 failures (0 unexpected) in 93.321 seconds
```

SwiftLint clean on both touched files. 11 new tests in `ThorChainTokenPriceSourceTests`: bRUNE → RUNE parity; the pool lookup targets `THOR.BRUNE` before falling back; cold-cache feed fetch; feed fetched once per batch not once per contract; bRUNE still reports 0 when the feed can't be resolved at all; ybRUNE still takes the yield path and the pool is never consulted; TCY/RUJI/sRUJI pool-price regression guards; LQDY still zero; `poolAssetName` mapping; and a routing test asserting against `RateProvider.cryptoId` that only a coin *without* a declared feed reaches this source.

**Review:** three read-only `codex exec` passes (two during implementation, one whole-branch before pushing). Four findings across the first two rounds, all fixed — the cold-cache single-coin refresh, the stranded non-USD rates, the per-contract duplicate feed fetch, and a test that bypassed production routing. The final whole-branch pass returned no findings.

- **Agent**: Claude Code
- **Issue**: #5101
- **Confidence**: high on the fix and the blast radius (pool-priced assets are provably untouched, and the live pool probes are in the table above); medium on the *reproduction story* — the stale-`priceProviderId` explanation is derived by reading the code paths, not by observing a real affected vault.
- **Needs human review for**: (1) confirmation from someone with an affected vault that bRUNE now shows ~RUNE instead of $0; (2) whether the stale-`priceProviderId` re-sync deserves its own issue — I was not authorized to file it; (3) the deliberate USD-only choice, if anyone wants non-USD wallets covered sooner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved THORChain token price resolution with fallback pricing when pool data is unavailable or invalid.
  * Preserved valid zero-price results and improved handling for derived assets and yield tokens.
  * Added more reliable pricing for bRUNE, RUNE, TCY, and RUJI assets.

* **Tests**
  * Expanded coverage for token routing, fallback behavior, asset mapping, and price-fetching scenarios.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/TokenPriceSource.swift
+++ b/VultisigApp/VultisigApp/Core/Services/TokenPriceSource.swift
@@ -237,37 +237,150 @@ struct MayaChainTokenPriceSource: TokenPriceSource {
 /// Prices THORChain assets from THORChain pools, with special-casing for the
 /// yield tokens (yRUNE / yTCY / ybRUNE) and sanitisation of `x/` and `STAKING-`
 /// prefixes before the pool lookup.
+///
+/// A pool lookup only answers for an asset that actually has an L1 pool.
+/// THORChain *derived* assets have none — `/thorchain/pool/THOR.BRUNE` replies
+/// `"asset: THOR.BRUNE is a derived asset"` and `getAssetPriceInUSD` reports
+/// `0.0` — so when the pool has no price this falls back to the CoinGecko feed
+/// the token's curated `TokensStore` entry declares via `priceProviderId`.
+/// bRUNE names RUNE's own feed, which is what prices it at RUNE parity, the
+/// same result vultisig-android reaches by pricing every `priceProviderID`
+/// -bearing token from CoinGecko and never asking the pool. The pool stays
+/// first so every asset that does have one (TCY, RUJI and their staking
+/// receipts) keeps pricing exactly as it does today.
+///
+/// This path is only reached for a coin whose stored `priceProviderId` is
+/// empty: `RateProvider.cryptoId(for:)` routes a coin that carries one to the
+/// provider-id price fetch instead. A coin auto-discovered before its curated
+/// entry existed was persisted without one and is never re-synced, which is how
+/// a held bRUNE ends up priced from a pool that cannot exist.
 struct ThorChainTokenPriceSource: TokenPriceSource {
+    /// NAV-derived USD price for a yield token's contract denom.
+    typealias YieldPrice = (Chain, String) async -> Double?
+    /// USD price for a fully qualified THORChain pool asset (e.g. `THOR.TCY`).
+    typealias PoolPrice = (Chain, String) async -> Double
+    /// Cached USD rate for a curated token's declared CoinGecko feed.
+    typealias ProviderRate = (CoinMeta) -> Double?
+    /// Fetches and persists one CoinGecko provider id's rates.
+    typealias ProviderFeedFetch = (String) async -> Void
+
     let chain: Chain
+    private let yieldPrice: YieldPrice
+    private let poolPrice: PoolPrice
+    private let providerRate: ProviderRate
+    private let fetchProviderFeed: ProviderFeedFetch
+    private let logger = Log.chain.service
+
+    init(
+        chain: Chain,
+        yieldPrice: @escaping YieldPrice = { chain, contract in
+            await ThorchainServiceFactory.getService(for: chain).fetchYieldTokenPrice(for: contract)
+        },
+        poolPrice: @escaping PoolPrice = { chain, assetName in
+            await ThorchainServiceFactory.getService(for: chain).getAssetPriceInUSD(assetName: assetName)
+        },
+        providerRate: @escaping ProviderRate = { coin in
+            RateProvider.shared.rate(for: coin, currency: .USD)?.value
+        },
+        fetchProviderFeed: @escaping ProviderFeedFetch = { providerId in
+            _ = try? await CoinGeckoRates.fetchAndSaveByIds(
+                [providerId],
+                httpClient: HTTPClient(),
+                logger: Log.chain.service
+            )
+        }
+    ) {
+        self.chain = chain
+        self.yieldPrice = yieldPrice
+        self.poolPrice = poolPrice
+        self.providerRate = providerRate
+        self.fetchProviderFeed = fetchProviderFeed
+    }
 
     func prices(contracts: [String], coins _: [CoinMeta]) async throws -> [Rate] {
-        let thorService = ThorchainServiceFactory.getService(for: chain)
+        let yieldTokens = TokensStore.TokenSelectionAssets
+            .filter { $0.chain == chain && ($0.ticker == "yRUNE" || $0.ticker == "yTCY" || $0.ticker == "ybRUNE") }
+            .map { $0.contractAddress }
+
+        // Two contracts can name the same declared feed (a stale TCY and sTCY both
+        // name `tcy`), so remember which feeds this batch already tried to fetch
+        // and don't issue the same request twice. A later refresh retries.
+        var attemptedFeeds: Set<String> = []
         var rates: [Rate] = []
         for contract in contracts {
-
-            let yieldTokens = TokensStore.TokenSelectionAssets.filter({ $0.chain == chain && ( $0.ticker == "yRUNE" || $0.ticker == "yTCY" || $0.ticker == "ybRUNE") }).map({$0.contractAddress})
-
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #5099, opened 2026-08-11
Select route offers THORChain/Maya routes below their recommended minimum


aminsato:
## Summary

The Select route picker (#4546 / PR #4570) lists THORChain/Maya candidate routes even when the swap amount is **below the node's `recommended_min_amount_in`** for that route. A user can manually pick such a route; the node's own numbers say the amount is under the recommended floor, so the output is visibly eaten by fees (and small swaps risk refund-level outcomes).

This is exactly the case in the PR's own reference scenario: for a ~$9.20 BTC→ETH swap, the picker shows a Maya protocol row quoting **$9.21 in → $7.90 out** (Fee $1.40, ~15% of the swap) — a route Mayanode itself recommends against at that size.

## Repro

1. Swap BTC → ETH with a small amount, e.g. 0.00014844 BTC (~$9.50).
2. Open Advanced swap → Select route.
3. A Maya protocol row is listed. The same quote request against Mayanode returns a minimum above the amount:

```
GET /mayachain/quote/swap?from_asset=BTC.BTC&to_asset=ETH.ETH&amount=14844&destination=<redacted>&streaming_interval=3&affiliate=va&affiliate_bps=30

"recommended_min_amount_in": "24068",   // 0.00024068 BTC ≈ $15.40 — above the 14844-sat amount
"fees": { "outbound": "75000", "affiliate": "1519", "total_bps": 1312 }
```

Note the minimum scales **inversely** with affiliate bps (the affiliate cut must clear outbound dust), so tier-discounted vaults (lower bps) see an even higher floor — 24068 sats at 30 bps vs ~14441 at 50 bps at the time of capture.

## Platform divergence

Android (route picker in vultisig-android#5580) drops a THOR/Maya candidate whose amount is below `recommended_min_amount_in` — the pre-existing minimum guard applies per candidate — so the same swap shows only SwapKit + THORChain there. The two platforms currently disagree on whether a below-minimum route is offerable.

## Possible resolutions (needs a product decision)

- Filter below-minimum routes out of the picker (Android's current behavior), or
- Keep listing them but mark the row (e.g. "below recommended minimum") and/or block selection.

Either way the platforms should converge on one rule.

gastonm5:
Closing as resolved, with a correction to the original diagnosis.

The picker was not retaining routes below `recommended_min_amount_in`: `SwapService` already rejects those provider quotes before ranking, for both THORChain and MayaChain.

[...]

Issue #4589, opened 2026-06-16
VULT discount tiers: migrate from holding VULT → staking VULT (mechanic + copy)


realpaaao:
### Platform: iOS

**Mechanic:** `VultTierService` (the resolver behind `Features/Swap/Interactor/SwapDiscountTierResolving.swift`) resolves the tier from **VULT balance + Thorguard NFT eth_call**. Point the balance read at sVULT `balanceOf`. The discount bps then flows through `SwapService.vultTierDiscount` as today.

**Copy:** `VultisigApp/Core/Localizables/en.lproj/Localizable.strings` keys `vultDiscountTiersDescription` ("Hold $VULT to unlock lower trading fees."), `unlockXTierDescription` ("By holding %@ $VULT…"), `unlockUltimateTierDescription` ("By holding %@ $VULT…") — and the same keys in **every other `*.lproj`** locale. Settings entry: `SettingsOption.vultDiscountTiers`.

Follow-up to staking issue #4588. Owner: Gaston.

---

## Migrate VULT discount tiers: holding VULT → staking VULT

Follow-up to the VULT staking work (iOS staking issue #4588). Now that staking exists, the **discount-tier / benefits system must switch its basis from holding VULT (raw ERC-20 balance) to staked VULT (sVULT balance)**, and all copy must change from "hold" to "stake". This covers **both the mechanic and the copy** — they must ship together so the numbers and the words agree.

### Mechanic change (the important part)
The tier resolver currently reads the vault's **VULT ERC-20 balance** on Ethereum. Change the source to the **staked sVULT balance**:
- Read `balanceOf(user)` on **sVULT** `0x11113d7311FB8584a6e82BB126aA11D92e5fB39B` (Ethereum) instead of the VULT token balance.
- sVULT is a **1:1 wrapper** of VULT, so the existing numeric thresholds — **1,500 / 3,000 / 7,500 / 15,000 / 100,000 / 1,000,000** (Bronze→Ultimate) — stay identical. Only the balance source moves.
- Everything downstream (discount bps → swap affiliate fee) is unchanged.

### ⚠️ Decision to confirm with product (flagged, not silently decided)
Per the instruction this is **staked-only**: a vault's tier is determined by sVULT held, not VULT held. That means **existing VULT holders who have not staked lose their tier until they stake** — this needs a comms/migration plan and possibly a grace window. Flagging here; default implementation = staked-only.

### NFT modifier — out of scope
[...]

Issue #3451, opened 2025-11-20
[ADD] Update Discount tiers


realpaaao:
## Reason 
To better reflect current and future price and discounts for holding $VULT in the Vaults we will adjust and introduce more tiers and gamification

## To-Dos

- [ ] Adjust and add the follwing tiers, see [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=51817-156821&t=HVz437V4DvFMRj9E-4) :
Bronze: 1,500 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)= 5 bps

Silver: 3,000 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)= 10 bps

Gold: 7,500 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)= 20 bps

Platinum: 15,000 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)= 25 bps

Diamond: 100,000 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)=35 bps

Ultimate: 1,000,000 [$VULT](https://x.com/search?q=%24VULT&src=cashtag_click)= Complete Vultisig fee waive, will also clear referral.

- [ ] Add the logic that if at least one [Thorguard](https://etherscan.io/token/0xa98b29a8f5a247802149c268ecf860b8308b7291) is held in the wallet, the vault gets upgraded one tier. But only until Platinum, if user is already Platinum Tier, the Thorguard will not have an effect.

PR #5133, opened 2026-08-13
fix(swap): stop a failed VULT tier read dropping the discount for the whole session

### LINKED ISSUE #5131: [Fix] A failed VULT balance read silently drops the tier discount for the entire app session
## Summary

A vault's VULT discount tier is resolved **once per app launch** and cached for the whole process. If that one resolution fails — or lands before the VULT balance has been fetched — the result is cached as **"no tier"** and never retried. Every swap for the rest of that session then sends the **undiscounted** `affiliate_bps`.

A Gold-tier holder (7,500 VULT, 20 bps discount) is charged **50 bps instead of 30** on every swap in that session, with no tier badge shown. The user has no way to recover except relaunching the app.

This affects **every swap provider** — `vultTierDiscount` is passed to THORChain, MayaChain, 1inch, LI.FI, KyberSwap, Jupiter and SwapKit alike (`SwapService.swift:310-400`) — so it is not specific to one route.

## Mechanism

1. `DefaultSwapInteractor.fetchQuote` (`:53`) and the screen warm-up (`:218`) call `VultTierService.resolveTierForSession(for:)`.
2. `SessionTierCache.resolve` (`VultTierService.swift:195-211`) caches the resolved `VultDiscountTier?` **for the process lifetime**. Its `Box` type exists specifically so a `nil` result is remembered as a final answer and never re-resolved — the file says so: *"a VULT/Thorguard balance change for the same vault isn't reflected until the next app launch."*
3. `fetchDiscountTier` → `fetchVultBalance` (`:126-144`). If the VULT balance reads `0`, no tier threshold is met and the tier is `nil`.
4. That `nil` is pinned. Every later quote reads it as "no discount".

Two things make step 3 more likely than it looks:

- **`fetchVultBalance` stamps its 3-minute freshness `CacheEntry` unconditionally** (`:136-138`), after `BalanceService.shared.updateBalance(for:)` runs — whether or not it succeeded. So one failed refresh tells the next 3 minutes of calls "we just fetched", and they read the stale `0`.
- `BalanceService.updateBalance(for coin:) async` (`BalanceService.swift:485`) returns `Void` and swallows its own errors, so **the caller cannot distinguish a failed fetch from a genuine zero balance**.

There is nowhere in the current design that separates *"this wallet holds no VULT"* from *"we could not find out"*. Both become "no discount, permanently".

## Why the discount matters beyond the fee

The `affiliate_bps` we send also moves THORChain/Mayanode's `recommended_min_amount_in`. On Mayanode that floor is an **affiliate-dust floor inversely proportional to `affiliate_bps`** — measured across a 5→50 bps sweep, `min_in × bps / 10000 = 72.71`, constant to two decimals. So a *higher* bps produces a *lower* floor.

Measured for BTC→ETH at 14844 sat (0.00014844 BTC):

| `affiliate_bps` sent | Maya `recommended_min_amount_in` | 14844 below floor? |
|---|---|---|
| 30 (Gold tier, correct) | 24236 | yes — candidate dropped |
| 50 (tier lost, this bug) | 14541 | no — candidate offered |

That is the visible symptom reported in #5099: a MAYAChain route quoting a **14.97%** fee is offered on iOS where Android drops it, for the same wallet. The route filter is working correctly on both platforms; iOS was simply sending the wrong `affiliate_bps`.

## Android does not have this bug

`GetDiscountBpsUseCase.invoke(vaultId, provider)` runs **fresh on every quote fetch** (`SwapQuotePipeline.kt:226`) — no session cache. If the balance is not ready on the first attempt, the next quote a few seconds later resolves correctly. Android self-heals; iOS cannot.

## Expected behavior

- A tier resolution that could not be determined is **never cached as "no tier"**.
- A late-arriving VULT balance is reflected on the next quote without relaunching the app.
- The 3-minute freshness stamp is only written when a refresh actually landed.

## Notes

- The tier **values** are correct and are not the problem — they match spec #3451 (Bronze 1,500 = 5 bps / Silver 3,000 = 10 / Gold 7,500 = 20 / Platinum 15,000 = 25 / Diamond 100,000 = 35 / Ultimate 1,000,000 = full waive), as do the Thorguard +1-tier boost and its Platinum cap.
- Reading the plain VULT ERC-20 balance is also correct: #4589 (move tiers to staked sVULT) was closed as won't-do.
- The per-session cache itself was requested deliberately in #4457, to keep the Thorguard `eth_call` off the per-quote critical path. That intent is sound; caching a *failed* resolution is not. A fix should preserve the optimization.

Related: #5099 (the symptom), #4457 (introduced the cache), #3451 (tier spec), #4589 (sVULT migration, won't-do).

gastonm5:
## Description

Fixes #5131

A vault's VULT discount tier was resolved **once per app launch** and cached for the whole process — and a resolution that *failed* was cached as "no tier" and never retried. Every swap for the rest of that session then sent the undiscounted `affiliate_bps`: a Gold-tier holder charged **50 bps instead of 30**, with no tier badge, recoverable only by relaunching.

`SessionTierCache`'s `Box` type existed specifically so a `nil` result was remembered as final. The file said so: *"a VULT/Thorguard balance change for the same vault isn't reflected until the next app launch."* Nothing anywhere distinguished *"this wallet holds no VULT"* from *"we could not find out"*.

### Why it looked like a route-picker bug

The `affiliate_bps` we send moves Mayanode's `recommended_min_amount_in`, which is an **affiliate-dust floor inversely proportional to that bps** — measured over a 5→50 bps sweep, `min_in × bps / 10000 = 72.71`, constant to two decimals. Higher bps ⇒ lower floor. For the swap in #5099 (BTC→ETH, 14844 sat):

| `affiliate_bps` sent | Maya `recommended_min_amount_in` | below floor? |
|---|---|---|
| 30 (Gold, correct) | 24236 | yes — candidate dropped |
| 50 (tier lost) | 14541 | no — candidate offered |

So #5099's "iOS offers a 14.97% Maya route that Android drops" was this bug, seen from the far end. The below-minimum filter is correct on both platforms and is **not touched here**.

## The fix

**Cache the thing that can't change, not the answer that can.** #4457 asked for the session cache to keep the Thorguard `eth_call` off the per-quote path; that intent is preserved.

- `SessionTierCache` (keyed `VultDiscountTier?`) → `ThorguardOwnershipCache` (keyed `Bool`, single-flight, **never stores an undetermined answer**). `checkThorguardBalance` returns `Bool?`. The tier is recomputed per quote from the in-memory balance, so a late balance is picked up by the next quote — the same self-healing Android gets by recomputing.
- `BalanceService.updateBalance(for:)` now reports whether a refresh landed, so the 3-minute freshness stamp is only written on a real fetch. It previously stamped unconditionally, so one failed refresh suppressed the next three minutes of retries.
- VULT is matched by **contract address** (`0xb788144DF611029C60b859DF47e79B7726C4DEBa`), not by ticker — an impostor ERC-20 tickered "VULT" could previously grant a tier.

The NFT cache is scoped to the quote path only, so the $VULT Tiers screen and `TierGate` keep their live `eth_call` — buying a Thorguard NFT still takes effect without relaunching.

**Not changed:** the tier thresholds and bps values (they match spec #3451), the balance source (#4589's sVULT migration was closed won't-do), and the below-minimum guard.

## Which feature is affected?
- [ ] Sending

**No swap tx link — deliberate, not an omission.** This changes which `affiliate_bps` value is sent, not how any payload is built, signed or broadcast. Reproducing it on-chain would mean signing a real swap from a tier-holding vault twice; the evidence is the quote-fetch measurement above, taken against live Mayanode/THORNode.

## Parity

- Android: `n/a: no equivalent defect` — `GetDiscountBpsUseCase.invoke(vaultId, provider)` runs fresh on every quote fetch (`SwapQuotePipeline.kt:226`), so there is no session cache to pin a failed resolution. Verified against `vultisig-android@main`.
- Windows + extension: `first: VULT tier discount resolution` — a code search for the discount surface returned no equivalent resolver; if one exists it was not audited here.
- SDK: `first: clients/cli/src/commands/discount.ts` — a discount surface exists; its caching behaviour was not audited in this PR.

## Checklist

`make test`: `** TEST SUCCEEDED **` — **6069 tests, 11 skipped, 0 failures**. SwiftLint 26, identical to `origin/main` (zero introduced).

## Additional context

Cross-model review, 3 Codex passes. The whole-branch pass caught what both per-commit passes cleared: the first cut kept the balance refresh *awaited*, so every quote past the 3-minute expiry blocked on a price fetch plus an Ethereum balance RPC **with the loading spinner held**. The shipped version answers from the already-landed balance and refreshes staleness in the background; only a vault with no landed balance waits. It also caught that annotating `resolveTierForSession` `@MainActor` bought nothing, because the nonisolated `async` helpers still read the SwiftData `@Model` vault off the main actor.

**Deferred:** full dependency injection into `VultTierService` so the real resolver is under test. It reaches `BalanceService.shared`, `EvmService`, `CoinService.addToChain` and a process-global `@AppStorage` stamp, so a test around it would be suite-order-dependent across 6069 tests — a worse outcome than the gap.

**Separately flagged, not fixed:** spec #3451 says the Ultimate tier "will also clear referral", but `ThorchainService.affiliateParams` sends `"10/0"`, so an Ultimate-tier referred swap still pays the referrer 10 bps. Same on Android. Cross-platform fee policy — needs product sign-off, so it is deliberately out of scope here.

## Agent Metadata
- **Agent**: Claude Code
- **Issue**: #5131
- **Needs human review for**: the `BalanceService.updateBalance` signature change (it ripples through the protocol and three test doubles), and on-device confirmation that a tier-holding vault shows its badge and sends the discounted bps after a cold start on a flaky connection.

## Summary by CodeRabbit

* **New Features**
  * VULT discount tiers now use the latest available balance when determining eligibility.
  * Thorguard ownership checks are cached during a session for faster quote updates.
  * Balance refreshes are coordinated to avoid duplicate requests and can retry when results are unavailable.

* **Bug Fixes**
  * Improved handling of stale or failed balance updates while preserving the most recent known balance.
  * More accurate tier boosts and affiliate discounts based on token identity and ownership status.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Services/BalanceService.swift
+++ b/VultisigApp/VultisigApp/Core/Services/BalanceService.swift
@@ -481,8 +481,18 @@ class BalanceService {
         try Storage.shared.save()
     }
 
+    /// Best-effort balance refresh. Returns whether a live **raw balance**
+    /// actually landed, so a caller that stamps its own freshness cache can tell
+    /// a failed fetch from a genuine zero — the two are otherwise
+    /// indistinguishable here, because a failure leaves the last known balance in
+    /// place rather than zeroing it. Only the raw balance counts: a price fetch,
+    /// a persistence error, or a failed auxiliary read that happens *after* the
+    /// balance (RUNE's bonded nodes, staked balances) leaves a live number on the
+    /// coin, and reporting those as a failure would make the caller refetch a
+    /// balance it already has.
     @MainActor
-    func updateBalance(for coin: Coin) async {
+    @discardableResult
+    func updateBalance(for coin: Coin) async -> Bool {
         // Fetch price
         do {
             try await cryptoPriceService.fetchPrice(coin: coin)
@@ -522,6 +532,8 @@ class BalanceService {
                 logger.warning("Update Balance error: \(error.localizedDescription)")
             }
         }
+
+        return update.rawBalance != nil
     }
 
     /// Fail-closed spendable-balance refresh: fetches the live raw balance and
--- a/VultisigApp/VultisigApp/Core/Services/BalanceServiceProtocol.swift
+++ b/VultisigApp/VultisigApp/Core/Services/BalanceServiceProtocol.swift
@@ -10,7 +10,11 @@
 import Foundation
 
 protocol BalanceServiceProtocol {
-    func updateBalance(for coin: Coin) async
+    /// Best-effort refresh. Returns whether a live balance actually landed, so a
+    /// caller stamping its own freshness cache can tell a failed fetch from a
+    /// genuine zero. Discardable — most callers only want the side effect.
+    @discardableResult
+    func updateBalance(for coin: Coin) async -> Bool
     /// Fail-closed refresh — throws if the live balance fetch fails. See
     /// `BalanceService.refreshSpendableBalanceOrThrow`.
     func refreshSpendableBalanceOrThrow(for coin: Coin) async throws
--- a/VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiChain/DefiChainMainScreen.swift
@@ -517,7 +517,7 @@ private extension DefiChainMainScreen {
             owner: nativeCoin.address,
             decimals: nativeCoin.decimals
         )
-        async let balance: Void = BalanceService.shared.updateBalance(for: nativeCoin)
+        async let balance: Bool = BalanceService.shared.updateBalance(for: nativeCoin)
         _ = await (rows, balance)
     }
 
--- a/VultisigApp/VultisigApp/Features/Send/Views/Screens/SendDetailsScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Send/Views/Screens/SendDetailsScreen.swift
@@ -355,7 +355,7 @@ struct SendDetailsScreen: View {
     }
 
     private func onRefresh() async {
-        async let bal: Void = BalanceService.shared.updateBalance(for: viewModel.coin)
+        async let bal: Bool = BalanceService.shared.updateBalance(for: viewModel.coin)
         async let pendingCheck: Void = viewModel.forceCheckPendingTransactions()
         _ = await (bal, pendingCheck)
     }
--- a/VultisigApp/VultisigApp/Features/Swap/Interactor/DefaultSwapInteractor.swift
+++ b/VultisigApp/VultisigApp/Features/Swap/Interactor/DefaultSwapInteractor.swift
@@ -45,11 +45,12 @@ struct DefaultSwapInteractor: SwapInteractor {
             throw SwapCryptoLogic.Errors.sameAsset
         }
 
-        // Read the per-session cached tier. The first call resolves it once
-        // (VULT balance + Thorguard NFT eth_call) and caches it for the wallet;
-        // every later quote reads the cached value, keeping the Thorguard
-        // eth_call off the per-quote critical path. The screen warms this cache
-        // on load, so by the time quotes run it's usually already populated.
+        // Resolve the tier fresh for this quote. The VULT balance is re-read
+        // every time, so a balance that arrives late — or a first read that
+        // failed — is reflected in the very next quote instead of costing the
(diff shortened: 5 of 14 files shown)


johnnyluo (approved):
LGTM

PR #5136, opened 2026-08-13
feat(defi): one verify-hero vocabulary for every function transaction

gastonm5:
## Description

**Stacked on #5089 — please review that one first.** Base is `fix/tcy-withdraw-5074`, not `main`.

#5089 built the mechanism that lets a transaction describe itself on the screens around signing, and instantiated it exactly once, for staked-TCY withdrawal. This generalises the verb across every DeFi function transaction, and fixes the hero title's alignment.

No linked issue — this came out of review feedback on #5089.

### Why `TCYUnstakePresentation` could not generalise

It conflated two independent concerns: **the verb** ("You're sending" is wrong for a withdrawal) and **the amount** (a memo-only `MsgDeposit` has `amount == "0"`, so the figure must be carried out-of-band). Only TCY/CACAO unstake need both — a bond or an addLP has a perfectly good amount and needs only the verb. Since the trigger was an *amount* field (`withdrawDisplayAmount`), an operation with a real amount could never opt in.

This splits them. `withdrawDisplayAmount` is untouched and keeps solving the amount problem; a new optional `FunctionTransactionKind` carries the verb. TCY stops being a special case — it becomes `.unstake` that also sets `withdrawDisplayAmount`.

## The builder audit is the substance

There are **28** `TransactionBuilder` conformers. **21 are stamped; 6 deliberately are not.** The rule applied per builder: *stamping must not change what number is announced, and must never let a placeholder become the announced figure.* The verb changes; the figure stays true.

**Three builders would have announced a zero under a correct verb**, and now carry a real figure:

- `UnbondTransactionBuilder` — `amount` is `"0"`; the figure lived only in the `UNBOND:<node>:<baseUnits>` memo.
- `CacaoUnstakeTransactionBuilder` — the TCY defect repeating on MAYAChain. `POOL-:<bps>` is memo-only; now takes `stakedAmount` from the form and quantises by bps exactly as TCY does.
- `SolanaUnstakeTransactionBuilder` — a deactivate carries no lamports. Now takes the row's **delegated** stake, deliberately not total lamports: the rent-exempt reserve was never staked. (The withdraw builder quotes the total, because a withdraw takes the reserve too.)

**Six left unstamped rather than announce a false figure**, each with reasoning at the site:

| Builder | Why |
|---|---|
| `CancelLimitOrder` | Has its own hero; the generic one would answer with the L1 route's donated dust |
| `BondMaya` | `amount` is a fixed 1-CACAO carrier; what is bonded is LP units. "You're bonding 1 CACAO" would be false |
| `RemoveLP` | `amount` is the 0.02 RUNE attach dust, not the liquidity removed |
| `TonUnstake` | `amount` is the 0.2 TON withdrawal signal; quoting the staked balance instead would erase the only place that charge is shown |
| `SolanaWithdraw` | Closes the whole stake account — Solana has no rewards-claim operation. Flagged rather than inventing a 13th `.withdraw` verb |
| `Mint` | Carries the coin *spent*, not the yCoin minted |

Deviation from the plan's table: **RUJILiquidBond/Unbond are `.stake`/`.unstake`**, not `.bond`/`.unbond` — both RUJI positions are reached from the same Stake sheet, and "bonding" here means a THORChain node bond.

## Alignment

`HeroContentView.send` centred the title while the `coinRow` beneath it was `.leading`; same in `.swap`. Both titles are now `.leading`. `.title` stays centred — it is a standalone caption block with no row to align against.

## Two things reviewers should know

**The Cosmos verbs are correct but inert.** `FunctionCallVerifyScreen` routes `cosmosStakingPayload != nil` to `CosmosStakingVerifySummaryView`, which renders its own headline and never consults the resolver — so a Cosmos delegate does *not* say "You're sending" today. Unifying that screen would also change its shipped copy ("You're staking" → "You're delegating"), which is a product call rather than a refactor. Solana staking *does* take the generic path, so its verbs are live.

**TCY's copy changed.** `tcyUnstakeVerifyTitle` ("You are withdrawing") is replaced by the shared "You're unstaking"; the now-unused key is deleted from all 8 locales. **The figure is unchanged** — all seven existing TCY tests pass with their amount assertions untouched.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

Presentation only. No signing, payload-building or broadcast path is touched; what is *announced* changes, what is *built* does not.

## Checklist

`make test`: `** TEST SUCCEEDED **` — **5365 tests, 1 skipped, 0 failures**. (Lower than a `main`-based branch because the base forked from `main` on 2026-08-07.) SwiftLint 22 at the base and 22 on the branch — zero introduced. **10** new locale keys, not 12: `youreStaking`/`youreUnstaking` already ship. All 8 files at 1745 entries, Korean included.

## Screenshots — wants a visual check

No screenshots attached; the heroes need seeded DeFi positions this environment does not have. Three surfaces change visually and deserve eyes:

- `.send` left-alignment now also reaches the **co-signer jetton hero** and the **Blockaid simulation hero**
- `.swap` left-alignment on `SwapVerifyScreen`
- AddLP's header loses its `→ <pool> LP` suffix (the pool is still shown in the rows below)

## Additional context

Cross-model review. The whole-branch pass caught a HIGH both per-commit passes cleared: the yVault **mint naming the coin it spends as the coin it mints**. An earlier pass caught that RUJI rewards are paid in **USDC** while the builder's coin is RUJI — so the verb is kept and the figure dropped, rather than rendering "claiming 12.34 RUJI" over a 12.34 USDC claim.

**Known gap, deferred with reason — the recommended next follow-up.** Four stamped builders legitimately have no figure (share-based redemptions, the Coin-less Cosmos claim). A zero guard suppresses the hero for them so they never announce "You're unstaking 0" — but suppression means they fall back to the screen's generic header, which still reads **"You're sending 0 X"**. The proper fix is a verb-only `.title` hero; it was deferred because it would change TCY auto-compound's behaviour, which #5089 decided and this PR was scoped to preserve.

One review finding was **rejected with reason**: that four verbs are production-unreachable. Recorded in the enum's doc comment rather than removed.

## Agent Metadata
- **Agent**: Claude Code
- **Issue**: none — follow-up to #5089
- **Needs human review for**: the TCY copy change ("You are withdrawing" → "You're unstaking"), the three visual surfaces above, and whether the deferred verb-only hero should land before or after this merges.


gastonm5:
Closing — superseded by the decoder in #5176, which turns out to cover the defect this stack was opened for on its own.

The question that settled it was what the reported bug actually needs. The decoder is wired into `JoinKeysignViewModel` and `KeysignViewModel`, and nowhere else: the **co-signer**. That is exactly where "You're sending 0 TCY" was seen, so #5176 fixes it standalone.

This PR was the **initiator's** verify screen — `FunctionTransactionKind`, its eighteen builder declarations, and `FunctionTransactionPresentation`. That is a real improvement, but it buys no part of the reported defect, and it is the largest and most invasive layer in the stack. Dropping it returns that screen to what ships today: the same "no improvement, never wrong" trade already accepted for the chains the decoder does not read.

What stays, and why it is not this PR:

- `TransactionHeroResolver` and `TCYUnstakePresentation` belong to #5089 — both are absent from `main` and from #5176. The TCY unstake hero and its provider registry are unaffected by this closure.
- `TCYUnstakePresentation` deliberately does **not** move into the decoder. `tcy-:5006` commits to a *fraction*, and nothing signed names the position, so a decoder can only ever render "50.06% of your staked TCY". The initiator reads `withdrawDisplayAmount` — the quantised figure its own form resolved — and renders the real amount. Moving it would put a percentage on the one screen that knows the number.

The stack is now `main → #5176 → #5089 → #5157`, rebased, with the full local gate green (6295 tests) and #5176 green on CI.

Not deleting the branch: `feat/verify-hero-verbs` still holds the eighteen declarations and `DeclaredKindRedundancyTests`, which is the record of which ones the decoder makes derivable. If the initiator side is picked up again, that is the starting point — and it should come back much smaller, as providers on #5089's existing registry rather than a declaration on every builder.

Issue #5132, opened 2026-08-13
[Fix] sTCY/yTCY default the Functions screen to a Custom form whose asset picker excludes them


gastonm5:
## Summary

On THORChain, `FunctionCallType.getDefault(for:)` routes **any** coin whose ticker *contains* "TCY" to the `.custom` function. `FunctionCallCustom.loadTokens` then builds the Custom form's asset picker with an *exact* match. The two disagree, so **sTCY and yTCY default the screen to a form whose asset picker cannot offer them.**

## The mismatch

`FunctionCallTypeEnum.swift:105-111`:

```swift
static func getDefault(for coin: Coin) -> FunctionCallType {
    switch coin.chain {
    case .thorChain:
        if coin.ticker.contains("TCY") {   // matches TCY, sTCY, yTCY
            return .custom
        }
        return .rebond
```

`FunctionCallCustom.swift:42-53`:

```swift
case .thorChain:
    let chainCoins = vault.coins.filter { $0.chain == .thorChain }
    for coin in chainCoins {
        let ticker = coin.ticker.uppercased()
        if ticker == "RUNE" || ticker == "RUJI" || ticker == "TCY" {   // exact
            tokens.append(.init(value: ticker))
        }
    }
    if tokens.isEmpty {
        tokens.append(.init(value: "RUNE"))
    }
```

`sTCY` uppercases to `STCY` and `yTCY` to `YTCY`; neither equals `TCY`. Both tickers ship in `TokensStore` (`:3195`, `:3215`).

## Result

Opening the Functions screen with sTCY or yTCY as the selected THORChain coin:

1. defaults to the Custom form **because** the coin is TCY-family, then
2. presents an asset picker containing only whichever of RUNE / RUJI / TCY the vault holds — or a bare `["RUNE"]` fallback if it holds none — so `preSelectToken(matching: coin)` has nothing to match.

The user lands on a form chosen on account of their asset, which then cannot represent that asset.

## Expected behavior

The two checks should agree. Either sTCY/yTCY are TCY-family for both purposes (default *and* asset list), or for neither. Whichever way it resolves, one predicate should decide it rather than two that disagree.

## Notes

- This is **not** the "screen opens on a function the dropdown does not offer" defect: both `.custom` and `.rebond` are in the THORChain function list (`FunctionCallTypeEnum.swift:63-72`), so the *function* dropdown is fine. It is the *asset* picker inside the resulting form that cannot represent the coin.
[...]

PR #5154, opened 2026-08-13
refactor(functions): drop the unreachable IBC arms and fix the stranded Function defaults

### LINKED ISSUE #5138: [Functions] Remove the unreachable Noble/Akash IBC arms and the broken Osmosis default
## Summary

Three reachability defects in `FunctionCallTypeEnum.swift`, all of which make the migration inventory lie about what actually ships.

## 1. Noble and Akash are unreachable

`getCases(for:)` returns `[.cosmosIBC]` for `.noble` and `.akash` (`FunctionCallTypeEnum.swift:95-98`), but neither chain is in `CoinAction.memoChains` (`Core/Services/Actions/Coin+ChainAction.swift:12-16`), so the Function button never appears there. Dead code — delete, do not migrate.

## 2. Osmosis opens on a form its own dropdown does not offer

`getCases` offers only `[.cosmosIBC]` for Osmosis, but `getDefault` falls through to `.custom` (`FunctionCallTypeEnum.swift:122-124`) and `FunctionCallInstance.getDefault` builds a `FunctionCallCustom` whose `loadTokens` handles only THORChain and MayaChain. The user lands on a token-less Custom form and has to manually pick "IBC Transfer".

## 3. THORChain chainnet / stagenet have no cases at all

Both are in `memoChains` but absent from `getCases` — empty selector, Custom fallback.

## Note on overlap

The default-selection half of this is the same class of bug as #5132 (sTCY/yTCY), and the action-list screen removes `getDefault` from the picture entirely. Delete the dead arms here regardless; fix the Osmosis/testnet defaults only if the list screen has not landed first.

## Acceptance

- `.noble` / `.akash` arms deleted; no chain in `getCases` is absent from `memoChains`.
- Every chain in `memoChains` either has a non-empty `getCases` or is removed from `memoChains`.
- A test pins `getCases(for:)` non-empty ⟺ `getDefault(for:)` ∈ `getCases(for:)` for every chain in `memoChains`.

Part of #5124.

gastonm5:
## Problem

Four tables decide what the Function (memo) screen shows, and **none is derived from any other**, so every disagreement between them is a silent runtime defect instead of a compile error:

| Table | Decides |
|---|---|
| `CoinAction.memoChains` | which chains show the Function button at all |
| `FunctionCallType.getCases(for:)` | what the dropdown lists |
| `FunctionCallType.getDefault(for:)` | which entry the dropdown opens on |
| `FunctionCallInstance.getDefault(for:vault:)` | which form body is actually built |

Three of them had drifted:

1. **Noble and Akash offered `.cosmosIBC` from `getCases` but are not in `memoChains`.** The Function button never appears on those chains, so the arms were unreachable. (Both also have an empty `Chain.ibcTo` table, so the form would have had no destination chain to pick even if you could reach it.)
2. **Osmosis opened on a form its own dropdown did not offer.** `getCases` returned `[.cosmosIBC]`, but *both* default factories fell through to `.custom` — and `FunctionCallCustom.loadTokens` only handles `.thorChain` / `.mayaChain`, so the token dropdown came up empty and `isFormValid` could never be true. The user landed on a dead-end Custom form labelled with a function the selector does not list.
3. **THORChain chainnet and stagenet were in `memoChains` with no cases at all** — an empty selector plus the same dead-end Custom fallback.

## Root cause

The two default factories are separate `switch`es over the same chain, and neither is derived from `getCases`. Nothing — not the compiler, not a test — held them in step. Osmosis is the proof this had already gone wrong unnoticed: `getCases` was updated for it, the defaults were not.

## Fix

- Deleted the `.noble` / `.akash` arms from `getCases`; folded `.osmosis` into the `.kujira` arm.
- Osmosis now defaults to `.cosmosIBC` in **both** factories (its `ibcTo` table has Gaia and Kujira destinations, so this is a real, working form).
- Removed `.thorChainChainnet` / `.thorChainStagenet` from `memoChains`. Wiring the mainnet case list to the forks was the alternative, but the sub-models behind those cases match `.thorChain` exactly (e.g. `FunctionCallReBond` rejects any other chain outright), so it would only have surfaced forms that reject the coin. The forks keep send / swap / buy / receive — those are built independently of `memoChains`.
- Documented the invariant at both factories and at `memoChains`.

## The durable part: `FunctionCallReachabilityTests`

Four tests pin the contract, and each one fails on one of the original defects:

| Test | Catches |
|---|---|
| `testEveryMemoChainHasCases` | chainnet/stagenet (button with an empty selector) |
| `testNoChainOutsideMemoChainsOffersCases` | Noble/Akash (unreachable arms) |
| `testEveryMemoChainDefaultIsOffered` | Osmosis (`.custom ∉ [.cosmosIBC]`) |
| `testBothDefaultFactoriesAgreeForEveryMemoChain` | either factory being fixed without the other |

The agreement test maps `FunctionCallInstance` back to `FunctionCallType` through an exhaustive `switch`, so a new sub-model case fails to compile until it is mapped. No network or persistence dependency.

### Relation to #5132 (not fixed here)

#5132 is the same class of bug one layer down: `FunctionCallType.getDefault` routes on `ticker.contains("TCY")` (case-sensitive substring) while `FunctionCallInstance.getDefault` routes on `uppercased() == "TCY"`, so `sTCY` / `yTCY` select `.custom` in the dropdown but build a `.rebond` form.

`testBothDefaultFactoriesAgreeForEveryMemoChain` **would have caught it** — that is exactly the mismatch it asserts against — but only if the sweep were fed a `sTCY`-family coin. It represents each chain by its *native* ticker (RUNE for THORChain), so the TCY branch is out of its range today. That is deliberate: asserting it now would make this PR red on a known, unfixed bug in someone else's scope. Extending the sweep to the TCY family is the natural one-line follow-up once #5132 lands; the gap is documented in the test.

## Verification

- **Full unit suite** at `2eb23482d` on a dedicated `gate-5138` simulator (iPhone 16, iOS 26.5), worktree-local `-derivedDataPath`:
  `** TEST SUCCEEDED **` — **6056 tests executed, 11 skipped, 0 failures**. All four new tests compiled and ran (`Test Suite 'FunctionCallReachabilityTests' passed`).
- **Disk** (checked because two sibling agents were building concurrently, so a stale-products false pass was a real risk): `41Gi` avail before, `36Gi` avail after — never near the pressure zone, so the marker is trustworthy.
- `7e811557c` is comment-only (no behaviour change) and was verified with `build-for-testing`: `** TEST BUILD SUCCEEDED **`.
- SwiftLint: 0 violations across every changed file (repo baseline of 26 pre-existing violations elsewhere is unchanged).

### Cross-model review ledger (Codex, read-only, 1 round)

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | Factory-agreement test does not cover THORChain's TCY branch | Low | **Rejected with reason** — that is #5132, explicitly out of scope; asserting it would make the suite red on a known unfixed bug. Gap documented in the test instead. |
| 2 | The new `memoChains` comment overclaimed ("every sub-model matches `.thorChain` exactly") — `FunctionCallWithdrawSecuredAsset` does recognise the forks | Low | **Fixed** in `7e811557c` — rationale narrowed to the accurate fact (zero selectable cases → unusable Custom fallback). |

Codex found no blocking regression, independently traced every consumer of `memoChains` and `getCases` (no missed production readers), and walked both factories chain-by-chain for all 15 `memoChains` entries.

Closes #5138

## Summary by CodeRabbit

* **Updates**
  * Excluded THORChain mainnet and stagenet from supported memo-action chains.
  * Added Cosmos IBC function-call support for Kujira and Osmosis.
  * Refined available function-call options and defaults for supported networks.
* **Bug Fixes**
  * Improved consistency between network options, default selections, and generated forms.
* **Quality Improvements**
  * Added coverage for function-call availability and default behavior across supported networks.


johnnyluo (approved):
LGTM

PR #5156, opened 2026-08-13
fix(model): publish @Transient-derived state so views re-render on change

### LINKED ISSUE #5110: [Fix] Views reading @Transient-derived state don't re-render when it changes [Vault / Coin]
# [Fix] Views reading `@Transient`-derived state don't re-render when it changes [Vault / Coin]

## Problem

`@Transient` properties on a SwiftData `@Model` are **not observation-tracked**. Writing one — even with a genuinely different value — publishes no change to SwiftUI. Measured in an `NSWindow` + `NSHostingView` harness: a deferred write of a *changing* value to a `@Transient` property produced **1** body evaluation, identical to the control that wrote nothing.

Two shipping computed properties derive entirely from `@Transient` storage, so any view reading them is stale until something unrelated forces a re-render:

- **`Vault.isFastVault`** (`VultisigApp/Model/Vault.swift:262`) reads `fastVaultEligibilityCheckedAt` and `fastVaultEligibility`, both `@Transient` (`:49-50`). It is written by `FastVaultEligibilityRefresher`, which runs asynchronously after launch — exactly the case that needs to publish.
- **`Coin.hasPendingBalance`** (`VultisigApp/Model/Coin.swift:142`) derives from `pendingRawBalance`, `@Transient` (`:35`). Read at `CoinDetailHeaderView.swift:61` to decide whether to show the pending line at all.

These currently *appear* to work, which is the dangerous part: a neighbouring `Storage.shared.save()` re-renders the tree via the root `@Query` at `ContentView.swift:17` and incidentally picks up the new value. So correctness depends on an unrelated save happening to fire nearby, not on the write itself.

That coupling is actively being removed — there is in-flight work to scope the root `@Query` and to guard redundant `@Model` writes, both of which reduce how often the tree re-renders for unrelated reasons. **Each of those otherwise-correct changes makes this bug more likely to become visible**, which is why it is worth fixing on its own rather than discovering it as a regression.

## Expected Behavior

A view reading `vault.isFastVault` or `coin.hasPendingBalance` re-renders when the underlying value changes, without depending on an unrelated `save()` elsewhere in the app.

Concretely: after `FastVaultEligibilityRefresher` resolves eligibility, any view branching on `isFastVault` reflects it. When an inbound balance arrives, `CoinDetailHeaderView` shows the pending line without waiting for another write to the store.

## Steps to Reproduce

1. Run on macOS with a vault whose FastVault eligibility is not yet cached.
2. Observe a view that branches on `vault.isFastVault` while the async refresher resolves it.
3. The view does not update on the write. It updates later, when an unrelated `save()` re-renders the tree.

Same shape for `hasPendingBalance`: mutate `pendingRawBalance` in isolation and observe that `CoinDetailHeaderView` does not re-render.

## Solution

Move the published state off `@Transient`, or publish it explicitly. Options, roughly in order of preference:

1. **Hold the derived state on an `@Observable` view model** rather than on the `@Model`. Transient per-session state arguably does not belong on the persisted entity in the first place, and this keeps SwiftData's observation semantics out of it.
2. **Make the properties persisted** if the values are worth storing. Note this puts them back into the redundant-write bug class, so any writer must then carry an equality guard (see the `LPPosition`/`StakePosition` guard shape).
3. **Publish explicitly** from the writer — have the refresher notify through a channel the view actually observes.

Option 1 is likely right for `fastVaultEligibility` (a cache with a `checkedAt` stamp, session-scoped by design). `pendingRawBalance` may want option 2, since a pending balance is arguably real state rather than a cache.

Worth auditing the other `@Transient` declarations for the same pattern while fixing — `Coin.bondedNodes` (`Coin.swift:37`) is `@Transient` too and currently has no reader, so it is latent rather than broken.

## Context

Found during the investigation of the macOS layout-loop report, as an incidental result rather than a cause — it is **not** related to that crash. Detail: `wiki/pages/projects/vultisig/macos-render-cycle-crash-4470/model-write-guard-audit.md`.

gastonm5:
## Description

Fixes #5110

`@Transient` properties on a SwiftData `@Model` are **not observation-tracked**: writing one publishes nothing to SwiftUI, even when the value genuinely changes. Two shipping computed properties derived entirely from `@Transient` storage, so views reading them were stale until something unrelated forced a re-render:

- `Vault.isFastVault` — written asynchronously by `FastVaultEligibilityRefresher` after launch, exactly the case that needs to publish.
- `Coin.hasPendingBalance` — read by `CoinDetailHeaderView` to decide whether to show the pending line at all.

⚠️ **These appeared to work, which is the dangerous part.** A neighbouring `Storage.shared.save()` re-rendered the tree through the root `@Query` and incidentally picked up the new value — so correctness depended on an unrelated save happening to fire nearby. In-flight work to scope that root `@Query` and to guard redundant `@Model` writes removes that coupling, and each of those otherwise-correct changes makes this bug visible.

## Approach

New `ObservedTransient<Value: Equatable>` (`Core/Models/ObservedTransient.swift`) — a plain `@Observable` box with an **equality-guarded setter**. Each model holds it in a `private @Transient` reference and exposes a computed property of the same name and type, so **no reader or writer call site changes**.

**Why not a free-standing view model**, which is what the issue's option 1 literally suggests: `isFastVault` has **~30 readers**, most of them views handed only a `Vault`. Threading a new view model to all of them is a large refactor that would also put "is this vault fast-signable" in two places. The state still moves *off* the `@Model`'s storage onto an `@Observable` object — the model just keeps an accessor facade.

Per-instance boxes rather than a shared store, deliberately: a store keyed by `pubKeyECDSA` / `coin.id` would cross-talk between vaults that have no key yet — the test fixtures share `"test-pub-ecdsa"` and an identical `coin.id`. There are tests pinning that isolation.

**No schema change, no migration.** The boxes are `@Transient`, the accessors are computed (SwiftData treats computed properties as transient), neither property was ever in `CodingKeys`, and `Vault+ProtoMappable` is untouched.

The issue's option 2 (persist the values) is **rejected explicitly for `pendingRawBalance`**: it is a mempool snapshot, worthless a launch later, and persisting it would be a schema change on a critical boundary *and* move it into the redundant-write bug class.

## `@Transient` audit

Four declarations across two files; three converted. **`Coin.bondedNodes` left as a plain `@Transient` deliberately** — no view body reads it. Its only readers are `UnbondTransactionViewModel.selectBondNode()` (called once from `onLoad()` and republished through its own `@Published`) and `Coin.hasBondedNodes`, which has **zero callers**. Converting it would be unverifiable churn on a critical boundary; it now carries a comment saying exactly that, and when to convert. It is the only plain `@Transient` left in `Model/`.

Adjacent and not touched: `Coin.fiatRateAvailable` reads the non-`@Observable` `RateProvider.shared`, which the repo already solved with the issue's option 3 (a `ratesDidChange` subject).

## Which feature is affected?
- [ ] Sending
- [ ] Swap

None of the four. This is SwiftData/SwiftUI observation plumbing on `Vault`/`Coin` plus the views reading them — no signing, broadcast or fee path is touched, and it is genuinely platform-specific (no cross-platform analogue to SwiftData observation semantics).

## Checklist

`make test`: `** TEST SUCCEEDED **` — **6063 tests, 11 skipped, 0 failures**. SwiftLint 26, exactly the baseline.

**The tests pin publication, not value.** 11 cases in `TransientObservationTests`, asserting through `withObservationTracking` on the same expression the view reads (`isFastVault`, `hasPendingBalance`, `pendingBalanceStringWithTicker`) rather than on the underlying storage. Includes a per-input sweep (`isFastVault` short-circuits on `checkedAt == nil`, so the stamp and the flag each get a case), redundant-write silence, and per-instance isolation.

**Verified against the parent rather than assumed:** reverting only `Model/Vault.swift` produces exactly 3 failures — the fast-vault publication cases. Reverting only `Model/Coin.swift` produces exactly 3 — the pending-balance ones. The value-asserting cases pass in both, which is the point: a value test cannot see this bug.

## Additional context

Cross-model review: three read-only Codex passes, 0 findings on both per-commit passes. The whole-branch pass raised one minor — *"a future view-facing plain `@Transient` still compiles and silently recreates the defect; the doc comments are good but not enforceable."* Accepted, and became the third commit.

⚠️ **That third commit edits `.claude/rules/swiftdata.md`**, a tracked repo rules file, adding the `@Transient`-publishes-nothing rule. It is markdown only — not compiled or tested — but it changes guidance for everyone working in this repo, so call it out in review. Happy to split it into its own PR if preferred.

## Agent Metadata
- **Agent**: Claude Code
- **Issue**: #5110
- **Needs human review for**: the `ObservedTransient` facade shape (accessor on the model vs. a separate view model), and whether the `.claude/rules/swiftdata.md` addition belongs in this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SwiftUI updates for session-only vault eligibility and pending-balance values.
  - Prevented unnecessary updates when transient values are written with unchanged data.
  - Preserved per-instance isolation and derived-value refresh behavior.

- **Tests**
  - Added coverage for transient state updates, redundant writes, instance isolation, derived values, and server-party scenarios.

- **Documentation**
  - Documented recommended handling for transient state that must trigger SwiftUI observation updates.

### DIFF
--- a/.claude/rules/swiftdata.md
+++ b/.claude/rules/swiftdata.md
@@ -8,3 +8,4 @@ paths:
 - Never access `@Model` classes off MainActor
 - Use value types (structs) to pass data across actor boundaries
 - Follow the three-phase architecture: load -> transform -> save
+- `@Transient` properties are **not observation-tracked**. Writing one publishes nothing to SwiftUI, even when the value genuinely changes, so a view reading it (or anything derived from it) never re-renders — it will *appear* to work whenever a neighbouring `Storage.shared.save()` happens to re-render the tree, which is an accident, not a mechanism. A plain `@Transient` is only safe for state that no view body ever reads. Otherwise hold the value on an `ObservedTransient` box (`Core/Models/ObservedTransient.swift`) behind a computed property of the same name, and cover it with a `withObservationTracking` test that asserts the write is *published*, not just stored.
--- a/VultisigApp/VultisigApp/Core/Models/ObservedTransient.swift
+++ b/VultisigApp/VultisigApp/Core/Models/ObservedTransient.swift
@@ -1,0 +1,50 @@
+//
+//  ObservedTransient.swift
+//  VultisigApp
+//
+//  Observation-tracked storage for per-session state that hangs off a SwiftData
+//  `@Model` but is deliberately kept out of the schema.
+//
+
+import Foundation
+import Observation
+
+/// A box holding one value that publishes through the Observation registrar.
+///
+/// SwiftData excludes `@Transient` properties from a `@Model`'s observation
+/// graph. Writing one publishes nothing — not even when the value genuinely
+/// changes — so a SwiftUI view reading anything derived from it is never
+/// re-evaluated. That was measured directly: a deferred write of a *changing*
+/// value to a `@Transient` property produced the same single body evaluation as
+/// a control that wrote nothing at all. Views built on such a property appeared
+/// to work only because an unrelated `Storage.shared.save()` re-rendered the
+/// tree soon afterwards, which is an accident of neighbouring code rather than a
+/// mechanism.
+///
+/// Holding the value on a plain `@Observable` object restores publication. The
+/// model keeps a `@Transient` reference to the box; that reference is assigned
+/// once and never reassigned, so the fact that reading it is untracked costs
+/// nothing, while the read of ``value`` inside a view body registers with the
+/// registrar exactly as a persisted property would. The state stays session
+/// scoped and stays out of the schema, so no migration is involved.
+///
+/// The setter is equality-guarded. Observation's generated setters notify
+/// without comparing new to old, and a redundant notification from a deferred
+/// writer is the shape that drives a runaway re-render, so the comparison lives
+/// here instead of being re-derived at every call site.
+@Observable
+final class ObservedTransient<Value: Equatable> {
+    private var storage: Value
+
+    var value: Value {
+        get { storage }
+        set {
+            guard newValue != storage else { return }
+            storage = newValue
+        }
+    }
+
+    init(_ value: Value) {
+        self.storage = value
+    }
+}
--- a/VultisigApp/VultisigApp/Model/Coin.swift
+++ b/VultisigApp/VultisigApp/Model/Coin.swift
@@ -32,8 +32,30 @@ class Coin: ObservableObject, Codable, Hashable {
     /// Transient because it describes the mempool, which is not a thing worth
     /// persisting: it is refreshed by the same fetch that sets `rawBalance`
     /// and is meaningless a launch later.
-    @Transient var pendingRawBalance: String = "0"
+    ///
+    /// Held on an `ObservedTransient` box rather than in a `@Transient` stored
+    /// property. SwiftData does not observation-track `@Transient`, and
+    /// `CoinDetailHeaderView` reads `hasPendingBalance` to decide whether to draw
+    /// the pending line at all — so an inbound payment landing published nothing
+    /// and the line stayed absent until an unrelated save re-rendered the tree.
+    /// The value is still per-instance, still session scoped, and still absent
+    /// from the schema; only the publication changes.
+    var pendingRawBalance: String {
+        get { pendingRawBalanceBox.value }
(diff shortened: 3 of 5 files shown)


johnnyluo (approved):
LGTM

PR #5157, opened 2026-08-13
fix(defi): redeem receipt shares at the amount typed, not the nearest whole percent

### LINKED ISSUE #5090: [Fix] Unstake truncates the percentage to whole percent on bRUNE, RUJI auto-compound and CACAO [Staking]
## Problem

`UnstakeTransactionViewModel` holds the withdrawal fraction as a `Double` (`percentageSelected`, `UnstakeTransactionViewModel.swift:20`) and then discards everything after the decimal point at every call site that builds a transaction:

| line | chain | expression |
|---|---|---|
| `:74` | TCY | `percentage: Int(percentageSelected ?? percentageFromAmount)` |
| `:82` | bRUNE | `percentage: Int(percentageSelected ?? percentageFromAmount)` |
| `:101` | RUJI auto-compound (`RUJILiquidUnbondTransactionBuilder`) | `percentage: Int(percentageSelected ?? percentageFromAmount)` |
| `:115` | CACAO | `bps: Int(percentageSelected ?? percentageFromAmount) * 100` |

The only exact path is RUJI **bonded** (`:107`), which passes `amount: amountField.value` and never routes through a percentage.

The consequence is that a user who types an amount gets a withdrawal quantised to 1% of their position, silently. On a 2002.74 position that is a ~20 token step. This was reported by a real user for TCY in #5074 — 1002.73 requested, 1001.37 withdrawn — and the arithmetic reproduces exactly:

```
1002.73 / 2002.74 = 50.0679%
Int(50.0679) = 50 → memo tcy-:5000 → 1001.37
```

Nothing about this is TCY-specific. It is the shared view model, so bRUNE, RUJI auto-compound and CACAO all carry it today.

**TCY is already fixed** by #5089, which converts straight to basis points on that arm only (error 1% → 0.01%) and quotes the amount that will actually be paid out. The other three arms are untouched.

This was previously observed and deliberately deferred during the bRUNE/ybRUNE work on the grounds that it was byte-identical to the shipped sTCY behaviour and carried no fund loss. #5074 is that deferral arriving as a user report, so it will keep arriving.

## Expected behavior

A typed amount withdraws as close to that amount as the underlying instruction can express, on every staking asset — not rounded to the nearest whole percent of the position.

### Must Do
- [ ] Carry the fraction at basis-point resolution (or better, where the instruction allows it) through all four call sites in `UnstakeTransactionViewModel`
- [ ] Apply the same treatment #5089 applied to TCY: convert the typed amount directly to the instruction's own units rather than via a truncated whole percent
- [ ] Where the withdrawal cannot honour the typed figure exactly, show the user the amount that will actually be paid out rather than the amount they typed
- [ ] Carry over the guards #5089 added, which apply identically here: never let rounding close a position that was meant to stay open, never build a zero-fraction instruction for a positive amount, and do not let `isMaxAmount` stay stale-true after a custom amount is typed

### Must NOT Do
- Don't change the RUJI bonded path (`:107`) — it already carries an absolute amount and is exact
- Don't alter which message each asset unstakes through; bRUNE/TCY compute an absolute amount locally while CACAO and TCY-bonded carry a fraction in the memo, and those differences are deliberate

## Acceptance Criteria
- [ ] A typed unstake amount on bRUNE, RUJI auto-compound and CACAO produces an instruction accurate to at least 0.01% of the position
- [ ] The confirmation screen states the amount that will actually be withdrawn
- [ ] A full-balance withdrawal still closes the position exactly, and a partial one never does
- [ ] Tests that fail against the parent commit for each of the three arms

## Context

Found while implementing #5074 (→ #5089). Verified against `main` @ `01e25f4f7`.

gastonm5:
## Description

Fixes #5090

`UnstakeTransactionViewModel` held the withdrawal fraction as a `Double` and truncated it to `Int` at each builder call site, so a typed amount was silently quantised to **1% of the position** — a ~20 token step on a 2002.74 position. That is how #5074 arrived as a user report; this PR fixes the two arms #5089 deliberately fenced off.

### Scope: two arms, not three

The issue lists bRUNE, RUJI auto-compound **and CACAO**. **CACAO was already fixed on this branch's base** by the last commit of #5089 — it is in `withdrawsInBasisPoints`, has a `withdrawDisplayAmount`, and is covered (`POOL-:5006`). The issue text predates that commit. Real scope is **bRUNE and RUJI auto-compound**.

### Deliberate deviation: not basis points

The issue asks for basis-point resolution. That is right for the memo-carrying arms and wrong for these two.

A `tcy-:<bps>` / `POOL-:<bps>` memo names a **fraction**, and cannot express finer than a ten-thousandth — which is why `WithdrawBasisPoints` exists. A `liquid.unbond` names **nothing**: the execute message is empty and the **attached funds are the instruction**, an absolute count of receipt base units. Routing these arms through basis points would re-impose a quantisation the instruction does not have.

So they get a sibling, `ReceiptShareRedemption`, converting the typed amount straight to base units. Accuracy goes from 1% of the position to **~1e-8** of it — three orders finer than the 0.01% the acceptance criteria asked for.

The two arms are not symmetric, which is why the conversion takes a position value rather than assuming one:

- **bRUNE** — the card renders the ybRUNE receipt balance directly, so the typed amount is already a share count.
- **RUJI auto-compound** — the card renders what the receipt is *worth* in RUJI, so shares follow from the ratio of the two balances.

### What the screen says it will pay out

RUJI auto-compound gains a `withdrawDisplayAmount` (`stakedAmount × redeemed / held`): that card quotes the position's value, so the ratio is real, and the verify screen named no figure at all before. It renders through the withdrawal provider #5089 already installed in `TransactionHeroResolver` — that provider keys on a builder-supplied figure rather than on TCY, so it describes this redemption too, which `testTheCompoundedRujiRedemptionQuotesTheRujiThatWillBePaidOut` pins end to end.

**bRUNE deliberately still names none**, reusing #5089's established call: its sheet is denominated in ybRUNE while the payout is bRUNE, and the bond ratio is not held here. What changed for bRUNE is that the figure in the field is now exactly the figure redeemed.

**Guards, per arm:** a full exit *pins* the whole held balance rather than deriving it; truncation means a partial can never close the position; the count is clamped to held; and below one base unit no payload is built. That base unit is the floor for these arms, in place of the one basis point the bps arms floor at — which is why they stay out of `withdrawsInBasisPoints` and `WithdrawMinimumAmountValidator`.

## What only the whole-branch review caught

**A near-MAX typed amount could close a position meant to stay open.** `AmountPercentageBinding.percentage` returns a `Double`, so an amount within ~1e-15 of the balance derives *exactly* 100 — the value a fresh sheet already holds. SwiftUI emits no `onChange`, `onPercentage` never runs, and `isMaxAmount` stays true. A full exit now additionally requires the amount to **be** the position: the balance, or the balance truncated to the field's own precision (what MAX prefills). Exact rather than a tolerance — MAX's prefill is already up to one display step short, so any window wide enough to admit it also admits the typed amounts the check exists to catch.

**A large position could become unwithdrawable at MAX.** The redemption is kept in `Decimal` rather than routed through `Decimal.toInt()`, which wraps outside signed 64-bit range while the receipt balance is parsed as `UInt64` — so a position above `Int.max` base units would render as a negative count, fail the positive guard, and build no payload at all.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

**No tx link.** Both arms are fund paths — the funds attached to the unbond are what change — and signing one needs a real staked THORChain position, which this environment does not have.

> ⚠️ **Wants an on-device mainnet check before merge.** Every commit message says so.

## Parity

- Android: `n/a: not implemented` — no `liquid.unbond` staking path in the app source.
- Windows + extension: `n/a: not affected` — `core/ui/vault/deposit/staking/resolvers/stcy-auto.ts:25-38` converts the typed amount straight to share units and attaches it as funds. It never routes through a percentage, so the quantisation fixed here does not exist there. (Its denomination differs — it treats the typed amount as a share count — which is a UI choice, not this defect.)
- SDK: `n/a: no amount entry` — the funds arrive already built.

## Testing

Full `make test` on this exact tree: **`** TEST SUCCEEDED **`, 0 failures.** SwiftLint clean.

`ReceiptShareRedemptionTests` covers both arms: the conversion and its guards, the MAX-flag closure, the `Decimal` range behaviour, the payout quote reaching the verify screen, and the refusal below one base unit. The precision limit is documented where it actually runs out rather than claimed away.

## Stack

```
main → #5176 → #5089 → #5157 (this)
```

Rebased onto #5089 directly. It previously sat on #5136, which has been **closed as superseded** — that PR added a declared `FunctionTransactionKind` to every builder for the initiator's verify screen, and the decoder in #5176 covers the reported defect without it. Nothing in this PR depended on it: the two `functionKind` declarations it carried were dropped, and the payout figure renders through #5089's registry instead.

Review #5176 and #5089 first.

## Checklist

## Agent Metadata

- **Agent**: Claude Code
- **Issue**: #5090
- **Confidence**: high on the conversion and guards; the residual risk is the two-read projection named below
- **Needs human review for**: the on-device mainnet check, and the RUJI payout quote — it comes from two reads taken close together but not simultaneously (the persisted DeFi card's `liquidSize` and a separate THORNode bank-balance read), so a stale card skews the quote by that same proportion. It is bounded above by the typed amount and never quotes more than was asked for.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Forms/AmountPercentageBinding.swift
+++ b/VultisigApp/VultisigApp/Components/Forms/AmountPercentageBinding.swift
@@ -25,8 +25,27 @@ import Foundation
 /// figure the user typed.
 enum AmountPercentageBinding {
 
+    /// How many decimals the function-transaction amount field renders, and
+    /// therefore the precision `AmountTextField.setupAmount()` writes when the
+    /// percentage control fills the field.
+    ///
+    /// Named rather than repeated because it is not only a display choice: a MAX
+    /// selection puts the balance TRUNCATED to this precision into the field, so
+    /// anything deciding "is this the whole position" has to know the figure it
+    /// is comparing against was rounded here. See
+    /// `ReceiptShareRedemption.isWholePosition`.
+    static let displayedDecimals = 4
+
     /// The percentage of `available` that `amount` comes to, clamped to 0…100,
     /// or `nil` when there is no balance to take a percentage of.
+    ///
+    /// ⚠️ **Lossy, and load-bearingly so.** The result is a `Double`, so a
+    /// fraction within about 1e-15 of the whole position comes back as exactly
+    /// `100`. When the field already held 100 that is not a change, SwiftUI emits
+    /// no `onChange`, and `UnstakeTransactionViewModel.onPercentage` never runs to
+    /// clear the MAX flag — leaving it stale-true over an amount the user meant to
+    /// be short of the balance. Nothing downstream may treat that flag as the sole
+    /// evidence of a full exit; see `ReceiptShareRedemption.baseUnits`.
     ///
     /// The clamp is load-bearing: the amount field accepts whatever is being
     /// typed, including a figure above the balance (rejecting that is the
--- a/VultisigApp/VultisigApp/Features/FunctionTransaction/Common/Screen/AmountFunctionTransactionScreen.swift
+++ b/VultisigApp/VultisigApp/Features/FunctionTransaction/Common/Screen/AmountFunctionTransactionScreen.swift
@@ -44,7 +44,10 @@ struct AmountFunctionTransactionScreen<CustomView: View, TopView: View>: View {
         availableAmount: Decimal,
         percentageSelected: Binding<Double?>,
         percentageFieldType: PercentageFieldType,
-        amountDecimals: Int = 4,
+        // Defaults to the precision the amount/percentage binding rounds to, so
+        // a screen that does not override it cannot drift from the figure that
+        // binding derives its percentage from.
+        amountDecimals: Int = AmountPercentageBinding.displayedDecimals,
         amountField: FormField,
         isContinueDisabled: Bool = false,
         customViewPosition: AmountTextField<CustomView>.CustomViewPosition = .balance,
--- a/VultisigApp/VultisigApp/Features/FunctionTransaction/SharedFunctions/Unstake/UnstakeTransactionViewModel.swift
+++ b/VultisigApp/VultisigApp/Features/FunctionTransaction/SharedFunctions/Unstake/UnstakeTransactionViewModel.swift
@@ -209,9 +209,22 @@ final class UnstakeTransactionViewModel: ObservableObject, Form {
             // `account.withdraw`. Both arrive here on the RUJI coin because the
             // compounded card maps sRUJI back to its bond coin.
             if isAutocompound {
+                // The redemption spends receipt shares, and unlike TCY/bRUNE the
+                // share balance is not what bounds the amount field — so nothing
+                // else stops a zero here. A zero means the read is still in flight,
+                // failed, or the position is empty; all three would build a wasm
+                // execute carrying no funds.
+                guard autocompoundBalance > 0 else { return nil }
+                // Carries the typed amount rather than a whole percentage of the
+                // position, for the reason its bRUNE sibling does — the
+                // redemption is FUNDED with an absolute share count. Here the
+                // amount is priced in RUJI and the funds are sRUJI shares, so
+                // `availableAmount` is what converts between them: it is the RUJI
+                // the shares are worth, read at the same moment as the shares.
                 return RUJILiquidUnbondTransactionBuilder(
                     coin: coin,
-                    percentage: Int(resolvedPercentage),
+                    withdrawAmount: amountField.value.toDecimal(),
+                    stakedAmount: availableAmount,
                     receiptShares: autocompoundBalance,
                     sendMaxAmount: isMaxAmount
                 )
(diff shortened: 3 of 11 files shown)


gastonm5:
Closing — folded into #5089 rather than abandoned. Every commit here is now on `fix/tcy-withdraw-5074`, and #5089 fixes both #5074 and #5090.

These were always one change. The truncating expression lives in the shared `UnstakeTransactionViewModel`, and it was split only because #5074's issue text said "don't modify other token withdrawal flows" — so #5089 fixed the memo-carrying arms and fenced off the receipt-share ones, and this PR came back for them. Reviewing them apart means reading the same view model twice, and they need the same on-device mainnet check before either can merge.

The practical reason to collapse now: this repo only runs CI on `main`-based PRs, so as the third layer of a three-deep stack this PR had **never had a CI run**. Every change to the base also meant three rebases and three force-pushes — four such cycles yesterday alone. One PR on top of #5176 gets CI as soon as that merges, and the stack stops costing anything.

Nothing was dropped in the fold: the six commits replay unchanged, and the full gate on the combined branch is green at **6720 tests**.

The branch `fix/unstake-precision` is left in place rather than deleted, in case you want the narrower diff to review this half on its own.

Issue #5124, opened 2026-08-12
[Epic] Migrate the legacy Functions screen (FunctionCall) into the DeFi tab


gastonm5:
## Summary

The app ships **two form architectures** for "do a non-Send chain operation", side by side:

| | LOC | Design-system violations |
|---|---|---|
| Legacy `Features/FunctionCall/` — the "Function" screen | 4,627 | **39** |
| New `Features/FunctionTransaction/` — entered from the DeFi tab | 7,729 | **0** |

They already converge on the same tail (`FunctionCallRoute.verify` → `FunctionCallVerifyScreen` → shared `SigningRoute` keysign), so only the **form** half is duplicated. The new architecture is clearly the better one — typed intent (`FunctionTransactionType`) → `TransactionBuilder` → one `buildSendTransaction(vault:)` — and its seams are under-used rather than missing.

This issue tracks moving the remaining legacy operations off the Functions screen and eventually deleting it.

## The blocker is chain coverage, not code

`CoinAction.memoChains` (where the Function button appears) and `CoinAction.defiChains` (where the DeFi tab exists) overlap on **THORChain and MayaChain only** — `Core/Services/Actions/Coin+ChainAction.swift:12-27`.

| | Chains |
|---|---|
| `memoChains` | THORChain (+ chainnet, stagenet), MayaChain, dYdX, Kujira, Gaia, Osmosis, BTC, BCH, LTC, DOGE, Ethereum, Avalanche, BSC, Base, Ripple |
| `defiChains` | THORChain, MayaChain, Tron, Terra, Terra Classic, QBTC, TON, Solana |

So nothing on dYdX, the Cosmos chains, or the nine L1 chains that do THOR LP / SECURE+ can simply be re-homed. Each needs either a new `defiChains` entry with a segment that makes sense there, re-homing under **THORChain's** DeFi tab (defensible — LP add and SECURE+ *are* THORChain operations initiated from another chain's wallet), or a non-DeFi home.

**Open decision:** do L1-initiated THORChain operations live under THORChain's DeFi tab, or does BTC/ETH get its own tab?

## Inventory — the 13 remaining operations

| Legacy case | Chains | DeFi / new equivalent | Legacy still the only route? |
|---|---|---|---|
| `bondMaya` | Maya | ✅ `.bond` → `BondMayaTransactionScreen` | **No — duplicated** |
[...]

PR #5158, opened 2026-08-13
fix(forms): recompute validForm when validators change, not only values

### LINKED ISSUE #5139: [Fix] Form.validForm goes stale when validators are installed asynchronously [FunctionTransaction]
## Summary

`Form.setupForm()` recomputes `validForm` **only** when a `field.$value` publishes (`Components/Form.swift:19-40`), and `validateErrors()` re-runs the validators without recomputing it. So installing a validator *after* the user has already typed leaves `validForm` reflecting the **old** validator set.

## Reproduction (shipping today)

`BondMayaTransactionViewModel.onAssetSelected` installs `LPUnitsValidator(availableUnits:)` once the balance arrives, and `UnbondMayaTransactionViewModel.fetchBondedLPUnits` does the same. Type the LP units first, pick the asset second: no `$value` publish follows the validator swap, so `validForm` stays `true` under the pre-swap validators and `transactionBuilder` — which guards on exactly that flag — lets an over-available LP-unit transaction through.

## Why it matters now

Every screen migrated off the legacy Functions tree inherits this stack. Fixing it before the migrations land means ten new screens get it for free; fixing it after means auditing ten screens for the same race.

## Scope

1. Recompute `validForm` whenever the validator set changes — either publish validator mutations through the same pipeline, or expose a `revalidate()` that `validateErrors()` also calls.
2. Fix the retain cycle in `BondTransactionViewModel.onLoad`: the installed `ClosureValidator` captures `self` strongly while the VM owns the field and the field owns the validator. Sibling validators already use `[weak self]`.

## Acceptance

- A test that sets a field value, then installs a stricter validator, and asserts `validForm` flips to `false` without a further keystroke.
- A test asserting the Maya bond VM rejects LP units above the available amount when the asset is chosen after typing.
- No retain cycle: a `weak var` reference to the VM deallocates after the screen is dismissed.

Part of #5124.

gastonm5:
## Problem

`Form.setupForm()` recomputed `validForm` **only** when a field's `$value` published, and `validateErrors()` re-ran the validators without touching the flag. Validators, though, routinely arrive *after* the user has typed — an available LP balance, a stakeable ceiling, a pool's minimum — with no keystroke behind them. So `validForm` kept answering for the validator set the form had *before* the swap, and every `transactionBuilder` guards on exactly that flag.

Shipping today on Maya bond/unbond: type the LP units, then pick the asset (which is what installs `LPUnitsValidator(availableUnits:)`), and an over-available bond builds and reaches the signer. The same shape is latent on the Kamino and Yield forms, which append their balance guard behind an async probe.

Second defect: `BondTransactionViewModel.onLoad` installed a `ClosureValidator` capturing `self` strongly — the view model owns the field, the field owns the validator, the validator owned the view model. Unbreakable cycle; every sibling `ClosureValidator` already used `[weak self]`.

## Root cause

`FormField.validators` is a plain `var`. Mutating it changes what the field *is* but publishes nothing, so the only trigger the pipeline had was `$value`. And `Form.validateErrors()` — the Continue-tap path — computed each field's validity for its visible error and then threw the answer away, so the tap could not repair a stale flag either.

## Fix

The invariant: **`validForm` always reflects the validators currently installed.**

- `FormField.validators` now publishes from `didSet` through `validatorsPublisher`. Deliberately **not** `@Published`: that fires from `willSet`, so a subscriber reading the validators back would judge the value against the rules being replaced.
- `setupForm()` merges both triggers per field. The value branch keeps its run-loop hop (`@Published` fires from `willSet`, so validating inline would judge the outgoing value); the validator branch takes effect **synchronously**, so a form is never briefly valid under rules it no longer has. `Publishers.MergeMany` replaces `combineLatest` — the question is "did anything change?", not "what is the latest of each?".
- `validForm` gains a single assignment site, `revalidate()`, which runs the side-effect-free `FormField.validate()` across the whole form. `validateErrors()` calls it too, so the flag cannot be stale by the time a builder reads it. `setupForm()` calls it before returning, so validators installed *before* the pipeline was wired (Kamino builds its from an awaited vault probe) are accounted for immediately rather than one run-loop turn later.
- Only the field that changed refreshes its own visible error, so an error assigned directly elsewhere — `notWhitelistedOnNode`, `noBondedAssetsOnNode` — is not wiped by typing in another field.
- `BondTransactionViewModel`: `[weak self]` on the operator-fee validator.
- `SolanaDelegateTransactionViewModel` had already grown the per-VM workaround this replaces (`validateErrors(); validForm = form.allSatisfy { $0.valid }` after each validator rebuild); the manual recompute is gone.
- `CombineLatest.swift` deleted — `setupForm()` was its only consumer.

Not touched, deliberately: **Continue is still not gated on `validForm`.** `FormScreen.swift`'s doc comment explains at length why (it advances focus, opens a picker, or reveals errors), and that enforcement lives in `transactionBuilder` returning `nil`. That reasoning has been re-learned twice.

Why now: ten screens are about to migrate onto this stack under #5124. Fixing it here means they inherit it; fixing it later means auditing ten screens for the same race.

## Tests

`VultisigAppTests/FunctionTransaction/FormValidatorInstallationTests.swift` (9 tests, all camelCase):

- a stricter validator installed after the value is typed flips `validForm` with **no further keystroke and no run-loop hop** — plus the within-ceiling control, and the wholesale-replacement variant;
- `setupForm()` answers for validators installed before it;
- the keystroke path still publishes through the run loop;
- `validateErrors()` republishes the flag with no pipeline running;
- the Maya bond ordering that used to let an over-available bond through — `transactionBuilder` is `nil` — plus the within-available control that proves the `nil` is the LP ceiling and not an unfinished form;
- `BondTransactionViewModel` deallocates after installing its closure validator (`weak var` is `nil`).

## Verification

- Full suite on a dedicated `iPhone 16` sim (`gate-5139`), worktree-local DerivedData: **`** TEST SUCCEEDED **`** — 6061 tests, 11 skipped, **0 failures**.
- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 0 violations in every file this PR touches (26 pre-existing violations elsewhere, unchanged).
- Disk (`/System/Volumes/Data`): `41Gi` free at branch start, `32Gi` free immediately before the gate (93% used), `30Gi` free after (93% used). Free space never approached zero, so the marker is not a stale-run artifact.
- Cross-model review: two read-only `codex exec` passes. Round 1 raised 3 findings (1 fixed — `setupForm()` now revalidates synchronously; 2 rejected with reasons, see below). Round 2: *"No new or unresolved issues."*

### Codex findings ledger

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Validator mutations made *before* `setupForm()` were not replayed, and setup did not establish the invariant synchronously (reachable on the Kamino forms). | **Fixed** — `setupForm()` ends with `revalidate()`; new test covers it. |
| 2 | `validForm` can disagree with a field's rendered `valid`/`error` for cross-field validators (password confirm, bond operator fee). | **Rejected** — pre-existing: the old pipeline validated only the changed field too, so the rendered state is exactly as stale as before. Refreshing every field's *visible* state would wipe the errors other code assigns directly (Maya whitelist / no-bonded-assets), a worse regression. |
| 3 | `validateErrors()` before `onLoad()` can now vacuously return a builder for 4 VMs whose fields start with no validators. | **Deferred** — no reachable path; every screen calls `onLoad()` from `.onLoad` before Continue can be tapped. Codex confirmed the same in round 2. Flagged here for a human call. |

## ⚠️ Needs human review

This changes **shared validation infrastructure** used by 25 form view models across FunctionTransaction, DeFi (Kamino/Yield), Tron, keygen and vault setup — all signing-adjacent. `SendDetailsViewModel` conforms to `Form` but never calls `setupForm()`, so the send flow is out of the blast radius. No on-device pass yet.

Plan page: `wiki/pages/projects/vultisig/functions-to-defi-migration/form-validform-stale-5139-plan.md` (mechanism, chosen fix, alternatives rejected).

Closes #5139

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved form validation to update reliably as users enter information or validation rules change.
  - Forms now calculate overall validity consistently, including during initial setup and after asynchronous updates.
  - Updated transaction forms to recheck entered values when network configuration changes.
  - Improved Maya bond transaction validation for LP unit amounts after asset selection.
  - Reduced the risk of lingering resources in transaction workflows.


johnnyluo (approved):
LGTM

Issue #5159, opened 2026-08-13
[Swap] Swap-only operations are unreachable in GB/JP/MY — SECURE+ mint is the first casualty


gastonm5:
## Summary

`SwapFeatureGate.canSwap()` returns `false` when the device region is **GB, JP or MY** (`Core/Services/Actions/SwapFeatureGate.swift:10-16`). `Chain.defaultActions` then filters `.swap` out of every chain's action list (`Core/Services/Actions/Coin+ChainAction.swift:59-66`) while leaving `.memo` (the Function button) in place (`:49-51`).

Everything that is only reachable through Swap is therefore unreachable in those three regions. That was survivable while the Function screen carried a duplicate of the operation. It stops being survivable as those duplicates are retired.

## Concrete case

Retiring the legacy `securedAsset` Function in favour of the Swap route (#5141) removes the last region-independent path to a **SECURE+ mint**. After that change:

- Outside GB/JP/MY: mint via Swap (hold BTC → pick secured BTC → "Mint (SECURE+)").
- Inside GB/JP/MY: no in-app route to a SECURE+ mint at all. The Function button still shows on those chains, but now offers only `addThorLP`.

The redemption direction (`withdrawSecuredAsset`, SECURE−) is unaffected — it lives on the THORChain Function screen and is not gated.

## The question for a maintainer

Is this the gate working as intended, or a hole being closed by accident?

A SECURE+ mint is a THORChain deposit, i.e. exactly the class of operation a regional swap restriction is presumably meant to cover — in which case the legacy Function form was an unintended bypass and losing it is a *fix*, not a regression. The alternative reading is that the gate is meant to cover market swaps specifically, and a ~1:1 wrap should stay available.

The same question applies to every later step of the Functions → DeFi/Swap migration (#5124): each time an operation's only remaining home is Swap, it silently becomes region-gated.

## Suggested outcomes (pick one)

1. **Intended** — document it on `SwapFeatureGate` so the next migration step does not have to re-derive it, and close.
2. **Not intended for non-market operations** — the gate needs to distinguish "market swap" from "same-underlying mint / wrap", so `mode == .securedMint` stays reachable while pool swaps do not.

Related: the remote action config (`https://api.vultisig.com/actions/default.json`, consumed by `CoinActionResolver`) can also disable `swap` per chain. It currently does so only for Solana, which never offered the Function — but it is the same shape of lever and would have the same effect on any chain it is pointed at.

Follow-up from #5141.

PR #5160, opened 2026-08-13
refactor(functions): retire the legacy SECURE+ mint form in favour of the Swap route

### LINKED ISSUE #5141: [Functions] Retire the legacy SECURE+ mint form in favour of the Swap route
## Summary

SECURE+ mint ships twice. Swap already performs it — `SwapCryptoLogic.isSameUnderlyingSecuredMint` -> `DefaultSwapInteractor.swift:173` -> `ThorchainRouterDepositBuilder.buildSecuredMintPayload` — and the legacy `securedAsset` function does the same thing with a second form.

## Why this one first

`securedAsset` appears in `getCases` for THORChain **and all nine L1 chains** (`FunctionCallTypeEnum.swift:60-79`). Removing it drops BTC / BCH / LTC / DOGE / ETH / AVAX / BSC / Base / XRP from two actions to **one** (`addThorLP`), which is what lets those chains skip the action-list screen entirely and route the button straight to the operation.

## Scope

- Delete `FunctionCallSecuredAsset` + its co-located `FormView` and its `FunctionCallSecuredAssetTests`.
- Remove the arm from `getCases` for all ten chains and from both `getDefault` factories (`FunctionCallTypeEnum.swift:105` and `FunctionCallInstance.swift:88`) — a removal that updates only one of the two reproduces the Osmosis breakage.
- Verify discoverability parity first: confirm Swap surfaces the mint for each of the ten chains the legacy form covered, and file a follow-up for any gap **before** deleting.

## Acceptance

- No route to `.securedAsset` remains; Swap still mints SECURE+ on every chain that previously offered the function.
- No chain is left with an empty `getCases` while still in `memoChains`.

Part of #5124.

gastonm5:
## Problem

SECURE+ mint ships twice. Swap already performs it: a same-underlying selection (hold BTC, pick secured BTC) is detected by `SwapCryptoLogic.isSameUnderlyingSecuredMint`, skips the pool quote for a synthetic ~1:1 "Mint (SECURE+)" quote, and builds the real deposit through `DefaultSwapInteractor.buildSwapKeysignPayload` → `ThorchainRouterDepositBuilder.buildSecuredMintPayload`. The legacy `securedAsset` Function built the identical `SECURE+:<thor address>` deposit from a second form, with its own copy of the inbound / router / halt handling.

## Parity evidence — checked per chain *before* deleting

A chain keeps the mint only if four independent things line up. None is derived from the others, so each was checked separately:

1. `Chain.isSwapAvailable` — the Swap action exists for that chain (`Model/Chain.swift:107`).
2. The destination picker lists the secured twin: `SecuredAssetTokenProvider` is registered at startup (`App/VultisigApp.swift:79`) and emits one `.thorChain` bucket from `SecuredAssetCatalog`; the destination side of the picker merges every registered `DestinationTokenProvider` (`SwapCoinSelectionViewModel.swift:267-273`), and the swap entry passes **no** source-chain filter to the chain carousel (`SwapCoinPickerView.swift:253-260`; only Limit Swap passes one). So THORChain + its secured assets are selectable as a destination from any source chain.
3. `isSameUnderlyingSecuredMint` matches `fromCoin.swapAsset` against `<securedChain>.<securedSymbol>` derived from the denom — a string match that has to hold chain by chain.
4. The payload builder is chain-agnostic (`buildTransfer` + memo) and repeats the deleted form's inbound resolution, including the halted-inbound and missing-router gates (`ThorchainRouterDepositBuilder.resolveInboundDestination`).

| Chain | `isSwapAvailable` | Secured denom (in the offline fallback list) | `fromCoin.swapAsset` | Detected as mint |
|---|---|---|---|---|
| Bitcoin | yes | `btc-btc` | `BTC.BTC` | yes |
| Bitcoin-Cash | yes | `bch-bch` | `BCH.BCH` | yes |
| Litecoin | yes | `ltc-ltc` | `LTC.LTC` | yes |
| Dogecoin | yes | `doge-doge` | `DOGE.DOGE` | yes |
| Ethereum | yes | `eth-eth` (+ `eth-usdc-…`, `eth-usdt-…`) | `ETH.ETH` | yes |
| Avalanche | yes | `avax-avax` (+ `avax-usdc-…`) | `AVAX.AVAX` | yes |
| BSC | yes | `bsc-bnb` | `BSC.BNB` | yes |
| Base | yes | `base-eth` (+ `base-usdc-…`) | `BASE.ETH` (native ETH is special-cased in `Coin.swapAsset`) | yes |
| Ripple | yes | `xrp-xrp` | `XRP.XRP` | yes |
| THORChain | yes | — | — | n/a — see below |

Each row is asserted by the new test, so a rename in any of those derivations fails the build rather than the user. The denoms listed are the **offline fallback** (`SecuredAssetCatalog.fallbackDenoms`) — the worst case when `/thorchain/securedassets` is unreachable; the live list is ~2.5× larger. Every retired chain keeps its native mint even in that degraded state.

ERC-20 sources are covered as well: `isSameUnderlyingSecuredMint` compares the contract case-insensitively (pinned by `SecuredMintRoutingTests`), and the mint builder synthesises the router-deposit shim + approve payload exactly as the deleted form did.

**THORChain is not a gap.** A SECURE+ deposit only mints when an L1 asset lands on a THORChain inbound. With a THORChain source the deleted form resolved `toAddress` to the vault's *own* address (`FunctionCallSecuredAsset.fetchInboundAddressAndSetupApproval`, THORChain early-return) and emitted a self-addressed deposit carrying `SECURE+:<own thor address>` — no L1 leg, nothing mints. Swap correspondingly never classifies a THORChain source as a same-underlying mint. So for THORChain the justification is "removing an operation that could not work", not parity. The redemption direction, `withdrawSecuredAsset` (SECURE−), is untouched and stays on THORChain.

### Two honest caveats

- **Regional (filed as #5159).** `SwapFeatureGate.canSwap()` is `false` in GB / JP / MY, which strips `.swap` from `Chain.defaultActions` while leaving `.memo`. In those regions the legacy form was the last in-app route to a SECURE+ mint, and this PR removes it. That reads as the gate working as intended — a SECURE+ mint *is* a THORChain deposit — rather than something to preserve, but it is a real behaviour change and needs a maintainer's call, so it is filed separately rather than buried here. The same shape of lever exists remotely (`api.vultisig.com/actions/default.json` can disable `swap` per chain; today it does so only for Solana, which never offered the Function).
- **Offline + not-yet-held tokens.** The deleted form minted whatever coin the chain-detail screen had selected and needed no catalog entry. Swap needs the secured twin to be *listed*. Online that is the full `/securedassets` universe; if that endpoint fails, the fallback carries all nine native twins but only four tokens (ETH USDC/USDT, Base USDC, AVAX USDC). So a securable token the vault does not already hold, on a failed catalog fetch, is not selectable. Note the form's version of this "capability" was unvalidated — it would happily build a SECURE+ deposit for an asset THORChain cannot secure, which the network refunds.

## What was deleted

- `FunctionCallSecuredAsset` + its co-located `SecuredAssetFormView` (306 lines) and its `FunctionCallSubModel` conformance.
- `FunctionCallSecuredAssetTests`, plus the `.securedAsset` entries in `FunctionCallContentViewTests` and `FunctionCallInstancePolymorphismParityTests`.
- The `.securedAsset` arm from `FunctionCallType` (case + `display()` + both `getCases` arms), from `FunctionCallInstance` (case + `model` dispatch), and from the two view switches (`FunctionCallContentView`, `FunctionCallDetailsScreen`).
- Seven now-orphaned locale keys — `Secured Assets`, `mintSecuredAsset`, `targetAsset`, `thorAddressAutoFilled`, `generatedMemo`, `erc20ApprovalRequiredMessage`, `mintTransaction` — from **all 8 locales** (Korean included). Every locale went 1803 → 1796 entries; `sort_localizable.py` re-run afterwards. Keys still used elsewhere (`erc20ApprovalRequired`, `approvalTransaction`, `thorAddressNotFound`) were kept.

**Both default factories.** `.securedAsset` was not the default of any chain, so neither `FunctionCallType.getDefault` nor `FunctionCallInstance.getDefault` needed re-pointing — but the pairing is now pinned by a test instead of by inspection, since nothing in the compiler couples them (the Osmosis breakage came from exactly that gap).

**No chain is left with an empty `getCases` by this change.** THORChain keeps five functions; the nine L1 chains keep `addThorLP`. (`thorChainChainnet` / `thorChainStagenet` sit in `memoChains` with empty cases *already on `main`* — pre-existing, and fixed by #5154, not here.)

## Why it matters beyond the deletion

The nine L1 chains drop from two functions to **one** (`addThorLP`). That is the precondition for the follow-on rule "a chain with a single action routes the button straight to the operation, skipping the action-list screen".

## Merge order — read before merging

Draft PR #5154 (`refactor/fn-dead-arms-5138`) edits the same `FunctionCallType.getCases` / `getDefault` switches and `CoinAction.memoChains`. It is not merged, and this branch is cut from plain `main`.

**Merge #5154 first, then rebase this branch.** Expect a small textual conflict in the same switch statements (both remove arms from the THORChain and Cosmos cases). #5154's `FunctionCallReachabilityTests` and this PR's `SecuredMintRetirementParityTests` are complementary, not duplicates: the former sweeps every chain in `memoChains` for the reachability invariant, the latter pins the Swap-side parity that justifies the deletion plus the exact post-deletion case lists. Nothing was branched off or cherry-picked from #5154.

## New test

`VultisigAppTests/FunctionCall/SecuredMintRetirementParityTests.swift`:

- `testSwapDetectsTheSecuredMintForEveryRetiredChain` — the same-underlying detection holds for all nine chains, with the destination built through the production `SecuredAssetMapper`, so a change to the denom → `CoinMeta` derivation fails here rather than in the field. Deliberately scoped to the detection; the provider/quote/payload legs are covered by `SecuredAssetTokenProviderTests` / `SecuredMintRoutingTests`, and the test says so.
- `testSwapIsAvailableOnEveryRetiredChain` / `testOfflineCatalogFallbackCoversEveryRetiredChain` — the two preconditions for the route being reachable, including with THORNode unreachable.
- `testRetiredChainsOfferAddThorLPOnly` / `testThorchainKeepsItsRemainingFunctions` — the post-deletion case lists, plus the assertion that each chain's default is offered by its own dropdown and that both default factories build the same function.

## Verification

- `make generate` re-run after the file deletions (`project.pbxproj` is generated; never hand-edited).
- SwiftLint: `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 26 violations, 0 serious, **none in any file this PR touches** (all pre-existing on `main`).
- Full unit-test suite, whole `VultisigApp` scheme (no subset), dedicated simulator, worktree-local DerivedData — run on both commits:
  - `0886abd`: `** TEST SUCCEEDED **` — Executed 6054 tests, 11 skipped, 0 failures.
  - `241008c` (final HEAD, after the review fixes): `** TEST SUCCEEDED **` — Executed 6054 tests, 11 skipped, 0 failures. All five new parity tests present and passing.
- Disk (`/System/Volumes/Data`): 37Gi free before the first run → 33Gi after; 31Gi before the re-run → 30Gi after. Never near zero, so neither result is a stale-artifact read. DerivedData deleted afterwards.

## Codex review ledger

Read-only `codex exec` review of `0886abd`, then triaged against the code:

| # | Finding | Verdict |
|---|---|---|
| 1 | High — SECURE+ becomes unreachable in GB/JP/MY, because `SwapFeatureGate` hides Swap while leaving the Function button | **Accepted.** Independently found before the deletion. Filed as #5159 for a maintainer's call; documented above. Not fixed in code: restoring the form would defeat the issue, and the gate plausibly *should* cover a THORChain deposit. |
| 2 | High — THORChain source has no Swap parity and the test silently excludes it | **Accepted in part.** The exclusion was intentional and is now explained where the parity table is defined: a THORChain-source SECURE+ could never mint, so there is nothing to hold parity with. Codex's own conclusion agrees the arm can go — on "invalid operation" grounds, which is how the PR now frames it. |
| 3 | Medium — token parity is weaker than the form's during a `/securedassets` outage | **Accepted, documented, not fixed.** Real but narrow (offline **and** token **and** not-yet-held). Expanding `fallbackDenoms` is a separate change; the live catalog is the primary path. |
| 4 | Medium — the test's names claim more than they assert ("routes" vs. detection only) | **Fixed** in `241008c`. Renamed to `testSwapDetects…` and the covered/uncovered surface written down. |
| 5 | Low — stale comments: sub-model protocol header counts 13 sub-models; interactor says the mint builder is shared with the Function-Call verify path | **Fixed** in `241008c` (12 sub-models; only the router-deposit shim is still shared, for LP adds). |

Codex also confirmed independently: no new default-factory mismatch and no new empty `memoChains` entry introduced; the seven removed locale keys have no remaining references and no still-referenced key was removed; no dangling `FunctionCallSecuredAsset` / enum-case / form-view reference remains. It flagged two **pre-existing** factory inconsistencies it wants on record — Osmosis (offers `.cosmosIBC`, both defaults fall through to `.custom`) and the `sTCY`/`yTCY` disagreement between `ticker.contains("TCY")` and `== "TCY"`. The first is fixed by #5154; the second is untouched by both PRs.

Closes #5141

## Summary by CodeRabbit

- **Removed Features**
  - Removed secured-asset minting from the Function screen and related flows.
  - Retained secured-asset withdrawals on THORChain.
  - Updated available function options across supported chains.

- **Localization**
  - Removed obsolete approval, minting, memo, target-asset, and automatic-address messages.

- **Documentation**
  - Clarified descriptions related to secured-asset deposits and liquidity-pool additions.

- **Quality Improvements**
  - Added coverage verifying the retired minting flow is unavailable where intended.


johnnyluo (approved):
LGTM

PR #5155, opened 2026-08-13
refactor(functions): delete the write-only contract-type selector

gastonm5:
## Problem

The legacy Functions screen rendered **two** dropdowns stacked on top of each other. The upper one — the contract-type selector — did nothing.

`selectedContractMemoType` was declared, defaulted in `setupForm()`, and rendered. It was read by nothing: it never reached form validation, the memo, `transactionType`, or signing. Selecting a value changed no output.

On top of that, `FunctionCallContractType.getCases(for:)` returned `[]` for every chain except THORChain, Maya and dYdX — and those three returned exactly one case. The dropdown's `isActive` guard is `items.count > 1`, so the row was **permanently disabled on every chain**, showing a fixed "message deposit" label that no user could ever change.

## Root cause

The enum dates back to the original memo screen, where the contract type was meant to pick the message envelope. That responsibility moved into `FunctionCallInstance` / `FunctionCallType` when the sub-models were introduced, but the selector was never removed — leaving a write-only state variable and a dead UI row wired into the screen.

## Fix

Deletes the layer whole, with no behaviour change:

- `Features/FunctionCall/Models/FunctionCallContractTypeEnum.swift` (49 lines)
- `Components/FunctionCallContractSelectorDropDown.swift` (98 lines)
- the `@State`, its default write in `setupForm()`, and the `contractSelector` view in `FunctionCallDetailsScreen`
- the now-orphaned `messageDeposit` / `messageVote` keys, removed from all 8 shipping locales (en, de, es, hr, it, ko, pt, zh-Hans) and re-sorted with `sort_localizable.py` — every locale went 1803 → 1801 entries, so parity is preserved

173 deletions, 0 insertions. `make generate` re-run after the file deletions; `project.pbxproj` untouched by hand.

The screen now renders one dropdown (the function selector) instead of two. The only visual delta is the removal of that 48pt row plus one 16pt `VStack` gap; every function form behaves identically.

## How I proved it was write-only

Before deleting anything, grepped the whole repo for `selectedContractMemoType`, `FunctionCallContractType` and `contractSelector` across `*.swift`, `*.strings` and `*.yml`. Every single hit was a write or a render, confined to the three files this PR touches:

- the enum's own declaration and its `getCases` / `getDefault` / `getDescription` members
- the dropdown component's own bindings
- the screen's `@State` declaration, its `setupForm()` default write, and the `contractSelector` render

[...]

PR #5161, opened 2026-08-13
fix(function-call): disclose the real network fee on Verify for EVM/UTXO/Cardano

### LINKED ISSUE #5140: [Fix] Verify discloses a zero network fee for EVM/UTXO function calls [FunctionTransaction]
## Summary

Function-call transactions on EVM / UTXO / Cardano chains show a **zero network fee** on the Verify screen. The user approves a fee they were never shown; signing then re-fetches a real one and charges it.

## Mechanism

`SendCryptoLogic.displayFee` deliberately reads `fee`, not `gas`, for EVM / UTXO / Cardano (`Features/Send/Logic/SendCryptoLogic.swift:312-315`).

- **Legacy path**: `FunctionCallDetailsScreen.loadGasInfo()` copies only `chainSpecific.gas` (`:244-251`), and `SendTransaction.empty` leaves `fee` at zero.
- **New path**: `TransactionBuilder.buildSendTransaction(vault:)` hardcodes `fee: .zero`, and `FunctionTransactionScreen` copies only `gas`.

The new pipeline is currently masked because every shipping DeFi builder is THORChain / Cosmos / Solana / TON, where `displayFee` returns `gas`.

## Why this blocks migration work

The moment **Add THORChain LP** or **SECURE+ mint** moves onto `FunctionTransaction`, the zero-fee disclosure moves with it — on the two operations that run on BTC and ETH. This must land **before** any EVM or UTXO operation is migrated.

## Scope

- Populate `fee` alongside `gas` wherever the function-call pipeline fetches chain-specific data, for every chain family `displayFee` reads `fee` for.
- Prefer fixing it once in the shared path rather than per builder.
- Cover the legacy screen too, so the disclosure is correct in the interim.

## Acceptance

- A test asserting a non-zero `displayFee` on Verify for an EVM function call and a UTXO function call.
- Manual check: THORChain LP add from ETH and from BTC both show a fee on Verify that matches what signing charges.

Part of #5124.

gastonm5:
## ⚠️ Needs human + on-device review

This changes **what a user is shown before they approve a signature**. Please check on device: a THORChain LP add from **ETH** and from **BTC** must both show a fee on Verify that matches what signing charges.

## Problem

A function-call transaction on **EVM, UTXO or Cardano** reached the Verify screen with a network fee of **zero**. The user approved a fee they were never shown, and signing then re-fetched a real one and charged it.

## Mechanism

`SendCryptoLogic.displayFee` reads **`fee`** for `chainType == .EVM | .UTXO | .Cardano` and **`gas`** for everything else; `FunctionCallViewModel.feesInReadable` derives the fiat figure from the same call, and `CryptoAmountFormatter.feesInReadable` reads `fee` on *every* chain.

Nothing downstream of `FunctionCallRoute.verify` re-resolves either figure for display — `FunctionCallVerifyScreen` renders `transaction.gasInReadable` off the `let transaction` it was handed, and `FunctionCallVerifyViewModel` only re-fetches chain-specific data inside `createKeysignPayload`, i.e. at signing. **The hand-off is the only chance to get the disclosure right.**

Three seams built that hand-off, and each copied `chainSpecific.gas` alone, leaving `fee` at the `.zero` it was constructed with:

| Seam | What it did |
|---|---|
| `FunctionCallDetailsScreen.loadGasInfo()` (legacy Functions screen) | fetched against an **empty probe** `SendTransaction.empty`, copied `.gas` into `toSendTransaction(coin:vault:gas:)` |
| `FunctionTransactionScreen.onVerify` (DeFi pipeline) | `TransactionBuilder.buildSendTransaction(vault:)` hardcodes `fee: .zero`; the screen then copied `chainSpecific.gas` |
| `DefiChainScreenModel.buildVerifyTransaction(for:)` | the same block again, with a comment claiming "gas is re-fetched during Verify". It is not. |

Two consequences, not one:

1. **EVM / UTXO / Cardano disclosed `0`** — `displayFee` reads `fee`, which nobody set.
2. **Even the `gas` that was set was priced wrong on EVM**, because the legacy probe carried no memo, amount or recipient, so the estimate sized a bare transfer rather than the contract call being signed.

The new pipeline is currently **masked**: every shipping DeFi builder is THORChain / Cosmos / Solana / TON, where `displayFee` returns `gas` and `BlockChainSpecific.fee == gas`. That masking is exactly why this must land **before** Add-LP or SECURE+ moves an EVM or UTXO operation onto `FunctionTransaction`.

## Fix, and why there

New `FunctionCallFeePricer` (`Features/FunctionCall/Services/`) resolves both figures for a built transaction and stamps them together. It is reached through **one** new `TransactionBuilder.buildPricedSendTransaction(vault:)` — a default protocol-extension method, so every builder present and future is covered without a per-builder change — and through `FunctionCallViewModel.pricedForVerify` for the legacy screen, which now prices the **real built transaction** on Continue (behind its existing loading overlay) instead of navigating with the probe's `gas`.

- **Off EVM**: `SendCryptoVerifyLogic.calculateFee`, the resolver the Send flow's own Verify screen uses. A UTXO/Cardano fee is `rate × the planned size` — the sat/vB rate on `chainSpecific` is a rate, not a fee — and every other chain quotes a flat per-unit gas that is the whole cost.
- **On EVM**: the same `BlockChainService.fetchSpecific(tx:)` the **signing** path makes. `fetchSpecificForEVM` floors every transaction at the 120,000 ERC-20 gas limit, while the Send resolver floors a *native* transfer at `Coin.feeDefault` (23,000 on Ethereum) — so pricing a native ETH call through the Send resolver understates by up to 5x against the limit the vault signs, and disagrees with the co-signer, whose `JoinKeysignGasViewModel` prices EVM straight off `payload.chainSpecific.fee`. This was caught by the review pass and fixed in the second commit.

**Why the hand-off and not the Verify screen.** Re-pricing on Verify (the way `SendVerifyScreen` calls `loadGasInfoForSending`) would be one place for every entry point — and wrong here. Several flows deliberately stamp a fee a fresh fetch does not produce: the Cosmos staking flow stamps `CosmosStakingConfig.scaledFeeAmountBigInt`, which the SignDoc resolver actually signs, and a THORChain limit-order cancel stamps `THORChainConstants.depositGasBaseUnits` because `thorchain.swift` hardcodes it at signing regardless of what was fetched. A blanket re-fetch would clobber both and re-break disclosures that were deliberately fixed.

Also: the Cosmos staking branch now stamps `fee` alongside `gas` (the same number — a Cosmos SignDoc's per-unit gas is its whole cost), so the fiat string beside that row stops reading `$0.00`.

## Residual gaps

- **A failed fee fetch still discloses whatever the flow already had**, which on EVM/UTXO/Cardano is still zero. This is unchanged pre-existing behaviour and is now logged rather than silently swallowed, but it is not *fixed*. Mitigating fact: after the EVM change, a pricing failure on those three families is the same fetch signing needs, so the signature cannot complete either. An honest "fee unavailable" state on the Verify row (with Sign gated on it) is a separate change across three screens — happy to do it in a follow-up if you'd rather have it here.
- **Adjacent, deliberately not touched** (both found by the review pass, both worth their own issue): the two QBTC governance-vote routes in `DefiChainScreenModel` build from `SendTransaction.empty` and navigate to this same Verify screen with both figures at zero; and `EditReferralDetailsViewModel` never fetches gas at all, so a referral **edit** discloses `0 RUNE` (referral **create** does fetch it). Both are a different mechanism — a fixed signer constant, not a fetch — in another feature.

## Tests

`VultisigAppTests/FunctionCall/FunctionCallFeePricerTests.swift`, 10 cases. They pin the **disclosed** figure (`SendCryptoLogic.displayFee`, the exact value the fee row and the fiat figure are derived from) to exact values, not `> 0`:

- EVM: `120,000 × 1.32 gwei = 0.0001584 ETH`, plus a case asserting the Send resolver is **never** consulted for an EVM function call.
- UTXO: the planned `3,000` sats, explicitly *not* the `12` sat/vB rate.
- Cardano: the planned `174,301` lovelace.
- THORChain: regression guard that the figure every shipping DeFi builder discloses today does not move.
- Memo/recipient forwarding on both EVM and UTXO — the fee is priced for *this* call, not a probe.
- The builder seam (`buildPricedSendTransaction`) for a stub EVM and a stub UTXO builder — the shape the Add-LP / SECURE+ migration will land on.
- The failure path, which documents the residual gap above.

## Verification

- Full scheme: `** TEST SUCCEEDED **` (`xcodebuild test -scheme VultisigApp`, iPhone 16 sim, worktree-local `-derivedDataPath`). No failures anywhere in the suite.
- Disk before the gate: `38Gi` avail → after: `36Gi` avail (`/System/Volumes/Data`). Nowhere near exhaustion, so the marker is trustworthy.
- SwiftLint clean on every changed file.
- `make generate` run after adding the two new files.

## Cross-model review ledger (Codex, read-only)

| Finding | Verdict |
|---|---|
| EVM disclosure ≠ the signed gas limit (23,000 vs 120,000 floor) | **Fixed** (commit 2) — priced off the signing fetch; test pins that the Send resolver is never consulted |
| No Cardano test | **Fixed** — `testACardanoFunctionCallDisclosesThePlannedFee` |
| Fail-open on a pricing failure still discloses zero | **Deferred**, documented above and in the wiki plan |
| Referral edit + QBTC governance routes unpriced | **Rejected for this PR** — real, but a different mechanism in another feature; listed as adjacent |
| Does UTXO `gas` becoming the planned total break a consumer? | **Confirmed none**: `SendTransaction.byteFee` has no production call site, `isAmountExceeded` reads `fee` for UTXO/Cardano, `canBeReaped` is effectively Polkadot-only, and the keysign payload carries a freshly fetched rate |
| Actor isolation / SwiftData / conventions | No violations found |

## Merge order

Branched off plain `main`. Draft #5155 (`refactor/fn-contract-type-5137`) also edits `FunctionCallDetailsScreen.swift`, in the contract-selector region rather than in `loadGasInfo()` / the Continue button, so a conflict is possible but should be trivial — **this may need a small rebase after #5155 lands**.

Plan and rationale: `wiki/pages/projects/vultisig/functions-to-defi-migration/fee-disclosure-5140-plan.md`.

Closes #5140

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/DefiChainScreenModel.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiChain/ViewModel/DefiChainScreenModel.swift
@@ -209,18 +209,11 @@ final class DefiChainScreenModel: ObservableObject {
         try await CoinService.addToChain(assets: type.coins, to: vault)
     }
 
-    /// Builds the unsigned tx and pre-fetches the chain-specific gas so Verify
-    /// shows the fee immediately. Mirrors `FunctionTransactionScreen.onVerify`:
-    /// the gas is re-fetched during Verify anyway, so a fetch failure here is
-    /// non-fatal and the un-gassed tx is returned unchanged.
+    /// Builds the unsigned tx and prices it, so Verify discloses the fee the
+    /// user is agreeing to pay. Mirrors `FunctionTransactionScreen.onVerify` —
+    /// both go through the one shared pricing step, because nothing downstream
+    /// of Verify re-resolves these figures for display.
     func buildVerifyTransaction(for builder: TransactionBuilder) async -> SendTransaction {
-        var sendTx = builder.buildSendTransaction(vault: vault)
-        do {
-            let chainSpecific = try await BlockChainService.shared.fetchSpecific(tx: sendTx)
-            sendTx = sendTx.copy(gas: chainSpecific.gas)
-        } catch {
-            // Non-fatal: gas is re-fetched during Verify.
-        }
-        return sendTx
+        await builder.buildPricedSendTransaction(vault: vault)
     }
 }
--- a/VultisigApp/VultisigApp/Features/FunctionCall/Services/FunctionCallFeePricer.swift
+++ b/VultisigApp/VultisigApp/Features/FunctionCall/Services/FunctionCallFeePricer.swift
@@ -24,23 +24,45 @@ private let logger = Log.send.service
 /// step instead of a `chainSpecific.gas` copy repeated at each navigation seam —
 /// there were three of them, and every one of them was wrong the same way.
 ///
-/// The figures come from `SendCryptoVerifyLogic.calculateFee`, the resolver the
-/// Send flow's own Verify screen uses, so a function call is priced the way a
-/// send is: an EVM fee is `maxFeePerGas × the limit estimated for THIS call`, a
-/// UTXO/Cardano fee is `rate × the planned size` (the sat/vB rate on its own is
-/// a rate, not a fee), and every other chain quotes a flat per-unit gas that IS
-/// the whole cost. Mirroring it is what makes "what Verify says" and "what
-/// signing charges" the same number.
+/// Off EVM the figures come from `SendCryptoVerifyLogic.calculateFee`, the
+/// resolver the Send flow's own Verify screen uses: a UTXO/Cardano fee is
+/// `rate × the planned size` (the sat/vB rate on `chainSpecific` is a rate, not
+/// a fee), and every other chain quotes a flat per-unit gas that IS the whole
+/// cost.
+///
+/// ⚠️ **EVM is priced off the chain-specific fetch the SIGNING path makes**, not
+/// off that resolver, and the difference is not cosmetic. A function call is
+/// signed from `FunctionCallVerifyViewModel.createKeysignPayload` →
+/// `BlockChainService.fetchSpecific(tx:)`, whose EVM branch floors every
+/// transaction at the 120,000 ERC-20 gas limit. The Send resolver instead
+/// floors a native transfer at `Coin.feeDefault` (23,000 on Ethereum). Pricing
+/// a native ETH call through the Send resolver therefore discloses as little as
+/// `maxFeePerGas × 23,000` for a transaction the vault signs with a limit of
+/// 120,000 — understating the fee the user approves, on the screen where they
+/// approve it. Reading the same fetch also makes this device agree with the
+/// CO-SIGNER, whose `JoinKeysignGasViewModel` prices EVM straight off
+/// `payload.chainSpecific.fee` (`maxFeePerGas × gasLimit`).
 ///
 /// It prices the REAL transaction — memo, amount and recipient included — not a
 /// bare probe. On EVM that is the difference between the gas limit of the
 /// contract call and the 21,000 of a plain transfer.
 @MainActor
 struct FunctionCallFeePricer {
-    private let logic: SendCryptoVerifyLogic
+    /// The chain-specific fetch the function-call signing path performs.
+    /// Injectable so the disclosure can be tested without a network.
+    typealias SigningChainSpecificFetch = (SendTransaction) async throws -> BlockChainSpecific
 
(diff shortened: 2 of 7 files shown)


johnnyluo (approved):
LGTM

Issue #5153, opened 2026-08-13
[Functions] Delete the legacy FunctionCall shell and rename the shared verify route


gastonm5:
## Summary

Final step. Once every operation has moved, delete the legacy tree and give the shared tail an operation-neutral name.

## Delete

- `FunctionCallDetailsScreen`, `FunctionCallContentView`, `FunctionCallRouteBuilder`
- `FunctionCallType`, `FunctionCallInstance`, `FunctionCallSubModel`
- the remaining sub-model files under `Features/FunctionCall/Models/` and their co-located `FormView`s
- the `preselected` parameter added to the route for the migration interim
- `CoinAction.memo` and its `memoChains` list, if nothing else consumes them by then

## Extract before deleting

`struct IdentifiableString` is declared **inside** `FunctionCallBondMayaChain.swift:15` and used by 8 other production files plus 5 test files. Extract it to its own file first, or the build breaks somewhere that looks unrelated to the change. (Already flagged in #5125.)

## Rename

`FunctionCallRoute` and `FunctionCallVerifyScreen` are **not** legacy — they are the shared seam for the DeFi tab, Referral and limit-order cancel, and they survive. The name misleads once `FunctionCall` no longer exists. Rename both to something operation-neutral in the same PR.

## Acceptance

- No type named `FunctionCall*` remains except the renamed shared route and verify screen.
- Every operation is reachable and signs identically to before.
- Full test suite green.

Part of #5124.

PR #5163, opened 2026-08-13
refactor(functions): migrate node REBOND to the FunctionTransaction architecture

### LINKED ISSUE #5144: [Functions] Migrate REBOND to the FunctionTransaction architecture
## Summary

Move rebond off `FunctionCallReBond` onto the shared architecture. THORChain only.

## Operation

Memo (`FunctionCallReBond.swift:62-71`):

```
REBOND:<currentNode>:<newNode>            // whole bond
REBOND:<currentNode>:<newNode>:<amount>   // partial, amount x 10^8, integer
```

Fields: current node address, new node address, optional amount.

## Defects to fix in this migration

- `validate(against:)` sets `customErrorMessage` but `isTheFormValid` never consults it (`:34,42`), and the conformance ignores the coin — the "REBOND requires RUNE" guard is display-only. Express it as a real validator on the new form.
- The optional-amount branch must keep its `x 10^8` integer conversion **exactly**; pin it with a builder test for both the whole-bond and partial-bond memos.

## Notes

Confirm the intended THORChain semantics of the operator-fee warning copy (`rebondNote`) before porting it — the current wording suggests behaviour the form does not enforce.

## Architecture contract

Part of #5124.

gastonm5:
## What

Migrates the **node REBOND** operation (THORChain only) off the legacy `FunctionCallReBond` sub-model onto the shared `Features/FunctionTransaction/` architecture: typed intent → ViewModel → pure `TransactionBuilder` → `buildSendTransaction(vault:)`, with the existing verify → keysign tail unchanged.

Step 8 of #5124, applying the recipe #5162 established. Written up in the wiki at `projects/vultisig/functions-to-defi-migration/migrate-rebond-5144-plan.md`.

> **Stacked on #5162** (`refactor/migrate-leave-5143`), which introduces the seam file, the grouped `break` arm and `FunctionCallMigrationSeamTests`. **Must merge after it.**

## The operation, pinned

REBOND has two memo shapes and the partial one is a fixed-point conversion, so both get a builder test.

| | Whole bond | Partial bond |
|---|---|---|
| Memo | `REBOND:<nodeAddress>:<newAddress>` | `REBOND:<nodeAddress>:<newAddress>:<amount × 10⁸>` |
| Attached amount | `"0"` | `"0"` |
| `memoFunctionDictionary` | `nodeAddress`, `newAddress`, `memo` (3) | + `rebondAmount` (4, human decimal) |
| `toAddress` | empty (`MsgDeposit` is addressed by memo) | empty |
| `transactionType` | `.unspecified` | `.unspecified` |

The conversion is carried over verbatim:

```swift
NSDecimalNumber(decimal: rebondAmount).multiplying(byPowerOf10: 8).int64Value
```

The literal `8` is legacy's, not `coin.decimals`. REBOND is THORChain-only, RUNE is 8-decimal, and THORChain memo amounts are 1e8 base units, so the three coincide — but the tests pin the exponent rather than trusting that:

| Input | Memo tail | Pinned by |
|---|---|---|
| *(empty)* | *(none)* | `testWholeBondMemoOmitsTheAmountSegment` |
| `100` | `:10000000000` | `testPartialBondMemoAppendsTheAmountInBaseUnits` (the legacy golden fixture) |
| `0.00000001` | `:1` | `testOneBaseUnitEncodesAsOne` |
| `1.234567891` | `:123456789` | `testSubBaseUnitPrecisionIsTruncatedNotRounded` |
| `0.000000005` | `:0` | `testDustBelowOneBaseUnitEmitsAZeroSegmentRatherThanTheWholeBond` |

`int64Value` **truncates**, so dust below 1e-8 scales to zero. The builder gates the segment on the *decimal* being `> 0`, not on the scaled integer — exactly as legacy did — so dust emits `:0` (a rebond of nothing) rather than falling through to the whole-bond shape, which would move the entire stake. **The safe fallback is the one that moves less.** The form rejects that input upstream anyway (below).

The attached amount stays the hardcoded `"0"`: `SendTransaction.amountInRaw` treats the builder's `amount` as a **human decimal** and multiplies by `10^decimals`, so anything non-zero would attach real RUNE to a `MsgDeposit` that has no return path. Legacy `FunctionCallReBond.amount` was `.zero` regardless of the rebond amount; `testAttachedAmountIsZeroForBothMemoShapes` pins it for both shapes.

## ⚠️ REBOND was THORChain's default — re-pointed in both factories

`FunctionCallType.getDefault(for:)` returned `.rebond` for every non-TCY THORChain ticker. The route-out fires on a **change** of selection, so a migrated default opens the legacy screen on a function that builds no form at all. #5162's `testNoChainDefaultsToAMigratedFunction` sweeps `Chain.allCases` and went red the moment `.rebond` entered the migration map — the test doing precisely its job.

**Both factories now default THORChain to `.custom`**, and the `TCY` ticker branch collapses away:

| Factory | Before | After |
|---|---|---|
| `FunctionCallType.getDefault` | `.custom` for TCY, `.rebond` otherwise | `.custom` |
| `FunctionCallInstance.getDefault` | `.custom(…)` for TCY, `.rebond(…)` otherwise | `.custom(…)` |

Why `.custom` and not another THORChain entry:

- It is in THORChain's `getCases`, so the dropdown can actually show it — the reachability invariant #5154 pins. New test: `testThorchainDefaultIsOfferedByTheDropdown`.
- It was **already** the THORChain default for TCY, so that half of THORChain users land exactly where they always did.
- Every alternative is disqualified: `.leave` is migrated, **`.merge` is being migrated concurrently by #5145** (picking it would turn that branch red), and `.unmerge` / `.securedAsset` / `.withdrawSecuredAsset` are niche operations that make a worse landing page than the general-purpose memo form.

A new test, `testThorchainDefaultsAgreeAcrossBothFactories`, pins that the two switches name the same function — they run back to back in `setupForm()`, and a disagreement puts one function's name over another's form. It uses an exhaustive `FunctionCallInstance` → `FunctionCallType` mapping, so adding an enum case will not compile until the mapping names it.

### 1. The "REBOND requires RUNE" guard was display-only

```swift
// deleted FunctionCallReBond.swift
func validate(against coin: Coin) {
    if coin.chain != .thorChain || !coin.isNativeToken {
        customErrorMessage = "rebondRequiresRune".localized   // ← written
    } ...
}
var isTheFormValid: Bool {          // ← never reads customErrorMessage
    rebondAmount >= 0 && …addresses…
}
```

`FunctionCallReBond: FunctionCallSubModel` bridged `isFormValid(for:)` straight to `isTheFormValid` and dropped the coin, so a non-RUNE selection showed the warning and submitted anyway.

It is a real form validator now — a `ClosureValidator` on the node-address field that ignores the field's content on purpose, because **no address the user can type makes a non-RUNE asset a legal REBOND**, so the gate has to close on the asset alone. The field goes invalid, `transactionBuilder` returns nil, nothing navigates. `testNonRuneAssetBlocksTheBuilderEvenWithValidAddresses` pins it, and `testRuneAssetWithTheSameInputsDoesBuild` is the control so it cannot pass vacuously.

Defence in depth: the seam pins the intent to THORChain's native asset from `TokensStore` (the same fail-closed resolution LEAVE uses), so a vault holding only TCY hits `resolvingCoin`'s shared "not in vault" error before the form is built at all.

### 2. Address validation

Tightened from `isValidThorMayaTON` (THOR **or** Maya **or** TON on either field) to `AddressValidator(chain: coin.chain)` + `RequiredValidator`, as LEAVE did.

## One deliberate tightening

The optional amount now rejects a typed value that scales to zero — `0`, junk, and dust below 1e-8. Legacy accepted all three, and turned a typed `0` into the **whole-bond** memo: a fund action reached by typing a number that reads as "none". The whole-bond shape stays reachable by leaving the field empty, which is what the field's own label already instructs ("Rebond Amount (Optional - leave empty to transfer all bonded RUNE)"). This is the one place the PR is not byte-parity with the deleted sub-model, and it is called out rather than buried.

## `rebondNote` — checked, accurate, ported unchanged

The issue asked to confirm the warning copy's semantics before porting it, on the suspicion that it described operator-fee behaviour the form does not enforce. **It does not describe operator fees.** All 8 locales say the same thing:

> "Note: REBOND transactions must send 0 RUNE. The amount above only goes in the memo."

That is a statement about the **attached amount**, and it is exactly what the builder does — `amount` is the hardcoded `"0"`, the value travels only in the memo segment. So the copy is accurate *and* enforced. Ported unchanged, moved off the legacy hardcoded `.orange` / `.caption` onto `Theme.colors.alertWarning` / `Theme.fonts.caption12`. No unverified claim is carried forward.

## Screen shape

Two `FormExpandableSection`s — both node addresses in the first, the optional amount plus the note in the second. `onContinue` follows LEAVE's, not Bond's: Bond's focus-advance dance (`if amountField.valid, !address.valid { focus address; return }`) assumes a *required* amount field. With an optional one — valid while empty — it would bounce focus between sections forever and never reach `transactionBuilder`, so the address errors would never be revealed.

## Deleted

- `Features/FunctionCall/Models/FunctionCallReBond.swift` (model + co-located `ReBondFormView`)
- `FunctionCallInstance.rebond` + its `model` arm, the `FunctionCallReBond: FunctionCallSubModel` conformance, the `FunctionCallContentView` arm
- the `.rebond` sub-model construction in `FunctionCallDetailsScreen.onChange` (folded into the grouped `case .leave, .rebond: break` arm)
- **`extractNodeAddress(from:)` entirely** — rebond was its last remaining case, so no legacy form holds a node address to carry over any more and the screen passes `nodeAddress: nil`. The parameter stays on the seam as the pre-fill hook for whatever replaces the dropdown.
- `VultisigAppTests/FunctionCall/FunctionCallReBondTests.swift`, `testRebondParity` in `FunctionCallInstancePolymorphismParityTests`, and `.rebond` from `FunctionCallContentViewTests`' dispatch array — **every assertion carried over first**.

Not deleted: `.rebond` in `FunctionCallType.getCases` (the seam fires on selection — removing it would make REBOND unreachable rather than migrated; `testRebondStaysSelectableOnThorchain` pins it), and `struct IdentifiableString` inside `FunctionCallBondMayaChain.swift` — 8 production files use it, extraction belongs to #5153.

## Tests

- `VultisigAppTests/FunctionTransaction/RebondTransactionBuilderTests.swift` — both memo shapes, the 1e8 exponent at one base unit, truncation, the dust `:0` shape, the attached `"0"` for both shapes, the 3-entry / 4-entry memo dictionary, and the `buildSendTransaction` boundary.
- `VultisigAppTests/FunctionTransaction/RebondTransactionViewModelTests.swift` — empty / garbage / cross-chain addresses rejected, missing destination rejected, both memo shapes accepted, `0` / junk / dust amounts rejected, smallest representable amount accepted, non-RUNE asset blocked **with the RUNE control**, gate re-closes on clear, pre-fill, address-result handling.
- `VultisigAppTests/FunctionCall/FunctionCallMigrationSeamTests.swift` — rebond → intent, the RUNE pin from a TCY selection, fail-closed when the vault holds no RUNE, `.rebond` still selectable, THORChain's default is offered by the dropdown, and the two default factories agree.

## Merge order / conflicts

- **Merges after #5162.** Base is `refactor/migrate-leave-5143`, not `main`.
- **#5145 (MERGE)** is stacked on the same base and touches five of the same files one line each: the migration map, the grouped case arm, `FunctionTransactionType`, `FunctionTransactionScreen`, and the locale files (this PR adds **no** new locale keys — every string it uses already ships in all 8 locales). Conflicts are mechanical. It must not re-point THORChain's default to `.merge`; this PR touches nothing merge-related.
- **#5154** pins "default ∈ getCases" — satisfied, and pinned here too.
- **#5155 / #5160** edit `FunctionCallTypeEnum.getCases`; this PR edits only `getDefault` in that file.

## Codex review

Two read-only `codex exec` rounds against the branch.

**Round 1** (commit `727d31d`) — asked specifically about memo/×10⁸ parity for both shapes, the default re-pointing across both factories, whether the RUNE guard actually blocks, and seam correctness.

| Finding | Disposition |
|---|---|
| Memo parity confirmed: identical conversion, identical truncation, identical gating on the unscaled decimal including the dust `:0`; dictionary keys, human amount and gating identical; attached amount equivalent to legacy zero at any precision | Confirmed, no action |
| Default re-pointing confirmed: both factories now `.custom`, `.custom` is in THORChain's `getCases`, dropping the TCY branch introduces no unintended ticker exception | Confirmed, no action |
| RUNE guard confirmed effective: validator order puts `AddressValidator` first, so a valid THOR address on TCY reaches the RUNE validator and throws, closing the aggregate. "Because the coin predicate is immutable, a non-native form can never first become valid" | Confirmed, no action |
| **`transactionBuilder` gates on a stale `validForm`** — the aggregate is republished through `RunLoop.main`, so an amount edited to `0` and submitted in the same turn could still build the **whole-bond** memo, defeating the tightening above | **Fixed** in `1a8c575` — the guard now also reads the freshly-validated fields, with two same-turn regression tests that assert the stale `validForm == true` precondition so they cannot pass vacuously |
| The comment left where `extractNodeAddress` used to be claimed no legacy sub-model holds a node address; Maya bond/unbond do. Deleting the extractor is still behaviour-preserving (it only ever read `.rebond`), but the comment was false | **Fixed** in `1a8c575` — comment corrected; wiring the Maya forms into the carry-over would be new behaviour, not a migration, so it is left alone |
| `DelayedFocus` does not cancel superseded callbacks, so an early tap can have focus stolen back after 0.5s | **Rejected — out of scope.** Shared component, unchanged by this branch, identical on Leave / Bond / Unbond and every other form screen. Round 2 re-verified the branch did not make it worse |

**Round 2** (whole branch `35a7423..1a8c575`) — verified every round-1 disposition (including that the two new same-turn tests genuinely fail without the guard), then raised two more:

| Finding | Disposition |
|---|---|
| The new guard fixed the stale aggregate in **one direction only**: keeping `validForm` in it meant a form completed and submitted in a single turn had its first Continue tap swallowed | **Fixed** in `55f223a` — the aggregate is derived from nothing but the fields, and `validateErrors()` refreshes them synchronously, so the guard now reads the fields alone. A mirror test covers the invalid→valid window |
| The new doc comments called REBOND "moving a bond from one validator to another" and named `NEWADDR` the destination validator; THORChain's memo documentation describes `NEWADDR` differently | **Fixed** in `55f223a` — see below |

### On `NEWADDR`

Codex is right that this PR should not assert protocol semantics it did not verify. Nothing in the change depends on the answer — the memo positions are pinned from the legacy sub-model either way — so the code no longer claims one. Builder fields are back to the legacy sub-model's `nodeAddress` / `newAddress` (which are also the memo dictionary keys), and the prose describes memo positions rather than entities. The **user-facing labels never changed**: they are the same `"nodeAddress"` / `"newAddress"` localized strings the legacy form used. **The exact meaning of `NEWADDR` on the protocol side is unverified by this PR and deliberately unasserted.**

## Verification

- `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 26 violations, all pre-existing in untouched files, **0 in any file this PR touches**.
- `make generate` run after adding/deleting files; `project.pbxproj` never edited by hand.
- **No new locale keys.** Every string the new screen uses (`Rebond`, `nodeAddress`, `newAddress`, `amount`, `rebondAmount`, `rebondAmountPlaceholder`, `rebondNote`, `rebondRequiresRune`, `emptyAddressField`, `invalidAmount`, `validAddressError`) already ships in all 8 locales, so the locale files are untouched — one less conflict surface with #5145.
- Full `xcodebuild test` on a dedicated simulator (`gate-5144`, iPhone 16), worktree-local `-derivedDataPath .build-dd`, re-run after every commit. Final run on `55f223a`: `** TEST SUCCEEDED **`, `Executed 6099 tests, with 11 tests skipped and 0 failures (0 unexpected)`. 31 of those are the new REBOND cases; the earlier runs on `727d31d` (6096) and `1a8c575` (6098) were also green.
- Disk before the final gate: `460Gi size, 399Gi used, 30Gi avail, 93% capacity`. After: `460Gi size, 400Gi used, 29Gi avail, 94% capacity`. Free space never approached zero, so the marker is trustworthy rather than a stale-product artifact. Worktree DerivedData deleted after the run.
- ⚠️ **CI on upper stack layers is unreliable in this repo — a missing check is not a passing one.** The local full-suite gate above is the trusted signal for this PR.

Closes #5144

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated THORChain rebond transaction flow.
  * Supports node and destination addresses, optional rebond amounts, validation, and prefilled node details.
  * Displays a warning that rebond amounts are encoded in the transaction memo.

* **Bug Fixes**
  * Improved migrated rebond handling with native RUNE selection and node address preservation.
  * Added safeguards for invalid addresses, unsupported assets, and amounts that round down to zero.

* **Changes**
  * Rebond operations now use the transaction flow instead of the previous function-call form.


johnnyluo (approved):
LGTM

PR #5164, opened 2026-08-13
refactor(functions): migrate Rujira MERGE to the FunctionTransaction architecture

### LINKED ISSUE #5145: [Functions] Migrate MERGE to the FunctionTransaction architecture
## Summary

Move THORChain merge off `FunctionCallCosmosMerge` onto the shared architecture.

## Operation

Memo: `merge:<denom>` (`FunctionCallCosmosMerge.swift:114-116`), with an amount.

The token list is the intersection of `ThorchainMergeTokens.tokensToMerge` with the vault's THORChain coins (`loadTokens`, matching on the denom with the `thor.` prefix stripped) — a static catalog filtered by holdings, not a discovered position. Resolve the contract from a typed descriptor in the ViewModel; do not resolve it in the view.

## Notes

`.mint` / `.redeem` in `FunctionTransactionType` are the yVault operations and are unrelated — do not reuse them. This needs its own case.

## Architecture contract

Part of #5124.

gastonm5:
Moves the THORChain Rujira **MERGE** function off the legacy `FunctionCallCosmosMerge` sub-model onto the shared `Features/FunctionTransaction/` architecture: typed intent → ViewModel → pure `TransactionBuilder` → `buildSendTransaction(vault:)`, keeping the existing verify/keysign tail.

Closes #5145. Part of #5124.

> **Stacked on #5162** (`refactor/migrate-leave-5143`) — base branch, **must merge after it**. #5162 built the route-out seam this PR plugs into; the diff against `main` will look larger than this PR's own change until that one lands.

## The three values that decide where the funds go

| | Legacy | Here |
|---|---|---|
| Memo | `"merge:" + selectedToken.value`, where `selectedToken.value` was `token.denom.uppercased()` | `merge:THOR.KUJI` — **lowercase prefix, uppercase denom**, pinned per catalog entry |
| Destination | `wasmContractAddress` of the selected denom, resolved **in the view** | same address, resolved from a typed descriptor **in the ViewModel** |
| Amount | `amount.formatToDecimal(digits:)` on the *selected token's* coin, `.thorMerge` | the form field string on the same coin, `.thorMerge` |

The memo casing is wire format, not display: the signer lowercases it again to derive the deposit denom (`buildThorchainWasmGenericMessage`), but the memo itself travels on the transaction, and a memo THORChain does not recognise is a *silently failed* merge rather than a rejected one. `MergeTransactionBuilderTests` therefore sweeps **all six** catalog entries rather than pinning one, asserting the memo and that `toAddress` is that same token's contract.

## Token list

Unchanged in substance: the static `ThorchainMergeTokens.tokensToMerge` catalog **intersected with the vault's THORChain holdings**, matched on the denom with its `thor.` prefix stripped and lowercased, in catalog order. It now lives on `ThorchainMergeAsset.mergeable(in:chain:)`, which pairs each catalog entry with the coin that funds it and the contract it deposits into — so switching tokens re-points the contract and the balance ceiling together instead of leaving one behind.

The picker reuses the existing `AssetSelectionListScreen` / `AssetSelectionDataSource` that Bond and Unbond Maya already use.

## Intent

```swift
case merge(coin: CoinMeta, denom: String?)
```

`coin` is THORChain's **native asset** — the chain anchor and the fee asset — not the coin being spent, which the form picks from holdings. A vault that cannot resolve it lands on `FunctionTransactionScreen`'s shared "not in vault" error instead of a form whose deposit it could not pay for: the same fail-closed posture #5162 settled on, and what the legacy `ensureRuneCoin()` was reaching for. `denom` optionally pre-selects a token — the legacy `preSelectToken()` tried the same thing but never fired, because the RUNE pin ran first — and gives the future action-list / DeFi merge card a "merge THIS token" hook.

## Gate

`FormScreen` deliberately does not disable Continue, so `transactionBuilder` returning nil is the enforcement. With no token picked there is no contract to deposit into, so the amount field carries a validator that fails with the existing `selectTokenToMerge` string until one is — the tap says what is missing rather than no-op'ing. Once a token is selected the validators are required + balance for **that** token, carrying the legacy submit-time predicate (`amount > 0`, `amount <= balance`, non-empty destination), evaluated against the balance read at submit.

## Deleted

`FunctionCallCosmosMerge` + its co-located `CosmosMergeFormView`, `FunctionCallInstance.merge` and its sub-model conformance, the content-view arm, the details-screen arm (its case name joins the grouped migrated `break` arm), and `FunctionCallCosmosMergeTests` — every assertion carried into the new builder / view-model tests first.

**Not** deleted: `.merge` in `FunctionCallType.getCases` — the route-out fires on *selection*, so removing it would make the operation unreachable rather than migrated; `ensureRuneCoin()`, still used by rebond; `IdentifiableString` (#5153); anything under `FunctionCallCosmosUnmerge` (#5146).

MERGE is **not** any chain's default, so neither `FunctionCallType.getDefault` nor `FunctionCallInstance.getDefault` changes here. `FunctionCallMigrationSeamTests.testNoChainDefaultsToAMigratedFunction` still sweeps `Chain.allCases` and stays green.

## No new strings

Title, labels and errors all come from keys that already ship in all **8** locales (`Merge`, `asset`, `amount`, `amountBalance`, `amountSelectToken`, `selectTokenToMerge`, `selectAsset`, `emptyAmountField`). Zero locale-file edits.

## Verification

- **Local gate: `** TEST SUCCEEDED **`, 6096 tests, 0 failures** (11 skipped), SwiftLint clean, on a dedicated simulator with worktree-local derived data.
- ⚠️ **CI on upper stack layers is unreliable in this repo — a missing check is not a passing one.** The local gate is the trusted signal here.
- Two rounds of read-only cross-model review (Codex). Round 1 raised four findings, all fixed in the second commit: the submit gate read the run-loop-deferred `validForm` instead of the flags `validateErrors()` had just written; the balance ceiling was a copy taken at selection where legacy read it at submit; the legacy non-empty-destination guard had been dropped; and the builder tests touched SwiftData models off the main actor. Round 2 confirmed those and raised one minor point — the ceiling was still stale in the *increase* direction — fixed in the third commit. Round 2 also flagged a pre-existing `contains("TCY")` vs `== "TCY"` disagreement between the two default factories for `sTCY`/`yTCY`; out of scope here, left alone.

## Conflicts

REBOND (#5144) is being migrated concurrently onto the same base and touches one line each in `FunctionCallTypeMigration.swift`, the grouped `break` arm, `FunctionTransactionType.swift`, `FunctionTransactionScreen.swift`, and the same `stillLegacy` array in the seam tests. All mechanical. That PR is the one that re-points THORChain's default in both factories; this one does not touch either.

Plan: `wiki/pages/projects/vultisig/functions-to-defi-migration/migrate-merge-5145-plan.md`

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated THORChain MERGE transaction flow.
  - Select eligible assets, enter an amount, review validation, and continue to transaction verification.
  - Automatically preselects supported assets and validates available balances.
  - Builds token-specific MERGE transactions with the correct destination and memo details.

- **Improvements**
  - MERGE operations now migrate directly into the transaction flow instead of using the legacy function-call form.

- **Tests**
  - Added coverage for asset selection, validation, balance limits, transaction construction, and migration behavior.


johnnyluo (approved):
LGTM

PR #5166, opened 2026-08-13
feat(defi): switch the Kamino attribution memo to 8k2mz

gastonm5:
## Description

Kamino specified the memo their attribution keys on: **`8k2mz`**, replacing the placeholder `vs` this app has been writing since #5117.

`KaminoAttribution.memoTag` is the single source of the literal, so the injector, the validator and the offline verify-screen decoder all move with it, and the derived `tagged` test fixtures follow automatically. The rest of the diff is the tests that deliberately do *not* derive it.

No issue — the change came from Kamino directly. Context: #5017, tag introduced in #5117.

### The tag is case-sensitive

It is compared as raw bytes and matched **whole**, so `8K2MZ` is a different memo that no downstream filter would count. The pinned-literal test asserts both the string and its exact bytes (`[0x38, 0x6b, 0x32, 0x6d, 0x7a]`) precisely because everything else in the suite derives from the constant — a drift would leave the whole suite agreeing with itself and green while the attribution silently matched nothing.

The near-miss rejection cases were re-derived from the new tag (`8k2m`, `8k2mz1`, `8K2MZ`, `8k2mz `), so prefix, suffix, case and trailing-whitespace variants are all still refused.

### Two things a tag change could have broken

**Packet size.** The tag grows by three bytes. Decoding all nine fixture vectors, the largest tagged transaction goes 799 → **802 bytes** against the 1,232-byte Solana cap — 430 bytes of headroom. `injectingMemo` also enforces the limit at runtime, so an outlier vault throws rather than producing an oversized transaction.

**Compute.** The pinned `KaminoComputeBudget` limits were measured on untagged transactions and carry >62k CU of margin; three more bytes of memo data is far inside it. Production also simulates the *final tagged* transaction before signing, so this fails loudly rather than silently.

### No migration, no dual-accept window

The old tag never shipped. `a6ba7f2b5` is not an ancestor of `v1.44.70` and nothing tagged has been broadcast, so there is no on-chain history keyed on the old bytes.

The one consequence worth naming: a device on an older `main` build that builds a `vs`-tagged transaction will have it rejected as unreadable by a co-signer on this build, and blocked from co-signing. That is the guard working — the app signs exactly one memo literal under the Memo program, and teaching the decoder a legacy tag would weaken it for a window that only exists between unreleased builds. Called deliberately; say the word if a dual-accept window is wanted instead.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

Kamino Earn deposit/withdraw on Solana. No tx link: the tagged form has not been broadcast yet, and the opt-in live test (`KAMINO_LIVE=1 testTheTaggedDepositStillFitsAndStillExecutes`) is still the thing that would measure size and compute against mainnet rather than argue them.

## Parity

- Android: `n/a` — no Kamino integration exists (grepped `*.kt`: no hits)
- Windows + extension: `n/a` — no Kamino integration exists (grepped `*.ts`/`*.tsx`/`*.go`: no hits)
- SDK: `n/a` — no Kamino integration exists
- `first:` Kamino Earn attribution memo. iOS is the only platform with a Kamino integration, so there is nothing to keep in sync today — but whoever implements Kamino on Android or Windows must write this same literal, byte for byte.

## Checklist

## Additional context

**Gate:** `** TEST SUCCEEDED **` — 6052 tests, 0 failures, 11 skipped (the opt-in live suite). SwiftLint clean on every touched file.

**Codex review** (read-only, whole-commit) raised two points:
1. `SolanaV0Transaction.maxMemoByteCount`'s doc comment still described the tag as "two bytes" — **fixed** in this commit.
2. The mixed-version co-sign behaviour described above — **rejected deliberately**, reasoning in this description.

Everything else it checked came back clean: no remaining `"vs"` / `[0x76, 0x73]` / `[118, 115]` hardcodes anywhere; the byte literals are correct ASCII for `8k2mz`; and there is no persisted raw-transaction or memo path, so nothing stored keys on the old tag.

- **Agent**: Claude Code
- **Issue**: none — request came from Kamino directly; context #5017, #5117
- **Confidence**: high
- **Needs human review for**: confirming the exact literal and its casing with Kamino (`8k2mz`, lowercase, five bytes) — it is compared byte-for-byte, so a casing mismatch would silently match nothing downstream

## Summary by CodeRabbit

* **Documentation**
  * Clarified supported memo text length and attribution matching behavior.

* **Bug Fixes**
  * Updated Kamino transaction attribution to use the new `8k2mz` memo tag.
  * Improved validation to reject incomplete, modified, uppercase, or whitespace-padded variants.

* **Tests**
  * Updated memo injection and attribution coverage for the new tag, including composition, preservation, and duplicate scenarios.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Solana/Signing/SolanaV0Transaction.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Solana/Signing/SolanaV0Transaction.swift
@@ -123,8 +123,8 @@ struct SolanaV0Transaction: Equatable {
     static let memoProgramId = "MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr"
 
     /// Bound on an injected memo. Nothing here needs a long one — the tag this
-    /// app writes is two bytes — and the ceiling keeps a caller from spending
-    /// the transaction's remaining size budget on a note.
+    /// app writes is a handful of bytes — and the ceiling keeps a caller from
+    /// spending the transaction's remaining size budget on a note.
     static let maxMemoByteCount = 64
 
     /// The 32 raw bytes of `computeBudgetProgramId`, decoded once. The input is
--- a/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoInstructionSequence.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoInstructionSequence.swift
@@ -278,7 +278,7 @@ enum KaminoInstructionSequence {
             return nil
         case .memo:
             // The tag is matched WHOLE, not as a prefix. A memo is free-form
-            // bytes, so "vs" followed by anything else is a different memo —
+            // bytes, so the tag followed by anything else is a different memo —
             // and the one thing this app is willing to sign under the Memo
             // program is its own attribution tag, exactly.
             return data == KaminoAttribution.memoTagBytes ? .attributionMemo : nil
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

Issue #5150, opened 2026-08-13
[Functions] Migrate SWITCH (Gaia to THORChain) to the FunctionTransaction architecture


gastonm5:
## Summary

Move the switch operation off `FunctionCallCosmosSwitch` onto the shared architecture. Gaia only.

## Operation

Memo: `SWITCH:<thorAddress>` (`FunctionCallCosmosSwitch.swift:120-122`), with an amount, sent to the live inbound address.

## Notes

- Resolve the inbound address at build time, not once at form load — the same failure mode as the Add-LP wrong-destination bug.
- Surface an explicit halt state: if the route is halted, say so rather than letting the user sign into a halted inbound.
- Like IBC, this may ultimately belong in a Bridge surface (see #5124); porting it first makes that move cheap.

## Architecture contract

Follow `Features/FunctionTransaction/` exactly — the same pattern every DeFi screen already uses. Do **not** port the legacy `FunctionCallSubModel` shape.

1. **Intent** — add a case to `FunctionTransactionType` (`Features/FunctionTransaction/Common/Model/FunctionTransactionType.swift`) carrying value types only (`CoinMeta`, `Decimal`, `String`), and extend its `coins` property so the caller's `needsCoinAddition` / `addCoins` resolve every coin the form touches.
2. **Route** — add the case to the `switch` in `FunctionTransactionScreen.body`, wrapped in `resolvingCoin(coinMeta:)`, so a coin the vault does not hold surfaces the shared error view instead of failing silently.
3. **ViewModel** — `final class <Name>TransactionViewModel: ObservableObject, Form`, owning its `FormField`s, exposing `form: [FormField]` and `validForm`, calling `setupForm()` from `onLoad()`. Async work inside a `Task` with `MainActor.run`; cancellables stored on the VM; `[weak self]` in every closure. No networking or business logic in the view.
4. **Screen** — `<Name>TransactionScreen` built on `FormScreen(title:onContinue:)` with `FormExpandableSection` / `CommonTextField` / `FunctionAddressField`, exposing `var onVerify: (TransactionBuilder) -> Void`. Continue is deliberately **not** gated on `validForm` — read the doc comment on `FormScreen` before changing that.
[...]

Issue #5142, opened 2026-08-13
[Functions] Replace the two dropdowns with an action list screen


gastonm5:
## Summary

Replace `FunctionCallDetailsScreen`'s dropdown-over-inline-form shape with a **list of the operations available on this chain**. Each row navigates to that operation's own screen. This is the enabling step for the whole migration: it lets each operation move to the `FunctionTransaction` architecture independently, and it makes the eventual DeFi-tab decision a matter of re-parenting rows rather than rewriting screens.

## Why a list, and why now

- The dropdown's default selection is a recurring bug source — Osmosis opens on a form its own dropdown does not offer, THORChain chainnet/stagenet open on an empty selector, and #5132 is the same class of defect. **A list has no default.**
- The row's job is to produce a `FunctionTransactionType` and navigate — exactly what `DefiChainMainScreen.onTransactionToPresent` already does. That makes the list and the DeFi tab two producers of the same intent.

## Most chains should not get a list

After the SECURE+ retirement and #5125, `getCases` per chain is:

| Chain | Actions |
|---|---|
| THORChain | 6 — rebond, leave, merge, unmerge, custom, withdrawSecured |
| MayaChain | 2 — leave, custom |
| Gaia | 2 — IBC, Switch |
| BTC, BCH, LTC, DOGE, ETH, AVAX, BSC, Base, XRP | 1 — addThorLP |
| dYdX, Kujira, Osmosis | 1 |

So: **exactly one action -> skip the list and route straight to the operation**, and label the entry button with the operation rather than "function". A list screen with one row is a wasted tap.

## Scope

1. **`FunctionActionDescriptor`** — a value type carrying title, subtitle, icon, availability and the destination. Deliberately shaped so the DeFi tab can render the same descriptors later without a rewrite; keep it free of any list-screen-specific concern.
2. **`FunctionActionsScreen`** — renders the descriptors for the coin's chain. No default selection, no dropdowns.
3. **Single-action passthrough** — when a chain resolves to one descriptor, the entry point navigates directly to it.
4. **Dual routing during the migration** — a descriptor points either at a `FunctionTransactionType` (migrated) or at the legacy screen pre-selected to that function. `FunctionCallRouteBuilder.swift:17` is the only construction site for `FunctionCallDetailsScreen`, so this is an optional `preselected: FunctionCallType?` on the route plus hiding the selector when it is set. Temporary — it dies with the last migration.
[...]

PR #5172, opened 2026-08-13
refactor(functions): migrate IBC transfer, and stop dropping memos that contain a colon

### LINKED ISSUE #5149: [Functions] Migrate IBC transfer to the FunctionTransaction architecture
## Summary

Move IBC transfer off `FunctionCallCosmosIBC` onto the shared architecture. Gaia, Kujira, Osmosis.

## Operation

Today the form colon-packs the whole transfer into one memo string (`FunctionCallCosmosIBC.swift:121-126`):

```
<destinationChainName>:<ibcChannel>:<destinationAddress>[:<userMemo>]
```

## Defects to fix in this migration

1. **A user memo containing `:` is silently dropped.** The signer splits on every colon and copies the app memo only when the component count is exactly 4 (`Blockchain/Cosmos/Signing/CosmosHelperStruct.swift:80-88`). Any memo with a colon — JSON forwarding payloads, exchange deposit tags — yields more than 4 components and signs with an **empty** memo. The transfer succeeds; the crediting does not. Carry destination chain, channel, address and user memo as **typed fields on the intent**, not as one packed string.
2. **The amount pre-fills with the entire balance** with no fee reserve (`:55`) — one tap from sending everything.

## Notes

Whether IBC ultimately belongs in Send/Bridge rather than the function list is an open question in #5124. This migration deliberately does not decide it: porting to the shared architecture is what makes either home cheap.

## Architecture contract

Part of #5124.

gastonm5:
Closes #5149. Part of #5124.

Moves IBC transfer off `FunctionCallCosmosIBC` onto `Features/FunctionTransaction/`, on Gaia, Kujira and Osmosis — **and changes a Cosmos signing path**, which is why it wants a careful human read.

## The defect this migration exists for

The form colon-packed the whole transfer into one memo string, and `CosmosHelperStruct`'s `.ibcTransfer` arm re-parsed it by splitting on **every** colon, copying the user memo only when the component count was exactly four.

So any memo containing a `:` — a JSON forwarding payload, an exchange deposit tag, a memo-required destination — produced more than four components and signed with an **empty** memo. The transfer settles on chain. The crediting does not.

## What could and could not change

The obvious fix is "carry the channel as a typed field instead of packing it". That would have broken co-signing:

- proto `CosmosSpecific` carries account number, sequence, gas, transaction type, denom trace and gas limit — **no source channel**;
- `commondata` is a *pinned remote SPM package*, so adding a field is another repo plus an Android/Windows rollout;
- peers do not blind-sign — each device rebuilds `KeysignPayload` from the proto and derives its own signing input.

The packed string is therefore the cross-device wire format, and it stays. What changes is that it is now written and read in exactly one place — `CosmosIBCTransferMemo` — with a **bounded** decode (`maxSplits: 3`), so the fourth component is the whole remainder, colons included. Destination chain, channel, address and user memo are typed fields everywhere above that boundary (intent → view-model → builder).

## Byte parity — the thing to check first

| memo shape | before | after | bytes |
|---|---|---|---|
| no user memo | channel `[1]`, memo empty | identical | **identical** |
| user memo, no `:` | channel `[1]`, memo `[3]` | identical | **identical** |
| user memo with `:` | memo **dropped** | memo preserved | changed — the fix |
| trailing `:` in memo | trailing `:` eaten | preserved | changed — same fix |
| < 3 components | **traps** (out-of-bounds `[1]`) | throws | changed — was a crash |
| empty channel | signs the destination address *as* the channel | throws | changed — was unbroadcastable |

Every difference is confined to a case that was already broken.

Evidence, two independent directions:

1. `CosmosIBCTransferMemoTests` re-implements the **previous** parse and compares serialized `CosmosSigningInput` as raw `Data` for the no-memo and colon-free-memo cases, on all three chains (each has its own `CosmosHelperConfig`).
2. The pre-existing golden `cosmos_osmosis_ibc_transfer_to_gaia` in `SigningGoldenVectors.json` — recorded before this change — is **untouched and still green**.

## Two more defects in the same arm

- `splitedMemo?[1]` was an out-of-bounds subscript on a memo with fewer than two components. Optional chaining does not save it and `?? ""` does not either — it traps. A peer can put any memo on a payload it marks `.ibcTransfer`, so this was a remotely triggerable crash on the co-signing device. It now fails closed.
- An empty channel (no known route) previously signed with the destination address as the source channel, because empty components were dropped. It now fails closed.

## Other producers / consumers

`FunctionCallCosmosIBC` was the only production producer of `transactionType: .ibcTransfer`. `QBTCHelper` was a **second consumer**, in two places — `buildTxBody` carried the identical `count == 4` memo bug, and `buildMsgTransfer` did its own unbounded split for the channel (so an empty component signed the destination address *as* the channel). Both now decode through the codec. QBTC stays unreachable from the UI (no `Chain.ibcTo` routes), so it is covered by the codec tests rather than a dedicated golden.

## The second defect: the amount

The form pre-filled the amount with the **entire balance** and no fee reserve — one tap from sending everything and then being unable to pay the fee. Not carried over:

- the field starts empty;
- the ceiling the percentage buttons scale reserves the chain fee for a native asset (a non-native token's fee comes from a different balance, so nothing is reserved from the token itself — but the chain's native coin is checked for gas, which Verify's own pre-flight skips entirely when the vault holds no native coin);
- a balance below the fee shows the shared `InsufficientFeeNotice` and disables Continue, instead of rejecting every input as "amount exceeded".

Amount parsing follows the lesson from #5165 and then goes one step further, because the Codex pass found the step was needed. Two refusals:

- **Wrong placement.** `NumberFormatter` reads en_US `1,5` as `15` — a ten-fold transfer from a paste. Refused.
- **A single group boundary with nothing to disambiguate it.** en_US `1,000` is one thousand if the comma groups and *one* if it was written by someone whose decimal separator is a comma. Both readings are well-formed, so the string does not say which, and guessing wrong is a **1000×** transfer. Refused. Two or more boundaries (`1,234,567`) cannot be a decimal point, and a string carrying the decimal separator too (`1,234.5`) has already named it — those are accepted.

That refusal would have broken the field's own percentage presets, which render through `Decimal.formatToDecimal(digits:)` and therefore group — 25% of a 4000 balance lands as `1,000`. So the view-model re-renders **app-generated** values without grouping (`IBCTransferAmount.plainSpelling`). Provenance is what makes that safe: `percentageSelected` is non-nil only for a value the app computed from a `Decimal` it already had, and the field clears it the moment the user types. The builder's own rendering is untouched, so byte parity is unaffected.

Tested with `en_US` and `de_DE` injected, both directions.

The builder's amount string is `parsed.formatToDecimal(digits: coin.decimals)` — the same call the legacy sub-model made — so `SendTransaction.amountInRaw` resolves to the identical base-unit value.

## Single-action passthrough — the first real exercise of #5142

Kujira and Osmosis have `.cosmosIBC` as their **only** operation, which the old selection-change seam could not migrate at all: a lone case is also the chain's `getDefault`, and a default never publishes a change. Both now pass through the action list straight to the new screen. Pinned per chain:

- `testTheSingleActionChainsPassThroughToTheIbcScreen` — Kujira and Osmosis resolve to `.action` with a `.transaction` destination;
- `testGaiaRendersAListWithTheIbcRowMigratedAndSwitchStillLegacy` — Gaia's two rows, one migrated and one still legacy.

Kujira's `getDefault` moves off the migrated type in **both** factories (`FunctionCallType` and `FunctionCallInstance`), keeping `testNoChainDefaultsToAMigratedFunction` green.

## Smaller things

- Destination address is validated against the **destination** chain (each of these three has its own WalletCore coin type, so it is a real check), re-applied whenever the route changes. Legacy fell back to a permissive THOR/Maya/TON check when nothing was selected.
- The Kujira **LVN** carve-out is preserved. Legacy expressed it as a `continue` inside its chain loop whose condition never depended on the loop variable, so it skipped *every* destination — LVN has never been IBC-transferable. Same outcome, now with a legible notice and a disabled Continue rather than a silently dead form.
- Two new strings in all 8 locales (Korean included); every locale at 1820 entries after `sort_localizable.py`.

## Merge order

- Lands after #5162 and after **#5167** (this branch's base, `feat/function-action-list-5142`).
- ⚠️ **#5150 (SWITCH) conflicts mechanically.** It re-points Gaia's default to `.cosmosIBC` in both factories because `.theSwitch` becomes migrated; this PR migrates `.cosmosIBC`. **Whichever lands second must re-point Gaia's default to `.custom`** — after both, Gaia offers only migrated operations and has no legacy default left to name. `testNoChainDefaultsToAMigratedFunction` fails loudly if it is missed. This branch does not touch `FunctionCallCosmosSwitch`.
- Nothing here edits `FunctionCallTypeEnum.getCases`, so no conflict with #5155 / #5160.

The `Closes #5149` above will not auto-link while the base is not the default branch — expected; keeping the line so it fires on re-target.

## Follow-ups (deliberately not here)

- **A first-class `CosmosSpecific.ibc_source_channel`**, dual-written like `RippleSpecific.destination_tag`, so routing data stops riding the memo at all. Needs commondata + Android/Windows; this PR makes it a one-line change on the iOS side.
- **Android and Windows have the same `count == 4` parser.** Between updated iOS devices a colon memo now round-trips correctly; a mixed-version or cross-platform pair fails the ceremony instead of broadcasting a memo-less transfer. Failing closed beats the current silent loss, but the sibling repos need the same codec.
- `AmountBalanceValidator`'s permissive grouping parse, app-wide.
- Whether IBC ultimately belongs in Send/Bridge rather than the action list is open in #5124. This migration deliberately does not decide it — porting to the shared architecture is what makes either home cheap.

Plan: `wiki/pages/projects/vultisig/functions-to-defi-migration/migrate-ibc-5149-plan.md`

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated IBC transfer flow with destination-chain selection, address validation, amount and memo fields, fee-aware balance checks, and transaction review.
  * Added locale-aware amount formatting and validation.
  * Added localized IBC transfer labels and empty-route messages in supported languages.
  * Migrated eligible chain actions to the new IBC transfer experience.

* **Bug Fixes**
  * Preserved user memos containing colons or JSON content.
  * Invalid or incomplete IBC routing data is now rejected before signing.


gastonm5:
⚠️ **Needs a human review before this leaves draft — it changes a Cosmos signing path.**

The specific things worth a second pair of eyes:

1. **The wire format decision.** The packed memo string is retained deliberately, because `CosmosSpecific` has no source-channel field and every co-signing device derives its own signing input from that proto. If that reasoning is wrong, the whole shape of the change is wrong.
2. **Byte parity.** `CosmosIBCTransferMemoTests` re-implements the previous parse and compares serialized `CosmosSigningInput` as raw `Data` on Gaia, Kujira and Osmosis; the pre-existing `cosmos_osmosis_ibc_transfer_to_gaia` golden is untouched and still green.
3. **The deliberate byte changes**, all in cases that were already broken: a memo containing a colon (previously signed empty), a memo shorter than routing data (previously a *trap* — reachable from a peer-supplied payload), and an empty channel (previously signed the destination address as the channel).
4. **Cross-platform.** Android and Windows still carry the `count == 4` parser. Between updated iOS devices a colon memo now round-trips; a mixed-version or cross-platform pair fails the ceremony instead of broadcasting a memo-less transfer. Failing closed beats the current silent loss, but the sibling repos need the same codec.

Gate: `** TEST SUCCEEDED **`, 6140 tests / 0 failures / 11 skipped, dedicated simulator, worktree-local derived data. Two Codex read-only rounds; ledger in the wiki plan.

PR #5181, opened 2026-08-16
fix(ui): round the asset picker's bottom corners and drop the macOS focus ring

gastonm5:
Two UI fixes reported from macOS screenshots.

## 1. The asset picker was missing its bottom corner radius

`AssetSelectionListScreen` applied `.cornerRadius(Theme.radius.xl)` to the **ScrollView**. A `ScrollView` takes the full height it is offered, so its rounded bottom corners sat at the bottom of the sheet while the *list content* ended higher up with square ones — a list that reads as flat and cut off at the bottom, rounded only at the top.

The rounded surface belongs to the rows, not to the scroll viewport. The list now uses the shared `commonListItemContainer(index:itemsCount:)` / `commonListContainer()` pair from `Components/List/CommonListItemContainer.swift` — the same pattern `FunctionActionsScreen` already renders correctly — so the group is rounded at both ends, carries the hairline border, and shows dividers between rows only.

### `SwapCoinCell` needed adapting, not just wrapping

`SwapCoinCell` draws its own bottom `Separator` **unconditionally**, including under the last row. Wrapping it in `commonListItemContainer` as-is would have put two lines between every pair of rows and left a stray line hanging under the last one, inside the newly rounded bottom.

It also draws its own `isSelected ? bgSurface2 : bgSurface1` background, which is the selected-row highlight — that is kept, and it now gets clipped correctly by the first/last row's corner mask instead of bleeding past the rounded corner.

So the cell gained a `showsSeparator: Bool = true` parameter. The default preserves today's behaviour for every existing caller; only the two pickers that hand divider duty to the container pass `false`.

### Screens covered

`AssetSelectionListScreen` is shared by seven Functions flows — one fix covers them all:

- Custom memo
- BondMaya
- UnbondMaya
- Merge
- Unmerge
- Add-LP
- SECURE− withdraw

None of the seven call sites overrides the container styling locally; they only pass bindings and a data source.

`IBCDestinationSelectionScreen` (also under `Features/FunctionTransaction/`, also built on `SwapCoinCell`) had the identical ScrollView-cornerRadius bug and is fixed the same way — that is the eighth asset picker under Functions.

### Not changed

- `StakingValidatorPickerScreen` — it also puts `.cornerRadius` on its ScrollView, but it is not the same bug. Its rows are `StakingValidatorCard`s: individually rounded (`Theme.radius.lg`), individually backgrounded, laid out in a `LazyVStack(spacing: 8)`. There is no grouped surface whose bottom corners can go missing, so there is nothing visibly cut off. Converting it to `commonListContainer` would redesign the sheet from spaced cards to one grouped list, which is out of scope here.
- `PreferredAssetSelectionView` (Referral) has the same ScrollView-cornerRadius mistake and the same cell, but it is not under Functions, so it is left for a follow-up rather than silently pulled in.
- `SwapCoinPickerView` — already applies its surface to the content rather than the ScrollView, so it renders correctly today. Untouched.

## 2. Back/close buttons showed a focus box on macOS

The circular ✕ and ‹ buttons rendered inside a blue rounded square. That is AppKit's **focus ring**, not a background. It is drawn *outside* the button's clip shape, which is why neither the existing `.buttonStyle(.plain)` nor the `.clipShape(Circle())` contained it.

Both buttons route through `Components/Toolbar/ToolbarButton.swift` — the back button via `NavigationBackButton` and `MacOSToolbarView` — so one component fixes ✕ and ‹ everywhere.

`.focusEffectDisabled()` is applied to the outer `Group` inside the `#if os(macOS)` branch, so it covers **both** macOS rendering paths (the macOS 26+ `glassEffect` branch and the pre-26 `customButton` branch) in one place. There is existing precedent in-tree at `PasscodeEntryView`, and the modifier is available on the macOS 15.0 deployment target.

### Accessibility

`focusEffectDisabled()` suppresses only the *visual* focus effect. The buttons stay focusable, so keyboard navigation and VoiceOver focus order are unchanged. `.focusable(false)` was deliberately **not** used — that would have taken the toolbar's only dismiss affordance out of keyboard and VoiceOver reach, which is a real accessibility regression rather than a cosmetic fix.

The iOS branch is outside the `#if os(macOS)` block and is untouched.

## Verification

⚠️ **Visually unverified — needs a desktop check.** This is a purely visual change, so a green suite proves very little on its own. Nothing here was seen rendering: reaching a Functions asset picker needs a set-up vault and roughly six UI interactions, which could not be driven headlessly. Please eyeball, on a Mac:

- one asset picker — rounded top **and** bottom, hairline border, dividers between rows only (no doubled lines, no stray line under the last row), selected row still highlighted;
- one toolbar ✕ or ‹ — no blue rounded square, and still reachable by keyboard focus.

What *was* verified:

- macOS build (`-destination platform=macOS,arch=arm64`) — `** BUILD SUCCEEDED **`
- Full iOS unit-test suite — `** TEST SUCCEEDED **`, 6610 tests, 11 skipped, 0 failures
- SwiftLint — 0 violations on all four changed files
- Two read-only cross-model review rounds. Round 1 raised two low-severity findings: a 1pt tap-target shift (the divider moving out of the button's label — accepted, it makes the row behave exactly like the shared `FunctionActionRow`), and a redundant re-evaluation of the filtering `filteredAssets` computed property once per row (fixed — the list now snapshots it once). Round 2 on the amended commit returned no findings.

## Bug Fixes
- Improved macOS toolbar focus behavior while preserving keyboard and VoiceOver accessibility.

## Style
- Updated destination and asset selection lists with consistent row styling, rounded containers, and cleaner separators.
- Added support for hiding separators in swap coin cells when needed.
- Improved asset list rendering by caching data before display.


gastonm5:
Valid in principle, and deliberately not taken here.

`.focusEffectDisabled()` does leave the button focusable without a visual indicator, so a keyboard user tabbing to it gets nothing. That's a real gap and the suggested shape — keep the suppression, add an `isFocused`-driven indicator in `Theme.colors.*` — is the right one.

Not bundling it into this PR for two reasons:

1. **The reported bug is the ring appearing when nobody focused anything.** macOS parks initial focus on the first focusable control, so the close/back button renders as a highlighted blue square on screen load, which reads as a selected state. That's what this change fixes, and it's a visual regression fix rather than a focus-model change.
2. **A bespoke focus indicator is a decision for the toolbar component as a whole**, not a side effect of a two-line visual fix. `ToolbarButton` has three rendering branches (macOS 26+ glass, macOS < 26 custom, iOS), and an indicator that works across all of them wants designing once — with the design system — rather than improvised here.

Recorded as a follow-up on the toolbar component. Focusability and VoiceOver order are unchanged by this PR; `.focusable(false)` was explicitly avoided for exactly that reason.

johnnyluo (approved):
LGTM

PR #5186, opened 2026-08-17
fix(functions): resolve Swift 6 warnings in Maya bond view models

johnnyluo:
## Summary
- Removed redundant `await MainActor.run { }` wrappers in `BondMayaTransactionViewModel` and `UnbondMayaTransactionViewModel` — both classes are already MainActor-isolated (via `Form`'s `@MainActor` conformance), so the enclosing `Task { [weak self] in }` bodies already run on the main actor. The wrappers only added an unnecessary `@Sendable` closure boundary, which is what triggered Swift 6's "implicit `self.`" warnings.
- Fixed the two "non-Sendable `Coin`" warnings in `BondMayaTransactionViewModel.fetchUserLPPositions()` by capturing `coin.address` (a `Sendable` `String`) once before the `async let`s, instead of letting the non-Sendable `Coin` cross into the child tasks.

## Test Plan
- [x] `swiftlint lint --config VultisigApp/.swiftlint.yml VultisigApp/` — 0 violations in the two changed files
- [x] `xcodebuild` (via XcodeBuildMCP `build_sim`) — build succeeds, all 30 warnings previously reported for these two files are gone; no new warnings introduced


johnnyluo:
@coderabbitai review please

PR #5188, opened 2026-08-17
fix(defi): denominate the CACAO withdraw ceiling in CACAO, not pool units

### LINKED ISSUE #5184: [BUG] CACAOPool withdrawal screen amount error
# [Fix] Display correct CACAO value instead of units on withdrawal screen [CACAOPool]

## Problem
The CACAOPool withdrawal screen displays incorrect amounts when removing liquidity. The screen shows raw "units" from the Maya API instead of the actual CACAO token value. This causes confusion as users see different amounts on the staked positions list (correct) versus the removal slider screen (incorrect), making it impossible to accurately determine how much CACAO they're withdrawing.

## Expected Behavior
The withdrawal screen should display the same CACAO token value shown on the staked positions screen, properly converting units to actual token amounts using the Maya API response data.

## Steps to Reproduce
1. Navigate to DeFi tab
2. Select Mayachain
3. Tap on Staked positions (observe correct CACAO quantity displayed)
4. Tap Remove on a CACAOPool position (observe incorrect units value instead of CACAO amount)

### Must Do
- [ ] Identify where the withdrawal screen retrieves the CACAO amount for display
- [ ] Update the withdrawal slider UI to use the converted CACAO value instead of raw units
- [ ] Ensure the conversion logic matches the calculation used on the staked positions list screen

### Must NOT Do
- Don't change the staked positions list display logic
- Don't modify Maya API integration beyond value conversion
- Don't alter withdrawal transaction logic or amounts sent to chain

## Acceptance Criteria
- [ ] `xcodebuild -scheme Vultisig build` succeeds
- [ ] CACAOPool withdrawal screen displays same CACAO amount as shown in staked positions list
- [ ] Slider value updates reflect actual CACAO amounts, not raw units
- [ ] Withdrawal transaction uses correct amount regardless of display fix

## Context
- Reported by **SamYap** (@samyap on Discord) on Discord
- Version: v1.44.70 on iOS
- Affects CACAOPool liquidity removal flow specifically

cc @realpaaao

gastonm5:
## Problem

The MayaChain CACAO staked card and its withdrawal sheet were denominated in two different scales. The staked-positions list showed the member's **CACAO value**; tapping **Remove** opened a slider whose maximum was the member's **pool units** — a different, and in an earning pool strictly smaller, quantity. Users saw one number on the list and another on the removal screen, with no way to tell how much CACAO they were actually withdrawing.

Reported by **SamYap** on Discord against v1.44.70 iOS.

## Root cause

`MayaChainStakeInteractor` split one position across two scales:

```swift
let stakedAmount = position.stakedAmount / pow(10, cacao.decimals)      // member.value — CACAO
let availableToUnstake = position.availableUnits / pow(10, cacao.decimals)  // member.units — POOL UNITS
```

`StakePositionData.amount` (the card) got the CACAO value; `StakePositionData.availableToUnstake` got the pool units. That second field travels `DefiChainScreenModel.unstakeTransactionType(for:)` → `FunctionTransactionType.unstake(…)` → `UnstakeTransactionViewModel.availableAmount`, which is what the slider, the Max button, the `AmountBalanceValidator` and the displayed balance all read.

A pool unit is worth `member.value / member.units` CACAO — a rate at or above 1 that only rises as the pool earns — so the two figures are never interchangeable.

Corroborating that CACAO is the intended denomination everywhere else: `BalanceService` already feeds `coin.stakedBalance` (and therefore `coin.stakedBalanceDecimal`, the sheet's *fallback* ceiling) from `position.stakedAmount`, the CACAO value. The explicit `availableToUnstake` was the only units-denominated figure in the entire path.

## Fix

Project the same CACAO value into both figures, in the interactor that produces them — not by branching on chain in the chain-agnostic screen model or view model. The mapping is extracted into a pure `static func MayaChainStakeInteractor.stakePositionData(position:coin:decimals:apr:unstakeMetadata:)` so it can be exercised without a network hop; `fetchStakePositions` keeps doing the I/O and is otherwise unchanged.

Also corrects the now-misleading comments on `MayaCacaoPoolPosition`, whose `availableUnits` field advertised itself as "units available for unstaking".

## ⚠️ Can the signed amount change?

**Answered precisely, and independently verified by a read-only cross-model review.**

`CacaoUnstakeTransactionBuilder` carries `amount = "0"` and `sendMaxAmount = false`, and `TransactionBuilder.buildSendTransaction(vault:)` copies both straight into `SendTransaction`. **A CACAO withdrawal signs no coin amount at all** — the entire instruction is the memo `POOL-:<bps>`, where

```
bps = Int(resolvedPercentage) * 100
resolvedPercentage = percentageSelected ?? (amountField.value / availableAmount * 100)
```

| path | effect of this change |
|---|---|
| **Slider / Max / percentage** (`percentageSelected != nil`) | `availableAmount` does not appear in the expression. **Byte-identical before and after.** This is the default: the sheet opens at 100%. |
| **Typed amount** (`percentageSelected == nil`) | `availableAmount` is the denominator, so the signed bps **does** change — and that *is* the defect being fixed. A typed CACAO figure was being read as a fraction of a pool-unit total, and any amount between the units total and the real CACAO total could not be entered at all. |

No user-entered absolute CACAO amount reaches `SendTransaction.amount` on either path.

## Tests

New `VultisigAppTests/Defi/MayaCacaoWithdrawCeilingTests.swift` (10 tests), pinning both halves:

- the projection puts the CACAO value — never scaled pool units — into `availableToUnstake`, at every unit/value ratio including the degenerate 1:1 that would hide the bug in a fixture;
- **money regression:** a slider-driven withdrawal signs the same transaction under the old units ceiling and the new CACAO ceiling at 100/50/25/10%, comparing every builder field that reaches `SendTransaction` (memo, `amount`, `sendMaxAmount`, `toAddress`, memo dictionary) — not just the memo;
- the typed path, driven end-to-end through the real projection: the whole card amount is now enterable, an amount between the two totals resolves to the right fraction (`POOL-:9100` for 1100 of 1200), and anything above the position is still refused.

## Verification status

- ✅ **Full unit suite green** on a dedicated simulator: `** TEST SUCCEEDED **`, 6501 tests executed, 11 skipped, 0 failures.
- ✅ `swiftlint` clean on all touched files.
- ✅ Read-only cross-model review of the branch diff; the fund-safety trace above was independently re-derived from the source rather than taken on trust.
- ❌ **Not verified on device.** Confirming the sheet now opens on the card's figure needs a real CACAOPool position — please treat this as an outstanding manual step before merge.

## Known residual (deliberately out of scope)

`DefiPositionsStorageService.upsert(stake:for:)` has no delete-stale, and a failed Maya fetch returns `[]`, so a **cached row written by a pre-fix build keeps its pool-unit `availableToUnstake` until a refresh succeeds**. In that window the reported symptom reproduces unchanged — never worse than current shipped behaviour, and self-healing on the first successful fetch (which the chain screen performs on appear). Closing it would mean a MayaChain-CACAO-specific backfill inside a chain-agnostic view model, or gating the withdraw CTA on a fresh fetch. Both are larger than this issue's stated scope ("don't alter withdrawal transaction logic") — flagging rather than doing it silently. Happy to add it here if a maintainer prefers.

## Note for reviewers

A textual conflict is plausible with the unmerged `feat/operation-coverage` / `feat/verify-hero-verbs` branches, which rework `UnstakeTransactionViewModel`'s CACAO arm and `CacaoUnstakeTransactionBuilder`. This change touches neither file. Those branches add a `withdrawDisplayAmount` derived from `availableAmount`, which would be *equally wrong* while the ceiling is pool units — so this is a prerequisite for them rather than a competitor.

Closes #5184

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Maya/API/MayaChainAPIService+Staking.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Maya/API/MayaChainAPIService+Staking.swift
@@ -26,8 +26,8 @@ extension MayaChainAPIService {
 
         return MayaCacaoPoolPosition(
             address: address,
-            stakedAmount: currentValue,  // Use value for display (includes earnings)
-            availableUnits: userUnits,   // Units available for unstaking
+            stakedAmount: currentValue,  // Member's CACAO value (includes earnings)
+            availableUnits: userUnits,   // Pool units held — NOT a CACAO amount
             userUnits: userUnits,
             netDeposit: netDeposit,
             lastWithdrawHeight: member.lastWithdrawHeight,
@@ -93,11 +93,17 @@ extension MayaChainAPIService {
 
 // MARK: - Staking Models
 
+/// A member's position in the CACAO pool, in the API's own scales.
+///
+/// The two scales are not interchangeable: `stakedAmount` is CACAO, the pool-unit
+/// fields are pool units, and a unit is worth `stakedAmount / userUnits` CACAO —
+/// a rate at or above 1 that only rises as the pool earns. Every figure the app
+/// shows a user, and every ceiling it lets a user type against, is CACAO.
 struct MayaCacaoPoolPosition {
     let address: String
-    let stakedAmount: Decimal      // Current value in CACAO (for display)
-    let availableUnits: Decimal    // Units available for unstaking
-    let userUnits: Decimal         // User's pool units
+    let stakedAmount: Decimal      // Current value in CACAO, base units (for display and withdrawal)
+    let availableUnits: Decimal    // Pool units held — NOT a CACAO amount
+    let userUnits: Decimal         // Pool units held (duplicate of `availableUnits`)
     let netDeposit: Decimal        // Deposit - Withdrawn
     let lastWithdrawHeight: Int64
     let lastDepositHeight: Int64
--- a/VultisigApp/VultisigApp/Features/Defi/DefiChain/Interactor/Stake/MayaChainStakeInteractor.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiChain/Interactor/Stake/MayaChainStakeInteractor.swift
@@ -35,22 +35,56 @@ struct MayaChainStakeInteractor: StakeInteractor {
             return []
         }
 
-        let stakedAmount = position.stakedAmount / pow(10, cacao.decimals)
-        let availableToUnstake = position.availableUnits / pow(10, cacao.decimals)
         let aprData = try? await mayaChainAPIService.getCacaoPoolAPR()
-
         let unstakeMetadata = await unstakeMetadata(for: position)
 
         return [
-            StakePositionData(
+            Self.stakePositionData(
+                position: position,
                 coin: cacao.meta,
-                type: .stake,
-                amount: stakedAmount,
-                availableToUnstake: availableToUnstake,
+                decimals: cacao.decimals,
                 apr: aprData?.apr ?? 0,
                 unstakeMetadata: unstakeMetadata
             )
         ]
+    }
+
+    /// Projects a fetched CACAO pool position onto the row the DeFi tab renders
+    /// and the withdraw sheet is opened from.
+    ///
+    /// Both figures are the member's CACAO **value**, and deliberately the same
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5190, opened 2026-08-17
fix(blockaid): net a same-mint SOL/WSOL diff pair instead of calling it a swap

### LINKED ISSUE #5187: Kamino SOL deposit renders as a bogus SOL to WSOL swap on the Blockaid hero
## Summary

`BlockaidSimulationParser.parseSolanaSwap` never compares the two sides' mints, and `buildSolanaCoin` maps native SOL onto the wrapped-SOL mint. A transaction that wraps SOL and spends it in the same instruction leaves its wSOL account holding only the rent-exempt residual, which Blockaid reports as a small incoming WSOL diff — so the hero renders the deposit as a swap into `0.00203928 WSOL` instead of stating the amount that is actually leaving.

## Details

`BlockaidSimulationParser.swift:183-184`:

```swift
let isNative = asset.type == "SOL"
let mint = isNative ? wrappedSolMint : asset.address
```

Native SOL and a real WSOL SPL diff therefore carry the identical mint. `parseSolanaSwap` (`:146-174`) destructures positionally and returns `.swap` whenever both sides carry a value:

```swift
let outCandidate = diffs[0]
let inCandidate = diffs[1]
...
return .swap(fromCoin: fromCoin, toCoin: toCoin, fromAmount: outAmount, toAmount: inAmount)
```

A SOL-out/WSOL-in pair is the same asset on both sides and can never be a genuine swap.

## Evidence

Blockaid's response for a Kamino Allez SOL deposit — the wrapped-SOL vault, `KaminoVaultRegistry.swift:105-115`, `tokenMint: wrappedSolMint`:

```json
{"account_assets_diff": [
  {"asset": {"type": "SOL", "decimals": 9, "symbol": "SOL", "address": null},
   "out": {"raw_value": "59435000"}},
  {"asset": {"type": "TOKEN", "address": "So11111111111111111111111111111111111111112",
             "symbol": "WSOL", "decimals": 9},
   "in": {"raw_value": "2039280"}}
]}
```

`2039280` is the rent-exempt minimum of a 165-byte token account. The deposit opens the wSOL ATA and never closes it — `KaminoDepositViewModel.swift:163-169` states it and ships it as a user-facing disclosure — so that residual is all that is left of the wrapped balance once the vault has taken the deposit. The wrap-in and the spend-out cancel, and never surface as diffs of their own.

Two diffs, so the `diffs.count == 3` fee filter at `:119-126` does not apply. `parseSolanaSwap` reads `diffs[0]` as the out side and `diffs[1]` as the in side and emits `.swap(SOL 0.059435, WSOL 0.00203928)`.

## Impact

The hero above the verify summary names a destination asset the user is not receiving, at an amount 29x smaller than the one leaving. `BlockaidSimulationInfo+HeroContent.swift` is shared by the initiator (`KeysignViewModel.swift:240`) and the co-signer (`JoinKeysignViewModel.swift:692`), so both devices show it.

The signed bytes are unaffected — this is a display defect — but the hero is the screen the approval decision rests on. It is not Kamino-specific either: any transaction that wraps SOL and spends it in one shot produces the same pair.

## Suggested remediation

Collapse the pair rather than emitting a swap: when both sides resolve to the same mint, the right hero is a transfer of the net movement, `out - in`.

Netting rather than keeping the out leg matters for the reverse direction, which is equally reachable. A withdraw that closes the wSOL account, or any unwrap, pairs a WSOL out with a native SOL in — a net inflow, where keeping the out leg alone would headline the residual and drop the amount received. Worth a fixture for each direction.

Android carries the same bug and is fixing it in vultisig-android#5620.

gastonm5:
## Problem

A Kamino Allez SOL deposit renders on the keysign hero as a **swap into `0.00203928 WSOL`** — a destination asset the user is not receiving, at an amount 29× smaller than the one leaving. The hero is the screen the approval decision rests on, and `BlockaidSimulationInfo+HeroContent.swift` is shared by the initiator (`KeysignViewModel`) and the co-signer (`JoinKeysignViewModel`), so **both devices show it**.

Blockaid's response for the deposit:

```json
{"account_assets_diff": [
  {"asset": {"type": "SOL", "decimals": 9, "symbol": "SOL", "address": null},
   "out": {"raw_value": "59435000"}},
  {"asset": {"type": "TOKEN", "address": "So11111111111111111111111111111111111111112",
             "symbol": "WSOL", "decimals": 9},
   "in": {"raw_value": "2039280"}}
]}
```

`2039280` lamports is the rent-exempt minimum of a 165-byte SPL token account. The deposit opens the wSOL ATA and never closes it (already disclosed by `KaminoDepositViewModel.showsWrappedSolNotice`), so once the vault has taken the deposit that residual is all that is left of the wrapped balance. The wrap-in and the spend-out cancel and never surface as diffs of their own.

## Root cause

`BlockaidSimulationParser.parseSolanaSwap` destructures the two diffs positionally and returns `.swap` whenever both sides carry a value — it never compares the two sides' mints. Meanwhile `buildSolanaCoin` maps native SOL onto the wrapped-SOL mint on purpose (so `TokensStore` lookups and the hero's fiat resolution work uniformly). Native SOL and a real WSOL SPL diff therefore carry the **identical** mint: a SOL-out/WSOL-in pair is the same asset on both sides and can never be a genuine swap.

Only two diffs, so the `diffs.count == 3` native-SOL fee filter does not apply.

Not Kamino-specific — any transaction that wraps SOL and spends it in one shot produces the same pair.

## Fix

When both legs of a two-diff Solana pair resolve to the same mint, collapse them into the **net** movement of that mint (`out - in`) instead of emitting a swap:

- net outflow → `.transfer` of the net, using the out leg's coin
- net inflow → `.receive` of the net, using the in leg's coin
- exact cancellation → `nil`, and the caller falls back to the generic title (there is no honest hero for "your balance does not change")

Mints are compared **exactly**: Solana addresses are base58 and case-sensitive, unlike the EVM path's checksummed hex. Comparing the *built* coins is what makes the check right — `buildSolanaCoin` already collapses native SOL onto the wrapped-SOL mint, which is precisely the "same asset" test wanted. The displayed coin comes from whichever leg the net follows, so the ticker stays truthful: both directions read **SOL**, not WSOL.

## The netting decision, and its reverse-direction consequence

Netting rather than simply keeping the out leg is a deliberate choice, and it is what the reverse direction needs.

A wrapped-SOL **withdraw** closes the payout account, and closing it is what unwraps wSOL into the native SOL the screen promises (`KaminoInstructionSequence` marks `closeTokenAccount` `required: isWrappedSolVault`). That yields the same pair reversed: a small **WSOL out** beside a large **native SOL in** — a net *inflow*. Keeping the out leg there would headline the `0.00203928` residual and drop the amount actually received, and reporting it as a `.transfer` would state an outflow for an inflow — a worse defect than the one being fixed.

`HeroContent` had no inflow shape, so this adds one:

- `BlockaidSimulationInfo.receive(coin:amount:)` → `HeroContent.receive(title:coin:)`, rendered by `HeroContentView` as the send row plus an explicit **"Receive"** label. Every other hero shape states money leaving, so an unlabelled row reads as a send. The label reuses the existing `"receive"` key, which already ships in **all 8 locales** — no new strings.
- Both keysign view models reach it through the same shared `heroContent(title:vaultCoins:)`, so initiator and co-signer cannot drift; each is pinned by its own fixture.
- `BlockaidSimulationInfo.fromCoin` / `fromAmount` / `fromAmountDecimal` are renamed `primaryCoin` / `primaryAmount` / `primaryAmountDecimal`. A `receive` has no from-side, and a name implying one is exactly how an inflow ends up displayed as an outflow. They are used only inside the model and the hero extension.

### Divergence from Android

[vultisig-android#5620](https://github.com/vultisig/vultisig-android/pull/5620) (open, not landed) fixes the same bug by keeping the **out leg** for the outflow direction and returning null for the inflow. This PR nets instead. The visible consequence: for the deposit above, iOS will read `0.05739572 SOL` (what leaves and does not come back) where Android reads `0.059435 SOL` (gross native outflow, rent included). The two platforms will disagree on that figure until one of them moves — flagged for a maintainer call.

## Tests

`BlockaidSolanaSimulationParserTests` (raw values from the captured response above):

- `test_parseSolana_solOutWsolIn_netsToTransfer` — deposit direction: `59435000` out / `2039280` in → `.transfer` of `57395720`, ticker **SOL** (not WSOL), native Solana logo.
- `test_parseSolana_wsolOutSolIn_netsToReceive` — withdraw direction: `2039280` out / `61474280` in → `.receive` of `59435000`, ticker **SOL**.
- `test_parseSolana_wsolOutSolIn_netsToReceive_regardlessOfDiffOrder` — diff order is not contractual.
- `test_parseSolana_sameMintExactCancellation_returnsNil` — exact cancellation nets to zero → no hero.
- `test_parseSolana_threeDiffs_afterSolFeeFilter_netsSameMintPair` — the fee filter runs first, so a same-mint pair can still be the survivor; the netting must apply to it.
- `test_parseSolana_sameMint_negativeRawValue_returnsNil` — `raw_value` decodes a *signed* integer; a negative reaching the subtraction would flip an approval headline's direction, so the netting fails closed.
- `test_parseSolana_sameMint_collapsedSources_netsWhenOtherDiffIsEmpty` / `..._returnsNil_whenOtherDiffCarriesBalance` — the in/out fallback can resolve both sides onto one diff. Netting a diff against itself is still that asset's net movement, so it stays allowed while the skipped diff holds nothing; with a second asset spending beside it the parser declines rather than headline a net it never looked at in full.
- `test_parseSolana_swap_nativeSolToUsdc_staysSwap` — **regression guard**: a genuine native SOL → USDC swap must still parse as `.swap`, because native SOL carries the wrapped-SOL sentinel and the guard must compare against the *other* leg's mint rather than treat every native leg as wrap noise. The pre-existing USDC → BONK swap fixture still passes unchanged.

Hero mapping, on both devices:

- `InitiatorAmountFiatTests.testInitiatorHeroReceiveMapsToReceiveHero`
- `JoinKeysignAmountFiatTests.testCoSignerHeroReceiveMapsToReceiveHero`

Both assert a `.receive` simulation produces a `.receive` hero — never a `.send` — with the ticker and fiat resolved.

## Verification

- Full unit suite on a dedicated simulator (iPhone 17 Pro Max, iOS 26.5): `** TEST SUCCEEDED **` — 6502 executed, 11 skipped, **0 failures**. All 11 new fixtures ran and passed.
- `swiftlint --config VultisigApp/.swiftlint.yml` clean on every touched file (0 violations).
- Codex read-only review of the branch, one round, five findings: two P1s **fixed** in the second commit (negative `raw_value`, collapsed sources dropping a leg — both with fixtures); one P2 **rejected with reason** (a deliberate 1:1 SOL↔WSOL wrap no longer renders as a swap — that is the accepted trade, and an exact cancellation only fires on *equal* legs, which a real wrap never has once rent and fee are in the native out); two P2s **deferred** as pre-existing and out of scope (the three-diff fee heuristic; the initiator Done screen has never carried the Blockaid hero for any shape). Codex separately confirmed every `HeroContent` switch is exhaustive, `shouldShowAmountRow` suppresses the duplicate amount row, no rename call site was missed, and the new fixtures would fail against the old parser.

## Risks / what needs a human

- **Display-only** — the signed bytes are untouched. But this is the transaction-approval hero on a signing path, and it renders on **both the initiator and the co-signer devices**: please review on device.
- The netted deposit figure (`0.05739572`) is smaller than the amount the deposit form said (`0.059435`); the difference is the rent residual the wSOL ATA keeps. That is the honest "what leaves and does not come back" number, but it will not match the deposit screen, and it disagrees with Android. Worth a maintainer call before this is un-drafted.
- **Not reproduced on device.** The outflow direction is proven against the captured raw values from the issue. The **inflow direction is derived** from the withdraw's instruction sequence, not from a captured Blockaid body for a withdraw — no such body exists anywhere in our records. The reverse fixture's raw values are constructed to match that shape.
- The `.receive` hero's label placement is new UI on the verify → sign → done screens and wants a design eye.
- The `diffs.count == 3` fee heuristic remains a separate pre-existing hazard (it would remove a real out leg as "likely the fee" if Blockaid ever returned the deposit as three diffs). Shared with the extension and Android; untouched here.

Plan: `wiki/pages/projects/vultisig/kamino-earn-5017/blockaid-sol-wsol-hero-5187-plan.md`

Closes #5187

## Summary by CodeRabbit

- **New Features**
  - Added clear receive transaction summaries showing the asset, amount, label, and optional title.
  - Added support for inbound assets, including wrapped-SOL withdrawals.
  - Improved transaction amount and fiat-value displays across transfer, receive, and swap simulations.

- **Bug Fixes**
  - Improved Solana transaction parsing for same-asset transfers, cancellations, fees, and mixed-asset scenarios.
  - Prevented redundant amount rows for receive transactions.

- **Tests**
  - Added coverage for receive simulations, wrapped-SOL handling, and Solana parsing edge cases.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
+++ b/VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
@@ -18,6 +18,8 @@ struct HeroContentView: View {
                 titleOnly(text: text, caption: caption)
             case .send(let title, let coin):
                 send(title: title, coin: coin)
+            case .receive(let title, let coin):
+                receive(title: title, coin: coin)
             case .swap(let title, let from, let to):
                 swap(title: title, from: from, to: to)
             }
@@ -50,6 +52,26 @@ struct HeroContentView: View {
                 .frame(maxWidth: .infinity, alignment: .center)
         }
         coinRow(coin, iconSize: 36)
+    }
+
+    /// The inflow twin of `send`. Every other hero shape states money leaving,
+    /// so the amount carries an explicit "Receive" label — without it the row
+    /// is indistinguishable from a send.
+    @ViewBuilder
+    private func receive(title: String?, coin: HeroCoinAmount) -> some View {
+        if let title {
+            Text(title)
+                .font(Theme.fonts.bodyMMedium)
+                .foregroundStyle(Theme.colors.textSecondary)
+                .frame(maxWidth: .infinity, alignment: .center)
+        }
+        VStack(spacing: 8) {
+            Text("receive".localized)
+                .font(Theme.fonts.caption10)
+                .foregroundStyle(Theme.colors.textTertiary)
+                .frame(maxWidth: .infinity, alignment: .center)
+            coinRow(coin, iconSize: 36)
+        }
     }
 
     @ViewBuilder
--- a/VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
+++ b/VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
@@ -17,9 +17,9 @@ struct HeroContentView: View {
             case .title(let text, let caption):
                 titleOnly(text: text, caption: caption)
             case .send(let title, let coin):
-                send(title: title, coin: coin)
+                amountRow(title: title, coin: coin, label: nil)
             case .receive(let title, let coin):
-                receive(title: title, coin: coin)
+                amountRow(title: title, coin: coin, label: "receive".localized)
             case .swap(let title, let from, let to):
                 swap(title: title, from: from, to: to)
             }
@@ -43,33 +43,27 @@ struct HeroContentView: View {
         .frame(maxWidth: .infinity, alignment: .center)
     }
 
+    /// The single-row heroes: send and receive render the same row and differ
+    /// only by `label`. Every hero shape but the receive states money leaving,
+    /// so the inflow carries an explicit label — without it the row is
+    /// indistinguishable from a send.
     @ViewBuilder
-    private func send(title: String?, coin: HeroCoinAmount) -> some View {
+    private func amountRow(title: String?, coin: HeroCoinAmount, label: String?) -> some View {
         if let title {
             Text(title)
                 .font(Theme.fonts.bodyMMedium)
                 .foregroundStyle(Theme.colors.textSecondary)
                 .frame(maxWidth: .infinity, alignment: .center)
         }
-        coinRow(coin, iconSize: 36)
-    }
-
-    /// The inflow twin of `send`. Every other hero shape states money leaving,
-    /// so the amount carries an explicit "Receive" label — without it the row
-    /// is indistinguishable from a send.
-    @ViewBuilder
-    private func receive(title: String?, coin: HeroCoinAmount) -> some View {
-        if let title {
-            Text(title)
-                .font(Theme.fonts.bodyMMedium)
-                .foregroundStyle(Theme.colors.textSecondary)
-                .frame(maxWidth: .infinity, alignment: .center)
-        }
-        VStack(spacing: 8) {
-            Text("receive".localized)
-                .font(Theme.fonts.caption10)
-                .foregroundStyle(Theme.colors.textTertiary)
-                .frame(maxWidth: .infinity, alignment: .center)
+        if let label {
+            VStack(spacing: 8) {
+                Text(label)
+                    .font(Theme.fonts.caption10)
+                    .foregroundStyle(Theme.colors.textTertiary)
+                    .frame(maxWidth: .infinity, alignment: .center)
+                coinRow(coin, iconSize: 36)
+            }
+        } else {
             coinRow(coin, iconSize: 36)
         }
     }
--- a/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
+++ b/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
@@ -107,6 +107,10 @@ enum BlockaidSimulationParser {
     /// filtered out as "likely the transaction fee". One remaining diff maps to
     /// `.transfer`, two remaining diffs map to `.swap` (or `.transfer` if only
     /// the out side has a value).
+    ///
+    /// Diverges from the extension in one place: a two-diff pair that resolves
+    /// to the same mint is netted rather than reported as a swap — see
+    /// `netSameMint(outCoin:outAmount:inCoin:inAmount:)`.
     static func parseSolana(
         response: BlockaidSolanaSimulationResponseJson
     ) -> BlockaidSimulationInfo? {
@@ -161,6 +165,16 @@ enum BlockaidSimulationParser {
            let inAmount = parseRawAmount(inRaw),
            let fromCoin = buildSolanaCoin(from: outSource.asset),
            let toCoin = buildSolanaCoin(from: inSource.asset) {
+            if let outMint = fromCoin.address,
+               let inMint = toCoin.address,
+               outMint == inMint {
+                return netSameMint(
+                    outCoin: fromCoin,
+                    outAmount: outAmount,
+                    inCoin: toCoin,
+                    inAmount: inAmount
+                )
+            }
             return .swap(fromCoin: fromCoin, toCoin: toCoin, fromAmount: outAmount, toAmount: inAmount)
         }
 
@@ -170,6 +184,46 @@ enum BlockaidSimulationParser {
             return .transfer(fromCoin: fromCoin, fromAmount: outAmount)
         }
 
+        return nil
+    }
+
+    /// Collapses an out/in pair that resolves to a single mint into the net
+    /// movement of that mint.
+    ///
+    /// Nothing was exchanged, so a swap hero would name a destination asset the
+    /// user is not receiving. Wrapping SOL and spending it in the same
+    /// transaction is the common source: the wrap-in and the spend-out cancel
+    /// and never surface as diffs of their own, leaving only the wSOL account's
+    /// rent-exempt residual as a small incoming diff beside the real native-SOL
+    /// out leg. `buildSolanaCoin` maps native SOL onto the wrapped-SOL mint, so
+    /// both legs arrive here carrying the same mint and are directly netted.
+    ///
+    /// The direction is decided by the net, not by which leg is which: the same
+    /// pair arrives reversed when a transaction unwraps wSOL back into native
+    /// SOL (a Kamino wrapped-SOL withdraw closes its payout account, which is
+    /// what unwraps it), where the out leg is only the residual being returned
+    /// and the in leg is the amount the user actually receives. The coin comes
+    /// from whichever leg the net follows, so the ticker stays truthful —
+    /// native SOL reads "SOL" even though it shares WSOL's mint here.
+    ///
+    /// An exact cancellation nets to zero. There is no honest hero for "your
+    /// balance does not change", so it returns nil and the caller falls back to
+    /// the generic title.
+    ///
+    /// Mints are compared exactly. Solana addresses are base58 and
+    /// case-sensitive, unlike the EVM path's checksummed hex.
+    private static func netSameMint(
+        outCoin: BlockaidSimulationCoin,
+        outAmount: BigInt,
+        inCoin: BlockaidSimulationCoin,
+        inAmount: BigInt
+    ) -> BlockaidSimulationInfo? {
+        if outAmount > inAmount {
+            return .transfer(fromCoin: outCoin, fromAmount: outAmount - inAmount)
+        }
+        if inAmount > outAmount {
+            return .receive(coin: inCoin, amount: inAmount - outAmount)
+        }
         return nil
     }
 
--- a/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
+++ b/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
@@ -167,14 +167,14 @@ enum BlockaidSimulationParser {
             if let outMint = fromCoin.address,
                let inMint = toCoin.address,
                outMint == inMint {
-                // Both sides can collapse onto one diff when the other carries
-                // neither leg. Netting a diff against itself is still the net
-                // movement of that asset — but only while the diff that was
-                // skipped holds nothing of its own. Otherwise the hero would
-                // state an authoritative net having never looked at a balance
-                // change the transaction makes, so decline and let the caller
-                // fall back to the generic title.
-                if outIndex == inIndex, carriesBalance(diffs[outIndex == 0 ? 1 : 0]) {
+                // The netting reads exactly two legs: one diff's out side and
+                // one diff's in side, which can collapse onto the same diff.
+                // Every other leg the response states goes unread, so netting
+                // past one would have the hero state an authoritative net
+                // having never looked at a balance change the transaction
+                // makes. Decline and let the caller fall back to the generic
+                // title.
+                guard readsEveryLeg(diffs, outIndex: outIndex, inIndex: inIndex) else {
                     return nil
                 }
                 return netSameMint(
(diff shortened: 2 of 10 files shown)


johnnyluo (commented):
Automated review findings (code-review skill).

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
```diff
@@ -149,18 +153,37 @@ enum BlockaidSimulationParser {
...
+               let inMint = toCoin.address,
+               outMint == inMint {
+                // Both sides can collapse onto one diff when the other carries
+                // neither leg. Netting a diff against itself is still the net
+                // movement of that asset — but only while the diff that was
+                // skipped holds nothing of its own. Otherwise the hero would
+                // state an authoritative net having never looked at a balance
+                // change the transaction makes, so decline and let the caller
+                // fall back to the generic title.
+                if outIndex == inIndex, carriesBalance(diffs[outIndex == 0 ? 1 : 0]) {
```
johnnyluo: **Correctness (confirmed):** the same-mint netting guard only fires when `outIndex == inIndex`, so it misses the case where the out and in legs come from *different* diffs and one of those diffs also carries its own extra, unused leg.

Example: `diffs = [SOL(out: 59435000), WSOL(out: 500000, in: 2039280)]`. `outIndex=0` picks SOL's out, `inIndex=1` picks WSOL's in — both resolve to the wrapped-SOL mint so same-mint netting fires — but WSOL's own `out: 500000` is never read by `outSource`/`inSource` and is silently dropped from the net. The reported outflow (57395720) understates the true net movement by exactly the ignored leg, on the transaction-approval hero. No existing test places `in` and `out` on a diff at a different index than the one supplying the opposite leg.

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
```diff
@@ -173,6 +196,60 @@ enum BlockaidSimulationParser {
...
+            return .transfer(fromCoin: outCoin, fromAmount: outAmount - inAmount)
+        }
+        if inAmount > outAmount {
+            return .receive(coin: inCoin, amount: inAmount - outAmount)
+        }
+        return nil
+    }
+
+    /// Whether a diff states a balance change at all, in either direction.
+    private static func carriesBalance(
```
johnnyluo, line 274: **Correctness (plausible, low severity):** `carriesBalance` treats a diff whose `out`/`in` field decodes to `raw_value: "0"` as carrying an actual balance change (`diff.out?.rawValue != nil` is true for `"0"`). A same-mint pair with an explicit zero leg on the unused diff would be declined instead of netted. Benign — falls back to the safe generic title rather than producing a wrong number — but unintentionally conservative.

VultisigApp/VultisigApp/Features/Keysign/ViewModels/BlockaidSimulationInfo+HeroContent.swift
```diff
@@ -11,9 +11,13 @@ import Foundation
...
+    /// Hero for this simulation: a single coin row for transfers and receives,
+    /// from/to rows for swaps. Fiat sub-lines are resolved best-effort against
     /// `vaultCoins` (see `heroFiat(for:amount:vaultCoins:)`).
+    ///
+    /// A `receive` maps to its own hero shape rather than reusing `.send`: the
+    /// two render the same row, and only the label separates money arriving
+    /// from money leaving.
     func heroContent(title: String?, vaultCoins: [Coin]) -> HeroContent {
         switch self {
         case .transfer(let coin, _):
```
johnnyluo, line 23: **Simplification:** `.transfer` and the new `.receive` case build byte-identical `HeroCoinAmount(amount:ticker:logo:fiat:)` values, differing only in which `HeroContent` case wraps them. A future change to how a coin row's fiat/ticker/logo is resolved must be edited in two (really three, counting swap's `from`) places and can silently drift if one is missed. A shared `coinAmount(for:amount:vaultCoins:) -> HeroCoinAmount` helper would remove the duplication.

VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
```diff
@@ -52,6 +54,26 @@ struct HeroContentView: View {
         coinRow(coin, iconSize: 36)
     }
 
+    /// The inflow twin of `send`. Every other hero shape states money leaving,
+    /// so the amount carries an explicit "Receive" label — without it the row
+    /// is indistinguishable from a send.
+    @ViewBuilder
+    private func receive(title: String?, coin: HeroCoinAmount) -> some View {
```
johnnyluo: **Simplification:** `receive(title:coin:)` duplicates `send(title:coin:)` wholesale, adding only a wrapping `VStack` with a localized label above the row. A single `amountRow(title:coin:label:)` with an optional label would collapse both into one implementation.

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
gastonm5: Fixed in `1cd024ee9`.

You are right that the guard only covered the collapsed case. Confirmed failing input: `[SOL(out: 59435000), WSOL(out: 500000, in: 2039280)]` gives `outIndex = 0`, `inIndex = 1`, both mints resolve to wrapped SOL so the netting fires — and `diffs[1].out` is never read by either source. The hero reported 57395720 out against a true net of 57895720, understating the outflow on the approval screen by exactly the ignored leg. The mirror drops the out-source's own `in` leg.

`BlockaidSimulationParser.swift` now replaces the index-equality check with `readsEveryLeg(_:outIndex:inIndex:)`, which asserts the two legs the netting actually reads are the only balance changes the response states — for every diff, not just the other one of the first two. That subsumes the collapsed case the old guard handled and holds for any diff count. When it fails, the parser declines and the caller falls back to the generic title, which is the posture this path already took: better no net than an authoritative net computed from part of the transaction.

Two tests in `BlockaidSolanaSimulationParserTests.swift` pin both mirror directions of the distinct-index case (`testParseSolanaSameMintDistinctSourcesReturnsNilWhenInSourceAlsoSpends`, `...WhenOutSourceAlsoReceives`).

VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
gastonm5: Fixed in `26d47a459` — and the fix went further than the report, because the predicate was wrong in both directions.

The half you flagged: an explicit `raw_value: "0"` on an unused leg read as a balance change, so a same-mint pair sitting beside a zero leg was declined and fell back to the generic title even though the net was the whole story. Zero moves nothing, so it no longer counts.

The half not in the report, and the more consequential one: `carriesBalance` was also failing **open**. `BalanceChange` keeps a `raw_value` it cannot decode as `nil` (see its `init(from:)`), so a leg that the response states but whose magnitude is unreadable looked exactly like an absent leg and *permitted* the netting — the hero would state an authoritative net across a balance change it could not account for. Unreadable is not zero, so a present leg with a nil or non-numeric magnitude now counts as a change and the netting fails closed.

`carriesBalance` is now four-state: absent → no change; decodes to 0 → no change; decodes to non-zero → change; present but unreadable → change. Four tests cover it: zero on the collapsed path, zero on the distinct-index path, `BalanceChange(rawValue: nil)`, and a non-numeric `raw_value`.

VultisigApp/VultisigApp/Features/Keysign/ViewModels/BlockaidSimulationInfo+HeroContent.swift
gastonm5: Fixed in `769b68c4c`.

`BlockaidSimulationInfo+HeroContent.swift` now builds every hero row through one `coinAmount(...)`, so ticker, logo and the `heroFiat` lookup resolve in a single place instead of three identical copies across `.transfer`, `.receive` and the swap's `from`.

The signature deliberately differs from the one suggested: `coinAmount(for:text:amount:vaultCoins:)` takes the display text and the priced Decimal as separate parameters. A single `amount:` would not fit, because the swap's destination row renders `heroToAmountText` while pricing `toAmountDecimal` — collapsing them would have silently changed which value that row prices, i.e. turned a refactor into a behavior change on the fiat sub-line. The three rows that genuinely share the leading amount go through a thin `primaryRow(for:vaultCoins:)`, which is the deduplication the comment was after.

No behavior change: `InitiatorAmountFiatTests` and `JoinKeysignAmountFiatTests` cover both fiat amounts and the receive direction, and pass untouched.

VultisigApp/VultisigApp/Components/Hero/HeroContentView.swift
gastonm5: Fixed in `769b68c4c`.

`HeroContentView.send(title:coin:)` and `receive(title:coin:)` are now one `amountRow(title:coin:label:)`. A `nil` label reproduces the send exactly; `"receive".localized` reproduces the receive, including the wrapping `VStack(spacing: 8)` that only the labelled variant needs. The switch in `body` passes the label, so the two cases stay distinct where the distinction actually lives.

Fonts, colors, alignment and the 36pt icon size are unchanged, so the rendered tree is identical for both shapes.

johnnyluo (approved):
LGTM

PR #5191, opened 2026-08-17
fix(kamino): read a withdraw built as withdraw_from_available

### LINKED ISSUE #5182: Kamino withdraw is unreadable: kvaultWithdraw is the wrong Anchor discriminator
## Summary

`KaminoTransactionDecoder` cannot read any Kamino Earn **withdraw**. It looks for the Anchor discriminator of the kVault `withdraw` instruction, but `POST /ktx/kvault/withdraw` builds `withdraw_from_available` instead. The instruction is therefore unrecognised, `decode` returns `nil`, and the verify screen shows:

> These bytes call the Kamino vault program but could not be read as a deposit or a withdrawal. Do not sign them.

with the Join button disabled. Deposits are unaffected.

## Details

`KaminoSolanaInstructions.swift:89`:

```swift
static let kvaultWithdraw: [UInt8] = [0xb7, 0x12, 0x46, 0x9c, 0x94, 0x6d, 0xa1, 0x22]
```

The two candidate Anchor discriminators, first eight bytes of `sha256("global:<name>")`:

| Instruction | Discriminator | |
|---|---|---|
| `withdraw` | `b712469c946da122` | what `kvaultWithdraw` holds |
| `withdraw_from_available` | `1383709baadc2239` | what the endpoint actually builds |
| `deposit` | `f223c68952e1f2b6` | matches `kvaultDeposit`, which is why deposits work |

`withdraw_from_available` appears nowhere in the repo.

Failure path: `KaminoInstructionSequence.kind(program:data:)` returns `nil` for the instruction → `decode` finds zero kVault positions → the `vaultPositions.count == 1` guard at `KaminoTransactionDecoder.swift:228-231` fails → `nil` → `KaminoVerifyPresentation.state(for:)` returns `.unreadable`.

## Evidence

A live `POST /ktx/kvault/withdraw` against the Steakhouse USDC vault (`HDsayqAsDWy3QvANGqh2yNraqcD8Fnjgh73Mhb3WRS5E`) for a fully staked position, decoded:

```
[0] AssociatedToken  accounts=6   data=01
[1] Farms            accounts=4   data=5a5f6b2acd7c32e1  (unstake, u128 arg)
[2] Farms            accounts=7   data=2466bb31dc248443  (withdraw_unstaked_deposits)
[3] AssociatedToken  accounts=6   data=01
[4] kVault           accounts=22  data=1383709baadc2239  <-- withdraw_from_available
```

Every other part matches `expected(operation: .withdraw, hasFarm: true, ...)` exactly — order, account-count minimums and payload lengths all clear. The unstake argument is `263479000000000000000000`, exactly `amount x 10^18` for the `263479` the kVault instruction burns, so all three bounds in `readArguments` pass. The discriminator is the only rejection.

This is not a recent change and is not specific to staked positions: an unstaked withdraw captured earlier (a wallet holding no position at all) carries the same `1383709baadc2239`.

## Why the tests pass

`KaminoStakedWithdrawTests` builds its fixtures from the `.kvaultWithdraw` **enum case** rather than from captured bytes, so the constant is never compared against a real Kamino response.

## Impact

No Kamino withdraw can be verified or co-signed on iOS, whichever platform initiated it, staked or unstaked. Nothing is at risk — the refusal is fail-closed and the signed bytes are correct — but the feature is unusable in the withdraw direction. Neither platform has shipped Kamino Earn.

## Suggested remediation

- Recognise `withdraw_from_available` (`1383709baadc2239`).
- Consider keeping `withdraw` as an accepted alternative rather than replacing it: a withdraw that exceeds the vault's liquid buffer may well use the other instruction, and that shape has not been captured. Worth confirming against a large withdraw before deciding.
- Derive the constants from `sha256("global:<name>")` in a test rather than trusting the literals, and add at least one fixture built from a real endpoint response so a decoder constant cannot drift from what Kamino sends.

For reference, vultisig-android now pins the same two discriminators and asserts they are the Anchor hashes of the instructions they name.

gastonm5:
> ⚠️ **This is a signing-path verify decoder.** It reads the raw bytes a user is
> about to co-sign and tells the verify screen what they do. It wants a human
> review and an on-device pass before merge. **Nothing was signed or broadcast in
> producing this change** — every capture below is a `POST` to
> `/ktx/kvault/withdraw`, which returns *unsigned* bytes for any address.

## Problem

`KaminoTransactionDecoder` could not read **any** Kamino Earn withdraw. The verify
screen showed

> These bytes call the Kamino vault program but could not be read as a deposit or
> a withdrawal. Do not sign them.

with Join disabled — on every device, whichever platform initiated, staked or
unstaked. Deposits were unaffected.

Nothing was ever at risk: the refusal is fail-closed and the bytes themselves are
correct. But the feature is unusable in the withdraw direction.

## Root cause

A kVault withdraw does not arrive as one instruction. It arrives as either
`kvault::withdraw` or `kvault::withdraw_from_available`, and which one depends on
whether the request fits the vault's liquid buffer.

| instruction | `sha256("global:<name>")[0..<8]` |
|---|---|
| `withdraw` | `b712469c946da122` — the only one the app knew |
| `withdraw_from_available` | `1383709baadc2239` — what the API builds today |
| `deposit` | `f223c68952e1f2b6` — unchanged, which is why deposits worked |

`KaminoInstructionSequence.kind(program:data:)` returned `nil` for the
unrecognised discriminator → `decode` found zero kVault positions → the
`vaultPositions.count == 1` guard failed → `nil` → `.unreadable`.

This is API drift, not a constant that was always wrong. The wiki's 2026-08-05
capture round recorded the kVault withdraw as `b712469c946da122` with 33
accounts; the *same* wallet, vault and amount re-requested on 2026-08-17 is
`1383709baadc2239` with 22.

## Fix

Accept **both** discriminators as the one existing `Kind.kvaultWithdraw`.

Collapsing them to a single `Kind` is deliberate. They are the same operation —
same `u64` shares argument, same authority, same share account burned, same
payout account credited — differing only in where the liquidity comes from, which
is a fact about the vault's balance sheet and not about what the user is being
asked to approve. A second `Kind` would need the step template to express
alternation (it cannot) and would fork `Layout(operation:)`, the validator's
dispatch and the paired-step machinery for a distinction the screen must not
draw.

`withdraw` is **kept**, not replaced — see below.

## What I proved about the `withdraw_from_available` account layout

This was the load-bearing risk: `KaminoInstructionAccounts.KvaultWithdraw` holds
positional indices (`user=0, vault=1, userTokenAccount=5, tokenMint=6,
userShareAccount=7, sharesMint=8`), and reusing them for a *different*
instruction would make the verify screen attest to accounts it has not proven —
worse than the refusal this removes.

**Four independent lines of evidence say the map transfers exactly.**

**1 — The program's IDL.** `src/idl/kvault.json` in `Kamino-Finance/klend-sdk`
(kamino_vault 2.2.2) declares `withdraw`'s accounts as a *nested group*:

| instruction | flattened accounts |
|---|---|
| `withdrawFromAvailable` | 14: `user, vaultState, globalConfig, tokenVault, baseVaultAuthority, userTokenAta, tokenMint, userSharesAta, sharesMint, tokenProgram, sharesTokenProgram, klendProgram, eventAuthority, program` |
| `withdraw` | 25: `withdrawFromAvailable.*` — the same 14, in the same order — then `withdrawFromReserveAccounts.*` (9), then `eventAuthority, program` |

So the shared prefix exists **by construction**, not by coincidence. The IDL's
names line up with the app's constants one for one, and every index the app reads
is ≤ 8.

**2 — A paired live capture.** Same wallet, same vault (RWA USDC), built minutes
apart at two amounts:

| amount | discriminator | instruction | accounts |
|---|---|---|---|
| `0.1` shares | `1383709baadc2239` | `withdraw_from_available` | 28 (14 + 14 remaining) |
| `22319` shares — the whole position | `b712469c946da122` | `withdraw` | 39 (25 + 14 remaining) |

Resolved through the address lookup table and compared **by pubkey**, the two
agree on slots 0–13 exactly and first diverge at slot 14 — precisely where the
IDL puts the reserve group. The trailing 14 `remaining_accounts` are identical
too. Slot 1 resolves to the RWA vault, slot 6 to the USDC mint, slot 8 to the RWA
shares mint, all matching `KaminoVaultRegistry`.

**3 — The builder's own branch.** `buildShareExitIxs` in klend-sdk picks
`withdrawFromAvailableIxs` when nothing has to come out of the reserves and
otherwise builds `withdrawIx`. That is why `withdraw` is kept: it is not legacy,
it is the exceeds-the-buffer path, and capture 2 above shows it live today.

**4 — Codex, independently.** The read-only review round corroborated the
ordering from Kamino's *program source* (`handler_withdraw.rs`) rather than from
the SDK's IDL copy, and confirmed nothing downstream branches on the raw
discriminator.

### What I did NOT prove

- **A withdraw that must exit MORE THAN ONE reserve.** `buildReserveExitIxs`
  loops over the reserves in the liquidity plan and emits one `withdraw` per
  reserve. The decoder's `vaultPositions.count == 1` guard refuses that, and the
  step template names only one `kvaultWithdraw` — so such a transaction stays
  **unreadable**, fail-closed, exactly as it is today and no worse than before
  this change. Left alone deliberately: summarising N instructions as one amount
  is a claim the screen has never made and needs its own design. Not observed in
  any capture — the 22,319-share whole-position withdraw above was still a single
  instruction.
- **Provenance of the checked-in captures.** As with any golden fixture, the
  base64 blobs and the lookup-table snapshot cannot prove *themselves*; they are
  inputs. Every assertion about them is against something else — a SHA-256
  derivation, the pinned registry, or a locally derived ATA.

## Tests

New `KaminoWithdrawFromAvailableTests` — 15 tests, all from captured bytes, none
built from the enum the production code reads:

- both withdraw discriminators recomputed from `sha256("global:<name>")` with
  **CryptoKit** (a different hash implementation from the app's), plus the
  deposit, plus distinctness;
- each capture's kVault instruction read off the wire: discriminator and `u64`
  share argument;
- `kind()` recognises all three; `decode` reads all three end-to-end (operation,
  descriptor, amount, signer);
- the pair `invokesKaminoVault == true` / `decode != nil` — the exact pair the
  screen keyed `.unreadable` off;
- **the shared-prefix proof**: both RWA captures resolved through the real lookup
  table agree on slots 0–13 by pubkey and differ at 14;
- **the guard on the map**: every index the app reads is `< sharedPrefixCount`,
  so adding one outside the proven region fails;
- both instructions name the vault, both mints and the signer's own ATAs;
- the 2026-08-05 `withdraw` and the 2026-08-17 `withdraw_from_available` for the
  same request carry the same accounts;
- all three run through `KaminoTransactionValidator` — the initiating device's
  pre-signing check, with the lookup table resolved;
- repointing the share source at the payout account is refused, asserting the
  whole `KaminoValidationError.accountNotOwnedByUser(role: "withdraw share
  source", actual:)` rather than "something threw";
- fail-closed preserved: a third real kVault discriminator and a one-byte edit
  are both still refused; the deposit path is untouched.

`KaminoTransactionValidatorTests.testAnchorDiscriminatorsMatchTheirInstructionNames`
extended to five names.

Three vectors added to `KaminoTransactionFixtures`, plus the RWA vault's lookup
table contents (read from mainnet 2026-08-17).

## Verification

| gate | result |
|---|---|
| full unit suite, iPhone 17 Pro Max sim | `** TEST SUCCEEDED **` — **6506 tests, 11 skipped, 0 failures** |
| swiftlint, 5 changed files | 0 violations |
| Codex read-only, 2 rounds | no production findings; 2 test-quality findings, both fixed |

Not yet run on device — please exercise a real Kamino withdraw on the verify
screen before merge.

Plan and full review ledger:
`wiki/pages/projects/vultisig/kamino-earn-5017/withdraw-discriminator-5182-plan.md`

Closes #5182

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing both Kamino vault withdrawal transaction variants.
  - Improved handling of withdrawals from available funds and reserve exits.
  - Enhanced transaction decoding and account resolution across both withdrawal paths.

- **Bug Fixes**
  - Improved validation for malformed or unrelated Kamino instructions.
  - Preserved existing deposit transaction handling while supporting the new withdrawal variant.

- **Tests**
  - Added comprehensive coverage for withdrawal decoding, validation, discriminator handling, and account compatibility.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoInstructionSequence.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoInstructionSequence.swift
@@ -274,7 +274,22 @@ enum KaminoInstructionSequence {
             return nil
         case .kvault:
             if Array(data.prefix(8)) == KaminoInstructionDiscriminator.kvaultDeposit { return .kvaultDeposit }
-            if Array(data.prefix(8)) == KaminoInstructionDiscriminator.kvaultWithdraw { return .kvaultWithdraw }
+            // TWO discriminators, one kind. `withdraw` and
+            // `withdraw_from_available` are the same withdraw — same `u64` share
+            // argument, same authority, same share account burned, same payout
+            // account credited — and which one the builder emits says only
+            // whether the vault's liquid buffer covered the request. That is a
+            // fact about the vault's balance sheet, not about what the user is
+            // being asked to approve, so drawing the distinction here would put
+            // a difference on the verify screen that means nothing to the person
+            // reading it.
+            //
+            // They may share the map because they share the accounts: see
+            // `KaminoInstructionAccounts.KvaultWithdraw`.
+            if Array(data.prefix(8)) == KaminoInstructionDiscriminator.kvaultWithdraw
+                || Array(data.prefix(8)) == KaminoInstructionDiscriminator.kvaultWithdrawFromAvailable {
+                return .kvaultWithdraw
+            }
             return nil
         case .memo:
             // The tag is matched WHOLE, not as a prefix. A memo is free-form
--- a/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoSolanaInstructions.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/Protocols/Kamino/KaminoSolanaInstructions.swift
@@ -72,7 +72,7 @@ enum KaminoAttribution {
 
 /// Instruction discriminators for the programs a Kamino Earn transaction uses.
 ///
-/// The six Anchor ones are `sha256("global:<name>")[0..<8]`, and each was also
+/// The seven Anchor ones are `sha256("global:<name>")[0..<8]`, and each was also
 /// read back out of a mainnet-simulated transaction built by the Kamino API, so
 /// the constant and the observation agree.
 ///
@@ -86,7 +86,29 @@ enum KaminoInstructionDiscriminator {
     static let kvaultDeposit: [UInt8] = [0xf2, 0x23, 0xc6, 0x89, 0x52, 0xe1, 0xf2, 0xb6]
     /// `kvault::withdraw(u64 shareAmount)` — the argument is in SHARES, the
     /// inverse of deposit's unit.
+    ///
+    /// One of TWO instructions a withdraw arrives as; see
+    /// `kvaultWithdrawFromAvailable`. This is the one the builder emits when the
+    /// request does not fit the vault's liquid buffer, because only this one
+    /// carries the accounts needed to pull the shortfall out of a lending
+    /// reserve.
     static let kvaultWithdraw: [UInt8] = [0xb7, 0x12, 0x46, 0x9c, 0x94, 0x6d, 0xa1, 0x22]
+    /// `kvault::withdraw_from_available(u64 shareAmount)` — the same withdraw,
+    /// served entirely out of the vault's liquid buffer.
+    ///
+    /// Which of the two arrives is a fact about the VAULT's liquidity at build
+    /// time, not about the user's request: Kamino's builder emits this one when
+    /// the reserves need not be touched and `withdraw` when they must, and both
+    /// are in live use. So both are accepted, and both mean the same thing to
+    /// the person signing — burn this many shares from this account, pay out to
+    /// that one.
+    ///
+    /// Reading one under the other's account map is safe, and that is checked
+    /// rather than assumed: the program's IDL declares `withdraw`'s accounts as
+    /// `withdraw_from_available`'s fourteen followed by the reserve group, so the
+    /// two share a prefix by construction. Every index in
+    /// `KaminoInstructionAccounts.KvaultWithdraw` lies inside it.
+    static let kvaultWithdrawFromAvailable: [UInt8] = [0x13, 0x83, 0x70, 0x9b, 0xaa, 0xdc, 0x22, 0x39]
     /// `farms::initialize_user` — creates the user's farm state. Absent when it
     /// already exists.
     static let farmsInitializeUser: [UInt8] = [0x6f, 0x11, 0xb9, 0xfa, 0x3c, 0x7a, 0x26, 0xfe]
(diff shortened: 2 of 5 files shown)


johnnyluo (approved):
LGTM

PR #5193, opened 2026-08-17
fix(wallet): stop the portfolio search bar scrolling itself off-screen

### LINKED ISSUE #5177: [Fix] Focusing portfolio search scrolls the search bar off-screen [Wallet / iOS]
# [Fix] Focusing portfolio search scrolls the search bar off-screen [Wallet / iOS]

## Problem

On iPhone, opening the chain search in the Wallet portfolio focuses the field and presents the keyboard, then the chain list jumps downward and pushes the search bar above the visible viewport. The user is left looking at mid-list results with the keyboard covering the lower half of the screen, but cannot see what they are typing or easily edit the query.

Chain list jumped below the hidden search bar while the iOS keyboard is open

`VaultMainScreen` currently waits 0.7 seconds after showing search, then centers a 300-point marker inside the chain list with `scrollTo(..., anchor: .center)`. That delayed forced scroll matches the visible jump and moves the focused search header off-screen. `DefiMainScreen` uses the same pattern and must be checked for the same bug.

## Expected Behavior

When search receives focus, the search field and Cancel action should remain visible above the keyboard. The results region may resize or scroll underneath, but focusing the field must not move the field itself out of view.

## Steps to Reproduce

1. Open a vault with enough enabled chains to create a long portfolio list.
2. Open the Wallet portfolio on an iPhone.
3. Tap the magnifier in the Portfolio section header.
4. Wait for the keyboard and delayed list animation.
5. Observe that the chain list jumps and the focused search bar is no longer visible.

## Solution

Remove the delayed center-scroll to the marker inside the chain list. If repositioning is needed, scroll the search header itself to a keyboard-safe top anchor and only by the minimum required amount. Apply the same behavior to the equivalent DeFi search flow, which currently duplicates the delayed marker-centering logic.

### Must Do

- [ ] Keep the Wallet portfolio search field and Cancel action visible while the keyboard is open.
- [ ] Remove the delayed post-focus jump caused by centering the chain-list marker.
- [ ] Keep the first matching result visible above the keyboard when a query is entered.
- [ ] Preserve or restore a predictable list offset when search is cancelled and reopened.
- [ ] Verify short and long chain lists on the smallest supported iPhone and a current large iPhone.
- [ ] Check and fix the same search-scroll pattern in `DefiMainScreen`.
- [ ] Add an iOS UI regression test that focuses search, waits past the former 0.7-second delay, and asserts the search field remains visible and hittable.

### Must NOT Do

- Don't dismiss or suppress the keyboard to avoid the layout issue.
- Don't add another timing delay or device-specific scroll offset.
- Don't change chain filtering, sorting, balances, or enabled-chain state.
- Don't break pull-to-refresh, the collapsing portfolio-balance header, or Cancel behavior.
- Don't refactor unrelated wallet navigation or business logic.

## Acceptance Criteria

- [ ] `make build-check` succeeds.
- [ ] `make test` succeeds.
- [ ] The targeted Wallet/DeFi search UI test succeeds.
- [ ] The search field remains fully visible and tappable before, during, and after keyboard presentation.
- [ ] No delayed scroll jump occurs after focus, including after waiting longer than one second.
- [ ] Typing filters the list while keeping the query and at least the first matching result visible.
- [ ] Cancelling search dismisses the keyboard and returns to a stable, predictable portfolio position.

gastonm5:
## Problem

Tapping the magnifier in the Wallet portfolio header focuses the chain search and raises the keyboard — and then, 0.7 seconds later, throws the search bar clean off the top of the screen. The user is left typing into a field they cannot see, looking at mid-list results. `DefiMainScreen` carried the same pattern.

## Runtime evidence

Confirmed in the simulator before writing anything, not reasoned about. A temporary DEBUG-only seeder put a 41-chain vault on a clean simulator, a temporary OSLog probe sampled the section header's global frame / scroll offset / keyboard frame / bottom safe-area inset at 10 Hz, and a temporary XCUITest drove the taps and read element frames. All of it was stripped before commit.

**iPhone 17 / iOS 26.5, 41 chains** — `headerY` is the section header's global y:

| t (s) | event | headerY | keyboard top | bottom safe area |
|---|---|---|---|---|
| 0.17 | focus set | 500 | — | 83 |
| 0.28 | `keyboardWillShow` | 500 | 539 | 335 |
| 0.70 | the app's 0.7 s timer fires | 500 | 539 | 335 |
| 0.91 | forced scroll running | −423 | 539 | 335 |
| 1.12 | settled | **−1074** | 539 | 335 |

The field ends 1074 points above the screen and leaves the accessibility tree entirely (`exists == 0`). The settled screenshot is a pixel match for the one attached to the issue.

Reproduced on both tab-bar paths: iPhone 16 / iOS 18.4 (custom tab bar) goes 497 → **−1116**, and there the keyboard additionally dismisses itself mid-scroll because the first responder has left the screen.

## Root cause

`VaultMainScreen` waited 0.7 s after focus and then called `scrollTo(marker, anchor: .center)`. That marker is a 300 pt `VStack` used as the **`.background` of `VaultMainChainListView`** — and a background is centred on its host, so the marker's centre is *the chain list's centre*. `anchor: .center` therefore scrolls the middle of a 2600 pt list to the middle of the viewport, which for any real vault is a scroll of well over a thousand points.

The trace confirms it exactly rather than approximately: the marker settles at `markerY = 150.3` and is 300 pt tall, so its centre lands at 300.3 — the centre of the keyboard-adjusted viewport (≈ 59 + (539 − 59)/2).

`DefiMainScreen` repeated the pattern against a zero-height `VStack {}` placed *after* its list — same bug, different wrong target.

`git log -S 'vaultMainScreenBottomContentId'` returns one commit, `67115f760` "feat: Defi Main Tab (#3204)", which is the commit that created these screens. **The delay was never patching a later regression**, so removing it does not revert anyone's fix.

One expectation the measurement corrected: the bottom safe-area inset moves 83 → 335 on keyboard show on *both* tab-bar paths, so SwiftUI keyboard avoidance is live here. The `.ignoresSafeArea(.all)` on `VultiTabBar.legacyTabBar` does not disable it, because the per-page `safeAreaInset` is published inside it.

## Fix, and why the smaller one was not enough

**Option A — just delete the forced scroll.** Measured: the header stays at 500 for the whole trace and the field is hittable, but typing `sol` puts the first matching row at y = 570 against a keyboard top of 539 — behind the keyboard, `isHittable == 0`. That fails the Must-Do "keep the first matching result visible above the keyboard", so Option A does not stand on its own.

**Option B — shipped.** Both screens now scroll on the `showSearchHeader` change itself, with no timer, to an anchor that parks the section header one `contentInset` below the top of the viewport. `scrollTo(anchor: .top)` on its own would put the header *behind* the floating home header, so the anchor is a clear box `2 * contentInset` taller than the 42 pt header slot, used as its `.background`: a background is centred on its host and takes no part in layout, so its top edge sits exactly `contentInset` above the header. That is the same clearance `VaultMainScreenScrollView`'s `topInset` already gives the rest of the content — no new constant, no device-specific offset, no delay.

(A first attempt using a zero-height anchor with negative top padding was measured and rejected: SwiftUI resolved the scroll against the padded container, the header landed at y = 62 behind the home header, `isHittable == 0`.)

**Option C** — lifting the header into `.safeAreaInset(edge: .top)` — was not needed.

Also removed: DeFi's now-dead `searchScrollTask` and its `onDisappear` cancel, and `VaultMainScreen`'s unused `@State var frameHeight`.

### After the fix

| device / OS | list | field on open | first match after typing | keyboard top |
|---|---|---|---|---|
| iPhone 17 / iOS 26.5 (native tab bar) | 41 chains | 151 | 458 | 539 |
| iPhone 16 / iOS 18.4 (custom tab bar) | 41 chains | 148 | 315 | — |
| iPhone SE 3rd gen / iOS 26.5 | 3 chains | 145 | 327 | 407 |

All hittable. Cancel is hittable throughout; cancelling dismisses the keyboard; reopening returns to a byte-identical offset (151 → 151). Pull-to-refresh and the collapsing balance header are untouched.

### The DeFi tab, measured separately

DeFi ships identical code, but its geometry is not Wallet's, so it was measured on its own rather than assumed: `CoinAction.defiChains` has **8 entries**, so that list can never exceed 8 rows, and DeFi's top section is only `DefiMainBalanceView` + a separator where Wallet also carries `CoinActionsView` and `BannersCarousel`. The section header starts at y ≈ 316 on DeFi versus 500 on Wallet.

**Before the fix, iPhone 17 / iOS 26.5, 8 DeFi chains:** the field settles at **y = −30.3**, `isHittable = 0`, Cancel `isHittable = 0`. The same defect, two orders of magnitude smaller than Wallet's −1074 purely because the list is capped at 8 rows.

**After the fix:**

| device / OS | header before tap | header settled | field | first match | keyboard top |
|---|---|---|---|---|---|
| iPhone 17 / iOS 26.5 | 316 | 140 | 150.7 | 395.7 | 539 |
| iPhone 16 / iOS 18.4 | 258 | 137 | 147.7 | 324.7 | 516 |
| iPhone SE 3rd gen / iOS 26.5 | 275 | 98 | 108.5 | 336.5 | 407 |

Field, Cancel and first match hittable in every row; the header settles by t ≈ 0.4 s and does not move again through t = 1.8 s, well past the old timer.

**Short and empty DeFi lists** — the case where the scroll cannot reach the anchor: with 1 chain and with 0 chains (the customize-chains empty state), the header stays at 316 and `onScrollOffsetChange` never fires at all, because the content is not scrollable and `scrollTo` is a no-op. That is the right outcome rather than a failure — the field sits at 326.7 against a keyboard top of 539, first match at 395.7, Cancel hittable, and nothing jumps. The anchor only moves the header when there is something to scroll.

## Tests — and whether they actually execute

`VultisigAppUITests` **has never compiled.** It referenced `AccessibilityID`, but `project.yml` only fed it `VultisigAppUITests/`, and a UI-test bundle links nothing from the app — so every UI test was *unrunnable*, not merely skipped, since `8e4d61086` "feat: ui testing foundation (#4049)". CI never noticed: the only workflow does not run `ui_test`. This PR adds `AccessibilityID.swift` to that target's sources (separate products, so no duplicate-symbol risk).

New `PortfolioSearchTests` covers Wallet and DeFi: open search, type a character and assert the field's value (proving focus without depending on the software keyboard), settle for 1.5 s — past the delay the removed scroll used — then assert the field and Cancel are present, inside the window, and hittable, and that cancelling restores the search control.

Honest status:

- **On a clean CI simulator these SKIP, they do not pass.** Every home-screen test is guarded by `guard isOnHomeScreen else { throw XCTSkip(...) }` and a clean simulator has no vault. Verified on a freshly erased simulator: `** TEST SUCCEEDED **`, both new tests reported `skipped`.
- **On a simulator carrying a vault they execute and pass**, and both **fail against the old code** — verified by restoring each screen's delayed centre-scroll in turn. Wallet fails on the presence assertion (the field has left the viewport within a second of the tap); DeFi fails on the window-bounds assertion (`DeFi search field at (52.0, -30.33, 235.67, 21.0) is outside the window (0.0, 0.0, 402.0, 874.0)`).
- **The DeFi case executes on iOS 26, it does not skip.** `AccessibilityID.Home.defiTab` genuinely does not exist on the iOS 26 native tab bar (measured: `exists = 0`), but `app.tabBars.buttons["DeFi"]` does and switches tabs; on iOS 18.4 the identifier exists and works instead. `resolvedDefiTab()` tries both, so the tab-bar skip gate does not fire on either OS.
- **A caught false pass.** The first version of this suite typed the focus-proving character *before* asserting visibility. On Wallet that was harmless, but on DeFi it made the test green against the bug: the field settles at y = −30 but is still in the accessibility tree, and typing filters the 8-row list to empty, which shrinks the content, clamps the scroll back and returns the field to view before the assertion runs. The assertions now run on the untouched focused layout first, then typing proves focus, then they run again.
- Coverage is weak on a *short* vault: with few chains the scroll clamps against the content's end before the field can leave the screen. Documented in the test file.

**Proposal, not built here:** a DEBUG-only seeded-vault launch argument would turn these from skips into real CI coverage. Happy to do it in a follow-up if you want it — I used exactly such a seeder as throwaway scaffolding to get this measured.

## Gates

- `make build-check` — `** BUILD SUCCEEDED **`
- `make test` — `** TEST SUCCEEDED **`, 6480 test cases passed, 0 failures
- `make ui_test` on a clean simulator — `** TEST SUCCEEDED **` (new tests skip, as above)
- `swiftlint --config VultisigApp/.swiftlint.yml` — 0 violations on all touched files
- Read-only Codex review, 2 rounds — one P2 on the test's determinism, addressed; no blocking findings remaining

## What is not verified

- **macOS and iPad.** The change is vertical-scroll-only in shared code, but neither was exercised at runtime.
- **Two pre-existing UI-test failures, unrelated to this change.** Now that the target compiles, `testHomeScreenShowsWalletTab` and `testNavigateToSettings` fail *when a vault exists* — the first because the iOS 26 native tab bar carries no accessibility identifiers and `HomePage.assertTabsExist()` also asserts an "Agent" tab that `HomeTab` does not have; the second at `SettingsPage.assertVisible()`. Both skip on a clean simulator, so `make ui_test` is green there. Left alone deliberately as out of scope — worth a follow-up issue.

Plan and full measurements: `wiki/pages/projects/vultisig/vault-chain-list/search-focus-scroll-5177-plan.md`.

Closes #5177

## Summary by CodeRabbit

* **Accessibility**
  * Added accessibility identifiers for Wallet and DeFi search controls.
  * Search fields now support improved automated accessibility.

* **Bug Fixes**
  * Improved search behavior so results scroll into position immediately when search opens.
  * Search focus and cancellation now behave more consistently across Wallet and DeFi screens.

* **Tests**
  * Added UI coverage for search focus, scrolling, visibility, cancellation, and restoration in Wallet and DeFi.

### DIFF
--- a/VultisigApp/VultisigApp/Components/TextField/SearchTextField.swift
+++ b/VultisigApp/VultisigApp/Components/TextField/SearchTextField.swift
@@ -12,18 +12,21 @@ struct SearchTextField: View {
     @Binding var isFocused: Bool
     let showPasteButton: Bool
     let placeholder: String
+    let accessibilityIdentifier: String?
     @FocusState var focusedState: Bool
 
     init(
         value: Binding<String>,
         isFocused: Binding<Bool> = .constant(false),
         showPasteButton: Bool = false,
-        placeholder: String = "search".localized
+        placeholder: String = "search".localized,
+        accessibilityIdentifier: String? = nil
     ) {
         self._value = value
         self._isFocused = isFocused
         self.showPasteButton = showPasteButton
         self.placeholder = placeholder
+        self.accessibilityIdentifier = accessibilityIdentifier
     }
 
     var showClearButton: Bool {
@@ -45,6 +48,9 @@ struct SearchTextField: View {
                 .colorScheme(.dark)
                 .padding(.horizontal, 8)
                 .focused($focusedState)
+                .unwrap(accessibilityIdentifier) { view, id in
+                    view.accessibilityIdentifier(id)
+                }
 
             HStack(spacing: 8) {
                 clearButton
--- a/VultisigApp/VultisigApp/Core/Utils/AccessibilityID.swift
+++ b/VultisigApp/VultisigApp/Core/Utils/AccessibilityID.swift
@@ -17,6 +17,18 @@ enum AccessibilityID {
         static let balanceLabel = "home.balanceLabel"
     }
 
+    enum VaultMain {
+        static let searchButton = "vaultMain.searchButton"
+        static let searchField = "vaultMain.searchField"
+        static let searchCancelButton = "vaultMain.searchCancelButton"
+    }
+
+    enum DefiMain {
+        static let searchButton = "defiMain.searchButton"
+        static let searchField = "defiMain.searchField"
+        static let searchCancelButton = "defiMain.searchCancelButton"
+    }
+
     enum Settings {
         static let container = "settings.container"
         static let languageCell = "settings.languageCell"
--- a/VultisigApp/VultisigApp/Features/Defi/DefiMain/DefiMainScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiMain/DefiMainScreen.swift
@@ -23,11 +23,11 @@ struct DefiMainScreen: View {
     @State var showSearchHeader: Bool = false
     @State var focusSearch: Bool = false
     @State var showChainSelection: Bool = false
-    @State private var searchScrollTask: Task<Void, Never>?
     @State private var clearSearchTask: Task<Void, Never>?
     @State private var refreshTask: Task<Void, Never>?
 
-    private let scrollReferenceId = "DefiMainScreenBottomContentId"
+    private let searchAnchorId = "defiMainSearchHeaderAnchor"
+    private let sectionHeaderHeight: CGFloat = 42
     private let contentInset: CGFloat = 78
     private let horizontalPadding: CGFloat = 16
 
@@ -58,16 +58,10 @@ struct DefiMainScreen: View {
             .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
             .background(MainBackgroundWithNotification())
             .onChange(of: showSearchHeader) { _, showSearchHeader in
-                if showSearchHeader {
-                    focusSearch = true
-                    searchScrollTask?.cancel()
-                    searchScrollTask = delayedTask(after: .milliseconds(700)) {
-                        withAnimation {
-                            scrollProxy?.scrollTo(scrollReferenceId, anchor: .center)
-                        }
-                    }
-                } else {
-                    searchScrollTask?.cancel()
+                guard showSearchHeader else { return }
+                focusSearch = true
+                withAnimation {
+                    scrollProxy?.scrollTo(searchAnchorId, anchor: .top)
                 }
             }
             .crossPlatformSheet(isPresented: $showChainSelection) {
@@ -86,7 +80,6 @@ struct DefiMainScreen: View {
             refresh()
         }
         .onDisappear {
-            searchScrollTask?.cancel()
             clearSearchTask?.cancel()
             refreshTask?.cancel()
         }
@@ -102,7 +95,8 @@ struct DefiMainScreen: View {
                 }
             }
             .transition(.opacity)
-            .frame(height: 42)
+            .frame(height: sectionHeaderHeight)
+            .background(searchScrollAnchor)
             .padding(.bottom, 16)
 
             DefiChainListView(
@@ -110,15 +104,24 @@ struct DefiMainScreen: View {
                 viewModel: viewModel,
                 onCustomizeChains: onCustomizeChains
             )
-            VStack {}
-                .background(
-                    // Reference to scroll when search gets presented
-                    VStack {}
-                        .frame(height: 300)
-                        .id(scrollReferenceId)
-                )
         }
         .id(vault.id)
+    }
+
+    /// Where the section header is parked when search opens.
+    ///
+    /// A background is centred on its host and does not affect layout, so a box
+    /// `2 * contentInset` taller than the header has its top edge exactly
+    /// `contentInset` above the header's. Scrolling *that* edge to the top of
+    /// the viewport leaves the header one floating-home-header height down —
+    /// the same clearance `VaultMainScreenScrollView`'s `topInset` gives the
+    /// rest of the content — so the field lands below that header rather than
+    /// behind it, with the results underneath it and above the keyboard.
+    var searchScrollAnchor: some View {
+        Color.clear
+            .frame(height: sectionHeaderHeight + 2 * contentInset)
+            .id(searchAnchorId)
+            .allowsHitTesting(false)
     }
 
     var defaultBottomSectionHeader: some View {
@@ -136,6 +139,7 @@ struct DefiMainScreen: View {
             CircularAccessoryIconButton(icon: .magnifier) {
                 toggleSearch()
             }
+            .accessibilityIdentifier(AccessibilityID.DefiMain.searchButton)
             CircularAccessoryIconButton(icon: .housePen, type: .secondary) {
                 showChainSelection.toggle()
             }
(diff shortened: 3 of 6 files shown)


johnnyluo (approved):
LGTM

PR #5197, opened 2026-08-18
fix(wallet): stop offering DeFi receipts in the token-selection sheet

### LINKED ISSUE #5194: [Fix] ybRUNE DeFi receipt is offered in the token-selection sheet
## Problem

Right after a vault is imported, opening THORChain → **Select tokens** lists **ybRUNE** among the selectable tokens.

ybRUNE (`x/staking-x/brune`) is the auto-compounding receipt for bonded bRUNE — a DeFi position, not a spendable wallet token — so it should not be offered as one. Enabling it from this sheet puts a DeFi receipt into the wallet token list.

**sTCY** is listed on the same sheet, and it is the other ticker the app already classifies as DeFi-only.

## Steps to reproduce

1. Import a vault (first import / fresh install).
2. Open the THORChain chain screen.
3. Open the token-selection sheet ("Select tokens").
4. ybRUNE is listed alongside bRUNE, TCY, sTCY. Searching `yb` also returns it.

## Expected

ybRUNE is never offered in the token-selection sheet, matching how it is already excluded from the chain detail list, the swap coin picker, and the vault's fiat total.

### Screen Record

https://github.com/user-attachments/assets/f0178635-da94-43a1-9085-51192d52dbdb

## Actual

It is offered as an ordinary selectable token and can be enabled into the wallet list.

## Notes

- `Coin.isDefiOnly` / `Coin.defiOnlyTickers` (`["STCY", "YBRUNE"]`) is consulted by the chain detail list, the swap coin selection, and the array fiat total. The token-selection surface does not consult it — which matches what the recording shows for both tickers.
- Behaviour in the recording: ybRUNE and bRUNE are both present when the sheet is first opened after import; after saving a selection and reopening the sheet, neither is in the list. So the leak is most visible on the first open after import, which is also the moment a user is most likely to be picking tokens.
- Recorded on the macOS build. The sheet is shared SwiftUI, so iOS is expected to behave the same — worth confirming on device before closing.

## Related

- Android: vultisig/vultisig-android#5583 (same receipt, hidden from the wallet token list)

gastonm5:
## Description

Right after a vault import, THORChain → **Select tokens** listed **ybRUNE** (`x/staking-x/brune`) — the auto-compounding receipt for bonded bRUNE — as an ordinary selectable wallet token, so a DeFi position could be enabled into the wallet token list. `Coin.isDefiOnly` / `defiOnlyTickers` (`["STCY", "YBRUNE"]`) already gated the chain-detail list, the swap picker's held tokens and the vault fiat total; this surface never consulted it.

Fixes #5194

### Three paths, not one

Closing any single one of these changes nothing:

1. **Held** — THORChain discovery (`ThorchainService.fetchTokens`) adds whatever denoms the address carries, and its ybRUNE/sTCY handling is *casing-only*, excluding nothing. A staker's vault genuinely holds a ybRUNE `Coin`, so it arrives as a held token. This is the path the issue's recording shows.
2. **Curated preset** — `TokensStore.ybrune` is in `TokenSelectionAssets`, so it arrives again via `preExistingTokens`.
3. **Catalog breadth** — `BundledTokensProvider` is registered into `TokenCatalogRepository` and its tokens are `.curated`, so they auto-surface. Dropping the receipt from the presets alone simply re-admits it as `browseProviderTokens`.

All three producers in `TokenSelectionLogic` now filter it, which covers browse and search together since the search pool is merged from exactly those lists. `defiOnlyTickers` moves onto `CoinMeta` (the selection surfaces are `CoinMeta`-typed end to end) with `Coin` forwarding to it, so the two can never disagree.

bRUNE and TCY are unaffected — the rule matches a **whole ticker, not a prefix**, which is worth stating because the `ThorchainService` filter next door matches these same tickers by anchored prefix. Pinned by a test.

### The part that isn't the filter

Hiding the row left a way to delete the coin. `CoinService.saveAssets` removes every held coin absent from the selection it is handed, and that selection is shared state seeded by `ChainDetailScreen`'s **periodic refresh** — not by opening the sheet. So a receipt discovery added since the last seed was deleted and hidden by a save the user made about entirely different tokens.

That hazard pre-existed for every token, but the row used to render (unchecked), so it was visible and recoverable by hand. Once the receipt cannot render at all, the same save is silent — for exactly the coins that represent staked funds. `TokenSelectionLogic.selectionPreservingDefiPositions` now carries held positions through at save time, so the guarantee comes from the save path rather than from when the seed last ran. The sheet has **two** exits and both are covered (`TokenSelectionScreen` and the custom-token screen, which saves the same shared selection).

The on-chain position was never at risk; what was lost is the local receipt coin and the wallet's tracking of it — including the DeFi card that unwinds it, which reads the held coin.

### Reachability is unchanged

The receipt stays enable-able through the DeFi position picker, which builds its list from `TokensStore` directly and never through the token catalog (`DefiPositionsService.stakeCoins`). Pinned by a test. sTCY has lived under exactly this arrangement since it was added.

## Which feature is affected?

## Checklist

## Tests

14 new tests. The ones that matter:

- The issue's repro at view-model level — a vault holding ybRUNE loads `.thorChain`; the receipt is absent from browse *and* from a search for `yb`, while bRUNE stays.
- One per producer (held / curated preset / provider breadth), plus a generic one over **every** ticker in `defiOnlyTickers` present in `TokenSelectionAssets`, so a third DeFi ticker is covered the day it is added rather than needing a new test.
- The save path driven for real through `CoinService.saveAssets`, from both sheet exits — **with a control asserting the raw selection genuinely deletes and hides the receipt**, so the protection is provably load-bearing rather than passing by construction.
- bRUNE is not DeFi-only despite being a prefix of the receipt it mints.

Gate: `** TEST SUCCEEDED **` — **6530 tests, 11 skipped, 0 failures** (6516 on the base + 14). All 14 confirmed present and passing in the log rather than inferred from the marker.

## Additional context

Three whole-branch `codex exec --sandbox read-only` passes; the final pass is clean. The review is what surfaced the save-path defect above and the fact that the sheet has two exits — the first version of this fix covered only one, and its "non-destruction" test was vacuous (it would have passed with the whole fix reverted). One Major was rejected with reason and the rejection independently re-verified: the native-token guard is outcome-neutral because `findAllCoinsToRemove` already removes every coin on a chain whose native is absent.

**Known, deliberately out of scope — the swap picker leaks the same receipts.** `SwapCoinSelectionViewModel.assemble` filters `isDefiOnly` on the **vault-held** tokens only; its `externalTokens` come from `TokenSearchService.loadTokens` → the same catalog that carries the curated `ybrune`, unfiltered, on both the source and destination sides. So ybRUNE/sTCY are still offered as swap assets. Not fixed here to keep this PR to the reported surface; happy to fold it in or file it separately.

Not verified on device: the issue is labelled `mac` and asks for an iOS confirmation before closing. The sheet is shared SwiftUI so behaviour should be identical, but that is reasoning, not a device run.

Related: android [#5583](https://github.com/vultisig/vultisig-android/issues/5583) hid the same receipt from the wallet token list.

- **Agent**: Claude Code
- **Issue**: #5194
- **Confidence**: high
- **Needs human review for**: the save-path carry-through (`selectionPreservingDefiPositions`) — it changes what a token-selection save persists, and the shared `CoinSelectionViewModel` seeding it is stale-prone in ways that predate this PR.

## Summary by CodeRabbit

- **New Features**
  - Improved DeFi token handling in wallet token selection.
  - Held DeFi positions remain preserved when saving token selections or adding custom tokens.
  - DeFi-only receipts are excluded from general browsing and search while remaining available in the DeFi position picker.

- **Bug Fixes**
  - Prevented held DeFi positions from being unintentionally removed when updating token selections.
  - DeFi token classification now consistently recognizes supported tickers regardless of letter casing.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/CoinExtension.swift
@@ -15,10 +15,14 @@ extension Coin {
         return chain.coinType
     }
 
-    static let defiOnlyTickers: Set<String> = ["STCY", "YBRUNE"]
+    /// Forwards to `CoinMeta.defiOnlyTickers`, which is the single definition.
+    /// A `Coin` and the `CoinMeta` it came from must never disagree about
+    /// whether a ticker names a DeFi position — the selection surfaces reason in
+    /// `CoinMeta`, everything downstream of a save reasons in `Coin`.
+    static var defiOnlyTickers: Set<String> { CoinMeta.defiOnlyTickers }
 
     var isDefiOnly: Bool {
-        Coin.defiOnlyTickers.contains(ticker.uppercased())
+        CoinMeta.defiOnlyTickers.contains(ticker.uppercased())
     }
 
     /// The balance in base units, read exactly.
--- a/VultisigApp/VultisigApp/Features/Wallet/TokenSelection/Screens/TokenSelectionScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Wallet/TokenSelection/Screens/TokenSelectionScreen.swift
@@ -102,14 +102,21 @@ struct TokenSelectionScreen: View {
     }
 
     private func persistSelection() {
-        // Held DeFi positions are carried through explicitly: this sheet never
-        // renders them, so the selection cannot speak for them, and a save is
-        // otherwise a removal for anything the selection omits.
-        let selection = TokenSelectionLogic.selectionPreservingDefiPositions(
-            selection: coinViewModel.selection,
-            vaultCoins: vault.coins
-        )
         Task {
+            // Held DeFi positions are carried through explicitly: this sheet
+            // never renders them, so the selection cannot speak for them, and a
+            // save is otherwise a removal for anything the selection omits.
+            //
+            // Read immediately before the save rather than before this task:
+            // token discovery is the concurrent producer at the heart of this
+            // bug, and a union formed earlier would not carry a coin that
+            // landed in between.
+            let selection = await MainActor.run {
+                TokenSelectionLogic.selectionPreservingDefiPositions(
+                    selection: coinViewModel.selection,
+                    vaultCoins: vault.coins
+                )
+            }
             await CoinService.saveAssets(for: vault, selection: selection)
             await MainActor.run { isPresented = false }
         }
--- a/VultisigApp/VultisigApp/Features/Wallet/TokenSelection/Screens/TokenSelectionScreen.swift
+++ b/VultisigApp/VultisigApp/Features/Wallet/TokenSelection/Screens/TokenSelectionScreen.swift
@@ -102,8 +102,15 @@ struct TokenSelectionScreen: View {
     }
 
     private func persistSelection() {
+        // Held DeFi positions are carried through explicitly: this sheet never
+        // renders them, so the selection cannot speak for them, and a save is
+        // otherwise a removal for anything the selection omits.
+        let selection = TokenSelectionLogic.selectionPreservingDefiPositions(
+            selection: coinViewModel.selection,
+            vaultCoins: vault.coins
+        )
         Task {
-            await CoinService.saveAssets(for: vault, selection: coinViewModel.selection)
+            await CoinService.saveAssets(for: vault, selection: selection)
             await MainActor.run { isPresented = false }
         }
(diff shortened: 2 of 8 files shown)


johnnyluo (approved):
LGTM

PR #5201, opened 2026-08-18
fix(defi): fade the banner over its own height, not half of it

### LINKED ISSUE #5198: [Fix] DeFi tab banner fades out mid-scroll, leaving a blank gap
## Problem

On the **DeFi tab**, the portfolio banner fades out roughly **halfway through its own height**. If you stop scrolling there, the banner is already invisible but still occupying its full layout height, so the screen shows a blank gap between the top bar and the portfolio list.

The banner should stay visible until it has actually scrolled out from behind the top bar — i.e. the fade should finish when the banner is fully hidden, not at mid-scroll.

## Steps to reproduce

1. Open the **DeFi** tab.
2. Scroll down slowly, about a third of the banner's height.
3. Stop.

**Expected:** the banner is still partly visible, fading in step with how much of it is left on screen.
**Actual:** the banner is fully transparent while ~80pt of it is still on screen — a blank gap that stays until you scroll further.

## Cause

`HeaderCollapseProgress` (`Features/Home/Model/HeaderCollapseProgress.swift`) drives the collapse crossfade off the scroll offset:

- `defaultDistance = 110` — the scroll distance the whole collapse plays over.
- `expandedOpacity = max(1 - value / 0.5, 0)` — the content side fades out over the **first half** of that, so it reaches zero after **55pt** of scroll.

`DefiMainScreen` applies `.headerCollapseFade(collapse, tab: .defi)` to `DefiMainBalanceView`, which is a fixed **`.frame(height: 135)`** card. 55pt of scroll ≠ 135pt of banner, so the card is fully transparent with ~80pt still laid out on screen. `.opacity()` does not reclaim layout, hence the blank space rather than the list moving up.

## Why the Wallet tab looks fine

`VaultMainScreen` fades `VaultMainBalanceView`, which has no fixed height — it is text, and short enough that the 55pt ramp roughly coincides with it leaving the viewport. The DeFi banner is a 135pt illustrated card, so the same constant reads as a bug there.

## Suggested direction

Tie the expanded-side ramp to the height of the thing being faded rather than to a single global constant, so the fade completes exactly as the banner finishes scrolling under the top bar. Worth checking against the Wallet tab so its current feel is preserved.

## Notes

- Cosmetic; no funds or signing involved.
- Reported on macOS; the screen is shared SwiftUI, so iOS is expected to behave the same — worth confirming on device.

gastonm5:
## Description

On the DeFi tab the portfolio banner faded to fully transparent about halfway down its own height. `.opacity` does not reclaim layout, so stopping mid-scroll left the card's 135pt still occupying the screen with nothing drawn in it — a blank gap between the top bar and the portfolio list.

Fixes #5198

### Cause

The collapse ramp was a single global 110pt (`HeaderCollapseProgress.defaultDistance`), and its expanded half — the one that fades the content's balance out — therefore finished after 55pt. That suits the wallet tab, whose balance is text roughly that tall. The DeFi banner is a fixed 135pt illustrated card, so the same constant emptied it with ~80pt still on screen.

### What changed

The two sides of the collapse measure different things, and no longer share one length:

- the **expanded** ramp spans the height of whatever that tab fades out — the banner, on DeFi;
- the **collapsed** ramp is a fixed `barFadeDistance`, *added* after it, so the top bar's entrance is the same length on every tab.

`midpoint` is derived from the two rather than fixed at one half.

The first attempt simply doubled the shared distance to 270pt, and review caught that this trades one visual bug for another: the top bar's **opaque background** rides the collapsed ramp, so the separator and the portfolio header would have scrolled under a still-translucent bar from ~155pt all the way to 270pt. Adding a fixed bar ramp instead of scaling the whole distance keeps that window at 55pt — the same as the wallet tab has always had.

### The wallet tab is untouched

Not by inspection but by arithmetic: `(110 - 55) / 110` is exactly one half in binary floating point, which is the midpoint the old fixed constant used. Every existing wallet expectation passes unchanged, and the exactness matters — `headerCollapseOpacity` gates hit-testing and VoiceOver on `opacity > 0` / `opacity == 0` comparisons.

## Which feature is affected?

## Checklist

## Tests

7 new, on top of the existing suite for this type:

- The banner is still drawn at every offset below its own height, and reaches zero exactly at it.
- At the old ramp's zero point (55pt) it is still more than half visible — the blank gap, stated as an assertion.
- **The bar comes in over the same distance on every tab** — the property the first attempt broke.
- The wallet's midpoint is still exactly `0.5` over the same 110pt.
- The suite's existing never-both-legible sweep now runs against the DeFi ramp too.
- The ramp is pinned to `DefiMainBalanceView.bannerHeight`, so the two cannot drift apart silently.

One pre-existing test (`testStoreKeepsTheTabsIndependent`) asserted a value derived from the old shared constant and moved with the change; review confirmed that is a legitimate expectation change, not a masked regression (reverting the production code while keeping the tests fails it).

Gate: `** TEST SUCCEEDED **` — **6523 tests, 11 skipped, 0 failures**. SwiftLint clean on every touched file.

## Additional context

Two `codex exec --sandbox read-only` passes. Round 1 found the merge-blocking Major described above; round 2 found no Major or Minor, only two documentation nits (both fixed — the type doc still described "halves", and a test comment credited the wallet's stability to a 55pt content height that nothing establishes).

Not verified on device: reported on macOS, and the screen is shared SwiftUI so iOS should match, but this is a scroll-feel change and is worth a look on both before merge.

- **Agent**: Claude Code
- **Issue**: #5198
- **Confidence**: high
- **Needs human review for**: the scroll feel itself — the DeFi collapse now runs over 190pt instead of 110pt, which is a deliberate but subjective change.

## Summary by CodeRabbit

* **UI Improvements**
  * Improved the DeFi header collapse transition with a longer banner fade window.
  * Kept the DeFi banner visible while it scrolls offscreen, preventing overlap with the collapsed balance.
  * Preserved existing collapse behavior for Wallet and Camera tabs.
  * Standardized the DeFi banner height for consistent display.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Defi/DefiMain/View/DefiMainBalanceView.swift
+++ b/VultisigApp/VultisigApp/Features/Defi/DefiMain/View/DefiMainBalanceView.swift
@@ -8,6 +8,15 @@
 import SwiftUI
 
 struct DefiMainBalanceView: View {
+    /// Height of the banner card.
+    ///
+    /// Named because the home header's collapse ramp is derived from it: the
+    /// banner is faded out by scroll, and `.opacity` does not reclaim layout, so
+    /// a ramp shorter than this height empties the card while it is still on
+    /// screen and leaves a blank gap. `HeaderCollapseProgress` carries the
+    /// matching distance, and a test pins the two together.
+    static let bannerHeight: CGFloat = 135
+
     @ObservedObject var vault: Vault
     @EnvironmentObject var homeViewModel: HomeViewModel
 
@@ -30,7 +39,7 @@ struct DefiMainBalanceView: View {
                 )
             }
         }
-        .frame(height: 135)
+        .frame(height: Self.bannerHeight)
         .background(
             Theme.colors.bgSurface1
                 .overlay(gradientBackground.clipped())
--- a/VultisigApp/VultisigApp/Features/Home/Model/HeaderCollapseProgress.swift
+++ b/VultisigApp/VultisigApp/Features/Home/Model/HeaderCollapseProgress.swift
@@ -11,13 +11,19 @@ import Foundation
 /// the scroll offset, plus the two fade ramps that value drives.
 ///
 /// The large balance in the scroll content and the balance in the top bar are
-/// faded over **non-overlapping** halves of the progress: everything belonging
-/// to the expanded state fades out over `0…0.5`, everything belonging to the
-/// collapsed state fades in over `0.5…1`. `expandedOpacity` and
-/// `collapsedOpacity` are therefore never both greater than zero, so the two
-/// balances can never be legible at the same time — and because the whole
-/// transition is a function of the scroll offset it scrubs with the drag and
-/// reverses with it, instead of running on its own wall-clock timeline.
+/// faded over **non-overlapping** stretches of the progress, meeting at
+/// `midpoint`: everything belonging to the expanded state fades out over
+/// `0…midpoint`, everything belonging to the collapsed state fades in over
+/// `midpoint…1`. `expandedOpacity` and `collapsedOpacity` are therefore never
+/// both greater than zero, so the two balances can never be legible at the same
+/// time — and because the whole transition is a function of the scroll offset it
+/// scrubs with the drag and reverses with it, instead of running on its own
+/// wall-clock timeline.
+///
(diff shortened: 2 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5202, opened 2026-08-18
fix(defi): open the unstake sheet against the amount the card showed

### LINKED ISSUE #5199: [Fix] THORChain unstake sheet opens with no balance for bonded positions (TCY, RUJI)
## Problem

On THORChain DeFi positions, the **unstake sheet opens with no balance** — the available amount reads zero and the percentage slider has nothing to work against — even though the position card it was opened from shows a real stake. TCY and RUJI both reproduce it.

The card and the sheet disagree about the same position: the card enables **Unstake**, and the sheet it opens says there is nothing to unstake.

## Steps to reproduce

1. Hold a THORChain stake (TCY or RUJI — the plain **bonded** position, not the auto-compounding card).
2. DeFi tab → THORChain → the position card shows a non-zero staked amount.
3. Tap **Unstake**.

**Expected:** the sheet opens with the same amount the card showed as the ceiling.
**Actual:** the amount is zero / blank.

## Cause

Two different quantities are used for "how much can be unstaked", and the sheet is handed the wrong one.

`DefiChainScreenModel.unstakeTransactionType(for:)` routes a `.stake` position as:

```swift
return .unstake(
    coin: position.coin,
    isAutocompound: false,
    availableToUnstake: position.availableToUnstake   // <- nil for THORChain
)
```

`THORChainStakeInteractor` never sets `availableToUnstake` — every `StakePositionData` it builds (TCY, RUJI, and the rest) carries `amount` only. So the sheet receives `nil` and falls back to the other branch of:

```swift
availableAmount = availableToUnstake ?? coin.stakedBalanceDecimal
```

`coin.stakedBalanceDecimal` is a *different* field, written by `BalanceService.fetchStakedBalance` on the wallet refresh cycle — not by the DeFi interactor that produced the number the card is rendering. Whenever it hasn't been written for that coin yet, the sheet opens at zero while the card shows the real position.

Two things make this specifically a **bonded**-position bug:

- The `.compound` branch right below passes `availableToUnstake: position.amount` explicitly, so auto-compounding cards work.
- `StakePosition.canUnstake` — which decides whether the card's Unstake button is enabled at all — already uses `availableToUnstake ?? amount`. The gate and the sheet disagree, which is exactly why the button is live and the sheet is empty.

## Suggested direction

Hand the sheet the same quantity the gate used and the card displayed:

```swift
availableToUnstake: position.availableToUnstake ?? position.amount
```

That keeps `availableToUnstake` meaningful where an interactor does set it (Maya sets it deliberately — see the comment in `MayaChainStakeInteractor`), and stops the sheet silently falling through to a field maintained on a different schedule by a different service.

Worth a test that pins the invariant rather than the symptom: **if `canUnstake` is true, the sheet's ceiling is non-zero.**

## Notes

- No funds at risk — the sheet refuses to build rather than building a zero-amount transaction.
- Reported on THORChain; the same routing is shared by every non-TON chain, so other chains whose interactor omits `availableToUnstake` are affected the same way.

gastonm5:
## Description

A THORChain bonded stake position's card showed a real stake, and the unstake sheet it opened offered nothing to unstake. Reproducible on TCY and RUJI.

Fixes #5199

### Cause

The card and the sheet were reading two different numbers.

`StakePosition.canUnstake` — which decides whether the Unstake button is even live — resolves `availableToUnstake ?? amount`. The route handed the sheet `availableToUnstake` alone:

```swift
case .stake:
    return .unstake(coin: position.coin, isAutocompound: false,
                    availableToUnstake: position.availableToUnstake)   // nil on THORChain
```

Only interactors that distinguish a withdrawal ceiling from the position's size set that field — Maya does, THORChain does not — so it arrived `nil` and the sheet fell through to its own fallback, `coin.stakedBalanceDecimal`. That field belongs to `BalanceService`'s refresh cycle rather than to the DeFi read that produced the position, so the two could disagree, and did. Review confirmed the fallback can be genuinely unpopulated, stale, *or* falsely zero: the per-chain refresh updates only native RUNE, a failed TCY read is swallowed as zero, a failed RUJI read becomes `"0"`, and RUJI's path picks the first `stakingV2` entry rather than selecting RUJI explicitly the way the DeFi interactor does.

The `.compound` branch immediately below already passed `position.amount`, which is why auto-compounding cards worked and bonded ones did not.

### What changed

Route the same expression the gate uses:

```swift
availableToUnstake: position.availableToUnstake ?? position.amount
```

An explicit ceiling still wins where an interactor sets one, and a genuinely empty position still routes zero rather than being papered over — the sheet's own guard is what should refuse there.

### The ceiling is the right quantity

This is signing-adjacent — the builders derive a withdrawal *fraction* from the ceiling — so it was checked rather than assumed. TCY's `amount` comes only from the TCY staker response (`x/staking-tcy` is a separate sTCY position), and RUJI's comes specifically from `bonded.amount` (`liquidSize` and the sRUJI receipt shares stay separate). Neither includes an auto-compound receipt, so the ceiling cannot over-state what `account.withdraw` / `tcy-:<bps>` can actually withdraw. Slider and default percentages are unchanged because `percentageSelected` wins; a typed amount changes basis points only where the old fallback disagreed with the card — i.e. where the denominator was already wrong.

## Which feature is affected?

## Checklist

## Tests

5 new, pinning the invariant rather than the symptom:

- **A card that offers Unstake always opens a non-zero sheet** — the gate and the route resolving the same quantity is the actual contract, and its violation is the bug.
- A bonded position routes the position's amount.
- An explicit ceiling (Maya) still wins and is not overwritten by the fallback.
- A genuinely empty position still routes zero.
- The `.compound` branch is unchanged.

Gate: `** TEST SUCCEEDED **` — **6521 tests, 11 skipped, 0 failures**. All 5 new tests confirmed present and passing in the log. SwiftLint clean on touched files.

## Additional context

One `codex exec --sandbox read-only` pass; **no defect found in this change**. It did surface two pre-existing, signing-adjacent defects in the *builders* downstream, which this PR deliberately does not touch — but which this fix makes reachable, since the sheet now actually opens with a balance:

1. **Typed TCY withdrawals truncate to whole percentages.** `Int(resolvedPercentage) * 100` — typing 12.5 TCY against a 100 TCY ceiling signs `tcy-:1200`, i.e. 12%. Already addressed by the open PR #5192, which rewrites exactly this path.
2. **Bonded RUJI can serialize a fractional base-unit amount.** `RUJIUnstakeTransactionBuilder` emits `amount.toDecimal() * 10^8` via `.description` without truncating, producing values like `7032429999.999999` where `account.withdraw.amount` expects integer base units. **Not covered by #5192** — that PR touches `RUJILiquidUnbondTransactionBuilder` (the compound arm), not this one. Currently unfiled.

Not verified on device — needs a THORChain vault holding a bonded TCY or RUJI position.

- **Agent**: Claude Code
- **Issue**: #5199
- **Confidence**: high
- **Needs human review for**: nothing in this diff specifically; but the two builder defects above are worth a decision, particularly the bonded-RUJI one, which is unfiled and not covered by #5192.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unstaking accuracy across bonded, compound, TON, Solana, and other supported positions.
  * Ensured eligible withdrawal amounts are displayed and applied consistently.
  * Preserved explicit limits, zero values, legacy data, and autocompound behavior.
  * Automatically updated existing positions with missing withdrawal amounts.

* **Tests**
  * Added coverage for unstake routing, migration handling, position scenarios, and supported staking networks.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Migration/AppMigrationService.swift
+++ b/VultisigApp/VultisigApp/Core/Migration/AppMigrationService.swift
@@ -101,7 +101,8 @@ struct AppMigrationService {
             TonGramRebrandMigration(),
             PromoBannerDismissalMigration(),
             RujiAutoCompoundPositionMigration(),
-            TonGramDefiPositionsMigration()
+            TonGramDefiPositionsMigration(),
+            StakePositionCeilingMigration()
         ]
     }
 }
--- a/VultisigApp/VultisigApp/Core/Migration/Migrations/StakePositionCeilingMigration.swift
+++ b/VultisigApp/VultisigApp/Core/Migration/Migrations/StakePositionCeilingMigration.swift
@@ -1,0 +1,55 @@
+//
+//  StakePositionCeilingMigration.swift
+//  VultisigApp
+//
+
+import SwiftData
+
+/// Gives every stake position persisted before the withdrawable amount was
+/// required one, so no card can offer Unstake against a row that never said how
+/// much could come out.
+///
+/// The unstake sheet takes its ceiling from `StakePosition.availableToUnstake`.
+/// Producers must now state it, but rows already on disk predate that and carry
+/// `nil` — and nothing corrects them until a refresh both runs and succeeds. A
+/// cached row paints an enabled card immediately, the refresh behind it is not
+/// awaited, and a read that fails writes nothing at all, so a user who is
+/// offline or hitting a failing endpoint could hold a `nil` row indefinitely
+/// while its card invited them into an empty sheet.
+///
+/// `amount` is the right value to backfill because it is exactly what those rows
+/// meant: every producer that existed when they were written treated the
+/// position's size as the withdrawable quantity, and the old routing read the
+/// same field. This records that reading once, in the data, rather than leaving
+/// every future reader to re-derive it — which is the reason the field was made
+/// mandatory in the first place.
+///
+/// Rows written after this are unaffected: they already carry a stated ceiling,
+/// and the guard skips them.
+struct StakePositionCeilingMigration: @MainActor AppMigration {
+    /// Surfaced when the store cannot be read. Throwing leaves the migration
+    /// version un-bumped so `AppMigrationService` retries on the next launch
+    /// rather than marking the backfill as done with no data.
+    private enum MigrationError: Error {
+        case missingModelContext
+    }
+
+    let version: Int = 5
+
+    let description: String = "Recording how much can be withdrawn from existing stake positions"
+
+    @MainActor
+    func migrate() throws {
+        guard let modelContext = Storage.shared.modelContext else {
+            throw MigrationError.missingModelContext
+        }
+
+        let descriptor = FetchDescriptor<StakePosition>()
+
+        for position in try modelContext.fetch(descriptor) where position.availableToUnstake == nil {
+            position.availableToUnstake = position.amount
+        }
+
+        try Storage.shared.save()
+    }
+}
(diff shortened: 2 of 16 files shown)


johnnyluo (approved):
LGTM

PR #5297, opened 2026-08-31
feat: add app-configured widget watchlist

### LINKED ISSUE #5286: [Add] Configurable watchlist [Widgets settings]
# [Add] Configurable watchlist [Widgets settings]

## Problem

The Watchlist widgets need an app-owned configuration flow so users can control which assets appear. Without shared persisted selection, the widget can only show a fixed list and cannot distinguish an untouched default from a deliberately cleared watchlist.

## Solution

Add Settings → Widgets → Watchlist with icon, ticker, name, and on/off controls. Persist an ordered selection through the App Group, default untouched users to CoinGecko's top five assets, and render the first three selections in Medium widgets and up to five in Large widgets. Use the existing shared HTTP transport for widget market requests.

Implementation: #5285

### Must Do

- [ ] Add second-level Widgets → Watchlist settings navigation.
- [ ] Default an unconfigured watchlist to the top five assets.
- [ ] Persist explicit selections and an explicitly empty selection through the App Group.
- [ ] Reload widget timelines after selection changes.
- [ ] Render up to three assets in Medium and five in Large Watchlist widgets.
- [ ] Use downloaded CoinGecko asset icons and the shared `TargetType` / `HTTPClientProtocol` transport.
- [ ] Use `withLoading` and center the empty-selection message.

### Must NOT Do

- Don't add separate per-widget watchlists.
- Don't add manual reordering or long-tail pagination in this iteration.
- Don't edit the generated Xcode project directly.

### Out of Scope

- Search and long-tail asset pagination.
- Manual watchlist reordering.
- Independent selections for each widget instance.

## Acceptance Criteria

- [ ] Users can enable or disable assets under Settings → Widgets → Watchlist.
- [ ] Medium and Large widgets reflect the stored ordered selection and size limits.
- [ ] Untouched users see the default top five; a deliberately cleared list stays empty.
- [ ] Market data and icons load through the shared networking abstraction.
- [ ] Widget tests, the full app test suite, and iOS/macOS builds pass.

gastonm5:
## Summary

- restore Medium and Large Watchlist widgets using the existing market-row and centered gradient sparkline presentation
- add Settings → Widgets → Watchlist configuration with asset icons, ticker, name, and on/off toggles
- share the ordered selection through the App Group, capped at five assets; Medium renders the first three and Large renders up to five
- default an unconfigured Watchlist to CoinGecko's top five while preserving an explicitly cleared selection
- reload the Watchlist timeline after changes and use the shared `withLoading` presentation in Settings
- reuse `TargetType` and `HTTPClientProtocol` for widget market API communication

## Context

This replaces #5285. That stacked PR was merged into the shared-resources feature branch rather than `main`, and its merge commit was subsequently reverted before #5284 landed. The watchlist commits have now been replayed directly onto current `main`.

## Verification

- `make generate`
- `make test`: 6,885 tests passed, 11 skipped, 0 failures
- SwiftLint: no serious violations and no new warnings (27-warning repository baseline)
- `git diff --check`: passed

## Follow-ups

- search and long-tail pagination
- manual reordering
- per-widget Watchlist selections

Closes #5286

## Summary by CodeRabbit

* **New Features**
  * Added a customizable crypto watchlist widget with medium and large layouts, refreshes, empty states, and unavailable-data messaging.
  * Added watchlist settings for selecting and managing up to five assets.
  * Added cached market data support for faster loading and improved refresh behavior.
* **Localization**
  * Added watchlist and widget translations in English, German, Spanish, Croatian, Italian, Korean, Portuguese, and Simplified Chinese.
* **Bug Fixes**
  * Improved network request handling and preserved existing selections when market data refreshes fail.


VultisigApp/VultisigWidgets/CryptoWatchlistWidget.swift
```diff
@@ -0,0 +1,221 @@
...
+                currency: currency,
+                hasSelection: true,
+                isStale: result.isStale
+            )
+        } catch {
+            return CryptoWatchlistEntry(
+                date: Date(),
+                assets: [],
+                currency: currency,
+                hasSelection: true,
```
johnnyluo: **Bug:** the catch block always sets `hasSelection: true`, even when the request only ran because no watchlist has ever been configured (`hasStoredSelection == false`).

Fresh install, user never opens Widget Watchlist settings, and the `.top(limit:)` fallback network call fails (offline/timeout/proxy error). The widget renders with `hasSelection: true`, so `unavailableKey` resolves to "widget.marketDataUnavailable" ("temporarily unavailable") instead of "widget.watchlist.configure" ("Choose assets in Vultisig Settings"), telling the user to wait out an outage when they actually need to open Settings and configure the widget.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+    @Published var isLoading = false
+    @Published private(set) var loadFailed = false
+
+    private let marketClient: any WidgetMarketRemote
+    private let defaults: UserDefaults?
+    private var hasLoaded = false
+    private var hasStoredSelection: Bool
+
+    init(
+        marketClient: any WidgetMarketRemote = WidgetMarketClient(),
```
johnnyluo, line 22: **Efficiency:** `load()` calls `WidgetMarketClient.markets()` directly, bypassing `WidgetMarketService`'s actor-level cache (`Shared/WidgetShared/WidgetMarketService.swift`) that the widget timeline provider already uses for the same catalog data.

Each time the Settings > Widget Watchlist screen is opened (a fresh `@StateObject` with `hasLoaded=false` every navigation), it re-fetches all 50 catalog assets from the network with no caching or stale-fallback, duplicating work the widget's cache layer already does and adding avoidable latency/proxy load.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+final class WidgetWatchlistSettingsViewModel: ObservableObject {
+    @Published private(set) var selectedAssets: [WidgetWatchlistAsset]
+    @Published private(set) var catalogAssets: [WidgetWatchlistAsset] = []
+    @Published var isLoading = false
+    @Published private(set) var loadFailed = false
+
+    private let marketClient: any WidgetMarketRemote
+    private let defaults: UserDefaults?
+    private var hasLoaded = false
+    private var hasStoredSelection: Bool
```
johnnyluo, line 19: **Correctness risk:** `hasStoredSelection` is cached as an ivar, manually kept in sync with `WidgetSharedStorage.hasStoredWatchlist(in:)` (set in init, set true in `persistSelection`) instead of just calling through to storage.

Any future write to the watchlist key that doesn't go through this view model's `persistSelection` (a second call site, a migration, a debug tool) desyncs the cached flag from actual UserDefaults state, causing `load()` to wrongly re-seed a user's explicitly-cleared watchlist with the top-5 catalog on next open.

VultisigApp/VultisigApp/Shared/WidgetShared/WidgetSharedStorage.swift
```diff
@@ -24,4 +27,87 @@ enum WidgetSharedStorage {
...
+    init(_ asset: WidgetMarketAsset) {
+        self.init(
+            id: asset.id,
+            symbol: asset.symbol,
+            name: asset.name,
+            imageURL: asset.imageURL
+        )
+    }
+
+    var iconLogo: String {
```
johnnyluo, line 100: **Duplication:** `WidgetWatchlistAsset.iconLogo` duplicates the id -> local-asset-name switch (bitcoin/ethereum/tether/binancecoin/solana) already defined verbatim in `WidgetMarketAsset.iconLogo` (`WidgetMarketAsset.swift:19`).

Adding, renaming, or fixing a hardcoded coin-icon mapping requires remembering to update both switch statements; they will silently drift as one is edited and the other is forgotten.

johnnyluo (commented):
Automated code review findings.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    func setSelected(_ selected: Bool, asset: WidgetWatchlistAsset) {
+        if selected {
+            guard !isSelected(asset), canSelectMore else { return }
+            selectedAssets.append(asset)
+        } else {
+            selectedAssets.removeAll(where: { $0.id == asset.id })
+        }
+
+        persistSelection(reloadWidget: true)
```
johnnyluo, line 93: **Efficiency:** `setSelected()` calls `persistSelection(reloadWidget: true)` synchronously on every toggle, unconditionally calling `WidgetCenter.shared.reloadTimelines(ofKind:)` with no debouncing.

A user rapidly toggling several assets while building their watchlist fires one `reloadTimelines` call per toggle; each triggers a full timeline recompute, which iOS may throttle/coalesce unpredictably instead of one reload after the user settles.

VultisigApp/VultisigApp/Features/Settings/Views/WidgetWatchlist/WidgetWatchlistSettingsScreen.swift
```diff
@@ -0,0 +1,142 @@
...
+                    assetSection
+                }
+            }
+        }
+        .screenTitle("watchlist".localized)
+        .withLoading(isLoading: $viewModel.isLoading)
+        .task { await viewModel.load() }
+    }
+
+    private var partialLoadError: some View {
```
johnnyluo, line 27: **Duplication:** `partialLoadError` (here) and `statusView`'s `loadFailed` branch (line ~81) both render the same "widgetWatchlistLoadError" text plus a retry button calling `viewModel.load(force: true)`, in two different layouts.

A copy or behavior fix to the retry flow (wording, accessibility label, button type) has to be applied in two places and can silently diverge between the two render paths.

VultisigApp/VultisigApp/Features/Settings/Views/Widgets/WidgetsSettingsScreen.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  WidgetsSettingsScreen.swift
+//  VultisigApp
+//
+
+import SwiftUI
+
+struct WidgetsSettingsScreen: View {
```
johnnyluo: **Simplification:** `WidgetsSettingsScreen` is an intermediate list screen with exactly one row (Watchlist) that just navigates to `WidgetWatchlistSettingsScreen`, adding an extra route, an extra tap, and extra code for a menu with a single, non-optional entry.

The PR's own "Follow-ups" section defers per-widget selection to future work, so today this screen has nothing else to list; `SettingsMainScreen` could navigate straight to `SettingsRoute.widgetWatchlist` and this container screen could be reintroduced once a second widget type actually exists.

VultisigApp/VultisigApp/Core/Networking/HTTPClient.swift
```diff
@@ -8,6 +8,14 @@
 import Foundation
 import OSLog
 
+private let defaultHTTPClientLogger: Logger = {
+    #if WIDGET_EXTENSION
```
johnnyluo, line 12: **Architecture:** `HTTPClient.swift` and `HTTPClientProtocol.swift` now branch on `#if WIDGET_EXTENSION` to pick a widget-specific `Logger` subsystem, reversing the deleted comment that the widget extension "cannot link the application's TargetType/HTTPClient graph" by threading the extension's needs directly into the app's core shared networking files.

Any future change to `HTTPClient`/`HTTPClientProtocol` (new imports, app-only dependencies, auth headers) must now stay widget-extension-safe or it silently breaks the `VultisigWidgets` build target — a constraint that used to be structurally impossible to violate and is now enforced only by programmer awareness of the `#if`.

VultisigApp/VultisigApp/Shared/WidgetShared/WidgetSharedStorage.swift
```diff
@@ -24,4 +27,87 @@ enum WidgetSharedStorage {
...
+              let assets = try? JSONDecoder().decode([WidgetWatchlistAsset].self, from: data) else {
+            return []
+        }
+        return normalizedWatchlist(assets)
+    }
+
+    static func hasStoredWatchlist(
+        in defaults: UserDefaults? = WidgetSharedStorage.defaults
+    ) -> Bool {
+        defaults?.object(forKey: watchlistKey) != nil
```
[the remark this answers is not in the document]
gastonm5: Fixed in 71acae065. hasStoredWatchlist now decodes the payload, treats corrupt data as absent, and preserves valid explicitly empty selections. Regression coverage verifies both the corrupt-state result and top-five reseeding. The full suite passes 6,931 tests with 0 failures.

VultisigApp/VultisigWidgets/CryptoWatchlistWidget.swift
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+        .padding(contentMargins)
+        .foregroundStyle(WidgetTheme.primaryText)
+    }
+
+    private var header: some View {
+        VStack(spacing: 8) {
+            HStack(alignment: .firstTextBaseline, spacing: 6) {
+                Text("widget.watchlist")
```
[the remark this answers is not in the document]
gastonm5: Reviewed and leaving this unchanged intentionally. The widget extension uses SwiftUI LocalizedStringKey and LocalizedStringResource directly; the app String.localized extension is not part of the widget target. This keeps localization native to WidgetKit without adding an app-only dependency, and both widget and app targets compile in the full test gate.

VultisigApp/VultisigWidgets/CryptoWatchlistWidget.swift
gastonm5: Fixed in 71acae065. The failure entry now carries hasStoredSelection, so a failed default top-market request on an unconfigured install renders the centered configure prompt, while a configured watchlist still reports unavailable market data.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
gastonm5: Fixed in 71acae065. Settings now uses WidgetMarketService, starts cache and remote reads concurrently, renders cached catalog data immediately, then applies the remote refresh. Remote failure retains cached rows and exposes the retry state. Added tests for both in-flight cache display and stale fallback.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
gastonm5: Reviewed and keeping the in-memory flag as intentional view-model state. It is initialized from storage and updated after every local persist; during the cache-first flow it also prevents the remote refresh from re-seeding after cached data has just seeded and persisted defaults. External mutation while this screen is alive is not currently supported. If another writer is introduced, the storage boundary should become observable rather than repeatedly reading UserDefaults.

VultisigApp/VultisigApp/Shared/WidgetShared/WidgetSharedStorage.swift
gastonm5: Fixed in 71acae065. The local id-to-asset mapping now lives in WidgetMarketAsset.localIconLogo(for:), and WidgetWatchlistAsset delegates to it after checking its remote image URL.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
gastonm5: Reviewed and keeping immediate reload semantics. Watchlist edits are explicit user actions capped at five assets, and each persisted toggle should make the widget eligible to reflect that exact state. WidgetKit still controls when timelines are actually scheduled. A debounce would add lifecycle and cancellation state for a very small bounded interaction.

VultisigApp/VultisigApp/Features/Settings/Views/WidgetWatchlist/WidgetWatchlistSettingsScreen.swift
gastonm5: Reviewed. The two paths are intentionally separate presentations for different states: an inline warning when cached rows remain usable, and a centered empty-state error when there are no rows. They share the localized copy and force-refresh behavior; extracting them now would still require divergent layout and button sizing. I would consolidate if retry behavior gains additional logic.

VultisigApp/VultisigApp/Features/Settings/Views/Widgets/WidgetsSettingsScreen.swift
gastonm5: Fixed in 71acae065. Settings now links directly to Watchlist, and the intermediate Widgets screen, route, builder method, enum case, and unused accessibility identifier were removed.

VultisigApp/VultisigApp/Core/Networking/HTTPClient.swift
gastonm5: This architecture is intentional for this PR: the widget reuses the existing TargetType and HTTPClientProtocol stack instead of maintaining a duplicate networking implementation. WIDGET_EXTENSION only selects the logger declaration, and the full app plus widget build passed in the 6,931-test gate. Extracting the networking primitives into a dedicated shared package remains a reasonable follow-up if this shared surface grows.

VultisigApp/VultisigApp/Features/Settings/ViewModels/WidgetWatchlistSettingsViewModel.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        async let cachedResult = marketService.cachedResult(query: query, currency: currency)
+        async let refreshedResult = marketService.load(
+            query: query,
+            currency: currency,
+            downloadsIcons: false
+        )
+
+        if let cachedResult = await cachedResult {
+            apply(cachedResult.assets)
```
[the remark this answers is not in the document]
gastonm5: Fixed in 269c3e252. Cached catalog data now hydrates the UI only. Default selection is persisted only from a non-stale refresh, so a failed remote refresh cannot freeze outdated rankings. The regression test verifies that no selection exists during cache hydration and that the fresh asset becomes the stored default. Full suite passes 6,931 tests with 0 failures.

johnnyluo (approved):
LGTM

PR #5314, opened 2026-09-01
fix(swap): explain min-output failures in transaction history

### LINKED ISSUE #5310: [Fix] Failed swap history: min-out reverts should say the price moved [Swap / History]
When a swap reverts on-chain with `Return amount is not enough` or `Insufficient output`, history looks like a generic failed swap. The user retried three times and paid gas each time.

We already have the right copy for quote-time THOR failures: `swapSlippageToleranceTooTight` = "Price moved beyond your slippage tolerance. Increase the slippage setting in swap settings, or try a smaller amount."

Use that (or the same idea) on the history / error leg when the receipt reverts with those strings.

Evidence: Kyber https://robinhoodchain.blockscout.com/tx/0xdf7169b3a449d30d0b873138a6094deebf6da85abaeb0c0c669ec90a064c1dba and LI.FI https://robinhoodchain.blockscout.com/tx/0xa1a1af0ea814e00e0027822b749cd8e76f7afa55813bd91a8980da59992ccdef

**Don't** change the swap form or add a new settings screen.

gastonm5:
## Problem

Failed EVM swaps whose on-chain revert reason reports `Return amount is not enough` or `Insufficient output` appear in transaction history with provider wording. That does not tell users that the quoted minimum output was missed because the price moved, which can lead to an ineffective retry and another gas payment.

## Root cause

The transaction-status path correctly extracts and persists the raw revert reason, but the history card and detail sheet rendered that stored value directly. There was no shared presentation mapping for actionable swap failure guidance.

## Implementation

- Add one pure transaction-history failure-reason presenter.
- Recognize both known minimum-output signatures case-insensitively, including inside wrapped RPC text.
- Render the existing localized `swapSlippageToleranceTooTight` guidance on both the compact card and detail sheet.
- Preserve unknown non-empty provider reasons unchanged and suppress nil/empty reasons.
- Keep raw stored reasons unchanged; this is presentation-only and resolves localization when rendered.
- Add focused coverage for both signatures, casing and wrapper text, unknown passthrough, nil, and empty input.

## Verification

- `make test`: 6,929 tests, 11 skipped, 0 failures (`** TEST SUCCEEDED **`)
- Changed-file SwiftLint: 0 violations
- `git diff --check`: clean
- Required cross-model review completed; final whole-branch pass reported no actionable findings

## Manual validation

Not performed locally. The focused validation target is a failed EVM swap row carrying either known revert signature: the card and detail sheet should show the same localized slippage guidance while the stored raw reason remains intact.

Closes #5310

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction history error messages by replacing technical insufficient-output errors with clear slippage-tolerance guidance.
  * Suppressed empty or missing failure messages.
  * Preserved meaningful details for other transaction errors.
  * Applied consistent error-message formatting across transaction cards and detail views.

* **Tests**
  * Added coverage for recognized, unknown, wrapped, empty, and missing failure reasons.

### DIFF
--- a/VultisigApp/VultisigApp/Features/TransactionHistory/Models/TransactionHistoryFailureReasonPresentation.swift
+++ b/VultisigApp/VultisigApp/Features/TransactionHistory/Models/TransactionHistoryFailureReasonPresentation.swift
@@ -1,0 +1,24 @@
+//
+//  TransactionHistoryFailureReasonPresentation.swift
+//  VultisigApp
+//
+
+enum TransactionHistoryFailureReasonPresentation {
+    private static let minimumOutputSignatures = [
+        "return amount is not enough",
+        "insufficient output"
+    ]
+
+    /// Converts a stored provider reason into the copy shown in transaction history.
+    /// Raw reasons remain unchanged in storage so localization is resolved when rendered.
+    static func displayText(for rawReason: String?) -> String? {
+        guard let rawReason, !rawReason.isEmpty else { return nil }
+
+        let normalizedReason = rawReason.lowercased()
+        if minimumOutputSignatures.contains(where: normalizedReason.contains) {
+            return "swapSlippageToleranceTooTight".localized
+        }
+
+        return rawReason
+    }
+}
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5321, opened 2026-09-03
fix(blockaid): net complete Solana asset diffs

### LINKED ISSUE #5195: Blockaid Solana: three-diff native-SOL filter can reverse the approval direction (mirrors vultisig-sdk#2091)
Tracking issue for the iOS side of vultisig/vultisig-sdk#2091.

**Summary**

`BlockaidSimulationParser.parseSolana` drops a native-SOL asset diff purely because the response contains exactly three diffs, assuming it is the transaction fee:

```swift
// VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift:119
if diffs.count == 3,
   let solIndex = diffs.firstIndex(where: { $0.asset.type == "SOL" || $0.assetType == "SOL" }) {
```

Nothing establishes the removed leg is fee-sized or outgoing. On a `native SOL in` + `WSOL residual out` + `receipt token out` shape, the principal inflow is filtered away and the parser falls through to `.transfer` of the WSOL residual — the approval screen states money *leaving* a transaction that nets to SOL *received*.

**Why the fix belongs in the SDK, not here**

This code deliberately mirrors `parseBlockaidSolanaSimulation` in `vultisig-sdk` (see the doc comment on `parseSolana`), and `vultisig-android` mirrors it too. Changing only iOS would diverge what the three approval screens say about the same transaction. The SDK issue carries the full analysis and the counterexample trace.

**Scope for this repo**

Mirror whatever resolution lands in vultisig/vultisig-sdk#2091. If the SDK closes it as a safe simplification, replace the implicit assumption here with an explicit comment stating the shapes it relies on.

**Not a regression**

Predates #5190 (which fixed an unrelated same-mint netting bug in this same function). Surfaced by a cross-model review pass on that PR.

**Caveat**

Derived from reading the code, not from a captured Blockaid response with this shape. Worth confirming against real payloads before treating it as user-facing.

gastonm5:
## Summary

- replace the positional three-diff native-SOL deletion heuristic with complete per-mint balance aggregation
- canonicalize native SOL and WSOL to the same mint, net every incoming and outgoing leg, and prefer native SOL metadata when it participates in a displayed direction
- fail closed to the generic approval title when the complete result has multiple net spends or malformed/inconsistent data
- add regressions for withdrawal and deposit direction, shuffled diff order, complete same-mint totals, ambiguous fee/principal shapes, and invalid legs

## Why

Blockaid can return a Solana withdrawal as native SOL in, a WSOL residual out, and a receipt token out. The old exactly-three-diffs heuristic deleted the native SOL leg unconditionally, then interpreted the remaining entries positionally. That could reverse the approval headline into a WSOL outflow even though the vault receives SOL.

This implementation does not guess whether a native movement is a fee from its magnitude. If the response contains two independent net outflows, the parser declines to produce a specialized hero instead of hiding a possible principal spend.

This changes approval presentation only; transaction construction and signing bytes are untouched.

## Linked work

- `vultisig-sdk#2091` remains open. SDK PR `#2108` remains open with an unresolved review noting that an absolute fee ceiling can still discard a small principal spend; PR `#2203` was closed without merge.
- `vultisig-android#5683` is closed by merged PR `#5730`; Android was already structurally immune and added the direction regression.
- `vultisig-windows#4728` remains open and depends on the SDK parser.
- iOS PR `#5190` introduced same-mint SOL/WSOL netting; this PR generalizes it across the full diff set.

## Verification

- `swiftlint lint --no-cache --config VultisigApp/.swiftlint.yml <changed files>` — 0 violations
- focused `BlockaidSolanaSimulationParserTests` — 35 tests, 0 failures
- `make test` — 6,970 tests, 11 skipped, 0 failures (`** TEST SUCCEEDED **`)
- Claude review skipped at the request of the maintainer

Closes #5195

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction simulation results by accurately netting incoming and outgoing asset movements.
  * Preserved native SOL details when processing SOL and wrapped SOL activity.
  * Prevented ambiguous, invalid, unsupported, or zero-value balance changes from being reported as transfers or swaps.
  * Improved handling of multi-leg transactions, inconsistent decimals, and multiple outgoing movements.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
+++ b/VultisigApp/VultisigApp/Core/Security/Providers/Blockaid/BlockaidSimulationParser.swift
@@ -102,15 +102,11 @@ enum BlockaidSimulationParser {
 
     // MARK: - Solana
 
-    /// Parses a Solana simulation response. Mirrors `parseBlockaidSolanaSimulation`:
-    /// when three diffs come back and one is native SOL, the SOL diff is
-    /// filtered out as "likely the transaction fee". One remaining diff maps to
-    /// `.transfer`, two remaining diffs map to `.swap` (or `.transfer` if only
-    /// the out side has a value).
-    ///
-    /// Diverges from the extension in one place: a two-diff pair that resolves
-    /// to the same mint is netted rather than reported as a swap — see
-    /// `netSameMint(outCoin:outAmount:inCoin:inAmount:)`.
+    /// Parses every Solana balance leg into complete per-mint net changes.
+    /// Native SOL and WSOL share the wrapped-SOL mint sentinel, so wrapping
+    /// noise is netted while distinct assets remain independent. A hero is
+    /// emitted only when the complete result is one transfer/receive or one
+    /// unambiguous swap; multi-spend shapes fail closed.
     static func parseSolana(
         response: BlockaidSolanaSimulationResponseJson
     ) -> BlockaidSimulationInfo? {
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5330, opened 2026-09-07
feat(history): swap/limit rows show FROM → TO pair and both amounts

### LINKED ISSUE #5325: [Update] Transaction History swap cards: show From → To and both amounts [History]
## Context
From Figma review on 2026-09-06. Transaction History → Swaps tab.

Figma: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=74765-110319

Parity: will be cross-linked to Android and Windows siblings.

## Today
Completed swap rows in `TransactionHistoryCardView.swift` collapsed layout show one amount (from-side) plus a `via {provider}` corner badge. They do not show the destination ticker or a From → To pair. Send rows already have a `to 0x…` pill; swaps do not get the equivalent.

## Change
1. On a completed swap or limit row, show both legs: `+{toAmount} {toTicker}` and `-{fromAmount} {fromTicker}`.
2. Replace the provider-as-pair cue with a pill `FROM → TO` (e.g. `USDC → SOL`).
3. Keep Successful / Error (incl. "Slippage tolerance exceeded") on the top-right. Send/receive rows stay as they are.

## Invariant
A completed swap row always names both assets. The user should not have to open the row to learn what they swapped into.

## Acceptance
Open Transaction History → Swaps after a USDC→SOL swap: the card shows +SOL, -USDC, and an `USDC → SOL` pill, matching the Figma frame.

gastonm5:
## Problem

Completed swap and limit rows in Transaction History showed only the from-side amount plus a `via {provider}` corner badge — no destination ticker, no From → To pair (send rows already have a `to 0x…` pill; swaps didn't get the equivalent). Figma review flagged the mismatch: [Transaction History - Swaps](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=74765-110319) (frame `74765-110319`).

## Fix

- `collapsedContent` (the completed-row layout) now branches to a `swapLegsColumn` (`+toAmount` primary / `-fromAmount` tertiary, same size, differing only by colour — confirmed against the Figma node) and a `swapPairPill` (`FROM → TO`, in the same pill slot/geometry `addressPill` already uses for a send row's `to 0x…`) whenever the row is a completed `.swap`/`.limit` with a known to-side.
- The `via {provider}` badge and its reserved bottom padding are now conditional (`showsViaBadge`), not unconditional on `swapProvider != nil`. It's **gone only** on a completed row already showing the new legs — every other swap-provider row keeps it exactly as before:
  - An **in-progress** swap/limit row (still on `expandedContent`, the from→to diagram) keeps the badge — that layout names both coins but never the provider, and the issue/Figma only cover completed rows.
  - A **native-source limit order** (`TransactionHistoryRecorder.recordLimitOrder` — an order placed from RUNE/BTC/etc. rather than via the swap screen) never resolves `toCoinTicker`/`toAmountCrypto` at all (the co-signer only sees the `=<` memo, never a real `Coin` — see the doc comment on that recorder), so it keeps the old single-amount layout *and* the badge, since it can't show a pair it doesn't have.
- Two static routing functions, `showsSwapLegs`/`showsViaBadge`, carry these decisions so they're pinnable by unit tests without rendering SwiftUI — matching the file's existing `shouldExpand`/`isLimitTerminal` pattern.
- The two legs rebuild their ticker from `toCoinTicker`/`coinTicker` via a shared `stripTicker` helper rather than trusting `toAmountCrypto`/`amountCrypto` to already carry it: the co-signer path (`TransactionHistoryRecorder.recordFromKeysignPayload`) records `toAmountCrypto` as a **bare number with no ticker suffix**, unlike `SwapDoneScreen`'s pre-formatted `"{amount} {ticker}"` — caught by cross-model review, fixed, and pinned by a test (`testStripTickerIsANoOpWhenTheSuffixIsAbsent`). `stripTicker` matches only the space-delimited suffix (not a bare `hasSuffix`) so a numeric-looking custom-token ticker (e.g. `"50"`) can't chew digits off an amount like `"200.50"`.

## Known caveats (not fixed here — flagging for follow-up / reviewer awareness)

- **`toAmountCrypto` is the swap quote's *expected* output**, frozen once at record time (`TransactionHistoryRecorder`/`SwapDoneScreen`) and never refreshed with what actually landed — `TransactionHistoryViewModel.updateStatus` and `SwapKitTrackingService.applyResponseToTx` both carry it forward verbatim. The in-progress row labels this same value "expectedPayout" for exactly this reason; the completed row this PR adds has no such label and reads as fact. Built as the issue specifies (the detail sheet already renders the same value) — **worth its own follow-up issue**, not addressed here.
- **Type pill wording**: the Figma frame reads "Swapped"/"Limit"; the code renders `"swap".localized` (currently "Swapped" in en, presumably matching, but not verified across all wording) via `TransactionHistoryTypePill.swift`. Pre-existing, unrelated to this issue's change list — left alone for a reviewer to decide.
- **Deferred (cross-model review, MINOR)**: `showsSwapLegs`/`showsViaBadge` route off `isExpanded` (the coarse `transaction.status`), while a limit order's top-right status line routes off the authoritative `isLimitTerminal` (`LimitOrderStatusDisplay.effectiveUiStatus`). These two signals were already independent before this PR (`body`'s `isExpanded`-gated layout choice vs `topRow`'s `isLimitTerminal`-gated status line) and can diverge in a narrow, pre-existing window — on the initiating device, if `THORChainLimitTrackingService.write()`'s row mirror (`storage.updateSwapTrackingStatus`) fails after the authoritative `LimitOrder` write already succeeded (the two are atomic on a co-signer, which holds no local order). This PR reuses `isExpanded` specifically because that's the same signal that already decides `collapsedContent` vs `expandedContent` — routing the badge/legs off `isLimitTerminal` instead would introduce a *new* mismatch within the collapsed layout itself. Fixing the underlying divergence means changing what drives the in-progress limit card, which is out of scope for this issue and has no Figma coverage here. Worth its own issue if the mirror-failure window is judged worth designing for.

## Out of scope (explicitly, per the issue)

Detail sheet, `expandedContent` (in-progress diagram), send/receive/approve/trust-line rows, tab filtering.

## Testing

- Unit tests: `VultisigAppTests/Features/TransactionHistory/TransactionHistoryCardViewTests.swift` — 10 new camelCase tests covering `showsSwapLegs`, `showsViaBadge`, and `stripTicker` (including the bare-number and ambiguous-numeric-ticker regressions caught by cross-model review). File pre-existed so no `make generate` was needed for registration; ran anyway as a safety check.
- Cross-model review (`codex-review-loop`): round 1 found 1 MAJOR (ticker dropped on the co-signer path — fixed), round 2 found 1 MINOR (over-strip on an ambiguous numeric ticker — fixed), whole-branch pass found 1 MINOR (deferred, see above).
- Full gate: `make test DESTINATION='platform=iOS Simulator,id=D63D0F2D-8734-43F5-A630-53982AADC558'` — `** TEST SUCCEEDED **`, `Executed 7030 tests, with 11 tests skipped and 0 failures (0 unexpected)` (17 in `TransactionHistoryCardViewTests`, up from 7 — the new tests actually ran). `df -h /System/Volumes/Data`: 34G free before, 17G free after — never approached zero.
- **Visual acceptance against the Figma frame in the simulator was NOT performed** — this PR is verified by unit tests and the Figma `get_design_context` token/geometry data only. A reviewer should eyeball a real completed swap and limit row before merge.

Closes #5325

## Summary by CodeRabbit

- **Enhancements**
  - Completed swap and limit transactions now show separate FROM and TO amounts with a clear “FROM → TO” indicator when destination details are available.
  - Provider badges are hidden for these completed layouts, while remaining visible for other applicable transaction history entries.
- **Bug Fixes**
  - Ticker labels now remove only a space-delimited suffix, preserving ambiguous or unaffected labels.
  - Transaction history display is more consistent across completed, incomplete, expanded, and provider-related states.

### DIFF
--- a/VultisigApp/VultisigApp/Features/TransactionHistory/Views/TransactionHistoryCardView.swift
+++ b/VultisigApp/VultisigApp/Features/TransactionHistory/Views/TransactionHistoryCardView.swift
@@ -482,7 +482,11 @@ struct TransactionHistoryCardView: View {
 
     private var collapsedContent: some View {
         HStack(spacing: 12) {
-            coinIcon
+            if showsSwapLegs {
+                swapPairIcon
+            } else {
+                coinIcon
+            }
 
             if transaction.type == .trustLineActivation {
                 trustLineColumn
@@ -530,6 +534,44 @@ struct TransactionHistoryCardView: View {
             ticker: transaction.coinTicker,
             tokenChainLogo: transaction.coinChainLogo
         )
+    }
+
+    /// A completed swap/limit row's icon: the FROM token's left half against
+    /// the TO token's right half, meeting as one 24pt circle split down the
+    /// middle — so the row names both assets before the eye reaches the
+    /// amounts. Send/receive rows keep the single `coinIcon` above.
+    private var swapPairIcon: some View {
+        HStack(spacing: 0) {
+            iconHalf(
+                logo: transaction.coinLogo,
+                ticker: transaction.coinTicker,
+                keeping: .leading
+            )
+            iconHalf(
+                logo: transaction.toCoinLogo ?? "",
+                ticker: transaction.toCoinTicker ?? "",
+                keeping: .trailing
+            )
+        }
+        .frame(width: 24, height: 24)
+    }
+
+    /// One half of `swapPairIcon`. The logo renders at its FULL 24pt inside a
+    /// 12pt-wide clip, so it keeps its own circular mask and true proportions
+    /// rather than being squashed to half width; `keeping` selects which side
+    /// survives the clip.
+    ///
+    /// No chain badge here: at 12pt it would be sliced down the middle, and
+    /// the `FROM → TO` pill already names both assets.
+    private func iconHalf(logo: String, ticker: String, keeping: Alignment) -> some View {
+        AsyncImageView(
+            logo: logo,
+            size: CGSize(width: 24, height: 24),
+            ticker: ticker,
+            tokenChainLogo: nil
+        )
+        .frame(width: 12, height: 24, alignment: keeping)
+        .clipped()
     }
 
     private var amountColumn: some View {
--- a/VultisigApp/VultisigApp/Features/TransactionHistory/Views/TransactionHistoryCardView.swift
+++ b/VultisigApp/VultisigApp/Features/TransactionHistory/Views/TransactionHistoryCardView.swift
@@ -25,6 +25,27 @@ struct TransactionHistoryCardView: View {
         TransactionHistoryFailureReasonPresentation.displayText(for: transaction.errorMessage)
     }
 
+    /// Whether the collapsed row shows the completed-swap layout (two amount
+    /// legs + a `FROM → TO` pill) instead of today's single fiat/crypto
+    /// amount. See the static twin below for the routing rule.
+    private var showsSwapLegs: Bool {
+        Self.showsSwapLegs(
+            type: transaction.type,
+            toAmountCrypto: transaction.toAmountCrypto,
+            toCoinTicker: transaction.toCoinTicker
+        )
+    }
+
+    /// Whether the "via {provider}" badge belongs on this row. See the
+    /// static twin below for the routing rule.
+    private var showsViaBadge: Bool {
+        Self.showsViaBadge(
+            hasProvider: transaction.swapProvider != nil,
+            isExpanded: isExpanded,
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5334, opened 2026-09-07
fix: stabilize vault navigation and scroll state

gastonm5:
## Description

Switching vaults could retain Home's previous scroll position and collapsed header, hiding the new vault's large balance. Give each vault its own Wallet/DeFi scroll viewport, publish its initial geometry, and clear the previous vault's open search and query. If the new vault has no DeFi tab, select the remaining Wallet tab so the header follows the displayed content. Same-vault refreshes preserve scrolling, and a still-available DeFi selection stays selected.

Selecting a vault with a navigation-restart request also changed two observed signals, resetting Home twice. Consume only the restart request, avoid unchanged root-state writes, and leave an empty navigation path untouched while always clearing route history.

The chain-picker offset reducer used one shared clock across every scroll view and reduction pass: identical inputs could produce `-80` followed by `0`. Use deterministic additive reduction with a zero identity, and skip unchanged offset-binding writes.

Related to #4470. These correct independently measured defects. The original macOS layout hang remains unlocalized, so the issue should stay open.

## Which feature is affected?

- Home Wallet/DeFi tabs, vault selection, and navigation back to Home.
- [x] Swap — chain-picker scrolling only; transaction construction and signing are unchanged. No swap transaction was submitted for these UI corrections.

## Parity

- Android: n/a — SwiftUI navigation, scroll-view lifecycle, and preference reduction.
- Windows + extension: n/a — SwiftUI navigation, scroll-view lifecycle, and preference reduction.
- SDK: n/a — app UI only; no shared transaction, signing, or fee behavior changed.

## Validation

Verified at `bdd4d6fd78deff2db45e762803b92ff5ad52a699`:

- Full iOS suite: **7,036 tests; 7,024 passed, 1 expected failure, 11 skipped, 0 unexpected failures**. iPhone 17 simulator, iOS 26.5, English/en_US. Fresh xcresult reports Passed.
- 16 new regression tests: seven navigation/reducer tests and nine hosted runtime tests using actual Wallet/DeFi/TabBar views with isolated in-memory fixtures.
- Hosted tests cover switching while scrolled, short/long content from the top, same-vault refresh preservation, initial header geometry, clearing Wallet/DeFi search, removing the selected DeFi tab, and every collapse publication across tab/vault transitions.
- Before/after: original scroll code fails the Wallet, DeFi, and initial-geometry tests; the added search/tab cases reproduced both review findings before their corrections. All nine current runtime tests pass.
- Native macOS scroll-component probe: switching retains a 220-point offset before the correction and returns to 0 afterward. Earlier native navigation/picker probes confirm two Home resets become one and identical `-80` reductions remain `-80`.
- macOS build and static analysis passed. SwiftLint matches the 27-warning baseline, with no new warnings or warnings in changed files.
- Per-commit and whole-branch read-only Claude reviews completed; findings verified and triaged, with search/tab corrections covered by runtime tests.

These isolated runtime checks do not reproduce the original app hang.

**Remaining manual check:** physical 120 Hz iPhone picker scrolling, haptic feel, and performance before release, including per-offset work-item allocation/cancellation. The existing 150 ms debounce covers scroll settlement, not all per-offset processing.

## Checklist

## Agent Metadata

- **Agent**: Codex; read-only cross-model review by Claude
- **Issue**: Related to #4470; remains open
- **Confidence**: High for the measured navigation, picker, and vault-switch defects; the original hang's cause remains unknown
- **Needs human review for**: 120 Hz iPhone picker scrolling, haptics, and performance before release

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation reset behavior when returning to the home screen.
  - Fixed tab selection so it falls back to the first available tab when the current tab is no longer present.
  - Improved scroll position and header behavior when switching between vaults.
  - Search state now clears appropriately when changing vaults.
  - Improved initial scroll and header positioning for a smoother experience.

### DIFF
--- a/VultisigApp/VultisigApp/App/ContentView.swift
+++ b/VultisigApp/VultisigApp/App/ContentView.swift
@@ -10,7 +10,7 @@ import SwiftData
 
 private let logger = Log.app.other
 
-enum RootRoute {
+enum RootRoute: Equatable {
     case home(showingVaultSelector: Bool)
     case createVault
 }
@@ -198,10 +198,8 @@ struct ContentView: View {
                 BackgroundTransactionPoller.shared.resumePendingTransactions()
             }
         }
-        .onChange(of: appViewModel.selectedVault) { _, _ in
-            guard appViewModel.restartNavigation else { return }
-            navigateToHome()
-        }
+        // A selection and its restart request change together. Observe the
+        // request once; observing the selection too resets the stack twice.
         .onChange(of: appViewModel.restartNavigation) { _, newValue in
             guard newValue else { return }
             navigateToHome()
@@ -334,9 +332,14 @@ struct ContentView: View {
     }
 
     func navigateToHome() {
-        appViewModel.showSplashView = false
-        rootRoute = .home(showingVaultSelector: appViewModel.showingVaultSelector)
-        navigationRouter.navPath = NavigationPath()
+        if appViewModel.showSplashView {
+            appViewModel.showSplashView = false
+        }
+        let homeRoute = RootRoute.home(showingVaultSelector: appViewModel.showingVaultSelector)
+        if rootRoute != homeRoute {
+            rootRoute = homeRoute
+        }
+        navigationRouter.navigateToRoot()
     }
 
     @ViewBuilder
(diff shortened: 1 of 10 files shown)


johnnyluo (approved):
LGTM

Issue #5351, opened 2026-09-08
[Fix] MayaChain swap fee rows scale by 1e8 instead of the asset's decimals [Swap]


gastonm5:
## Problem

Two swap fee-row helpers scale MayaChain amounts with THORChain's 1e8 fixed-point instead of the asset's own decimals.

`Features/Swap/ViewModels/SwapCryptoLogic.swift`:
```swift
// :593-598  affiliateFeeFiat
case let .thorchain(q), …, let .mayachain(q):
    let feeDecimal = q.fees.affiliate.toDecimal() / pow(10, 8)

// :649-654  outboundFeeFiat
case let .thorchain(q), …, let .mayachain(q):
    return toCoin.fiat(decimal: q.fees.outbound.toDecimal() / pow(10, 8))
```

Both fold `.mayachain` into the THORChain branch and hard-code `pow(10, 8)`.

The rest of the codebase treats Maya as the explicit exception to that convention — `Core/Services/Actions/Coin+Swaps.swift:12-25`:
```swift
/// THORChain fixed-point exponent. THORChain quotes amounts and LIM in 1e8
/// fixed-point regardless of a chain's natural decimals. …
static let thorchainFixedPointExponent = 8

var thorswapMultiplier: Decimal {
    switch chain {
    case .mayaChain:  return pow(10, decimals)      // <- Maya uses its own decimals
    default:          return pow(10, Coin.thorchainFixedPointExponent)
    }
}
```

If Maya's quote fees are denominated the same way its amounts are, the two helpers are wrong by `10^(decimals − 8)`:

| Asset | decimals | correct divisor | used | effect on the displayed fee |
|---|---|---|---|---|
| CACAO | 10 | 1e10 | 1e8 | **overstated 100×** |
| 4-decimal Maya asset | 4 | 1e4 | 1e8 | **understated 10,000×** |

Display-only — this is the fiat on the Swap Fee row, not a signed value. But a fee row wrong by two to four orders of magnitude on the screen where a user approves a swap is worth fixing.

## Confirm this first

[...]

Issue #5341, opened 2026-09-08
[Fix] Bump commondata pin — carry slippage_bps, SwapKit swap_fee and 1inch sub_provider to the co-signer [Swap / Keysign]


gastonm5:
## Problem

Three commondata fields that exist to keep a co-signer's swap Verify screen honest are unavailable to iOS because the pin predates them, and consequently unwritten:

| Payload | Field | Effect on an iOS joiner today |
|---|---|---|
| `THORChainSwapPayload` | `optional uint32 slippage_bps = 14` | no Price Impact row on native swaps |
| `SwapKitSwapPayload` | `swap_fee = 12`, `swap_fee_chain = 13`, `swap_fee_token_id = 14`, `swap_fee_decimals = 15` | **no swap fee at all** on non-EVM SwapKit routes |
| `OneInchSwapPayload` | `sub_provider = 7` | reads bare `SwapKit` where the initiator reads `SwapKit (NEAR)` |

iOS pins commondata at `63a65c37c50f6a944f6791d36eb905810af82760` (`VultisigApp/project.yml:88`). The three proto commits landed after it:

- `d782a4b` — swapkit `swap_fee` + coin context (commondata#103)
- `4ccedcb` — 1inch `sub_provider` (commondata#104 series)
- `882cf31` — thorchain `slippage_bps` (commondata#104)

Current `main` tip is `08e17eb`. All three changes are **additive and optional**, so older senders round-trip unchanged.

## Solution

Bump the pin to `08e17eb`, then populate and read the three fields.

**Price impact must be carried, not re-derived.** A joiner that fetches its own quote prices a pool that has moved since the initiator quoted it, which would make price impact the only term on the Verify screen where the two devices legitimately disagree. `optional uint32` means an unset field reads as absent, not `0` — hide the row rather than claiming a zero-impact route.

Before bumping, verify as previous pin bumps did (see the comment block at `project.yml:60-88`) that no *other* `.proto` changed between `63a65c37` and `08e17eb`, and record that in the PR.

### Must Do
- [ ] Bump the commondata pin to `08e17eb`; document the intervening proto diff in the PR body.
- [ ] Write + read `THORChainSwapPayload.slippage_bps` (thorchain and mayachain); render a Price Impact row on the joiner.
- [ ] Write + read `SwapKitSwapPayload.swap_fee` and its three coin-context fields; render the swap fee on the joiner for non-EVM SwapKit routes.
[...]

PR #5352, opened 2026-09-08
fix(swap): carry the native swap fee to the co-signer

### LINKED ISSUE #5340: [Fix] Native swap keysign payload omits the swap fee — a co-signer sees a total that understates the swap [Swap / Keysign]
## Problem

A device **joining** a keysign renders its swap Verify screen entirely from the `KeysignPayload` — it holds no quote. On a native THORChain/MayaChain swap, iOS never writes the swap fee onto that payload, so the joiner shows the network fee alone and a Max. Total Fee that understates what the swap costs. The initiator, holding the quote, shows both.

`THORChainSwapPayload` in commondata has carried `string fee = 13` since well before iOS's current pin — **the field is available today and simply never written**:

- `Features/Keysign/Services/KeysignMessage+ProtoMappable.swift:324-337` (thorchain) and `:339-352` (mayachain) write 12 fields and omit `fee`.
- `Features/Keysign/Services/Swap/THORChainSwapPayload.swift:35-47` does not model `fee` at all, so it is neither written nor read.
- The read side (`:226-254`) reconstructs the payload without it.

Windows measured the user-visible effect on a BTC → TRX THORChain swap, both devices on the same build:

| Row | Initiator | Joiner |
|---|---|---|
| Network Fee | 0.00001035 BTC / $0.80 | 0.00001035 BTC / $0.80 |
| Swap Fee | Protocol Fee $0.47 (+$0.01 product fee) | **$0.00** |
| Max. Total Fee | **$1.28** | **$0.80** |

A co-signer is approving a swap it has been told costs $0.48 less than it does, on the screen whose only purpose is for both parties to confirm the same swap.

## Expected Behavior

An iOS-initiated native THORChain/Maya swap carries its fee on the keysign payload, and an iOS joiner renders the swap fee and a Max. Total Fee that includes it — matching the initiator.

## Solution

Add `fee` to `THORChainSwapPayload`, populate it from the quote when building the payload, write it in both the thorchain and mayachain proto branches, read it back on the joiner, and render it on Verify.

**No dependency bump is required** — field 13 exists at the currently pinned commondata revision (`VultisigApp/project.yml:88`, `63a65c37c50f6a944f6791d36eb905810af82760`). Do not bump the pin in this issue; `slippage_bps` and the SwapKit/1inch fields are tracked separately.

iOS already resolves which coin the fee is denominated in correctly and by construction (`SwapCryptoLogic.swapFeeCoin`, `Features/Swap/ViewModels/SwapCryptoLogic.swift:862`) — the portfolio-lookup half of [windows#4816](https://github.com/vultisig/vultisig-windows/pull/4816) does **not** apply to iOS. Reuse that resolver so the initiator's and joiner's fiat agree by construction.

### Must Do
- [ ] Add `fee` to `THORChainSwapPayload` (model + proto write + proto read, thorchain **and** mayachain branches).
- [ ] Populate from the quote at payload-build time.
- [ ] Render the swap fee and a fee-inclusive Max. Total Fee on the joiner's Verify.
- [ ] Absent/empty `fee` (legacy sender) hides the row — never renders `$0.00`.
- [ ] Tests: round-trip through the proto mapper; joiner renders the initiator's figure; legacy payload with no `fee` hides the row.

### Must NOT Do
- Don't bump the commondata pin (separate issue).
- Don't touch `OneInchSwapPayload` / `SwapKitSwapPayload` (separate issue).
- Don't change how the initiator computes or displays the fee.

## Acceptance Criteria
- [ ] A native THORChain swap payload built on iOS carries a non-empty `fee`.
- [ ] Joiner Verify shows the same Swap Fee and Max. Total Fee as the initiator.
- [ ] A payload without `fee` still parses and hides the row.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [windows#4362](https://github.com/vultisig/vultisig-windows/issues/4362) (open parent) · reference impl [sdk#2315](https://github.com/vultisig/vultisig-sdk/pull/2315)

gastonm5:
## Problem

A device **joining** a keysign renders its swap confirm screen entirely from the serialized `KeysignPayload` — it holds no quote. Two gaps meant it could not state what a swap costs:

1. **The swap-fee row existed only for 1inch-shaped payloads.** On `main`, `JoinKeysignSwapFeeViewModel.resolveSwapFee` opens with `guard case let .generic(payload) = swapPayload else { return nil }`, so a native THORChain/MayaChain swap resolved no fee at all. iOS also wrote 12 of `THORChainSwapPayload`'s 13 fields, omitting `fee`, so there was nothing to resolve even in principle.
2. **There was no Total Fee row at all** on that screen — provider, an optional swap fee, and the network fee, nothing more.

So on a native swap the joiner showed the network fee alone while the initiator itemized a protocol fee and a product fee on top of it. Measured on Windows, BTC → TRX, both devices on the same build:

| Row | Initiator | Joiner |
|---|---|---|
| Network Fee | 0.00001035 BTC / $0.80 | 0.00001035 BTC / $0.80 |
| Swap Fee | Protocol Fee $0.47 (+$0.01 product fee) | **absent** |
| Max. Total Fee | **$1.28** | **absent** |

A co-signer was approving a swap it had been told cost $0.48 less than it does, on the screen whose only purpose is for both parties to confirm the same swap.

**No commondata pin bump is needed or included.** `string fee = 13` already exists at the pinned revision (`VultisigApp/project.yml:88` → `63a65c37c…`) and was simply never written.

## Change

Two additions, not a correction of an existing row:

- **Native THORChain/MayaChain routes gain a swap-fee row the co-signer never had.** `THORChainSwapPayload` gains `fee`, written and read in both proto branches; the builder populates it from the quote; the resolver gains a native branch.
- **The Total Fee row is new** on the confirm screen, for every payload shape that can produce one — not native-only. Hidden unless the network leg prices and a non-zero swap leg prices, so an unpriced fee is never absorbed as free.

The fee is resolved against the payload's own `toCoin`, scaled by the same `thorswapMultiplier` the rest of the native path uses (1e8 for THORChain, the coin's own decimals for MayaChain) — the denomination `vultisig-sdk` writes and `vultisig-windows` reads. `totalFee` already ships in all 8 locales: **no new strings, no locale files touched.**

`JoinKeysignGasViewModel.getCalculatedNetworkFee` was restructured around a private `resolveNetworkFee` so the fee's fiat is available as a `Decimal` to sum. Behaviour-preserving across every branch (dApp-supplied Cosmos fee, EVM incl. the `EVMSwapFee` reconciliation, Solana ComputeBudget, UTXO, Cardano, default) — the dropped `gasAmountToDisplay` ternary always selected the same value as `feeToUse`.

## Found on device: a stated zero is not an absent one

Testing a real two-device **RUNE → RUJI** swap, the co-signer showed *neither* row. RUJI is same-chain THORChain, so `fees.outbound` is 0 and the affiliate cut rounds to 0 — and the initiator still renders both rows at `$0.00`, deliberately (`showAffiliateFeeRow` is unconditional for a market route; `showProtocolFeeRow` tests the *formatted* string, and `"$0.00"` is non-empty).

The first iteration collapsed three states into two. `fee` has implicit presence, so unset and `"0"` are different bytes and always were distinguishable — Codex verified the encoding against the generated source: unset omits field 13, `"0"` serializes as `6a 01 30`. Four sites threw the distinction away, and all four are fixed: the builder returns `"0"` rather than `nil` when nothing is charged (reserving `nil` for a component too malformed to sum, each component checked so a negative can't offset a positive into a false zero); the writer no longer strips `"0"`; the resolver accepts `>= 0`; and `getSwapTotalFee` treats a zero leg as worth zero at any price, so it no longer demands a rate for the destination coin — which alone would still have suppressed the total on exactly this swap, since a co-signer that doesn't hold RUJI has no rate for it.

## Decision: `affiliate + outbound`, not `fees.total`

Deliberate, and a **known divergence from `vultisig-sdk` / `vultisig-windows`**, which write `quote.fees.total`.

iOS's initiator itemizes `affiliateFeeFiat` + `outboundFeeFiat` and sums exactly those into `totalFeeString`, excluding the liquidity component of `fees.total` on purpose — the quoted output already reflects it, so folding it in double-counts. Windows makes the opposite call (`protocol = total - affiliate`, so its rows sum to `fees.total`). The issue forbids changing the initiator and requires the joiner to match it, so iOS carries what the iOS initiator shows. Consequence: a Windows joiner on an iOS-initiated swap sees the iOS initiator's figure — the guarantee that matters — but not what a Windows initiator would have shown.

## Correction: the fee coin is `toCoin`, not `swapFeeCoin`

The issue text says to resolve the fee coin with `SwapCryptoLogic.swapFeeCoin`. **That is wrong for native routes and is not used.** `swapFeeCoin` disambiguates via `quote.swapFeeTokenContract`, which returns `nil` for every `.thorchain*`/`.mayachain` case (`SwapQuote.swift:226-233`), so it falls through to `feeCoin` — the *source-chain gas coin* (BTC on a BTC → TRX swap) — while a native route charges in the *destination* asset. Using it would have produced exactly the disagreement it was meant to prevent. Native fees need no coin context on the wire: the fee coin **is** `toCoin`, which already travels on the payload.

## Signing safety — display only

**`fee` cannot enter any pre-signing hash. Verified twice, independently.** A repo-wide search by the cross-model reviewer found no signer reading `swapPayload.fee`; separately I grepped `Blockchain/` and `Features/Keysign/` and the only consumers are the two proto writers and the display resolver (the `TransactionHistoryRecording` `payload.fee.crypto`/`.fiat` hits a naive grep surfaces are a *different type's* pair).

Native pre-sign inputs are built from `KeysignPayload` (`coin`, `toAddress`, `toAmount`, `memo`, `chainSpecific`) via `THORChainSwaps.getPreSignedInputData`; the swap payload supplies only `fromCoin.chain` for routing and `fromCoin`/`fromAmount` for the `MsgDeposit` asset. `KeysignMessage.payloadID` is a relay content hash the joiner downloads by id, not a signing input.

## Known adjacent instance, deliberately not widened into

The **1inch/generic path has the same zero-collapse** (`KeysignMessage+ProtoMappable.swift:381` writes only when `swapFee != "0"`), and it is pre-existing on `main` — it predates this PR. An EVM swap by a user on the Ultimate tier shows `$0.00` on the initiator and nothing on the co-signer.

It is **not** fixable by relaxing that guard. `EVMQuote.Transaction.init(from:)` decodes a missing `swap_fee` as `?? "0"`, so the initiator cannot tell a stated zero from an unmentioned one before it writes; relaxing the guard would make the app assert a zero the provider never quoted, which is worse than showing nothing. The fix belongs at the quote model (`swapFee` as `String?`) and ripples through `evmSwapFeeBigInt`, `swapFeeString`, `affiliateFeeFiat` and the LiFi/Kyber branches. Filed separately.

Two further wrinkles for that work: **LiFi Solana** hard-codes `swapFee: "0"` and charges via `integratorFee` as a fraction of output, so a stated zero there would be a lie — the amount is not expressible in `swap_fee`; and **Jupiter / SwapKit-generic** never populate the fee coin context at all in `SwapPayloadBuilder`, so their fees reach no co-signer regardless.

## Verification

Full gate on a dedicated `en_US` simulator (since deleted), against the exact pushed commit:

```
HEAD:      c018d66940ca48bc1eb57392629367b4e95151b5
DF_BEFORE: /dev/disk3s5  460Gi  406Gi   25Gi   95%
** TEST SUCCEEDED **
Executed 7102 tests, with 11 tests skipped and 0 failures (0 unexpected) in 130.533 seconds
GATE_EXIT=0
DF_AFTER:  /dev/disk3s5  460Gi  406Gi   24Gi   95%
```

0 failed cases; all 24 new cases named individually as `passed`; free space never approached zero, so the marker is not a stale-run artifact. SwiftLint: 0 violations across the changed files.

Cross-model review (`codex exec --sandbox read-only`) at every step: no P0/P1 and no Critical/High/Medium at any stage. The original production commit drew zero findings. Later passes raised test-quality issues — a circular total-fee expectation, a parity test that bypassed the wire, an `XCTAssertNotEqual` satisfiable by an always-`nil` implementation, process-wide rate state, incomplete negative validation, and a serialization test that never serialized — all fixed.

## Tests

`VultisigAppTests/Keysign/NativeSwapFeeProtoMappingTests.swift`, 24 cases. Two carry the load:

- `testCoSignerSwapFeeFiatMatchesTheInitiatorItemization` builds the payload from a live quote, **serializes and parses it back**, and asserts the co-signer's fiat equals `affiliateFeeFiat + outboundFeeFiat` from that quote.
- `testKnownZeroFeeSurvivesSerialization` and `testAbsentAndKnownZeroAreDistinguishableEndToEnd` assert the *distinction* through real bytes — field 13 present as `6a 01 30` for a stated zero, reparsing to unset when absent — and run through `getSwapFee`, the accessor the confirm screen binds to. A regression that rendered everything, or nothing, fails both.

Also covered: round-trip on both branches; pre-field-13 wire bytes parse and relay without gaining a fee; destination-coin denomination on both networks (1e8 THORChain, CACAO's own 1e10); negative and malformed components rejected; and the Total Fee row present for a stated zero on an unpriced destination coin, absent when either leg is genuinely unknown.

### The review pattern worth carrying forward

Three separate findings across this branch turned out to be one defect: **an assertion whose name claimed a boundary its body never crossed.** A parity test named for the wire that stayed in memory; a total named for correctness that recomputed its expectation through the function under test; a serialization test named for bytes that read an in-memory property. Each was individually plausible and collectively a blind spot.

The name is where the claim lives, so the name is what should be checked against the body. That is reviewable in a way "write better tests" is not.

## Review flags

- ⚠️ **Touches a keysign payload on a signing path — needs human review and on-device acceptance.** **On-device acceptance was NOT performed by me.** The stated-zero bug was found by Gaston on a real two-device RUNE → RUJI swap; re-running that is the acceptance test. A green gate only proves unit-level semantics.
- **The co-signer shows one row where the initiator shows two** (Swap Fee = affiliate, Protocol Fee = outbound). That mismatch survives this PR for non-zero routes and may need a second commondata field. Note it is *invisible* on the RUNE → RUJI retest, since both components are zero there — a non-zero native swap is needed to see it.
- Pre-existing, **not** fixed here: `affiliateFeeFiat`/`outboundFeeFiat` hard-code `pow(10, 8)` where `thorswapMultiplier` uses the coin's own decimals, so the *initiator* misprices a MayaChain-destination fee. Filed as #5351. The wire denomination here is the cross-platform-correct one.
- Out of scope by design: the commondata pin, and `OneInchSwapPayload` / `SwapKitSwapPayload` (#5341).

Closes #5340

## Summary by CodeRabbit

- **New Features**
  - Swap confirmations now show combined network and swap fees in fiat when pricing is available.
  - Native THORChain and MayaChain swap fees are included in fee calculations.
  - Network fees across supported chains use consistent crypto and fiat values.
  - Affiliate and outbound charges are included in applicable swap fee calculations.
  - Explicit zero-fee routes display as $0.00, while unavailable fees remain hidden.

- **Bug Fixes**
  - Fee handling now distinguishes valid zero fees from missing, malformed, or negative data.
  - Legacy swap payloads preserve absent fee information.

### DIFF
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
@@ -220,10 +220,11 @@ extension ERC20ApprovePayload {
     }
 }
 
-/// Receivers read an absent or empty `fee` as "legacy sender, render no row",
-/// so a nil or a zero must leave the field unset rather than write `"0"`.
+/// `fee` has implicit presence, so an unset field and a `"0"` are distinguishable
+/// on the wire and mean different things: unknown versus a route that charges
+/// nothing. Only a nil leaves the field unset; `"0"` is written like any value.
 private func writeNativeSwapFee(_ fee: String?, to proto: inout VSTHORChainSwapPayload) {
-    guard let fee = fee?.nilIfEmpty, fee != "0" else { return }
+    guard let fee = fee?.nilIfEmpty else { return }
     proto.fee = fee
 }
 
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
@@ -220,10 +220,8 @@ extension ERC20ApprovePayload {
     }
 }
 
-/// Explicit presence for the native swap fee: receivers read an absent or empty
-/// `fee` as "legacy sender → render no fee row", so a nil or a zero must leave
-/// the field untouched rather than write `"0"`. That is also what stops a
-/// relayed legacy payload from gaining a fee its sender never stated.
+/// Receivers read an absent or empty `fee` as "legacy sender, render no row",
+/// so a nil or a zero must leave the field unset rather than write `"0"`.
 private func writeNativeSwapFee(_ fee: String?, to proto: inout VSTHORChainSwapPayload) {
     guard let fee = fee?.nilIfEmpty, fee != "0" else { return }
     proto.fee = fee
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/KeysignMessage+ProtoMappable.swift
@@ -220,6 +220,15 @@ extension ERC20ApprovePayload {
     }
 }
 
+/// Explicit presence for the native swap fee: receivers read an absent or empty
+/// `fee` as "legacy sender → render no fee row", so a nil or a zero must leave
+/// the field untouched rather than write `"0"`. That is also what stops a
+/// relayed legacy payload from gaining a fee its sender never stated.
+private func writeNativeSwapFee(_ fee: String?, to proto: inout VSTHORChainSwapPayload) {
+    guard let fee = fee?.nilIfEmpty, fee != "0" else { return }
+    proto.fee = fee
+}
+
 extension SwapPayload {
     init(proto: VSKeysignPayload.OneOf_SwapPayload) throws {
         switch proto {
@@ -236,7 +245,8 @@ extension SwapPayload {
                 streamingInterval: value.streamingInterval,
                 streamingQuantity: value.streamingQuantity,
                 expirationTime: value.expirationTime,
-                isAffiliate: value.isAffiliate
+                isAffiliate: value.isAffiliate,
+                fee: value.fee.nilIfEmpty
             ))
         case .mayachainSwapPayload(let value):
             self = .mayachain(THORChainSwapPayload(
@@ -251,7 +261,8 @@ extension SwapPayload {
                 streamingInterval: value.streamingInterval,
                 streamingQuantity: value.streamingQuantity,
                 expirationTime: value.expirationTime,
-                isAffiliate: value.isAffiliate
+                isAffiliate: value.isAffiliate,
+                fee: value.fee.nilIfEmpty
             ))
         case .oneinchSwapPayload(let value):
             // `has*` guards distinguish "legacy sender, context unknown"
@@ -335,6 +346,7 @@ extension SwapPayload {
                 $0.streamingQuantity = payload.streamingQuantity
                 $0.expirationTime = payload.expirationTime
                 $0.isAffiliate = payload.isAffiliate
+                writeNativeSwapFee(payload.fee, to: &$0)
             })
         case .mayachain(let payload):
             return .mayachainSwapPayload(.with {
@@ -350,6 +362,7 @@ extension SwapPayload {
                 $0.streamingQuantity = payload.streamingQuantity
                 $0.expirationTime = payload.expirationTime
                 $0.isAffiliate = payload.isAffiliate
+                writeNativeSwapFee(payload.fee, to: &$0)
             })
         case .generic(let payload):
             return .oneinchSwapPayload(.with {
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/Swap/THORChainSwapPayload.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/Swap/THORChainSwapPayload.swift
@@ -45,13 +45,10 @@ struct THORChainSwapPayload: Codable, Hashable {
     let streamingQuantity: String
     let expirationTime: UInt64
     let isAffiliate: Bool
-    /// Provider-side swap fee, so a co-signer holding only this payload can show
-    /// what the swap costs instead of the network fee alone. A raw integer in the
-    /// destination coin's native-swap fixed point (`toCoin.thorswapMultiplier`) —
-    /// native routes charge in the output asset, so `toCoin` is the fee coin and
-    /// no separate coin context has to travel. `nil` from a sender that quotes no
-    /// fee (limit orders, synthesized deposits) or pre-dates the field; consumers
-    /// render no row rather than a definite zero. Display only: no signer reads it.
+    /// Provider-side swap fee, for a co-signer that holds no quote. A raw integer
+    /// in the destination coin's native-swap fixed point (`toCoin.thorswapMultiplier`):
+    /// native routes charge in the output asset, so `toCoin` is the fee coin and no
+    /// coin context travels. `nil` means unknown, and renders no row. Display only.
     var fee: String? = nil
 
     var toAddress: String {
--- a/VultisigApp/VultisigApp/Features/Keysign/Services/Swap/THORChainSwapPayload.swift
+++ b/VultisigApp/VultisigApp/Features/Keysign/Services/Swap/THORChainSwapPayload.swift
@@ -48,7 +48,9 @@ struct THORChainSwapPayload: Codable, Hashable {
     /// Provider-side swap fee, for a co-signer that holds no quote. A raw integer
     /// in the destination coin's native-swap fixed point (`toCoin.thorswapMultiplier`):
     /// native routes charge in the output asset, so `toCoin` is the fee coin and no
-    /// coin context travels. `nil` means unknown, and renders no row. Display only.
+    /// coin context travels. `"0"` is a route that charges nothing and renders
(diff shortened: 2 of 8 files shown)


johnnyluo (approved):
LGTM

PR #5353, opened 2026-09-08
fix(send): reject non-plain-decimal amounts at every external boundary

### LINKED ISSUE #5339: [Fix] Scientific-notation amounts expand into the signed value — parseInput accepts 1e5 [Send / Deeplink]
## Problem

`String.parseInput()` (`VultisigApp/Core/Extensions/StringExtension.swift:114`) parses amounts with `NumberFormatter(numberStyle: .decimal)`, which **accepts scientific notation**. Verified against the exact iOS implementation on 2026-09-08:

```
"1e5"   -> 100000        "1E5"  -> 100000
"2.5e3" -> 2500          "1e-3" -> 0.001
"0x1F"  -> nil                              (hex is correctly rejected)
```

`parseInput` is the app-wide numeric validator: `toDecimal()` (`:170`), `formatToFiat()` (`:181`), `formatToDecimal()` (`:189`) all funnel through it, and `SendCryptoLogic.amountDecimal` expands the send amount from it.

The deeplink path is unguarded end to end. `DeeplinkViewModel.swift:194` takes the `amount` query item verbatim (`removingPercentEncoding` only), publishes it as `sendAmount` (`:94`), and `HomeScreen.swift:521` hands it to the send form. A crafted `vultisig://…?amount=1e5` therefore expands into a signed amount three to five orders of magnitude larger than the string suggests.

`isDecimalInput()` (`:154`) *does* reject `e` by character set, but it is documented as gating **macOS keyboard entry only** — it is not on the deeplink path.

## Expected Behavior

Only plain decimal amounts are accepted anywhere an amount can reach signing. A scientific-notation or otherwise non-plain-decimal amount is rejected before it can be staged, not silently expanded.

## Steps to Reproduce

1. Open `vultisig://send?…&amount=1e5` (or paste `1e5` where `parseInput` is the only validator).
2. Observe the staged/signed amount is `100000`.

## Solution

Add a plain-decimal guard and apply it at the choke points, rather than changing `NumberFormatter` behaviour globally (locale parsing for comma-decimal locales must keep working — see the two-attempt logic at `:117-135`).

Mirrors [sdk#973](https://github.com/vultisig/vultisig-sdk/pull/973) "Reject non-decimal send amounts" and [android#5214](https://github.com/vultisig/vultisig-android/pull/5214). Note the Android issue ([android#5186](https://github.com/vultisig/vultisig-android/issues/5186), under open tracker [sdk#979](https://github.com/vultisig/vultisig-sdk/issues/979)) is scoped to the XRP parser only; **the iOS exposure is broader** and this issue is written from the iOS evidence above.

### Must Do
- [ ] Add a plain-decimal predicate (ASCII digits + a single locale/`.`/`,` decimal separator + optional grouping; no `e`/`E`, no sign exponent, no `0x`).
- [ ] Reject at the deeplink boundary (`DeeplinkViewModel.swift:194` / the `amount` extraction) so a bad value never becomes `sendAmount`.
- [ ] Confirm and, if needed, close the same hole on any other path where an amount string arrives from outside the app (deeplink, QR payload, pasted address+amount URIs).
- [ ] Tests covering `1e5`, `1E5`, `2.5e3`, `1e-3`, `0x1F`, plus plain-decimal and comma-locale values that must still parse.

### Must NOT Do
- Don't change `parseInput`'s locale fallback behaviour for legitimate comma-decimal input.
- Don't refactor the send form or `SendCryptoLogic`.
- Don't touch `isDecimalInput` (macOS entry gating) beyond reusing it if convenient.

## Acceptance Criteria
- [ ] A deeplink carrying `amount=1e5` is refused (or clamped to a rejected/empty state) rather than staging 100000.
- [ ] `pt_BR`-style `"1.234,56"` still parses to 1234.56.
- [ ] New unit tests cover every case listed above.

### Verification note for the implementer
Before writing the fix, confirm what the **Verify screen** renders for such an amount. The 2026-07-09 sweep claimed Verify shows the raw string while the signed value is expanded (a display/signature mismatch); that half was **not** re-verified on 2026-09-08. If Verify does show the expanded value, this is still a bug but a lesser one — say so in the PR rather than repeating the stronger claim.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [android#5186](https://github.com/vultisig/vultisig-android/issues/5186) · [sdk#979](https://github.com/vultisig/vultisig-sdk/issues/979)

gastonm5:
Closes #5339

> **Flagged for human review.** This touches the amount that reaches a signing path. **On-device acceptance was not performed** — the acceptance test is opening a `vultisig://send?...&amount=0.005` deeplink on a device set to a comma-decimal locale (de/es/hr/it/pt) and confirming the form stages `0,005` and not `5`. Nobody has run that.

## The headline defect is not the one the issue was filed about

**A canonical payment deeplink overstates the amount by 1000× in five of the eight shipping locales, and this is pre-existing on `main` — it is not introduced by this branch.**

`String.parseInput()` is locale-aware, and the boundary handed it a machine-defined amount. A deeplink or payment-URI amount is dot-decimal by specification, so under a comma-decimal locale `NumberFormatter` reads the `.` in `0.005` as a *grouping* separator:

| locale | `amount=0.005` parses as |
|---|---|
| `pt_BR`, `de_DE`, `es_ES`, `it_IT`, `hr_HR` | **5** |
| `en_US`, `fr_FR`, `ko_KR`, `zh_Hans_CN` | 0.005 |

Five of the eight locales this app ships. `1.234` becomes `1234` the same way. A reviewer should read this as a latent defect that predates the branch and is fixed here, not as a regression introduced by it.

The scientific-notation case the issue was filed about is the **lesser** defect: `NumberFormatter(numberStyle: .decimal)` also accepts `1e5` → 100000, `2.5e3` → 2500, `1e-3` → 0.001 (`0x1F` was already rejected).

## What Verify actually shows

The issue asked for this to be settled before writing the fix, since an older sweep's claim had not been re-verified.

`SendVerifyScreen` passes `tx.amount` into `SendCryptoVerifySummary`, which hands it to `CoinAmountFiatLabel`, which is a plain `Text(amount)` — no formatting hop anywhere on the path. So Verify renders the raw string `1e5` while the signed value is 100000.

The nuance, which the older claim missed: the fiat sub-line underneath (`SendCryptoVerifyViewModel.amountFiat`) derives from `transaction.amountDecimal`, so it *does* reflect the expanded value. The mismatch is real on the amount row but **not total** — the fiat line betrays it.

## Two unguarded external boundaries, not one

1. **Deeplink** — `DeeplinkLogic.processSendDeeplink` took the `amount` query item verbatim (`removingPercentEncoding` only) and published it as `sendAmount`; `HomeScreen` prefills the send form with it.
2. **Scanned QR / payment URI** — `Utils.parseCryptoURI` reads `?amount=` out of a `bitcoin:`/`ton://` URI via `AddressResult.fromURI`, and the send form's address handler assigns it straight into the amount field. This one was not in the issue. It is live on **iOS as well as macOS**: the type sits outside the `#if os(macOS)` guard, and both the iOS camera scan and QR-from-image paths build through it.

## A correction worth stating plainly

The first pass hardened `StyledFloatingPointField`. That component has **zero call sites** — it is dead code. The live gate on function-call and DeFi amount fields is `AmountBalanceValidator`, whose bare `NumberFormatter` still accepted `1e5`, across ~19 forms (bond, unbond, stake/unstake, Cosmos delegate/undelegate/redelegate, Kamino deposit/withdraw, yield deposit/withdraw, TRON freeze, Solana delegate, secured withdraw, mint, redeem, withdraw-rewards, TON stake). Its balance ceiling is not a backstop for an inflated amount — a large enough balance clears it. Now fixed.

## The fix

- **`String.isValidDecimal(locale:)`** requires the character set `isDecimalInput()` allows before `parseInput` rules on the value. Reusing that test on a signing gate made its ASCII-only digit assumption load-bearing, so it now tests for a **Unicode decimal digit (Nd)** — `NumberFormatter` renders in the locale's numbering system (`ar_EG` formats 1234.5 as `١٢٣٤٫٥`), and an amount the app formatted for the user has to validate again or a Max preset would block sending outright. Nd is deliberately narrower than `isNumber`, which also admits `²`, `½`, `Ⅷ`, `②`.
- **`String.externalAmount(locale:)`** reads a boundary amount with fixed ASCII semantics — BIP-321's grammar, `*digit [ "." *digit ]` — then re-renders it with the device's decimal separator, because the form and `toDecimal()` parse with `Locale.current` and passing the raw string through would only move the misparse one step later.
  - `1.` and `.5` are **accepted**: the grammar admits a bare separator on either side and `parseInput` already reads both (`1.` → 1, `.5` → 0.5), so rejecting them would drop a spec-legal QR amount. `""` and `"."` stay rejected — no digit, no amount.
  - A comma-bearing amount is **rejected rather than guessed at**, on the authority of the same spec text: *"All amounts MUST contain no commas and use a period (.) as the separating character."*
- **`validateAmountFormat()`** joins the composed `validateForm()` pipeline as the backstop, ordered *after* the non-zero rule so an empty field still reports `positiveAmountError` (`1e5` is non-zero once expanded, which is exactly why the non-zero rule waves it through). It reuses the existing `decimalAmountError` key, so **no locale files are touched**.

`parseInput`'s locale-fallback behaviour is unchanged: `pt_BR` `"1.234,56"` still parses to 1234.56.

## Verification

`make test` on a dedicated `en_US` simulator: `** TEST SUCCEEDED **`, `Executed 7105 tests, with 11 tests skipped and 0 failures (0 unexpected)`. Baseline `main` is 7078 and this branch adds 27 test functions counted from the diff — 7078 + 27 = 7105, so the suite is provably running the new tests. Disk was flat at ~38Gi throughout, so the result is not a stale-products artifact.

Reviewed commit-by-commit by `codex exec` (read-only) across four rounds; six findings, all fixed or resolved with a recorded reason.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved amount validation to reject scientific notation, hexadecimal-like values, malformed, negative, and unrepresentable amounts.
  * Added support for locale-specific separators and Unicode decimal digits.
  * Standardized amount handling for deeplinks and scanned payment requests while preserving associated addresses and memos.
  * Added inline validation errors for invalid send amounts.

* **Tests**
  * Expanded coverage for locale formatting, external amounts, deeplinks, scanned URIs, Unicode digits, and send form validation.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Validators/AmountBalanceValidator.swift
+++ b/VultisigApp/VultisigApp/Components/Validators/AmountBalanceValidator.swift
@@ -35,10 +35,8 @@ struct AmountBalanceValidator: FormFieldValidator {
     }
 
     func validate(value: String) throws {
-        // `formatter` is a plain `.decimal` NumberFormatter, which reads scientific
-        // notation: "1e5" would validate as 100000 and ride into the transaction
-        // builder. The balance ceiling below is not a backstop for that — a large
-        // enough balance, or a small exponent, clears it.
+        // `formatter` reads scientific notation, and the balance ceiling below is no
+        // backstop for it: a large enough balance clears the expanded value.
         guard value.isValidDecimal() else {
             throw ValidationError.invalidAmount
         }
--- a/VultisigApp/VultisigApp/Components/Validators/AmountBalanceValidator.swift
+++ b/VultisigApp/VultisigApp/Components/Validators/AmountBalanceValidator.swift
@@ -35,6 +35,14 @@ struct AmountBalanceValidator: FormFieldValidator {
     }
 
     func validate(value: String) throws {
+        // `formatter` is a plain `.decimal` NumberFormatter, which reads scientific
+        // notation: "1e5" would validate as 100000 and ride into the transaction
+        // builder. The balance ceiling below is not a backstop for that — a large
+        // enough balance, or a small exponent, clears it.
+        guard value.isValidDecimal() else {
+            throw ValidationError.invalidAmount
+        }
+
         guard
             let number = Self.formatter.number(from: value),
             let amount = Decimal(string: number.stringValue)
--- a/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
@@ -254,6 +254,41 @@ extension String {
         return number >= 0
     }
 
+    /// Normalizes an amount that arrived from outside the app — a deeplink query
+    /// item, a payment URI's `amount=` — into the representation everything
+    /// downstream reads back, or nil when it is not a plain dot-decimal number.
+    ///
+    /// The shape is checked with fixed ASCII semantics, never the device locale.
+    /// An external amount is dot-decimal by specification, and a locale-aware
+    /// parse of one is actively wrong: under `pt_BR` or `de_DE`, `NumberFormatter`
+    /// reads the `.` in `0.005` as a grouping separator and yields 5 — a
+    /// thousandfold error on a value heading for a signature, larger than the
+    /// scientific-notation case this guard was first written for.
+    ///
+    /// The accepted digits are then re-rendered with the locale's decimal
+    /// separator, because the send form and `toDecimal()` parse with
+    /// `Locale.current`: handing `0.005` to a `pt_BR` device unchanged would just
+    /// move the same misparse one step later.
+    func externalAmount(locale: Locale = .current) -> String? {
+        let candidate = trimmingCharacters(in: .whitespaces)
+        let parts = candidate.split(separator: ".", omittingEmptySubsequences: false)
+        guard !candidate.isEmpty,
+              parts.count <= 2,
+              parts.allSatisfy({ !$0.isEmpty && $0.allSatisfy { $0.isASCII && $0.isNumber } })
+        else {
+            return nil
+        }
+
+        guard let separator = locale.decimalSeparator, separator != "." else {
+            return candidate
+        }
+
+        let localized = candidate.replacingOccurrences(of: ".", with: separator)
+        // The form applies its own gate to whatever is prefilled; a value that
+        // could not pass it must not be staged in the first place.
+        return localized.isValidDecimal(locale: locale) ? localized : nil
+    }
+
     var isValidEmail: Bool {
         let regex = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
         let predicate = NSPredicate(format: "SELF MATCHES %@", regex)
--- a/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
+++ b/VultisigApp/VultisigApp/Core/Extensions/StringExtension.swift
@@ -258,34 +258,41 @@ extension String {
     /// item, a payment URI's `amount=` — into the representation everything
     /// downstream reads back, or nil when it is not a plain dot-decimal number.
     ///
-    /// The shape is checked with fixed ASCII semantics, never the device locale.
-    /// An external amount is dot-decimal by specification, and a locale-aware
-    /// parse of one is actively wrong: under `pt_BR` or `de_DE`, `NumberFormatter`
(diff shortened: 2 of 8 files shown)


johnnyluo (approved):
LGTM

Issue #5355, opened 2026-09-08
[Fix] Swap Verify: refresh races signing — the signed payload can differ from the transaction shown [Swap / Keysign]


gastonm5:
## Problem

On the swap Verify screen, the quote-refresh timer and the signing path both read and write `transaction` with no mutual exclusion. Signing can therefore build its payload from a quote that a concurrent refresh is in the middle of replacing — **and the transaction shown to the user and recorded in history can differ from the payload actually signed.**

All verified against `origin/main` @ `92109a83f`.

**The two racers.**

`Features/Swap/Views/Screens/SwapVerifyScreen.swift:64` fires the refresh asynchronously:
```swift
.swapRefreshTick {
    Task {
        await verifyViewModel.updateTimer(vault: vault)
```
`Features/Swap/ViewModels/SwapVerifyViewModel.swift:162` commits the result:
```swift
transaction = updated
```
while signing independently reads the same property at `:264`:
```swift
return try await interactor.buildSwapKeysignPayload(transaction: transaction, vault: vault)
```

Start signing just before a tick and the payload captures the old quote while the refresh commits the new one.

**It gets worse after the payload is built.** `SwapVerifyScreen.swift:29` re-reads the property for the navigation context:
```swift
private var currentTransaction: SwapTransaction { verifyViewModel.transaction }
```
and uses it again after payload construction (`:379`, `:388`). So the transaction the user sees on the next screen, and the one written to history, can be a *different* transaction from the one whose bytes were signed.

**The existing guard does not cover it.** `.disabled(isLoadingFees)` gates the Verify button, but the FastVault password sheet and the security-warning sheet are already presented by that point and call signing directly from their callbacks — bypassing the disabled state entirely.

## Expected Behavior

A signing operation is atomic with respect to refresh: one transaction snapshot serves the halt check, the payload build, and the navigation/history context. A refresh landing mid-sign either waits, or is discarded, or aborts the signing attempt with an explicit message — but never silently swaps the quote underneath it.

## Solution

Two parts, and the second is cheap and valuable on its own:

1. **Mutual exclusion** between refresh and signing (an actor, a signing-in-progress flag the refresh respects, or cancelling the timer for the duration of a sign).
2. **Snapshot once.** Capture the transaction at the start of the signing flow and pass that same value to the halt check, the payload build, and the navigation context, instead of re-reading `verifyViewModel.transaction` at each step. That alone removes the displayed-vs-signed divergence even before the race itself is closed.

[...]

Issue #5354, opened 2026-09-08
[Fix] A failed quote refresh leaves the stale swap quote signable [Swap]


gastonm5:
## Problem

When a swap quote refresh fails, the previous quote stays on screen **and stays signable**. Its memo/calldata can reach the signing path.

Verified against `origin/main` @ `92109a83f` (not a branch):

**1. The failure doesn't clear the quote.** `Features/Swap/ViewModels/SwapDetailsViewModel.swift:814-821` — the `catch` re-raises nothing and drops nothing:
```swift
} catch {
    if error is CancellationError || (error as? URLError)?.code == .cancelled {
        return
    }
    self.error = error          // <- the only effect
}
```
The previously fetched quote and fees remain in place. This includes structural failures such as `routeUnavailable`, not just transient network errors.

**2. Continue doesn't consider `error`.** `Features/Swap/Views/Screens/SwapDetailsScreen.swift:260`:
```swift
let isDisabled = !isFormValid || detailsViewModel.isLoading || detailsViewModel.isLoadingFees
```
`error` is not a term. Loading ends, the button re-enables.

**3. So the stale quote is captured and signed.** `makeTransaction()` takes whatever quote is current, and `Features/Swap/Interactor/DefaultSwapInteractor.swift:188` reuses its memo/calldata on the sign path.

The refresh that fails need not be the periodic one — a slippage or recipient settings change with an unchanged pair and amount re-fetches through the same path.

Recipient mismatches do fail closed later, but an ordinary refresh failure or a changed slippage setting is unprotected.

## Expected Behavior

A quote that failed to refresh is not signable. Either the stale quote is dropped and Continue blocks, or the user is required to acknowledge that they are proceeding on a quote of known age — but signing must not silently reuse payload the app failed to revalidate.

## Solution

The shape is a product decision, which is why this is filed rather than folded into an existing PR.

Today's behaviour is effectively stale-while-revalidate, and that is defensible for a *transient* network blip on a quote seconds old — dropping the quote on every hiccup would make the form unusable on a poor connection. It is not defensible for a structural failure like `routeUnavailable`, or for a quote that has aged past its own validity window.

A plausible split: distinguish transient from structural failures; keep the stale quote for the former but gate Continue on quote age; drop and block on the latter. Whatever the rule, `error` should become a term in the Continue gate.

### Must Do
- [ ] Decide and document the stale-while-revalidate policy for a failed refresh.
[...]

PR #5356, opened 2026-09-08
fix(swap): keep a manually picked swap route across quote refreshes

### LINKED ISSUE #5345: [Fix] A manually picked swap route reverts to Best on every quote refresh [Swap / Advanced Swap]
## Problem

A manually picked swap route does not stick. `Features/Swap/ViewModels/SwapDetailsViewModel.swift:73` says so outright:

```swift
/// Manual provider override. `nil` means "use Best". Reset on every refresh.
var selectedQuote: SwapQuote?
```

and the comment block at `:63-66` confirms the intent: *"A refresh resets `selectedQuote` so it re-defaults to Best."* Resets happen at `:83` (the `quote` setter), `:687`, `:779` and `:806`.

So a user who opens Advanced Swap → Select route and picks a non-default provider has that choice silently discarded on the next quote refresh — within 60 seconds, often much sooner — and the swap reverts to the auto-selected winner with nothing telling them. They can then sign a route they did not choose.

## Expected Behavior

A manual route pick survives quote refreshes for as long as that provider is still a valid candidate for the current pair and amount. It is cleared only when the user changes something that invalidates it (pair, amount, or the provider dropping out of the candidate set), and when it is cleared the UI says so rather than silently reverting.

## Solution

Key the pick to the **provider identity** for the current pair/amount rather than to the quote instance, so a refresh that returns a fresh `SwapQuote` object for the same provider re-attaches the selection instead of dropping it.

Windows hit the same product bug from a different cause — their pick was stored against a fingerprint that included `expiresAt`, which is rebuilt from `Date.now()` on every fetch, so a 1 ms shift dropped it ([windows#4798](https://github.com/vultisig/vultisig-windows/pull/4798)). The iOS cause is an explicit reset, so the fix is iOS-specific; the behavioural target is the same.

### Must Do
- [ ] Preserve `selectedQuote` across a refresh when the same provider is still among `allQuotes` for the unchanged pair/amount.
- [ ] Re-resolve the selection to the *fresh* quote object for that provider, so signing uses current numbers, never a stale quote.
- [ ] Clear the pick deliberately on pair change, amount change, or the provider leaving the candidate set — and surface that to the user rather than reverting silently.
- [ ] Audit all four reset sites (`:83`, `:687`, `:779`, `:806`) and keep the ones that are genuinely invalidating.
- [ ] Tests: pick survives a refresh; pick re-points at the refreshed quote; pick clears on pair/amount change; pick clears when the provider disappears.

### Must NOT Do
- Don't make the manual pick part of `advancedSettings` — `:46` and `:265` document that exclusion deliberately.
- Don't let a stale quote reach signing in order to preserve a selection.

## Acceptance Criteria
- [ ] Selecting a non-default route and waiting through two refresh cycles keeps the route selected.
- [ ] The signed swap uses the selected provider's current quote.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [windows#4797](https://github.com/vultisig/vultisig-windows/issues/4797) · reference impl [windows#4798](https://github.com/vultisig/vultisig-windows/pull/4798)

gastonm5:
## Problem

A manually picked swap route did not stick. A user who opened Advanced Swap → Select route and chose a non-default provider had that choice silently discarded on the next quote refresh — within 60 seconds, often sooner — and could then sign a route they never chose.

## Root cause

`SwapQuote` is `Hashable` **by its full payload**, so the quote a refresh returns for a given provider is never `==` the one it returned a moment earlier. The pick was stored as `selectedQuote: SwapQuote?`, which therefore could never have survived a fetch. The unconditional reset on every landing — documented as intentional (*"Every refresh re-defaults to Best"*) — was papering over an identity problem, not expressing a policy.

Fixing only the details screen turned out not to be enough. `SwapTransaction` carried the chosen quote but not the fact that it *was* chosen, so the verify screen's own 60s re-quote installed the fresh auto winner over it — the same silent substitution, one screen later and immediately before signing, with the confirmation checkboxes still ticked. That is worse than losing the pick early: the user has already read and confirmed a summary for a specific route.

## Fix

- **`SwapQuote.routeIdentity`** — a payload-free identity, one case per row of the route picker. THORChain network variants stay distinct (separate `SwapProvider`s, separate display names); the SwapKit `subProvider` is deliberately excluded, since SwapKit is one row whichever protocol it routes through and that choice can legitimately flip between fetches.
- **Details screen** — `updateQuotes` re-resolves the pick against the fresh candidate set by identity and re-points it at the quote object out of `result.allQuotes`, so signing uses current numbers and never the quote the user tapped.
- **Verify screen** — `SwapTransaction` carries an optional `selectedRouteIdentity` (its own field, **not** part of `advancedSettings`, which is the form's re-fetch trigger — a route pick must never re-quote). `refreshData` re-resolves it the same way. Auto, with nothing pinned, still follows a changed winner. When the picked route is genuinely gone the refresh falls back to Auto, raises the banner, and clears the three confirmations — they were given for a route that no longer exists.
- **Notices** — every production path that drops a pick goes through one `dropRouteSelection(_:)`, so a selection can never disappear silently. Two messages, since the causes differ (route withdrawn vs. the swap itself changed), in all 8 shipping locales.
- **Equivalent amounts** — quote ownership stamped the raw amount *text* and compared it byte-for-byte, so editing `1` to `1.` or `1.0` counted as a new swap. It now stamps the parsed `Decimal`, parsed once and reused for both the provider request and the stamp so the two cannot disagree.

### Reset-site audit

Three of the four sites were genuinely invalidating and stay: the `quote` setter (a wholesale write of the slot; nothing in the app writes it), `clearQuoteState` (any pair or amount change, which already runs before the fetch), and the secured-mint branch (the candidate set collapses to one synthetic quote). Only the refresh reset was the bug.

### One trap worth knowing about

`SwapTransaction.with(...)` rebuilds the struct field by field. Omitting `selectedRouteIdentity` there would have silently un-pinned the route on the very first verify refresh — the fix would have read as correct while doing nothing. It is carried explicitly, and `testRefreshKeepsThePickedRouteAndRepointsIt` fails if either `with(...)` call drops it.

## Tests

**+16 net** (`SwapProviderSelectionTests` 9 → 22, new `SwapVerifyRouteSelectionTests` ×3). `testRefreshResetsSelectionBackToBest` asserted the old behaviour and is replaced — an intentional behaviour change, not a regression.

Covers: the pick survives a refresh and re-points at the refreshed quote (and that is what reaches `makeTransaction()`); it clears with a notice when the route leaves the candidate set, on pair change, on amount change, and when an external recipient prunes it; the advanced-settings re-fetch preserves it; a re-typed equivalent amount keeps it; `makeTransaction` pins the identity only when the user picked one; the verify refresh keeps and re-points a pick, substitutes Auto with cleared confirmations when the route is gone, and leaves the Auto path free to follow a new winner.

The equivalent-amount fixture is built from `Locale.current.decimalSeparator`, not a hard-coded `"1.0"` — a dot is the *grouping* separator in five of the eight shipping locales, where `"1.0"` parses as ten.

## Verification

Full gate on an `en_US`-pinned simulator at `d603fffd3`:

```
** TEST SUCCEEDED **
Executed 7094 tests, with 11 tests skipped and 0 failures (0 unexpected)
```

7094 = the 7078 baseline + the 16 added tests, and corroborated per case (22 + 3 passing names matching the `func test` counts) so nothing was silently dropped from the generated project. Disk 34 Gi free before and after, so not a stale-products run. SwiftLint clean on every touched file. Locale entry counts equal across all 8 files (1921 → 1923).

**On-device acceptance was not performed** — the two acceptance criteria that need a real device (selecting a non-default route and waiting through two refresh cycles; confirming the signed swap uses the selected provider's current quote) are covered by unit tests only.

## Known, out of scope, tracked separately

Both were surfaced by review of this branch, both are pre-existing, and neither is introduced or widened here.

- **A failed refresh leaves the previously-fetched quote on screen and signable** — #5354. `main`'s `catch` block in `updateQuotes` is byte-identical and also kept `bestQuote`.
- **The verify screen's refresh can race signing** — #5355. The timer commits a refreshed `transaction` while signing reads `transaction` independently, and the navigation context re-reads it after the payload is built, so the recorded transaction can differ from the payload signed. Every element predates this branch (`main`: `transaction = updated` at `:162`, `buildSwapKeysignPayload(transaction: transaction)` at `:264`, `currentTransaction` at `SwapVerifyScreen:29`); the only change here to that screen is a 3-line banner modifier. Note the confirmation reset added here lands *after* the commit, so it does not mitigate this race — it is not broader protection than it looks.
- **A route banner already on screen finishes its ~1.5s dismissal after switching to the Limit tab.** The binding gates when a notice *starts*; the shared `BannerViewModifier` owns its own visibility and ignores the text going nil. Closing it means changing a modifier six other screens use.

## Review notes

Touches a signing path: the quote a pick resolves to is what gets signed. The re-attach only ever takes an object out of the freshly-returned `result.allQuotes`, never a carried-over one. Wants a human read and on-device acceptance.

Closes #5345

## Summary by CodeRabbit

- **New Features**
  - Manual swap-route selections persist through quote refreshes while available.
  - Unavailable routes automatically fall back to the refreshed default, with localized notices in Market swaps.
  - Route selections remain consistent when transactions are rebuilt.
  - Fee loading now uses a dedicated overlay, with refresh countdowns hidden during loading.

- **Localization**
  - Added route-reset notices in English, German, Spanish, Croatian, Italian, Korean, Portuguese, and Simplified Chinese.

- **Tests**
  - Added coverage for route persistence, fallback behavior, refresh concurrency, and loading states.

### DIFF
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -1584,6 +1584,8 @@
 "swapReferralDiscount" = "Empfehlung: %@";
 "swapRouteEta" = "~%lds";
 "swapRouteNotAvailable" = "Tauschroute nicht verfügbar";
+"swapRouteResetToAuto" = "Route auf Auto zurückgesetzt";
+"swapRouteUnavailableResetToAuto" = "Route nicht verfügbar – auf Auto zurückgesetzt";
 "swaps" = "Swaps";
 "swapSecuredAssetBadge" = "Secured";
 "swapSecuredAssetInvalidDestination" = "Ziel des Secured-Asset-Swaps ist keine gültige %1$@-Adresse (%2$@).";
--- a/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
@@ -1584,6 +1584,8 @@
 "swapReferralDiscount" = "Referral: %@";
 "swapRouteEta" = "~%lds";
 "swapRouteNotAvailable" = "Swap route not available";
+"swapRouteResetToAuto" = "Route reset to Auto";
+"swapRouteUnavailableResetToAuto" = "Route unavailable — reset to Auto";
 "swaps" = "Swaps";
 "swapSecuredAssetBadge" = "Secured";
 "swapSecuredAssetInvalidDestination" = "Secured-asset swap destination is not a valid %1$@ address (%2$@).";
--- a/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/es.lproj/Localizable.strings
@@ -1584,6 +1584,8 @@
 "swapReferralDiscount" = "Referido: %@";
 "swapRouteEta" = "~%lds";
 "swapRouteNotAvailable" = "Ruta de intercambio no disponible";
+"swapRouteResetToAuto" = "Ruta restablecida a Auto";
+"swapRouteUnavailableResetToAuto" = "Ruta no disponible: restablecida a Auto";
 "swaps" = "Intercambios";
 "swapSecuredAssetBadge" = "Secured";
 "swapSecuredAssetInvalidDestination" = "El destino del swap de activo asegurado no es una dirección %1$@ válida (%2$@).";
--- a/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/hr.lproj/Localizable.strings
@@ -1584,6 +1584,8 @@
 "swapReferralDiscount" = "Preporuka: %@";
 "swapRouteEta" = "~%lds";
 "swapRouteNotAvailable" = "Ruta zamjene nije dostupna";
+"swapRouteResetToAuto" = "Ruta vraćena na Auto";
+"swapRouteUnavailableResetToAuto" = "Ruta nije dostupna – vraćena na Auto";
 "swaps" = "Zamjene";
 "swapSecuredAssetBadge" = "Secured";
 "swapSecuredAssetInvalidDestination" = "Odredište zamjene osiguranog sredstva nije valjana %1$@ adresa (%2$@).";
--- a/VultisigApp/VultisigApp/Core/Localizables/it.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/it.lproj/Localizable.strings
@@ -1584,6 +1584,8 @@
 "swapReferralDiscount" = "Referral: %@";
 "swapRouteEta" = "~%lds";
 "swapRouteNotAvailable" = "Percorso di scambio non disponibile";
+"swapRouteResetToAuto" = "Percorso reimpostato su Auto";
+"swapRouteUnavailableResetToAuto" = "Percorso non disponibile – reimpostato su Auto";
 "swaps" = "Scambi";
 "swapSecuredAssetBadge" = "Garantito";
 "swapSecuredAssetInvalidDestination" = "La destinazione dello swap dell'asset protetto non è un indirizzo %1$@ valido (%2$@).";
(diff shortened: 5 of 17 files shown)


johnnyluo (approved):
LGTM

Issue #5362, opened 2026-09-08
[Fix] 17 of 18 SwapKitError cases render a domain-correct body under "Unexpected Error" [Swap]


gastonm5:
## Problem

Almost every `SwapKitError` renders a domain-correct message body under a generic **"Unexpected Error"** title. The user reads an accurate explanation of what went wrong, headed by a title that says the app doesn't know.

Mechanism:

- `SwapKitError` does **not** conform to `SwapErrorPresentable`.
- `normalizedSwapKitError` maps exactly **one** case, `.amountBelowProviderMinimum`.
- So **17 of the 18 cases** fall through to the generic presentation — including ones the user is most likely to hit, such as `swapRouteNotFound` and `insufficientBalance`.

`Features/Swap/Models/SwapErrorPresentable.swift` exists precisely to stop a title and a body being resolved through independent ladders, and this is that failure at scale.

## Why it is filed rather than fixed in #5343

[PR for #5343](https://github.com/vultisig/vultisig-ios/issues/5343) adds two more `SwapKitError` cases (`contradictoryResponse`, `responseEchoMismatch`). Giving *those two* a proper title while `swapRouteNotFound` and the other fifteen keep "Unexpected Error" would make the inconsistency worse, not better — a user hitting two adjacent SwapKit failures would see two different presentation qualities for no discernible reason.

The real fix is one of:
- conform the whole `SwapKitError` enum to `SwapErrorPresentable`, or
- add an `Errors` case plus its strings across all 8 shipping locales,

and either is its own change with its own localization footprint. Half-doing it inside a response-validation PR would have been worse than leaving it visibly open.

### Must Do
- [ ] Decide the mechanism: conform the enum, or add a presentation case.
- [ ] Give every `SwapKitError` case a title consistent with its body — not just the ones a recent PR happened to add.
- [ ] Locale strings for whatever new titles that requires, across all **8** shipping locales (`de`, `en`, `es`, `hr`, `it`, `ko`, `pt`, `zh-Hans`) — verify equal entry counts afterwards.
- [ ] Tests pinning that a representative sample of cases resolve title and body through the same ladder.

### Must NOT Do
- Don't special-case individual errors; that is how the current state arose.
- Don't change error *semantics* — this is presentation only.

## Acceptance Criteria
- [ ] Build succeeds; full gate green.
- [ ] No `SwapKitError` case renders a domain-specific body under a generic title.
- [ ] All 8 locale files reconcile to equal entry counts.

---
[...]

Issue #5359, opened 2026-09-08
[Investigate] A co-signer signs a swap payload it never revalidates [Keysign / Swap]


gastonm5:
## Problem

A co-signer rebuilds a swap payload from the wire and signs it **without revalidating any of the initiator's claims**. Validation that exists on the quoting device does not exist on the signing device, so a payload that this build would refuse to construct, it will still sign when a peer sends it.

Verified for SwapKit, which is where it surfaced:

- `Features/Keysign/Services/KeysignMessage+ProtoMappable.swift:305-318` reconstructs `SwapKitSwapPayload` from protobuf with `targetAddress`, `inboundAddress` and `txPayload` intact.
- Nothing validates any of it before `KeysignMessageFactory` produces pre-sign hashes.
- The new cross-field agreement guard added in [#5343](https://github.com/vultisig/vultisig-ios/issues/5343) is invoked only from the quote-fetch path (`SwapKitService.swift:197`).

Net effect: that guard protects swaps **this device quotes**, not swaps **this device signs**. A swap initiated by an older iOS build, by Android, or by the extension bypasses it entirely — and the co-signer has every field needed to run the same check.

## Why this is filed as its own issue rather than fixed in #5343

Three reasons, and the third is the important one:

1. It edits a signing path, which needs human and on-device review rather than a unit-tested patch.
2. A peer-side rejection that is *wrong* breaks a live co-signing ceremony with another client, mid-ceremony, in front of a user. No unit test derisks that.
3. **The peer trusting the initiator's payload is not a SwapKit property.** It is how swap payloads cross the wire generally. A SwapKit-shaped patch would close one instance and make the broader gap *look* handled — which is worse than leaving it visibly open, because the next reader sees a guard on the signing path and assumes the class is covered.

## What to decide

- **Where does validation belong?** Hardening every peer against every initiator is one answer; fixing the initiating platforms (Android and the SDK are both already in the cross-platform epic) is another, and cheaper. Probably both, with peer-side checks limited to what cannot be wrong.
- **What is safe to reject peer-side?** A rejection must be certain. Self-contradiction within the payload is a candidate: it needs no network call and no trust in the peer's provider. Anything requiring a re-quote is not, because the pool has moved.
- **Which payload types share the shape?** SwapKit is verified. `THORChainSwapPayload`, `OneInchSwapPayload` and the generic shapes need the same audit — this issue should not close until each is either confirmed safe or has its own follow-up.

### Must Do
[...]

Issue #5358, opened 2026-09-08
[Investigate] SwapKit deposit routes: can sellAmount corroborate tx[0].amount? [Swap / SwapKit]


gastonm5:
## Problem

[#5343](https://github.com/vultisig/vultisig-ios/issues/5343) added a cross-field **destination** agreement check for SwapKit deposit routes, plus `tx[]` cardinality validation. The matching **amount** check was deliberately deferred, because the evidence available did not support building a fail-closed guard on it.

SwapKit states the deposit size in more than one place, and the question is which fields may be compared against which. The candidate comparison was `(depositAmount ?? sellAmount)` vs `tx[0].amount`. The `sellAmount` half is the problem.

## Why it was deferred

Three independent implementations decline to treat `sellAmount` as the deposit size:

- **The SDK never consults it.** `getSwapKitQuote.ts:getTransferAmount` resolves `depositAmount` → `tx[0].amount` → the caller's own amount.
- **Android signs the user's typed amount**, `request.tokenValue`, not `sellAmount` (`SwapKitQuoteSource.kt`).
- **iOS has no production consumer for it at all** — it is decoded, then read only by request models and unrelated LimitSwap UI.

Against that: exactly **one** recorded fixture in which `sellAmount` happens to track the deposit size.

One fixture against three implementations is not a basis for a check that **fails closed** on a money path. A false reject would block a route that would have settled fine, and the most likely victim is a jetton-source TON route — precisely the case with the least fixture coverage.

## What would settle it

Either of:

- A statement in SwapKit's API documentation defining `sellAmount`'s relationship to the deposit amount, or
- A second recorded fixture — **ideally a jetton-source TON route** — showing whether `sellAmount` tracks the deposit size when the source asset is not the native coin.

With either in hand the comparison is a small additive change to the existing `SwapKitResponseAgreement` guard.

### Must Do
- [ ] Obtain one of the two forms of evidence above.
- [ ] If `sellAmount` is confirmed to track the deposit size: extend the agreement check to the amount, matching the destination leg's fail-closed behaviour.
- [ ] If it is not: record that finding so the question is not reopened, and consider dropping the unused decode.

### Must NOT Do
- Don't ship the amount comparison on the strength of the single existing fixture.
- Don't change the destination or cardinality checks — those shipped in #5343 and are independently justified.

## Acceptance Criteria
- [ ] The relationship between `sellAmount`, `depositAmount` and `tx[0].amount` is documented with evidence.
[...]

PR #5363, opened 2026-09-08
fix(swap): refuse a SwapKit deposit response that disagrees with itself

### LINKED ISSUE #5343: [Investigate] SwapKit deposit route: are targetAddress / depositAddress / tx[] cross-checked before signing? [Swap / SwapKit]
## Problem

SwapKit's `/v3/swap` response states the deposit **destination** in up to three independent fields (`targetAddress`, `depositAddress`, the `tx[]` transfer array) and its **size** in two (`depositAmount`, `tx[0].amount`). Taking the first field that is present, and never comparing it against the rest, signs whichever field happens to win precedence when a response disagrees with itself — a provider bug, an API change, or a tampered payload. On a deposit route that is a lost deposit.

[sdk#2296](https://github.com/vultisig/vultisig-sdk/pull/2296) made `buildTransferTx` refuse a self-contradicting response outright, on the grounds that neither field is authoritative enough to pick a winner from.

**iOS status is unconfirmed.** `Blockchain/Swaps/SwapKit/SwapKitSwapResponse.swift:67` resolves through `resolvedTargetAddress` and handles a destination-tag suffix, but a search of `Blockchain/Swaps/SwapKit/` on 2026-09-08 surfaced **no cross-field agreement check**. Absence was not proven — this issue is to settle it first and fix only if the gap is real.

### Must Do — step 1, investigate (timebox ~half a day)
- [ ] Determine exactly which response fields iOS reads for the SwapKit deposit destination and amount, and in what precedence.
- [ ] Determine whether any existing guard would reject a response whose fields disagree.
- [ ] **Report the finding before writing a fix.** If iOS already agrees the fields, close this issue with the evidence — do not manufacture a change.

### Must Do — step 2, only if the gap is real
- [ ] Refuse a response that disagrees with itself on destination or amount.
- [ ] Compare TON destinations as **parsed accounts** — `EQ…`, `UQ…` and raw `workchain:hex` are one account and only the flag byte and checksum separate the first two.
- [ ] Compare every other chain **byte-for-byte** — base58 is case-sensitive, so normalising case would let two different addresses pass as one.
- [ ] Reject a multi-entry `tx[]`: only `tx[0]` is ever built, so signing it would under-deliver.
- [ ] Re-spell an agreed TON destination in its bounceable (`EQ…`) form. The agreement check ignores bounceability on purpose, but the signer reads the bounce flag off the spelling — a `UQ…` or raw field winning precedence would send the deposit non-bounceable, where a rejecting contract absorbs it.
- [ ] Tests mirroring sdk#2296's block: divergent `tx[0].address`; divergent `depositAddress`; **accept** one TON account spelled `EQ…` in one field and `UQ…` in another; reject a non-TON divergence differing only by case; divergent `tx[0].amount`; **accept** a decimal `depositAmount` agreeing with a base-unit `tx[0].amount`; reject a multi-entry `tx[]`.

### Must NOT Do
- Don't normalise base58 case.
- Don't change how iOS already handles the destination-tag suffix (`SwapKitSwapResponse.swift:53-68`).
- Don't invent a fix if step 1 shows iOS is already correct.

## Acceptance Criteria
- [ ] The investigation result is stated explicitly, with file:line evidence, either in a closing comment or the PR body.
- [ ] If a fix ships: build succeeds, full gate green, and all seven test cases above pass.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [sdk#2188](https://github.com/vultisig/vultisig-sdk/issues/2188) under open epic [sdk#2181](https://github.com/vultisig/vultisig-sdk/issues/2181) · reference impl [sdk#2296](https://github.com/vultisig/vultisig-sdk/pull/2296)

gastonm5:
## Problem

SwapKit's `/v3/swap` response states the deposit destination in three independent fields — `targetAddress`, `inboundAddress`, and the `tx[]` transfer array — and the payload builder read only `targetAddress`. A response whose halves diverged was signed to that field while the others named a different account, and nothing downstream could tell.

**On the affected chains this was the only check there was.** `SecurityScannerTransactionFactory.swift:35-45` routes Blockaid swap scanning for **EVM and Solana sources only**, so TON, XRP, Cardano, UTXO, Sui and Tron SwapKit deposits are never scanned — and the signer never decodes `txPayload` for TON, so a co-signer could not spot a divergence either.

The realistic trigger is not a tampered payload but an unversioned wire change. This provider demonstrably does that: the decoder already absorbs `SOLANA` → `SERIALIZED_BASE64` and `CARDANO` → `CBOR` renames shipped mid-flight without versioning (`SwapKitSwapResponse.swift:210-215, 262-274`). Repurposing a destination field the same way would be a lost deposit.

`tx[]` cardinality was **completely unvalidated**: 0, 2 or 10 entries built and signed identically, because only `tx[0]` is ever built.

## Investigation first (the issue asked for this before any fix)

iOS has **no precedence chain at all** — unlike the SDK, it cannot pick the wrong field, because it only ever reads `targetAddress`. But it never compares it against anything either, and `grep` for `transfers.count` / `transfers.first` / `guard transfers` across `VultisigApp/VultisigApp/` returned **zero** production hits.

The issue (and [sdk#2296](https://github.com/vultisig/vultisig-sdk/pull/2296)) named `depositAddress` and `depositAmount` as the corroborating fields. **Neither appears in any of the 17 recorded fixtures** — the SDK models them defensively. What iOS actually receives is `inboundAddress` (present and byte-identical to `targetAddress` on every non-EVM route, absent on EVM) and `tx[0]`. So the guard checks the fields this client is actually sent.

## What changed

**1. Destination self-agreement (`319d5b9e9`)** — `targetAddress`, `inboundAddress` and `tx[0].address` must agree, or the response is refused at the single gate every quote already passes. Neither field is authoritative enough to arbitrate from, so a divergence is refused rather than resolved by precedence. A field that is present but blank is refused too: it corroborates nothing, and nothing is not agreement.

- **TON compares parsed accounts** — `EQ…`, `UQ…` and raw `workchain:hash` are one account. Two values that fail to parse as TON accounts are never reported as naming the same one.
- **Every other chain compares byte for byte.** Base58 is case-sensitive; folding case would let two different addresses pass as one. Bech32 and CashAddr *are* case-insensitive and could be normalised safely, but all 17 fixtures are byte-identical across `targetAddress`/`inboundAddress` (including BCH — no CashAddr prefix variance), so that would be untested surface in the code that most needs to be trusted. The cost is bounded and stated in the code: a refused route the user can retry, never a deposit signed to the wrong address.
- **A TON `tx[]` must state exactly one transfer.** Android already rejects an empty array (`SwapKitQuoteSource.kt:591-593`) and a blank `targetAddress` (`:511-513`); iOS did neither.
- **Destination tags are read as a tri-state**, not `UInt64?`. See below.

**2. Request echo (`86599e1b9`)** — `buildSwapTx` validated *nothing* about the response it returned. `JupiterService.swift:91-95` is the in-repo precedent for exactly this shape. The response must now echo the requested `routeId`, `sourceAddress` and `destinationAddress`. Verified against **all six** recorded quote/swap fixture pairs before relying on it.

**3. Three production comments corrected** — `KeysignMessageFactory.swift` (twice) and `KeysignViewModel.swift` claimed `keysignPayload.toAmount` is `tx[0].amount`. It is the amount the user typed. A comment asserting a check exists is how the next reader concludes it does.

## Behaviour changes worth calling out

- **`?dt=bogus&dt=5` no longer resolves to `5`.** It is now refused. The response states two tags, one unreadable; picking the readable one is the wallet inventing a fact the provider never established, and on XRP a wrong tag misattributes the deposit at a shared exchange address. Nothing else about destination-tag handling moved — `resolvedDestinationTag` and `resolvedTargetAddress` behave identically on every other input form.
- **`SwapKitRippleTests` deliberately constructs tag conflicts** (100 vs 200, suffix 300 vs meta 200) to pin `resolvedDestinationTag`'s precedence. Those tests still pass — the helper is untouched — but such a response can no longer reach it in production. The precedence remains live for non-conflicting inputs.

### The amount leg

**As specified, it would never fire.** The issue's Must-Do is `depositAmount` ↔ `tx[0].amount`, and iOS never receives `depositAmount` — it appears in none of the 17 recorded fixtures, and it isn't in the response model at all. Shipping that check would add a guard that cannot execute, which is worse than adding nothing: it shows up in review, in this description, and in the next reader's mental model as "the amount is checked", manufacturing confidence without protection.

**The version that *would* fire rests on one fixture, against three implementations that decline to make the same claim.** The only live pair on iOS is `sellAmount` (decimal) ↔ `tx[0].amount` (base units), and exactly one recorded response demonstrates them agreeing. Meanwhile the SDK's precedence is `depositAmount` → `tx[0].amount` → the caller's amount and never consults `sellAmount`; Android signs `request.tokenValue`, the user's typed amount; and on iOS `sellAmount` is decoded and then read by no production code whatsoever. A jetton-source TON route would state `sellAmount` in jetton units against a nano-TON `tx[0].amount`, and a fee-deducted deposit would diverge legitimately — either would refuse a healthy route.

**The two legs are not held to the same bar, deliberately.** A wrong destination loses the deposit outright, so that leg fails closed on the strength of fields present in every response. A wrong *amount* check doesn't lose anything — it refuses valid routes — so the bar for shipping it is higher, not lower, and inferring its semantics from a single sample doesn't clear it. Filed as **#5358**: a SwapKit doc statement that `sellAmount` is the deposit size, or a second fixture (ideally a jetton-source TON route), would settle it.

### Error presentation

Both new errors render a route-unavailable body under an "Unexpected Error" title. That is real, but not a regression: `SwapKitError` does not conform to `SwapErrorPresentable` and `normalizedSwapKitError` maps exactly one case, so **17 of 18 cases already do this**, `swapRouteNotFound` and `insufficientBalance` among them. Fixing only the two new ones would deepen the inconsistency that file exists to prevent. Filed as **#5362**.

### Peer-side revalidation

A co-signer reconstructs `SwapKitSwapPayload` from protobuf (`KeysignMessage+ProtoMappable.swift:305-318`) and signs without revalidating what the initiator claimed. Verified, and filed as **#5359** as the general question across payload types. So this guard is the sole check on those chains **for swaps quoted on this device**; a swap initiated elsewhere is #5359's scope.

## Verification

- `SwapKitResponseAgreementTests` (26), `TonAccountIdentityTests` (6), `SwapKitRequestEchoTests` (8) — divergent `tx[0].address`; divergent `inboundAddress`; **accepts** one TON account spelled `EQ…` in one field and `UQ…` in another, and raw against user-friendly; rejects a non-TON divergence differing only by case; rejects multi-entry and empty `tx[]`; rejects blank corroborating fields; rejects an unreadable or duplicated tag at each of the four sources; and **accepts every one of the 17 recorded real responses** through the full gate, which is what makes a fail-closed guard safe to ship.
- TON vectors are derived from the account SwapKit actually returns in `v3-real-ton-swap.json` (tag byte, workchain and CRC16 verified offline for each spelling), not synthetic ones.
- **The echo tests are built by mutating one field of a recorded response rather than hand-rolling JSON, deliberately.** Hand-written fixtures drifted from the real wire shape twice on this branch. The second time was worse than it looked: a `makeJSON` that declared `meta.txType = "TON"` with no `tx` key would have made `decodeTx` throw `keyNotFound(.tx)`, so every test in that file died at the decode line before reaching an assertion — a suite that could not have tested anything, and would have read as green intent in review. Starting from recorded wire data removes the class.
- `buildSwapTx` is driven end to end through an injected `HTTPClientProtocol` stub, asserting the error **type** that reaches the caller. Deleting the guard's call site, or letting the surrounding `catch HTTPError.statusCode` remap it to `.generic`, now fails a test instead of staying green. The agreement guard is pinned the same way through `validateSigningCapability`.
- **The gate caught a failure no cheaper check on this branch could have.** A `split/map/filter/map` chain in `extractTagSuffix` compiled to *"the compiler is unable to type-check this expression in reasonable time"* and the app target did not build. SwiftLint was clean on it — it walks a syntax tree and never invokes the type checker — so lint-clean was not compile evidence, and the fault was invisible until the build ran. Rewritten as the explicit loop the helper used before, which removes the chained generics rather than helping the solver through them.

## Review

**Needs human review.** This is a fail-closed guard on the signing path for chains that have no other protection, and it can refuse a route — a wrong rejection blocks a swap the user expected to work.

**On-device acceptance was NOT performed.** The gate is simulator-only: no real TON, XRP, Cardano or UTXO deposit was quoted, signed or broadcast against live SwapKit, and no cross-device co-signing ceremony was run. Every claim here rests on the 17 recorded fixtures and the unit suite.

Gate: `** TEST SUCCEEDED **`, `Executed 7119 tests, with 11 tests skipped and 0 failures (0 unexpected)`, `GATE_EXIT=0`, zero compile errors — 41 new tests on a 7078 baseline, predicted from the diff before the run.

Closes #5343

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation to ensure swap responses match requested routes and addresses.
  - Improved TON address comparison across equivalent formats.
  - Strengthened destination-tag handling, detecting invalid, ambiguous, or conflicting values.
  - Rejected responses with contradictory transfer details or unsupported destination data.
  - Added clearer diagnostic details for failed transaction construction.

- **Tests**
  - Added extensive coverage for response validation, TON address identity, destination tags, and request-echo mismatches.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitDestinationTag.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitDestinationTag.swift
@@ -7,36 +7,25 @@ import Foundation
 
 /// What one source says about an XRP destination tag.
 ///
-/// The tag is half the destination: a shared exchange address routes a deposit to an
-/// account by tag, so "states no tag" and "states a tag this client cannot read" are
-/// different answers that must never collapse into each other. A two-state `UInt64?`
-/// forces exactly that collapse, and a fail-closed guard built on it reads an unreadable
-/// tag as a tag-less payment — which credits nobody.
+/// "States no tag" and "states a tag this client cannot read" are different answers, and a
+/// `UInt64?` forces them to collapse — which sends a tag-less deposit to an address where the
+/// tag is what identifies the depositor.
 enum SwapKitDestinationTag: Hashable {
-    /// The source says nothing: the field is absent or null, or the address carries no
-    /// tag suffix.
     case absent
-    /// The source states exactly one readable tag.
     case tag(UInt64)
-    /// The source states something that is not one readable tag — a value that is not a
-    /// `UInt64`, or more than one `dt` parameter. Two answers is not one answer, and
-    /// picking one of them is a guess that misroutes a deposit.
+    /// A value that is not a `UInt64`, or more than one `dt` parameter. Two answers is not one
+    /// answer, and picking one is a guess.
     case unreadable(reason: String)
 
-    /// The tag to use, or nil when there is none to use.
-    ///
-    /// `.unreadable` yields nil so the resolution helpers behave exactly as they did
-    /// before this type existed. Refusing such a response is the agreement guard's job,
-    /// not theirs — resolution answers "what would we use", validation answers "may we
-    /// use this response at all".
+    /// `.unreadable` yields nil so resolution behaves as it did before this type existed;
+    /// refusing such a response is the agreement guard's job, not resolution's.
     var usableTag: UInt64? {
         guard case .tag(let value) = self else { return nil }
         return value
     }
 
-    /// Read a tag SwapKit may send as a number or as a numeric string (`meta.affiliateFee`
-    /// arrives string-wrapped despite being numeric, so both shapes are accepted). A key
-    /// that is present but holds neither is `.unreadable`, never `.absent`.
+    /// SwapKit sends the tag as a number or as a numeric string. Present-but-neither is
+    /// `.unreadable`, never `.absent`.
     static func decoded<Key: CodingKey>(
         from container: KeyedDecodingContainer<Key>,
         forKey key: Key
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitDestinationTag.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitDestinationTag.swift
@@ -1,0 +1,58 @@
+//
+//  SwapKitDestinationTag.swift
+//  VultisigApp
+//
+
+import Foundation
+
+/// What one source says about an XRP destination tag.
+///
+/// The tag is half the destination: a shared exchange address routes a deposit to an
+/// account by tag, so "states no tag" and "states a tag this client cannot read" are
+/// different answers that must never collapse into each other. A two-state `UInt64?`
+/// forces exactly that collapse, and a fail-closed guard built on it reads an unreadable
+/// tag as a tag-less payment — which credits nobody.
+enum SwapKitDestinationTag: Hashable {
+    /// The source says nothing: the field is absent or null, or the address carries no
+    /// tag suffix.
+    case absent
+    /// The source states exactly one readable tag.
+    case tag(UInt64)
+    /// The source states something that is not one readable tag — a value that is not a
+    /// `UInt64`, or more than one `dt` parameter. Two answers is not one answer, and
+    /// picking one of them is a guess that misroutes a deposit.
+    case unreadable(reason: String)
+
+    /// The tag to use, or nil when there is none to use.
+    ///
+    /// `.unreadable` yields nil so the resolution helpers behave exactly as they did
+    /// before this type existed. Refusing such a response is the agreement guard's job,
+    /// not theirs — resolution answers "what would we use", validation answers "may we
+    /// use this response at all".
+    var usableTag: UInt64? {
+        guard case .tag(let value) = self else { return nil }
+        return value
+    }
+
+    /// Read a tag SwapKit may send as a number or as a numeric string (`meta.affiliateFee`
+    /// arrives string-wrapped despite being numeric, so both shapes are accepted). A key
+    /// that is present but holds neither is `.unreadable`, never `.absent`.
+    static func decoded<Key: CodingKey>(
+        from container: KeyedDecodingContainer<Key>,
+        forKey key: Key
+    ) -> SwapKitDestinationTag {
+        guard container.contains(key), (try? container.decodeNil(forKey: key)) == false else {
+            return .absent
+        }
+        if let intTag = try? container.decode(UInt64.self, forKey: key) {
+            return .tag(intTag)
+        }
+        if let stringTag = try? container.decode(String.self, forKey: key) {
+            guard let parsed = UInt64(stringTag) else {
+                return .unreadable(reason: "\"\(stringTag)\" is not a destination tag")
+            }
+            return .tag(parsed)
+        }
+        return .unreadable(reason: "value is neither a number nor a numeric string")
+    }
+}
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
@@ -37,6 +37,11 @@ enum SwapKitError: Error, LocalizedError, Equatable {
     case isSanctionedAddress
     case addressScreeningFailed
     case unsupportedTxType(String)
+    /// The `/v3/swap` response disagrees with itself about where the deposit
+    /// goes, or states a transfer array the builder cannot honour. `detail` is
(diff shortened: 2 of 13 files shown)


johnnyluo (approved):
LGTM

Issue #5364, opened 2026-09-08
[Fix] A never-included TRON transaction stays pending forever — the observation window closes 55 min before expiry [Tron]


gastonm5:
## Problem

A TRON transaction that is broadcast but never included stays `notFound` forever. iOS has no way to reach a terminal state for it, because **the only observation window closes 55 minutes before the transaction expires.**

Measured, repo-internal:

- iOS sets `expiration = broadcast + 1 hour` — `Blockchain/Tron/Service/TronService.swift:94-95`.
- TRON's poll `maxWaitTime` is **300 s** — `Core/Models/TransactionStatus.swift:176`.
- So a live polling session ends **55 minutes before** the expiration it would need to observe.
- The only other observation point is a later app open, which performs exactly **one** check (`elapsed >= maxWaitTime` → one lookup, then `.stop`).

For a terminal expiry state to fire on its target case, the node would have to still be serving a raw transaction more than an hour after broadcast, having never included it.

## Why the obvious fix doesn't work

Deciding expiry from `gettransactionbyid` cannot work, because that endpoint answers for **included** transactions. Verified against a live transaction from the current head block through the app's own proxy:

```
txID db1423335757f48988983efb0d4a527ddf9b7754647c134b3c56d1ffa4d73262
gettransactionbyid     -> raw_data, raw_data_hex, ret, signature, txID
                          ret present (inclusion), raw_data.expiration = 1788924530336
gettransactioninfobyid -> blockNumber 86072202
```

`ret` present means it landed — and it still carries an `expiration` that will be in the past within the hour. So a rule of the form "raw transaction present + expiration passed ⇒ expired" **marks confirmed payments as terminally failed**, inviting a duplicate payment. A never-included transaction, meanwhile, returns `{}` and reads as `notFound`.

The state such a check can see is the one it must never act on; the state it needs is the one it cannot see.

## What a real fix requires

The node cannot supply an expiration for a transaction it has dropped, so **expiry has to be decided locally**. Two candidates:

1. **Persist the payload's `expiration`** on `StoredPendingTransaction` and compare against chain time. Correct, but a SwiftData schema change — which is why it belongs in its own PR rather than bolted onto a fee fix.
2. **Treat a row still `notFound` more than an hour after its persisted `createdAt` as expired.** Cheaper, but it terminalizes on *absence of evidence*, and it would be wrong for a cross-device payload whose initiator chose a different expiration.

[...]

PR #5302, opened 2026-08-31
fix(tron): remove phantom staking memo fee

gastonm5:
## Description

Stop charging TRON's 1 TRX memo fee for the app-local `FREEZE:` and `UNFREEZE:` routing markers. These markers select WalletCore staking builders but are never serialized into the transaction memo/data field.

The change shares the routing-marker detection with signing and adds regression coverage for both BANDWIDTH and ENERGY freeze/unfreeze operations while preserving the fee for real memos.

Fixes #5296

## Which feature is affected?

- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [x] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [x] New Chain / Chain related feature  -  Please attach tx link here

No transaction link: this draft requires an on-device freeze/unfreeze check before merge.

## Parity

- Android: linked: vultisig/vultisig-android#5481
- Windows + extension: n/a: the iOS WalletCore routing markers are not used there
- SDK: n/a: the iOS WalletCore routing markers are not used there

## Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

## Screenshots (if applicable):

Not applicable; fee behavior is covered by service-level tests.

## Additional context

The full local suite passed: 6,880 tests, 11 skipped, 0 failures. Claude review was skipped as requested.

## Agent Metadata (if applicable)

- **Agent**: Codex
- **Issue**: #5296
- **Confidence**: high
- **Needs human review for**: on-device freeze/unfreeze fee confirmation

PR #5298, opened 2026-08-31
fix(tron): accept successful contract receipts

gastonm5:
## Description

Fixes TRON transaction-status resolution so a contract receipt with result SUCCESS is confirmed instead of reported as failed. A nil receipt result also remains confirmed, while any other present result fails closed with the existing reason construction. Top-level FAILED keeps precedence.

Closes #5294

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [x] Sending - Live TRC-20 status validation still required before merge
- [ ] Swap
- [ ] New Chain / Chain related feature

## Parity

- Android: n/a: already treats case-insensitive SUCCESS as confirmed
- Windows + extension: n/a: use the SDK status resolver, which already accepts SUCCESS
- SDK: n/a: already accepts SUCCESS; it differs only on unknown future receipt codes

## Checklist

- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

## Screenshots (if applicable):

Not applicable.

## Additional context

The implementation follows the issue and Android fail-closed semantics for unknown non-SUCCESS values. Nine focused tests cover SUCCESS, lowercase success, nil, known failure, empty and unknown results, top-level failure precedence, pending, and not-found.

Verification: full local gate passed with 6,888 tests executed, 11 skipped, and 0 failures. Touched-file SwiftLint is clean.

Manual validation before merge: confirm a real successful TRC-20 transfer reaches the Done screen and history as confirmed.

## Agent Metadata (if applicable)
- **Agent**: Codex
- **Issue**: #5294
- **Confidence**: high
- **Needs human review for**: live TRC-20 status canary

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Tron transaction status handling.
  * Transactions are now marked successful when the result is missing or equals “SUCCESS” regardless of letter casing.
  * Other available result values are correctly treated as failed.

* **Tests**
  * Added coverage for confirmed, failed, pending, and not-found transaction states.

PR #5365, opened 2026-09-08
fix(tron): measure the bandwidth reserve from the transaction, and verify the broadcast txid

### LINKED ISSUE #5344: [Fix] TRON: memo-blind bandwidth reserve, and no expired terminal status [Tron]
## Problem

Two independent TRON defects.

**A — bandwidth is estimated from a constant, blind to the memo.** `Blockchain/Tron/Service/TronService.swift:36-37` declares fixed sizes:

```swift
private static let BYTES_PER_COIN_TX: Int64 = 300
private static let BYTES_PER_CONTRACT_TX: Int64 = 345
```

applied at `:387` in `getBandwidthFeeDiscount`, which then decides at `:391-400` whether the account has enough free bandwidth for the transfer to be free. A memo enlarges the serialized transaction, so a send iOS budgets as free can in fact consume paid bandwidth, and the displayed fee is wrong in the user's disfavour.

**B — there is no `expired` terminal status.** `Core/Services/TransactionStatus/Providers/TronTransactionStatusProvider.swift` maps only success / failed / pending / notFound. TRON transactions carry an `expiration` (iOS sets `now + 1 hour`, `TronService.swift:86-89`); once it passes without inclusion the transaction can never land, but iOS keeps polling it as `notFound` indefinitely.

## Expected Behavior

- The bandwidth estimate reflects the transaction actually being signed, memo included.
- A TRON transaction past its expiration that never landed is reported as terminally failed and polling stops.

## Solution

Size the bandwidth estimate from the serialized transaction plus the memo rather than a per-shape constant. Add an `expired` terminal state to the TRON status provider, keyed off the payload's `expiration` versus chain time, and stop polling on it.

While in this code, also confirm `TronAPI.broadcastTransaction` validates the txid the node returns against the locally computed hash — sdk#2318 found the node's value was being trusted. If iOS already validates, say so; if not, fix it here.

### Must Do
- [ ] Bandwidth estimate accounts for serialized size + memo.
- [ ] `expired` terminal status; polling stops; reported as failed with a clear reason.
- [ ] Verify (and fix if needed) broadcast txid validation against the locally computed hash.
- [ ] Tests: memo-bearing vs memo-less bandwidth estimate; an expired unincluded transaction reports expired and stops polling; a still-valid pending one keeps polling.

### Must NOT Do
- Don't change the fee-limit / energy path for TRC-20 (separate concern, already fixed in #5298–#5302).
- Don't move the simulation off the full node — iOS already uses it correctly (`TronAPI.swift:55`) and the SDK aligned to iOS here.

## Acceptance Criteria
- [ ] A memo-bearing TRX send no longer reports as free when it would consume paid bandwidth.
- [ ] An expired TRON transaction reaches a terminal failed/expired state instead of polling forever.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [sdk#2275](https://github.com/vultisig/vultisig-sdk/issues/2275), [android#5766](https://github.com/vultisig/vultisig-android/issues/5766), [android#5767](https://github.com/vultisig/vultisig-android/issues/5767) · reference impls [sdk#2318](https://github.com/vultisig/vultisig-sdk/pull/2318), [android#5776](https://github.com/vultisig/vultisig-android/pull/5776)

gastonm5:
## Problem

**A — the bandwidth reserve was a constant, blind to the memo.** `getBandwidthFeeDiscount` sized every native transfer at 300 bytes and decided from that whether the account's free bandwidth made the send free.

The issue predicted this under-reserves memo-bearing sends. Measuring showed the constant was **wrong in both directions**: a bare transfer is 263–267 bytes, so it *over*-reserved that too. Near the free/paid boundary the app charged a fee on sends that were genuinely free, as well as showing free sends that would burn paid bandwidth once a memo was attached.

**C — the broadcast txid was taken on trust.** The issue asked me to confirm whether iOS validates it. **It did not.** `TronAPIService.broadcastTransaction` returned `response.data.txid` verbatim and `KeysignViewModel` assigned it straight into `self.txid` — the hash the app then tracks, displays and links to an explorer. Both signers already compute the canonical hash (`TronHelper.getSignedTransaction` from WalletCore's output id, `SwapKitTronSigner` from `sha256(raw_data)`), so nothing forced the node's answer to describe the transaction we signed. Same gap sdk#2318 found.

## Fix

**A.** `TronHelper.nativeTransferBandwidthBytes` builds the signing input the signer will build and measures WalletCore's pre-signing payload, which is exactly the `raw_data` bytes. The signed envelope adds that field's length prefix, one length-delimited 65-byte signature, and java-tron's 64-byte per-contract result allowance. Ports the SDK's `getNativeTronBandwidthBytes`. The amount is part of those bytes, so it is threaded from `BlockChainService`.

**C.** The local hash is threaded into the broadcast call, compared case- and `0x`-tolerantly, and returned on success. A mismatch throws into the existing `handleBroadcastError` → `isAlreadyOnChain` net, which recovers the hash if the transaction really did land.

## B — the `expired` status was built, then removed before shipping

The issue's second half asked for an `expired` terminal state. It was implemented, reviewed, and **taken back out**, because the evidence says it cannot work against these endpoints. Refiled as #5364 with both fix options.

Two independent reasons:

- **It cannot see its target.** A live query against a head-block transaction returns `ret` alongside `raw_data.expiration` — so `gettransactionbyid` serves *included* transactions, with an expiration that will be in the past within the hour. A transaction that never landed is not served at all. The probe correlates with inclusion, not expiry: it would mark landed payments terminally failed while never firing on the stuck transactions it targets.
- **The timing rules it out regardless.** Transactions are built with a one-hour expiration (`TronService.swift:95`) while the poller's TRON window is five minutes (`TransactionStatus.swift:176`), so a live polling session ends fifty-five minutes before the expiration it is meant to detect. The only remaining observation is a single check on a later app open, by which time the node has long dropped a transaction that never landed. This argument stands without any assumption about TRON's pending-pool semantics.

The tests passed because they fabricated a raw response for a never-included transaction past its expiration — a response the chain does not produce. A terminal state that never fires while looking handled is worse than the known bug it replaced, and this one also misfires on landed payments, which invites a duplicate send.

## Three more things review turned up

**Send Max reserved nothing for the amount field.** `SendDetailsViewModel.refineMaxFromFee` quotes a fee with `amount: .zero`, and protobuf omits a zero scalar entirely — so the estimator reserved no bytes for a field every real transfer carries, and could show a max send as free when it crosses into paid bandwidth. Not in the issue; the same under-reservation by another road. Unknown amounts are now sized at the widest varint.

**The cached estimate ignored the amount.** `fetchSpecificForNonEVM` keyed its 60-second cache on the amount only for Terra Classic. Making TRON's estimate amount-sensitive meant a cached quote could be reused for a wider second amount — displayed free, actually charged. TRON now joins Terra Classic in the amount-keyed cache. (`FunctionTransactionVerifyViewModel` uses that cache; the standard Send interactor bypasses it.)

**The 300-byte fallback stopped being an upper bound.** Taken whenever the transaction cannot be serialized, it was sized for a bare transfer — but a real memo-bearing one is `267 + memo + 2`, so past ~30 memo bytes it under-reserved the exact case the measurement exists to catch. The memo is known even when the transaction cannot be built, so it is charged on top.

## How the byte counts were derived

The tests assert exact counts rather than a range, so the signature framing and the 64-byte allowance are pinned — a range would have passed with either removed. For the helper's fixture (`amount` 1,000,000), from TRON's `Transaction.raw` schema:

```
TransferContract  owner 1+1+21, to 1+1+21, amount 1+3            =  50
Any               type_url 1+1+45, value 1+1+50                  =  99
Contract          type 1+1, parameter 1+1+99                     = 103
raw_data          ref_block_bytes 1+1+2, ref_block_hash 1+1+8,
                  expiration 1+6, contract 1+1+103, timestamp 1+6 = 133
signed tx         raw_data 1+2+133, signature 1+1+65             = 203
+ MAX_RESULT_SIZE_IN_TX                                          =  64
                                                                 = 267
```

`ref_block_num` and `fee_limit` stay at their proto3 defaults and are not serialized (`wallet-core/src/Tron/Signer.cpp:402-414`). The service tests call `getBlockInfo` with no amount, which the estimator sizes at the widest varint: 267 + 6 = **273**.

Cross-checked against a real recorded transaction rather than taken on trust — the `raw_data_hex` in `Swap/SwapKit/__fixtures__/v3-tron-final-swap-fresh.json`. It opens `0a 02 8975` (ref_block_bytes, 4 bytes), `22 08 …` (ref_block_hash, 10), `40 …` (expiration, 7), carries its addresses as `0a 15 41…` / `12 15 41…` (21-byte payloads), ends `70 …` (timestamp, 7), and has no `ref_block_num` anywhere. Its `0a 31` type_url length of 49 is `type.googleapis.com/` (20) + `protocol.TriggerSmartContract` (29) — the same naming rule that gives `protocol.TransferContract` the 45 used above.

That fixture also supplied the sender address. The previous fixture address failed base58check, so WalletCore refused to build the transaction and every service assertion was silently measuring the 300-byte fallback instead of the measurement it meant to check. The fallback now has tests of its own.

## Reviewer notes

- The reserve is always shaped as a `TransferContract`. The Stake 2.0 routing memos sign smaller system contracts — none carries a `to_address` — so it is a strict upper bound for a freeze/unfreeze/claim: it can predict a fee for one that would have been free, but never under-reserves. Sizing each shape means duplicating four signing builders for ~14 bytes.
- The `TronAPIError.broadcastHashMismatch` message is English, deliberately: all three sibling cases and the `handleBroadcastError` wrapper around it are un-localized, so localizing only the new one would be inconsistent. Localizing that family is a clean follow-up.
- The status-provider test doubles dispatch on `target.path` alone and ignore `method` and `task`, so no test in that file — existing ones included — would notice a GET silently becoming a POST. A property of the double, not a missing case.
- The TRC-20 fee-limit/energy path (#5298–#5302) and the simulation node are untouched, per the issue.

## Verification

Full gate green, all tests, on an `en_US`-pinned simulator:

```
** TEST SUCCEEDED **
Executed 7098 tests, with 11 tests skipped and 0 failures (0 unexpected)
```

7078 on `main` + 21 new − 1 renamed = 7098, matching the count predicted from the diff. Disk was flat at 31 GiB free across the run, so this is not a stale-products pass.

Reviewed commit-by-commit by `codex exec --sandbox read-only`, plus two whole-branch passes. Findings fixed on the branch: a landed transaction could be reported expired (P1), `DUP_TRANSACTION_ERROR` accepted as a successful broadcast (P1), the Send Max zero-amount hole, the amount-blind cache key, the memo-blind fallback, a device-clock gate that suppressed expiry on a slow clock, and two tests that could not fail (a circular expectation and a fabricated chain response).

**Fee display changes what a user is shown before signing, so this wants human review. No on-device acceptance was performed.**

Closes #5344

## Summary by CodeRabbit

* **New Features**
  * Improved TRON transaction broadcast validation by confirming successful responses and matching returned transaction IDs with the expected transaction hash.
  * Enhanced TRON fee estimates using transaction size, transfer amount, and memo details.
  * Improved handling of duplicate, rejected, or mismatched transaction broadcasts.

* **Bug Fixes**
  * Prevented inaccurate fee-cache reuse for amount-dependent transactions.

* **Tests**
  * Added coverage for TRON broadcast validation, bandwidth estimates, memo handling, variable amounts, and invalid transaction data.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
@@ -69,28 +69,23 @@ struct TronAPIService {
 
     // MARK: - Broadcast
 
-    /// - Parameter expectedTxHash: the hash the signer computed locally, which
-    ///   the node's answer is checked against.
+    /// - Parameter expectedTxHash: the locally computed hash the node's answer
+    ///   is checked against.
     func broadcastTransaction(jsonString: String, expectedTxHash: String) async throws -> String {
         let response = try await httpClient.request(api(.broadcastTransaction(jsonString: jsonString)), responseType: TronBroadcastResponse.self)
 
-        // Only an explicit success counts. `DUP_TRANSACTION_ERROR` reads like
-        // one and used to be accepted as one, but TRON fills its duplicate
-        // cache before validating, so a transaction the node went on to reject
-        // answers with that code too — it is not evidence the transaction is
-        // on chain. Throwing routes it into `handleBroadcastError`, which looks
-        // the hash up on chain and recovers the ones that really did land.
+        // Only an explicit success counts. TRON fills its duplicate cache
+        // before validating, so `DUP_TRANSACTION_ERROR` is not evidence the
+        // transaction is on chain. Throwing routes it into
+        // `handleBroadcastError`, which verifies the hash against the chain.
         guard let txid = response.data.txid, response.data.result == true else {
             let errorMessage = response.data.message ?? response.data.code ?? "Unknown error"
             throw TronAPIError.broadcastFailed(errorMessage)
         }
 
-        // The hash the app then tracks, displays and links to an explorer must
-        // be the one it signed, not one the endpoint reports back. An honest
-        // node returns the same value — TRON's txid is `sha256(raw_data)`,
-        // which the signer already computed — so a mismatch means the response
-        // does not describe the transaction that was broadcast, and adopting it
-        // would follow someone else's transaction instead.
+        // TRON's txid is `sha256(raw_data)`, which the signer already computed,
+        // so an honest node returns the same value. A mismatch means the
+        // response does not describe the transaction that was broadcast.
         guard Self.txidMatchesLocalHash(nodeTxid: txid, localTxHash: expectedTxHash) else {
             throw TronAPIError.broadcastHashMismatch(expected: expectedTxHash, returned: txid)
         }
@@ -98,10 +93,8 @@ struct TronAPIService {
         return expectedTxHash
     }
 
-    /// Hex comparison tolerant of case and a `0x` prefix. An empty local hash
-    /// never matches: without one there is nothing to verify against, and
-    /// falling back to the node's value is exactly what this check exists to
-    /// prevent.
+    /// Tolerant of case and a `0x` prefix. An empty local hash never matches:
+    /// without one there is nothing to verify against.
     static func txidMatchesLocalHash(nodeTxid: String, localTxHash: String) -> Bool {
         let local = normalizedHash(localTxHash)
         guard !local.isEmpty else { return false }
--- a/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
@@ -69,19 +69,18 @@ struct TronAPIService {
 
     // MARK: - Broadcast
 
-    private static let DUP_TRANSACTION_ERROR_CODE = "DUP_TRANSACTION_ERROR"
-
     /// - Parameter expectedTxHash: the hash the signer computed locally, which
     ///   the node's answer is checked against.
     func broadcastTransaction(jsonString: String, expectedTxHash: String) async throws -> String {
         let response = try await httpClient.request(api(.broadcastTransaction(jsonString: jsonString)), responseType: TronBroadcastResponse.self)
 
-        // Accept success (result == true) OR duplicate transaction error (already broadcast)
-        // This matches Android behavior where DUP_TRANSACTION_ERROR is treated as success
-        let isSuccess = response.data.result == true
-        let isDuplicateTransaction = response.data.code == Self.DUP_TRANSACTION_ERROR_CODE
-
-        guard let txid = response.data.txid, isSuccess || isDuplicateTransaction else {
+        // Only an explicit success counts. `DUP_TRANSACTION_ERROR` reads like
+        // one and used to be accepted as one, but TRON fills its duplicate
+        // cache before validating, so a transaction the node went on to reject
+        // answers with that code too — it is not evidence the transaction is
+        // on chain. Throwing routes it into `handleBroadcastError`, which looks
+        // the hash up on chain and recovers the ones that really did land.
+        guard let txid = response.data.txid, response.data.result == true else {
             let errorMessage = response.data.message ?? response.data.code ?? "Unknown error"
             throw TronAPIError.broadcastFailed(errorMessage)
         }
--- a/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
@@ -71,7 +71,9 @@ struct TronAPIService {
 
     private static let DUP_TRANSACTION_ERROR_CODE = "DUP_TRANSACTION_ERROR"
 
-    func broadcastTransaction(jsonString: String) async throws -> String {
+    /// - Parameter expectedTxHash: the hash the signer computed locally, which
+    ///   the node's answer is checked against.
+    func broadcastTransaction(jsonString: String, expectedTxHash: String) async throws -> String {
         let response = try await httpClient.request(api(.broadcastTransaction(jsonString: jsonString)), responseType: TronBroadcastResponse.self)
 
         // Accept success (result == true) OR duplicate transaction error (already broadcast)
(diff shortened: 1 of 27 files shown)


VultisigApp/VultisigApp/Blockchain/Tron/Service/TronService.swift
```diff
@@ -88,7 +97,23 @@ class TronService {
         let oneHourMillis = Int64(60 * 60 * 1000)
         let expiration = nowMillis + oneHourMillis
 
-        let estimate = try await calculateTronFee(coin: coin, to: to, memo: memo, isSwap: isSwap)
+        let bandwidthBytes = nativeTransferBandwidthBytes(
```
johnnyluo, line 100: `getBlockInfo` unconditionally runs the WalletCore-backed bandwidth measurement (`nativeTransferBandwidthBytes`) even when `isSwap` is true, but `calculateTronFee`'s `isSwap` branch never reads `bandwidthBytes` — it's only consumed in the non-swap `else` branch via `calculateNativeTrxFee(coin:bandwidthBytes:)`. Every native-TRX swap quote triggers a full WalletCore FFI round-trip whose result is discarded. Consider skipping the computation (e.g. `isSwap ? 0 : nativeTransferBandwidthBytes(...)`) when `isSwap` is true.

VultisigApp/VultisigApp/Blockchain/Tron/Service/TronAPIService.swift
```diff
@@ -69,22 +69,44 @@ struct TronAPIService {
...
+
+    /// Tolerant of case and a `0x` prefix. An empty local hash never matches:
+    /// without one there is nothing to verify against.
+    static func txidMatchesLocalHash(nodeTxid: String, localTxHash: String) -> Bool {
+        let local = normalizedHash(localTxHash)
+        guard !local.isEmpty else { return false }
+        return normalizedHash(nodeTxid) == local
+    }
+
+    private static func normalizedHash(_ value: String) -> String {
```
johnnyluo, line 104: `normalizedHash` reimplements hex-prefix stripping (`if hash.hasPrefix("0x") { hash.removeFirst(2) }`) that already exists as `String.stripHexPrefix()` in `Core/Extensions/StringExtension.swift`. Could be `value.trimmingCharacters(in: .whitespacesAndNewlines).lowercased().stripHexPrefix()` — same behavior, avoids duplicating the shared helper.

johnnyluo (approved):
LGTM

PR #5370, opened 2026-09-09
fix(bond): bound the THORChain operator fee to 0…10000 basis points

### LINKED ISSUE #5347: [Fix] THORChain Bond operator fee is unbounded — an out-of-range basis-points value reaches keysign [DeFi / Bond]
## Problem

The THORChain Bond operator-fee field is validated only for parseability, not for range. `Features/FunctionTransaction/THORChain/Bond/BondTransactionViewModel.swift:56-68`:

```swift
if !value.isEmpty && Int64(value) == nil {
    throw HelperError.runtimeError("invalidOperatorFee".localized)
}
```

The value is basis points, so the valid range is `0...10000`. Nothing stops `50000` (or a negative) from being accepted, written into the memo by `BondTransactionBuilder.swift:19-30`, and carried into a keysign ceremony to be signed and broadcast.

The label (`operatorFeesLabel`) also does not name the unit, so there is nothing on screen telling the user the field is basis points rather than a percentage.

**Note:** iOS is already correct on the half that Android had wrong — `BondTransactionBuilder` passes the value through raw, with no `×100` scaling ([android#5757](https://github.com/vultisig/vultisig-android/issues/5757)). Do not introduce scaling.

## Expected Behavior

An out-of-range operator fee is rejected in the form and cannot reach a keysign ceremony. The field states its unit.

## Solution

Widen the validator to enforce `0...10000` basis points, make it blocking (the form refuses to build a transaction while the field errors), and relabel the field to name the unit across all 8 shipped locales.

### Must Do
- [ ] Enforce `0...10000` on the operator-fee field; reject non-numeric and decimal entries consistently.
- [ ] Make the validation blocking so an errored field cannot produce a `TransactionBuilder`.
- [ ] Relabel to state "basis points" in **all 8 locales** — `en`, `de`, `es`, `it`, `hr`, `nl`, `pt`, `ru`, **and `ko`**. Korean ships even though some docs omit it; verify the actual locale set in the project before editing.
- [ ] Tests: 0, 10000, 10001, negative, non-numeric, decimal, empty-with-provider-set.

### Must NOT Do
- Don't add `×100` scaling — iOS is already correct here.
- Don't change the Maya bond flow.
- Don't restructure the Bond form.

## Acceptance Criteria
- [ ] `50000` is refused in the form and cannot reach keysign.
- [ ] The label names the unit in every shipped locale; no raw keys leak.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [android#5757](https://github.com/vultisig/vultisig-android/issues/5757) · reference impl [android#5768](https://github.com/vultisig/vultisig-android/pull/5768)

gastonm5:
## Problem

The THORChain BOND form checked the operator-fee field for *parseability* and nothing else:

```swift
if !value.isEmpty && Int64(value) == nil {
    throw HelperError.runtimeError("invalidOperatorFee".localized)
}
```

The value is basis points, so `50000` — or `-5` — was accepted, interpolated verbatim into the `BOND:<node>:<provider>:<fee>` memo by `BondTransactionBuilder`, and carried into a keysign ceremony to be signed and broadcast.

## Root cause

Nothing in the form, the builder, or anywhere between them stated the range the memo's fee segment actually has. `Int64` parsing is the only rule the field had, and `Int64` is happy with `50000`.

## Fix

`BasisPointsValidator` states the `0...10000` range once and is installed on `operatorFeeField` alongside the existing cross-field rule (a provider with no fee is still refused). Distinct messages for the two failure modes: unparseable reuses `invalidOperatorFee`, out-of-range gets a new `operatorFeeRangeError`.

Two deliberate choices worth a reviewer's eye:

- **No whitespace trimming**, unlike `IntValidator`. The validator parses with plain `Int64` because `BondTransactionBuilder` does. A validator that accepted `" 500"` where the builder parses the same string back to `nil` would silently drop the fee segment from a memo the user believed carried it. Pinned by `testPaddedOperatorFeeBlocksTheBuilder`.
- **No clamping in the builder.** Clamping would sign a number the user never typed; the form refuses and says why instead.

**Blocking needed no new machinery.** `BondTransactionViewModel.transactionBuilder` already runs `validateErrors()`, which recomputes the form against the validators installed *right now*, and returns `nil` unless every one passes; `BondTransactionScreen.onContinue` guards on that builder. Any validator that throws is therefore already a hard block. The tests assert that end-to-end — through `transactionBuilder` being `nil`, not through the `validForm` flag — because `FormScreen` does not disable Continue on the flag, so the builder is the thing that actually stands between a bad value and a ceremony.

**The fee still reaches the memo raw.** No `×100` is introduced; iOS is correct on the half Android got wrong ([android#5757](https://github.com/vultisig/vultisig-android/issues/5757)). `testInRangeOperatorFeeReachesTheMemoUnscaled` pins `BOND:<node>:<provider>:500` so a future change that adds scaling fails here rather than on chain.

## Two corrections to the issue

- **The cited path is misspelled.** The issue says `Features/FunctionTransaction/THORChain/Bond/…`; the tracked directory is `.../Thorchain/Bond/...` (lowercase `chain`).
- **The relabel is already shipped, so this PR does not do it.** The issue asks to relabel the field because it "does not name the unit". `operatorFeesLabel` has read *"Operator-fees (Basis Points)"* since `f62cd95aa`, and names the unit in all 8 shipping locales already (`de` "Basis Punkte", `ko` "기준점", `zh-Hans` "基点", …). The locale work here is therefore **one new key**, `operatorFeeRangeError`, added to all 8 files — `en de es hr it ko pt zh-Hans`, the set in `scripts/sort_localizable.py` (the issue's list naming `nl` and `ru` is wrong). All 8 files reconcile at **1922 entries**; the diff is 8 insertions, 0 deletions, no re-sorting churn.

## Verification

Full gate, dedicated `en_US` simulator (iPhone 17 Pro Max, iOS 26.5), at `431ed33bd`:

```
** TEST SUCCEEDED **
Executed 7181 tests, with 11 tests skipped and 0 failures (0 unexpected) in 132.819 seconds
```

All 15 new tests appear by name in that log and passed. Disk stayed at 17–18 GB free throughout, so no stale-run risk. SwiftLint clean on every changed file.

Cross-model review: two per-commit `codex exec --sandbox read-only` passes plus a whole-branch pass. One P2 on the first commit ("boundary and blocking behaviour lack regression tests") — fixed by the test commit. No other findings at any severity.

**On-device acceptance NOT performed.** This touches a signing path (what goes into a signed memo), so it wants a human eye and a device run before merge.

Closes #5347

## Summary by CodeRabbit

* **New Features**
  * Added operator fee validation requiring values between 0 and 10,000 basis points.
  * Added localized range-error messages in supported languages.

* **Bug Fixes**
  * Improved handling of invalid, negative, fractional, oversized, nonnumeric, and whitespace-padded fee values.

* **Tests**
  * Added coverage for valid fee ranges, invalid input, provider validation, memo construction, and transaction-building behavior.

### DIFF
--- a/VultisigApp/VultisigApp/Components/Validators/BasisPointsValidator.swift
+++ b/VultisigApp/VultisigApp/Components/Validators/BasisPointsValidator.swift
@@ -1,0 +1,39 @@
+//
+//  BasisPointsValidator.swift
+//  VultisigApp
+//
+
+import Foundation
+
+/// Fails when a non-empty value is not a whole number of basis points in `0...10000`.
+///
+/// An empty value passes — pair it with `RequiredValidator`, or with whatever
+/// cross-field rule decides the field is mandatory.
+///
+/// ⚠️ **Parses with plain `Int64`, deliberately without trimming.** `IntValidator`
+/// trims whitespace; a validator that accepted `" 5"` where the memo builder parses
+/// the same string back to `nil` would drop the segment from a memo the user
+/// believed carried it. The validator and whatever encodes the value have to agree
+/// on what a number is.
+struct BasisPointsValidator: FormFieldValidator {
+
+    /// Ten thousand basis points is the whole of whatever the value is a fraction
+    /// of. Zero is a legitimate floor — it names the absence of a fee — unlike
+    /// `WithdrawBasisPoints.min`, where a withdrawal of nothing asks for nothing.
+    static let range: ClosedRange<Int64> = 0...10_000
+
+    let invalidMessage: String
+    let outOfRangeMessage: String
+
+    func validate(value: String) throws {
+        guard value.isNotEmpty else { return }
+
+        guard let basisPoints = Int64(value) else {
+            throw HelperError.runtimeError(invalidMessage)
+        }
+
+        guard Self.range.contains(basisPoints) else {
+            throw HelperError.runtimeError(outOfRangeMessage)
+        }
+    }
+}
--- a/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/de.lproj/Localizable.strings
@@ -1057,6 +1057,7 @@
 "OnboardingCard6Text4" = "";
 "openSettings" = "Einstellungen öffnen";
 "openSystemSettings" = "Systemeinstellungen öffnen";
+"operatorFeeRangeError" = "Betreibergebühr muss zwischen 0 und 10000 Basispunkten liegen";
 "operatorFeesError" = "Betreibergebühr darf nicht leer sein, wenn Anbieter ausgefüllt ist";
 "operatorFeesLabel" = "Betreibergebühren (Basis Punkte)";
 "optional" = "(optional)";
--- a/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
+++ b/VultisigApp/VultisigApp/Core/Localizables/en.lproj/Localizable.strings
@@ -1057,6 +1057,7 @@
 "OnboardingCard6Text4" = "";
 "openSettings" = "Open Settings";
 "openSystemSettings" = "Open System Settings";
+"operatorFeeRangeError" = "Operator-fee must be between 0 and 10000 basis points";
 "operatorFeesError" = "Operator-fee cannot be empty if provider is filled";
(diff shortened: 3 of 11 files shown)


johnnyluo (approved):
LGTM

PR #5373, opened 2026-09-09
fix(swap): give every SwapKitError case a title consistent with its body

### LINKED ISSUE #5362: [Fix] 17 of 18 SwapKitError cases render a domain-correct body under "Unexpected Error" [Swap]

gastonm5:
## Problem

Almost every `SwapKitError` renders an accurate, domain-correct sentence under the generic **"Unexpected Error"** heading. The user reads a correct explanation of what went wrong, titled by the app saying it does not know.

`SwapErrorPresentable` exists precisely to stop title and body being resolved through independent ladders. `SwapKitError` did not conform to it, and `SwapErrorPresentation.normalizedSwapKitError` mapped exactly one case — `.amountBelowProviderMinimum`.

**The issue's headline count is stale.** It says "17 of 18". On `main` the enum declares **21** cases (`providerNotEnabled`, `routeFiltered` and `malformedAmount` landed after the issue was written), so the real ratio is **20 of 21**.

## Fix

Conform `SwapKitError` to `SwapErrorPresentable` and delete the one-case normalization arm, so the enum resolves title *and* body through the single ladder rather than being translated into a second vocabulary by the view layer.

- **Titles for all 21 cases**, reusing an existing key wherever one already states the same verdict. Six new title keys where none did: `swapErrorAddressScreeningTitle`, `swapErrorApprovalRequiredTitle`, `swapErrorAssetBlockedTitle`, `swapErrorInvalidSourceTitle`, `swapErrorQuoteExpiredTitle`, `swapErrorRouteExpiredTitle`.
- **The five cases carrying an upstream string** (`unsupportedTxType`, `contradictoryResponse`, `responseEchoMismatch`, `malformedAmount`, `generic`) follow the rule the two response-validation cases already set in this file: the string stays in `errorDescription` for the log, and the body is the shared app copy *"This route is currently unavailable, try a different provider"* — the one statement true at all three sites they are thrown from (quote, payload build, signing). `SwapErrorPresentable`'s contract is explicit that `errorMessage` is "always localized app copy, never a raw upstream string".
- `.amountBelowProviderMinimum` renders **identically** to before: its title and body still match `SwapCryptoLogic.Errors.swapAmountTooSmall`, pinned by test rather than by inspection.

### Visible behaviour change

`generic(message:)` used to put its payload verbatim on screen — including app-side encoder failures such as `"SwapKit PSBT payload is not valid base64"`. It is now log-only. That is the fix, not a side effect, but it is the one case where a user sees different text for the same failure.

## Not done, deliberately

No error semantics changed: nothing new is thrown, nothing is thrown from anywhere new, and `init?(envelope:)` is untouched. No existing `swapKitError*` body was reworded — the issue is titles and the ladder, not a copy pass.

## Tests

New `SwapKitErrorPresentationTests` mirrors the double-exhaustive-switch pattern `SwapErrorPresentationTests` already uses, so a twenty-second case cannot compile without a sample and a recorded title decision. It pins: no case titled generically; title and body both resolving through `SwapErrorPresentation`; a sentinel string planted in every associated value never reaching the body but surviving in `errorDescription` / `refusalDetail`; `.amountBelowProviderMinimum` unchanged; non-swap errors still falling back to `localizedDescription`; and every title key in use resolving in all 8 shipped `.lproj` bundles with no English fallback.

## Localization

Six keys × **8** shipping locales (`de, en, es, hr, it, ko, pt, zh-Hans` — `ko` ships even though the repo CLAUDE.md's list omits it). Inserted in sorted position only, no whole-file re-sort; `sort_localizable.py` is a no-op on the result. All 8 files reconcile at **1927** entries (1921 + 6).

Closes #5362

## Gate

`** TEST SUCCEEDED **` — `Executed 7182 tests, with 11 tests skipped and 0 failures (0 unexpected)`, on a dedicated `en_US` iPhone 17 Pro Max simulator. That is `main`'s 7166 plus the 16 tests added here, all of which appear by name in the log. Cross-model review: Codex read-only pass per commit plus a whole-branch pass; four findings (three MAJOR, one MINOR) raised against the test commit, all triaged and fixed or rebutted.

On-device acceptance NOT performed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap error presentation with clearer, category-specific titles and messages.
  * Added handling for route availability, provider rejection, minimum amounts, invalid assets or addresses, and transaction-building issues.
  * Preserved relevant diagnostic and refusal details in displayed errors.
* **Localization**
  * Added localized swap error titles for address screening, required approval, blocked assets, invalid sources, and expired quotes or routes across supported languages.
* **Tests**
  * Expanded coverage to verify all swap error categories, fallback behavior, and localized content.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Swaps/SwapKit/SwapKitError.swift
@@ -27,9 +27,9 @@ enum SwapKitError: Error, LocalizedError, Equatable {
     /// `/v3/providers` snapshot shows at least one non-filtered provider
     /// enables both source and destination chains — meaning the pair is
     /// structurally supported, so the 404 must be amount-related rather
-    /// than a pair-coverage gap. The view layer normalizes this to
-    /// `SwapCryptoLogic.Errors.swapAmountTooSmall` so users see the same
-    /// "Amount Too Small" tooltip the THORChain path already produces.
+    /// than a pair-coverage gap. It presents under the same "Amount Too Small"
+    /// title and body as `SwapCryptoLogic.Errors.swapAmountTooSmall`, so users
+    /// see the tooltip the THORChain path already produces.
     case amountBelowProviderMinimum
     case blackListAsset
     case invalidSourceAddress
@@ -129,11 +129,9 @@ enum SwapKitError: Error, LocalizedError, Equatable {
             return "swapKitErrorNoRoutesFound".localized
         case .amountBelowProviderMinimum:
             // Reuse the existing THORChain "amount too small" copy rather than
-            // introducing a SwapKit-specific key — the user-facing meaning is
-            // identical and the view layer normalizes this case to
-            // `SwapCryptoLogic.Errors.swapAmountTooSmall` anyway. This
-            // `errorDescription` is only the fallback used if a caller logs
-            // the localized message without going through the tooltip view.
+            // introducing a SwapKit-specific key: the user-facing meaning is
+            // identical to `SwapCryptoLogic.Errors.swapAmountTooSmall`, whose
+            // title this case borrows too.
             return "swapErrorAmountTooSmallDescription".localized
         case .blackListAsset:
             return "swapKitErrorBlackListAsset".localized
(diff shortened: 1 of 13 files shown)


johnnyluo (approved):
LGTM

Issue #5348, opened 2026-09-08
[Add] TON jetton auto-discovery with verified / unverified / scam classification [TON]


gastonm5:
## Problem

iOS discovers no TON jettons and cannot tell a counterfeit from a real one.

`Core/Services/TokenDiscoverer.swift:121` returns `NoTokenDiscoverer()` for `.Ton` (alongside UTXO, Cardano, Polkadot, Tron), so a TON holder sees only the curated jetton list plus whatever they add by hand — while EVM, Solana, Cosmos, Sui and Ripple all have real discovery.

There is also no notion of jetton trust anywhere in the app. TON wallets get carpet-bombed with counterfeit jettons, fake USDT above all, and iOS currently presents whatever it is given with no signal.

## Expected Behavior

A TON account's held jettons are discovered automatically, and every jetton is classified so a counterfeit cannot pass as the real asset.

## Solution

Follow [sdk#2305](https://github.com/vultisig/vultisig-sdk/pull/2305).

**Discovery:** list every jetton with a non-zero balance through the Toncenter proxy in one paged, owner-filtered call — symbol, name, decimals and logo ride along in the same response, so there is no follow-up call per jetton. Return **verified jettons only**, so a counterfeit airdrop never auto-appears in the wallet.

**Verification tiers** (`verified` / `unverified` / `scam`):
- listed in the registry → `verified`, whatever it calls itself;
- unlisted, and the indexer flags it **or** its symbol/name collapses onto a verified jetton's → `scam`;
- otherwise → `unverified`.

The registry merges our curated TON tokens with Tonkeeper's community-reviewed `ton-assets` whitelist, fetched periodically and degrading to the curated list alone when unreachable. Cache only successful fetches, so an outage is retried rather than pinned.

Symbol normalisation must fold `USD₮`, Cyrillic `UЅDT`, `$USĐ₮` and full-width forms onto `USDT` — the fake-USDT pattern is only distinguishable by address.

**Critical finding from the SDK, do not skip:** Toncenter's own `token_info.is_scam` is honoured but is **not** the signal — all 40 tonapi-blacklisted jettons sampled from a busy wallet report `is_scam: false`. The whitelist plus the impersonation heuristic carry the classification.

### Must Do
- [ ] TON jetton discovery via the Toncenter proxy: paged, owner-filtered, zero-balance excluded, metadata from the same response.
[...]

PR #5374, opened 2026-09-09
feat(ton): discover held jettons and classify them verified / unverified / scam

gastonm5:
## Problem

`Core/Services/TokenDiscoverer.swift` mapped `.Ton` to `NoTokenDiscoverer`, so a TON holder saw only the two curated entries (`GRAM`, `USDT`) plus whatever they added by hand, while EVM, Solana, Sui, Cosmos, THORChain and Ripple all discover. There was also no notion of jetton trust anywhere in the app, and TON accounts are carpet-bombed with counterfeits — fake USDT above all.

## What this PR does

This is **PR A of two** for #5348: the discovery and classification core. It deliberately adds **no UI and no localized strings** — surfacing the tier on token rows and approval cards is PR B. PR A already delivers acceptance criteria 2, 3 and 4: verified jettons are discovered, a counterfeit is classified `scam` and never auto-appears, and a registry outage degrades to the curated list.

**Discovery** — one paged, owner-filtered Toncenter call (`exclude_zero_balance`, `limit`, `offset`), zero balances excluded, metadata taken from the same response. No follow-up call per jetton.

**Registry** — our curated TON tokens ∪ Tonkeeper's community-reviewed `ton-assets` whitelist, refreshed hourly, degrading to the curated list alone when unreachable. Only successful fetches are cached, so an outage is retried rather than pinned.

**Classification** — a pure `TonJettonClassifier` returning `verified` / `unverified` / `scam`, over a symbol fold that collapses `USD₮`, Cyrillic `UЅDT`, `$USĐ₮` and full-width forms onto `USDT`.

**Auto-discovery surfaces verified jettons only**, so a counterfeit airdrop never auto-appears.

## Notes for the reviewer

Three things here are iOS-specific and are not in [sdk#2305](https://github.com/vultisig/vultisig-sdk/pull/2305):

- **`.scam` joins the existing `TokenVerification`** rather than a parallel TON-only enum. iOS already ships `.curated` / `.verified(source:)` / `.unverified` with an `autoSurfaces` gate and a badge, so the SDK's tiers map onto a concept that exists. Ranking is `curated > scam > verified > unverified`: a positive identification survives another provider's ignorance, but nothing restored from the untrusted disk snapshot can suppress the in-code bundled list.
- **The canonical address key is user-friendly bounceable, not the SDK's raw `workchain:hex`.** WalletCore's `TONAddressConverter` has no raw converter — only `toBoc`/`fromBoc`/`toUserFriendly` — and user-friendly is what `CoinMeta.uniqueId` already dedups on, so it is also the only form that collides correctly with a coin the vault holds.
- **`isLikelySpam` would have dropped the genuine USDT.** It rejects any ticker containing a non-ASCII character, and Tether's on-chain symbol is `USD₮`. `TokenDiscoverer.vouchesForResults` lets a discoverer that matches every result against an address registry skip the heuristics; it defaults to `false`, so every other chain keeps the gate it has today. Same shape as the existing XRPL exemption in that function.

**Toncenter's `is_scam` is honoured but is not the signal** — the classifier tests assert the counterfeit is caught with `is_scam: false`, which is what all 40 blacklisted jettons the SDK sampled reported.

Discovery for `.UTXO`, `.Cardano`, `.Polkadot` and `.Tron` is unchanged, and a test asserts it.

## Verification

Full gate on `e61737bad`: `** TEST BUILD SUCCEEDED **`, `** TEST EXECUTE SUCCEEDED **`, `Executed 7244 tests, with 11 tests skipped and 0 failures (0 unexpected)` — 76 tests above the 7168 baseline, on a dedicated `en_US` simulator.

All network-dependent tests use fixtures recorded from the live services and then trimmed; nothing in this PR's tests touches the network. The fixtures preserve one detail that is easy to lose: Toncenter's `metadata` map carries `valid: true` entries of **both** `jetton_masters` and `jetton_wallets` type, and a test lists a wallet-typed entry *ahead* of the master one, so dropping the `type` filter fails the suite rather than silently classifying every jetton as unverified.

**The suite was verified test-by-test, not just run.** An earlier revision of this branch gated green at `Executed 7243 tests ... 0 failures` while a text-slice edit had silently deleted `testAFullPageThatSurvivesNoFilterIsStillWalkedPast` — the regression test for one of the review findings below. The green marker hid it; only checking each test name against the gate log caught it. It is restored and passing, and every test named in this PR was confirmed present in the final log by name.

Three review rounds all landed on the same object: the heuristic that tried to recognise a proxy replaying a page. Each variant — no new master, then no new wallet row, then two such pages in a row — also describes a legitimately overlapping page, so it is gone. Termination is guaranteed by `maxPages`, correctness by the master dedup.

## For the human reviewer

> **This decides which assets a user sees as legitimate.** A classification mistake either hides a real holding or lets a counterfeit pass. **No on-device acceptance was performed** — the evidence here is the test suite and recorded fixtures only.

**This is PR A of two, and issue #5348 stays open after it merges.** The follow-up PR B adds the tier to token rows and approval/verify cards; that one carries the linking keyword.

Refs #5348.

## Summary by CodeRabbit

- **New Features**
  - Added automatic discovery of verified TON jettons using curated and trusted registry data.
  - Added TON jetton metadata, canonical addresses, balances, images, symbols, and decimals.
  - Added scam and impersonation detection for safer token identification.
  - Added resilient registry updates with caching and fallback behavior.
  - Added support for retrieving an account’s TON jetton wallets with pagination and balance filtering.

- **Bug Fixes**
  - Improved discovery accuracy by filtering counterfeit, unverified, empty-balance, and duplicate jettons.
  - Preserved scam classifications while preventing suspicious tokens from appearing in automatic discovery.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Ton/Jetton/TonAssetsAPI.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ton/Jetton/TonAssetsAPI.swift
@@ -1,0 +1,42 @@
+//
+//  TonAssetsAPI.swift
+//  VultisigApp
+//
+
+import Foundation
+
+/// Tonkeeper's community-reviewed jetton whitelist — the list every major TON
+/// wallet treats as its verified tier. Compiled from the repo's YAML sources on
+/// every merge and served as a static file.
+///
+/// Host is injectable so tests drive the fetch without reaching the network.
+struct TonAssetsAPI: TargetType {
+    static let defaultHost = URL(string: "https://raw.githubusercontent.com")!
+
+    let host: URL
+
+    init(host: URL = TonAssetsAPI.defaultHost) {
+        self.host = host
+    }
+
+    var baseURL: URL { host }
+    var path: String { "/tonkeeper/ton-assets/main/jettons.json" }
+    var method: HTTPMethod { .get }
+    var task: HTTPTask { .requestPlain }
+    var headers: [String: String]? { ["Accept": "application/json"] }
+}
+
+/// One entry of `jettons.json`, narrowed to the fields the registry uses.
+///
+/// Everything a jetton can self-assert about its own trustworthiness is
+/// deliberately absent: being listed at all is the signal, and the tier is
+/// assigned in code (see the trust-boundary note on `TokenVerification`).
+struct TonAssetsJetton: Decodable, Sendable {
+    let address: String
+    let symbol: String?
+    let name: String?
+    let decimals: Int?
+    /// Present but explicitly `null` for the majority of entries.
+    let image: String?
+    let coingecko: String?
+}
--- a/VultisigApp/VultisigApp/Blockchain/Ton/Jetton/TonJettonAddress.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Ton/Jetton/TonJettonAddress.swift
@@ -1,0 +1,33 @@
+//
+//  TonJettonAddress.swift
+//  VultisigApp
+//
+
+import Foundation
+import WalletCore
+
+/// One spelling of a TON address, so every jetton lookup keyed by master
+/// address agrees on what the key is.
+///
+/// The same contract reaches us three ways: Toncenter answers in raw
+/// upper-case `0:HEX`, Tonkeeper's `ton-assets` lists raw lower-case, and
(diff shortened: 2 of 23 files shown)


johnnyluo (approved):
LGTM

PR #5393, opened 2026-09-12
fix(tron): price bandwidth on TRC20 sends, not just energy

### LINKED ISSUE #5389: [Fix] A TRC20 fee estimate has no bandwidth component at all, so a send with no quota left is quoted 0.345 TRX short [Tron / Fees]
# [Fix] A TRC20 fee estimate has no bandwidth component at all, so a send with no quota left is quoted 0.345 TRX short [Tron / Fees]

## Problem

`calculateTronFee` routes a token send to `calculateTrc20Fee` (`TronService.swift:250-256`), which is energy-only — it contains no reference to bandwidth. Bandwidth is computed solely in the native arm via `calculateNativeTrxFee` (`:268-275`). So a TRC20 send never prices the bandwidth the chain actually charges when the sender's quota is gone.

#5344 made the native estimate byte-accurate, which sharpened that arm and left the TRC20 arm with no bandwidth term whatsoever.

TRON charges bandwidth for a contract call on the same terms as any other transaction. In java-tron's `BandwidthProcessor.consume()` a `TriggerSmartContract` falls through `useAccountNet` → `useFreeNet` → `useTransactionFee`, which burns `getTransactionFee × bytes` for the whole transaction (`chainbase/src/main/java/org/tron/core/db/BandwidthProcessor.java`, the contract loop and `:172`). A TRC20 transfer runs about 345 bytes at the live `getTransactionFee` of 1,000 sun, so the omission is 0.345 TRX.

Confirmed on mainnet. TRC20 transfers (selector `0xa9059cbb`) whose bandwidth was NOT covered, each paying the burn as `net_fee` with zero energy fee, so the entire fee is the component iOS omits:

| tx | fee | net_fee | energy_fee |
|---|---|---|---|
| 0d7f71c98a7db4ce8115ba6d82572392f6f19b96fe5068315a211849cfc62441 | 345,000 | 345,000 | 0 |
| 8f9a854f983c4a378e0a1422ba40a015e8974092b55575ee0d183fcd3a4e4b7b | 345,000 | 345,000 | 0 |
| 058b712c4dc02aa31a7723144369d31533d76b17407ce0cb4f501695af7063fc | 345,000 | 345,000 | 0 |

And one that WAS covered, for contrast: https://tronscan.org/#/transaction/665a3a2373e754b98c67762389965020c5b945de82441fd678605d6b5483f325 carries `net_usage: 345` with no `net_fee`, because the sender's free quota absorbed it.

So the correct behaviour is conditional, not constant: charge it only when the quota cannot cover the transfer. Android does exactly that (`TronFeeService.calculateBandwidthFee`, applied to both shapes).

## Expected Behavior

A TRC20 send whose sender has no bandwidth quota left shows the ~0.345 TRX bandwidth burn in the estimate. A TRC20 send whose quota covers it still shows no bandwidth charge.

## Steps to Reproduce
1. Use a TRON account whose free daily bandwidth is already spent and which has no staked bandwidth.
2. Send any TRC20 token (USDT).
3. Compare Est. Network Fee against the on-chain fee — the estimate is 0.345 TRX short.

## Solution

Give the TRC20 arm the same bandwidth term the native arm has: fetch the account resource, size the contract transfer in bytes, and run it through `getBandwidthFeeDiscount` exactly as `calculateNativeTrxFee` does, then add the result to the TRC20 display fee. Keep it out of `feeLimit` — `fee_limit` caps energy burn only, so bandwidth does not belong in the ceiling.

Note this interacts with the pool-summing defect filed separately; fixing that one first makes this term correct rather than merely present.

### Must Do
- [ ] Add a bandwidth term to the TRC20 fee estimate, conditional on available bandwidth
- [ ] Keep bandwidth out of `feeLimit`

### Must NOT Do
- Don't change unrelated code
- Don't charge bandwidth unconditionally — a covered transfer must still show none

## Acceptance Criteria
- [ ] Build succeeds
- [ ] A TRC20 send from an account with exhausted bandwidth includes ~0.345 TRX in the estimate
- [ ] A TRC20 send from an account with its free quota intact includes no bandwidth charge
- [ ] `feeLimit` is unchanged by this work

---

Cross-platform: Android prices this correctly for both shapes; the equivalent Android work is https://github.com/vultisig/vultisig-android/pull/5861

johnnyluo:
## Summary
- `calculateTrc20Fee` priced only Energy, so a TRC20 send from an account with no free Bandwidth quota under-quoted the real fee by ~0.345 TRX (the bandwidth burn java-tron charges on every contract call, per `BandwidthProcessor.consume()`).
- Added `TronHelper.trc20TransferBandwidthBytes`, mirroring the byte-accurate native-transfer measurer for a `TransferTRC20Contract`.
- `TronService` now measures the real TRC20 transfer bytes and charges the bandwidth term only when the sender's quota can't cover it — a covered transfer still shows none. Kept out of `feeLimit`, which caps energy alone.

## Issue
Closes #5389

## Test Plan
- [x] `TronServiceFeeLimitTests` and `TronBandwidthEstimateTests` updated/added and passing
- [x] SwiftLint passes on changed files
- [x] `xcodebuild test` for the Tron suites succeeds

https://claude.ai/code/session_01LT9Ho8ZMxKRqinZbw3vtQt

## Summary by CodeRabbit

* **Improvements**
  * TRC20 transfer fee estimates now account for serialized transaction bandwidth, including memo data and the transaction’s fee limit.
  * Displayed fees include bandwidth and energy costs, with conservative estimates when token or resource data is unavailable.
  * Native TRON transfers continue using their existing fee calculation.
* **Bug Fixes**
  * Improved handling of zero-amount and invalid-recipient TRC20 transfers during fee estimation.

### DIFF
--- a/VultisigApp/VultisigApp/Blockchain/Tron/Signing/Tron.swift
+++ b/VultisigApp/VultisigApp/Blockchain/Tron/Signing/Tron.swift
@@ -289,6 +289,69 @@ enum TronHelper {
             + transactionResultBytes
     }
 
+    /// Placeholder TRC20 `amount` used when the real one is not known yet.
+    /// A zero amount serializes to 0 bytes and would under-reserve, so this
+    /// mirrors the native transfer's widest-`Int64` trick: token balances
+    /// realistically fit a `uint64`, so 8 bytes is the practical upper bound
+    /// rather than the full 32-byte `uint256` width.
+    private static let placeholderTrc20Amount = BigInt(Int64.max)
+
+    /// Bandwidth, in bytes, that a TRC20 transfer consumes once signed. Same
+    /// measurement approach as `nativeTransferBandwidthBytes`, but for
+    /// `TransferTRC20Contract` — java-tron bills bandwidth for a contract
+    /// call on the same terms as any other transaction.
+    /// See https://developers.tron.network/docs/resource-model#bandwidth-points.
+    static func trc20TransferBandwidthBytes(
+        ownerAddress: String,
+        toAddress: String,
+        contractAddress: String,
+        amount: BigInt,
+        memo: String?,
+        timestamp: UInt64,
+        expiration: UInt64,
+        blockHeaderTimestamp: UInt64, blockHeaderNumber: UInt64,
+        blockHeaderVersion: UInt64, blockHeaderTxTrieRoot: String,
+        blockHeaderParentHash: String, blockHeaderWitnessAddress: String
+    ) throws -> Int64 {
+        let contract = TronTransferTRC20Contract.with {
+            $0.ownerAddress = ownerAddress
+            $0.toAddress = toAddress
+            $0.contractAddress = contractAddress
+            $0.amount = (amount > .zero ? amount : Self.placeholderTrc20Amount).serialize()
+        }
+
+        let input = try TronSigningInput.with {
+            $0.transaction = try TronTransaction.with {
+                $0.transferTrc20Contract = contract
+                $0.timestamp = Int64(timestamp)
+                $0.expiration = Int64(expiration)
+                $0.blockHeader = try buildBlockHeader(
+                    timestamp: blockHeaderTimestamp, number: blockHeaderNumber,
+                    version: blockHeaderVersion, txTrieRoot: blockHeaderTxTrieRoot,
+                    parentHash: blockHeaderParentHash, witnessAddress: blockHeaderWitnessAddress
+                )
+                if let memo { $0.memo = memo }
+            }
+        }
+
+        let preSigningOutput = try TxCompilerPreSigningOutput(
+            serializedBytes: TransactionCompiler.preImageHashes(
+                coinType: .tron,
+                txInputData: try input.serializedData()
+            )
+        )
+        guard preSigningOutput.errorMessage.isEmpty else {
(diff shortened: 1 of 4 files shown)


VultisigApp/VultisigApp/Blockchain/Tron/Signing/Tron.swift
```diff
@@ -289,6 +289,69 @@ enum TronHelper {
...
+        blockHeaderParentHash: String, blockHeaderWitnessAddress: String
+    ) throws -> Int64 {
+        let contract = TronTransferTRC20Contract.with {
+            $0.ownerAddress = ownerAddress
+            $0.toAddress = toAddress
+            $0.contractAddress = contractAddress
+            $0.amount = (amount > .zero ? amount : Self.placeholderTrc20Amount).serialize()
+        }
+
+        let input = try TronSigningInput.with {
```
johnnyluo, line 326: TRC20 bandwidth sizing should match the transaction we actually sign. The real TRC20 path always writes `feeLimit` on the transaction, but this synthetic transaction leaves it out, so the measured byte count is a few bytes smaller than the broadcast transaction. For senders with no bandwidth quota left, that still underquotes the bandwidth burn this PR is trying to fix. Can we include the computed `feeLimit`, or a conservative placeholder, in this measurement?

PR #5412, opened 2026-09-15
fix(swap): fee-aware indicative estimate from the pair's last firm quote

### LINKED ISSUE #5346: [Fix] Swap indicative estimate ignores fees and overshoots the firm quote by 10-12% [Swap]
## Problem

While a firm quote resolves, the swap screen fills the "to" field with an indicative estimate. `SwapCryptoLogic.toAmountIndicative` (`Features/Swap/ViewModels/SwapCryptoLogic.swift:255-263`) computes it as a bare price ratio with no fees of any kind:

```swift
return amount * (fromPrice / toPrice)
```

This is the exact pre-fix shape Windows measured at 10–12% above the firm quote that replaces it:

```
$8.54  indicative -> $7.64  firm   (-10.5%)
$12.22 indicative -> $10.80 firm   (-11.6%)
```

The tell is that the fiat under both amounts is identical, which can only happen when nothing has been deducted. The user sees a materially better payout for about a second, and the correction reads as a worse rate arriving rather than an estimate settling.

## Expected Behavior

The indicative estimate approximates what the user will actually receive, so the firm quote lands close to it instead of visibly below it.

## Solution

The fee for a *new* amount cannot be known without quoting it — if it could, the estimate would be unnecessary. But the **previous firm quote for the same pair** describes what that pair costs, and native quotes carry a fee breakdown, so the two kinds of cost (proportional vs fixed) can be separated and re-applied on their own terms to the new amount. Follow [windows#4790](https://github.com/vultisig/vultisig-windows/pull/4790)'s model.

Keep the existing guarantees: display-only, greyed, `~`-prefixed, and it must **never** feed signing or validation (`:251-253`) — only the firm quote does.

### Must Do
- [ ] Deduct known fees from the indicative estimate, derived from the last firm quote for the same pair.
- [ ] Fall back to today's bare ratio when there is no prior firm quote for the pair, rather than showing nothing.
- [ ] Keep the estimate display-only and out of every signing/validation path.
- [ ] Tests: with a prior firm quote the indicative lands within a small band of the firm one; with no prior quote the old behaviour holds; the estimate never reaches `SwapPayloadBuilder`.

### Must NOT Do
- Don't let the indicative value influence validation, the payload, or the signed amount.
- Don't block the "to" field on a firm quote — the instant estimate is the point.

## Acceptance Criteria
- [ ] On a pair with a prior quote, indicative → firm no longer moves 10%+.
- [ ] The `~` prefix and greyed treatment are unchanged.

---
Source: [[cross-platform-alignment/2026-09-07-pr-sweep]] · upstream [windows#4785](https://github.com/vultisig/vultisig-windows/issues/4785) · reference impl [windows#4790](https://github.com/vultisig/vultisig-windows/pull/4790)

gastonm5:
## Description

While a firm quote loads, the swap "to" field shows a greyed `~` estimate computed from the spot ratio alone (`fromAmount × fromPrice / toPrice`). No fee of any kind is modelled, so on small trades the estimate lands 10–12% above the firm quote that replaces it — the user watches the number drop every time they edit the amount.

**Root cause.** `SwapCryptoLogic.toAmountIndicative` never had anything but prices to work with, and `fetchQuotes` blanks the previous quote on every amount edit, so at the moment the estimate is shown nothing on the view model remembers what the pair actually cost.

**Fix** (port of the model vultisig-windows shipped in vultisig/vultisig-windows#4790):

- `SwapPayoutModel` — one linear fit per pair from the last firm quote: `output = max(amount × rate × (1 − proportionalFeeFraction) − flatFee, 0)`. Native THORChain/Maya quotes split `fees.outbound` (flat) from `fees.total − fees.outbound` (proportional); aggregator quotes (1inch/Kyber/LiFi/Jupiter/SwapKit) report only a net payout, so the whole cost folds into the rate. Every term comes from the quote, never the price oracle, so the fit cannot drift with spot. Declines to fit (`nil` → spot fallback) on a non-positive amount or output, unreadable or inconsistent fee strings, or a proportional fee ≥ 50%. All `Decimal`; exact at the fitted amount by construction.
- `SwapDetailsViewModel` keeps `payoutFit: (pair, model)?`, refitted wherever a firm quote lands (network result, secured-mint synthetic quote) on the displayed quote (`selectedQuote ?? bestQuote`) at the pair/coin/amount captured before the await, and on a route pick. It deliberately survives the quote reset an amount edit performs and is scoped by pair at read time, so flipping A→B→A keeps A's fit and a new pair falls back to spot until its own first quote lands.
- `toAmountIndicative` uses the fit when the current pair matches; the `~` prefix, greyed treatment and the spot fallback are unchanged.

The estimate is display-only: nothing in `validateForm`, `makeTransaction()`, `SwapTransaction` or `SwapPayloadBuilder` reads it, and `testFittedIndicativeNeverSatisfiesValidationOrSigning` pins that.

**Also fixed at review (maintainer finding, pre-existing):** `updateQuotes` published `bestQuote` and stamped `quotedPair = currentPair` with the *post*-await pair. Both coin pickers assign the coin directly and the screen's `onChange` cancels the fetch one view update later, so a quote fetched for pair A could land in that window and be published as pair B. The silent-refresh check then kept it, `validateForm` only checks `quote != nil`, the verify screen does not re-quote on appear or sign, and `SwapPayloadBuilder` signs `quote.memo` verbatim. The success branch now guards `currentPair == requestedPair` before publishing anything; on a mismatch it only records the payout fit under the requested pair and returns, so the follow-up fetch runs as a full refresh. Both branches stamp `quotedPair = requestedPair`. Pinned by `testQuoteLandingAfterAPairChangeIsNeverPublishedAsTheNewPairsQuote`.

Closes #5346

## Which feature is affected?
- [ ] Sending
- [x] Swap — no tx link: the estimate is pre-quote display state. The quote guard added at review changes nothing about a correctly-paired quote; it only stops a superseded pair's quote from ever becoming the signable one.

## Parity

- Android: not checked in this PR.
- Windows + extension: linked: vultisig/vultisig-windows#4790 (shipped 2026-08-31; this is a port of that model, with stricter fee-consistency guards).
- SDK: n/a — the indicative amount is app-side form state; the SDK does not render one.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas (kept to the "why" only, per the surrounding-swap-code density feedback)
- [x] My changes generate no new warnings (`swiftlint` 0 violations on the touched files)

None. A before/after simulator run (type an amount on a pair that already has a firm quote, confirm the `~` value no longer drops 10%+ when the quote lands) was **not** performed: the gate simulator has no vault, and importing one to reach the swap screen with a live quote was out of bounded time. The behaviour is pinned by the view-model tests instead — worth a manual pass on device before merge.

## Additional context

**Verification**
- Full `make test` on iPhone 17 Pro Max / iOS 26.5 (en_US) at `112b5693a` (merged with `main` @ `bee02b855`): `** TEST SUCCEEDED **`, `Executed 7512 tests, with 11 tests skipped and 0 failures (0 unexpected)` (30 more than the previous run: the new test plus the suites `main` added), `SwapDetailsViewModelTests` 27/27, `SwapPayoutModelTests` 13/13. Earlier full run at `5260b8080`: `Executed 7482 tests, with 11 tests skipped and 0 failures (0 unexpected)`. New: `SwapPayoutModelTests` 13/13 (fit values on the issue's real ETH→BTC quote, exact at the fitted amount, half-amount payout `0.000060725` vs spot `0.000078125`, Maya divides by the coin's own decimals, aggregator rate-only, SwapKit human units, decline-to-fit cases, zero floor); `SwapDetailsViewModelTests` +5 (fit follows an amount edit, pair scoping, route pick, mid-flight pair change is stamped with the fetched pair, signing guardrail), +1 at review (a quote landing after a pair change is never published as the new pair's quote, and the follow-up fetch is a full refresh).
- Cross-model review (Codex, read-only) per step and whole-branch: 1 MAJOR fixed (the fit initially read the live `toCoin`/pair after the fetch's await — a chain-picker coin change lands one view update before `onChange` cancels the task, so an old-pair quote could have been fitted with the new coin's units and stamped as the new pair; now captured before the await, with a regression test), 2 MINORs fixed (fee-consistency guards, `output > 0` for every route), header comments trimmed. Review-round pass on the quote guard: 0 MAJOR, 0 MINOR, 1 NIT (the new test's `validateForm()`/`makeTransaction()` assertions are vacuous given the mock's zero fee; kept as intent documentation matching the existing guardrail test, since the `quote == nil` / `toAmountDecimal == 0` / second-fetch assertions carry the failure without the fix).

**Design notes**
- Single slot rather than a per-pair dictionary, matching Windows; the read-time pair check gives the A→B→A case for free.
- A quote that cannot be fitted drops the slot to `nil` (spot) rather than keeping a stale fit.
- The secured-mint synthetic quote fits to `rate = 1`, so no special case.
- The quote guard discards rather than re-stamps (`quotedPair = requestedPair` alone would also have made the next `fetchQuotes` clear it) so a wrong-pair quote is never observable state, even for one tick, and no fee fetch runs against a mismatched pair. The fit for the superseded pair is still recorded, so flipping back to it keeps its estimate.

- **Agent**: Claude Code
- **Issue**: #5346
- **Confidence**: high
- **Needs human review for**: the quote guard in `updateQuotes` (signing-adjacent: it decides which fetched quote may become `bestQuote`); a maintainer's eye on the native fee semantics (outbound = flat, total − outbound = proportional) for MayaChain, and the manual before/after check noted above.

## Summary by CodeRabbit

* **New Features**
  * Improved indicative swap output estimates using fee-aware calculations from recent quotes.
  * Supports proportional and flat fees across supported swap routes.
  * Preserves estimates for selected pairs and routes, with spot-price fallback when no matching quote is available.
  * Ensures estimated outputs never fall below zero.

* **Bug Fixes**
  * Prevented stale quote data from affecting estimates after swap pair changes.
  * Kept indicative estimates separate from swap validation and signing.

* **Tests**
  * Added coverage for fee calculations, route handling, pair changes, and quote-based estimates.


VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapDetailsViewModel.swift
```diff
@@ -89,6 +89,11 @@ final class SwapDetailsViewModel {
     /// Set when a route pick was dropped; the screen renders and clears it.
     var routeSelectionNotice: String?
 
+    /// Payout fit from the last firm quote shown, keyed by the pair it was fitted
+    /// on. Deliberately outlives the quote reset an amount edit performs: it is
+    /// what makes the `~` estimate for the next amount fee-aware instead of spot.
+    private(set) var payoutFit: (pair: SwapPairIdentity, model: SwapPayoutModel)?
```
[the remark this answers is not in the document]
gastonm5: Agreed — a firm quote landing on pair B evicted pair A's fit, so flipping back fell to spot.

Fixed in bc6ed3212: fits are keyed by pair (`payoutFits: [SwapPairIdentity: SwapPayoutModel]`, `SwapPairIdentity` is now `Hashable`), `currentPayoutModel` reads `payoutFits[currentPair]`, and a quote that cannot be fitted removes only that pair's entry. New test `testAFirmQuoteOnAnotherPairKeepsTheFirstPairsFit` lands a RUNE→ETH quote, flips back to RUNE→BTC, and asserts the fitted 1.97 estimate survives.

VultisigApp/VultisigApp/Features/Swap/ViewModels/SwapDetailsViewModel.swift
```diff
@@ -924,6 +949,7 @@ private extension SwapDetailsViewModel {
                 allQuotes = result.allQuotes
                 quotedPair = currentPair
```
johnnyluo: `bestQuote = result.quote` (line above) / `quotedPair = currentPair` here commit the fetch result using the post-await `currentPair`/`toCoin` instead of the `requestedPair`/`requestedToCoin` captured before the `await` two lines above `bestQuote` (used for `refitPayoutModel`). This is the same race the PR's own new test `testQuoteLandingAfterAPairChangeIsFittedToThePairItWasFetchedFor` sets up — the chain picker assigns `vm.toCoin` directly while a fetch is in flight, ahead of the screen's `onChange` → `updateToCoin` cancellation — but only the indicative `~` estimate got the fix; `bestQuote`/`quotedPair` are still exposed.

Concretely: a BTC-pair quote can land after `toCoin` has already moved to ETH, so `bestQuote` holds the BTC quote while `quotedPair`/`toCoin` claim ETH. `vm.quote` (and `toAmountDecimal`) then computes a BTC-shaped amount displayed/validated as ETH. `SwapCryptoLogic.validateForm` doesn't check that the quote's underlying asset matches `toCoin`, so `validateForm()` can pass and `makeTransaction()` can build a `SwapTransaction` with `kind: .market(quote)` targeting BTC while every UI field says ETH — until the next fetch overwrites it. The PR description flags this as "observed, not changed," but given the test already demonstrates the race and the fix for the indicative path is two lines away, it seems worth closing here too (stamp with `requestedPair`/`requestedToCoin` the same way).

gastonm5: Agreed, and it is worse than "display-only" implied: the silent-refresh check in `fetchQuotes` then keeps the mis-stamped quote (pair + amount match the new pair), `validateForm` only checks `quote != nil`, the verify screen does not re-quote on appear or on the sign tap, and `SwapPayloadBuilder` signs `quote.memo` verbatim. If the new pair's fetch fails or returns nil, the old-pair quote stays signable under the new pair's labels.

Fixed in 112b5693a by discarding rather than re-stamping: the success branch now guards `currentPair == requestedPair` before publishing anything. On a mismatch it only records the payout fit under `requestedPair` from `result.quote` and returns, so `quote` stays nil, `updateFees` is skipped, and the `onChange`-driven `fetchQuotes` runs as a full refresh instead of a silent one. Both branches stamp `quotedPair = requestedPair`. New test `testQuoteLandingAfterAPairChangeIsNeverPublishedAsTheNewPairsQuote` drives the same mid-flight race and asserts `quote == nil` / `makeTransaction() == nil` after the stale quote lands, then that `updateToCoin` starts a fresh fetch (not a silent refresh) which publishes normally.

Chose discard over `quotedPair = requestedPair` alone so a wrong-pair quote is never observable state, even for one tick, and no fee fetch runs against a mismatched pair. Full gate + Codex pass details in the PR body.

johnnyluo (approved):
LGTM

PR #5413, opened 2026-09-15
fix(swap): carry a stated-zero provider fee to the co-signer

### LINKED ISSUE #5366: [Fix] EVM/1inch swaps: a stated-zero provider fee is dropped, so the co-signer hides a row the initiator shows [Swap / Keysign]
## Problem

An EVM/1inch swap whose provider fee is genuinely **zero** shows `$0.00` on the initiator and **nothing** on the co-signer. Same defect fixed for native and SwapKit routes in [PR #5352](https://github.com/vultisig/vultisig-ios/pull/5352) / [PR #5360](https://github.com/vultisig/vultisig-ios/pull/5360), still open on the generic path — and **pre-existing on `main`**, not introduced by either.

It is reachable: `SwapCryptoLogic.affiliateFeeFiat` documents that a real affiliate route charging 0 keeps a `$0.00` row — "the Ultimate/waiver case". So a VULT Ultimate-tier user swapping on an EVM chain hits it.

Two guards collapse a stated zero into absent:

- `Features/Keysign/Services/KeysignMessage+ProtoMappable.swift:381` — `if payload.quote.tx.swapFee != "0" { … }` drops it before the wire, and takes the coin-context fields (`swap_fee_chain`, `swap_fee_token_id`, `swap_fee_decimals`) with it.
- `Features/Keysign/ViewModels/JoinKeysignSwapFeeViewModel.swift:71` — `guard let fee = BigInt(...), fee > 0 else { return nil }` hides it on read.

`string swap_fee = 7` is a plain proto3 string, so `"0"` is a non-empty value that serializes and **is** distinguishable from unset — the same property the native fix rests on (verified there at byte level: `6a 01 30`).

## Why the obvious fix is wrong

**Relaxing those two guards alone would make things worse.** `Features/Swap/Views/OneInch/EVMQuote.swift`:

```swift
// New fields with defaults for backward compatibility
self.swapFee = try container.decodeIfPresent(String.self, forKey: .swapFee) ?? "0"
```

A quote whose JSON simply **omits** `swap_fee` arrives as `"0"`, identical to a route that quoted a genuine zero. The initiator cannot tell them apart before it writes. Relaxing the guard would start writing `"0"` for both — the app asserting on the wire a zero the provider never stated, and a co-signer rendering a confident `$0.00` for a fee nobody quoted. That is worse than today's silence.

**The ambiguity has to be resolved at the quote model.** `EVMQuote.Transaction.swapFee` wants to be `String?` — `nil` when the key was absent, a value when the provider stated one — so the writer can distinguish them. That means touching every reader that currently treats `"0"` as a legitimate value: `SwapQuote.evmSwapFeeBigInt`, `SwapCryptoLogic.swapFeeString`, `affiliateFeeFiat`, and the LiFi/Kyber branches.

## Two wrinkles that make the quote-model fix insufficient on its own

1. **LiFi Solana hard-codes `swapFee: "0"`** in `LiFiService` and charges via `integratorFee` as a fraction of the output instead. Even with a `String?` model, a stated `"0"` there would be a **lie** — the fee is real and simply isn't expressible in `swap_fee`. Needs a carve-out or a different field.
2. **Jupiter and SwapKit-generic never populate the fee coin context at all** in `SwapPayloadBuilder`, so their fees reach no co-signer today regardless of zero semantics. Jupiter's `platformFee` is denominated in `toCoin` — the same shape as the LiFi problem.

So this is not a two-guard change. Scope it accordingly.

### Must Do
- [ ] Make `EVMQuote.Transaction.swapFee` able to express "no fee quoted" distinctly from "zero quoted"; update every reader.
- [ ] Then write a stated `"0"` with its coin context, and render `$0.00` on the co-signer — reserving absence for genuinely unknown.
- [ ] Decide what LiFi Solana and Jupiter should write, given their fees are not expressible in `swap_fee`.
- [ ] Tests asserting absent and stated-zero produce **different** UI, not merely that zero renders.

### Must NOT Do
- Don't relax `:381` / `:71` without fixing the decode default first.
- Don't write `"0"` for a quote that never mentioned a fee.

## Acceptance Criteria
- [ ] An Ultimate-tier EVM swap shows the same fee rows on both devices.
- [ ] A quote omitting `swap_fee` still renders no row on the co-signer.

---
**Provenance:** the native instance was found by @gastonm5 running a real two-device RUNE → RUJI swap; investigating it surfaced the same collapse in the SwapKit and generic paths. SwapKit was fixed in #5360 (both read and relay-write). The generic path was deliberately **not** widened into — a guard-only fix was started and reverted once the `?? "0"` decode default was found. There is a tripwire test on #5360, `testGenericPathStillHidesAZero`, which will fail when this is fixed; that is intended, and it should be updated to the new truth rather than relaxed.

gastonm5:
## Description

**Problem.** On EVM/aggregator swaps the initiator can show a `$0.00` "Vultisig Fee" row (KyberSwap at 0 bps for the Ultimate tier) while the co-signer shows no fee row at all. `EVMQuote.Transaction.swapFee` defaulted to `"0"`, so an omitted provider fee and a genuine stated zero were the same value; the keysign proto writer dropped any `"0"` together with its coin context, and the co-signer resolver hid a zero on the generic path because it could not tell it from "never quoted".

**Root cause.** The model collapsed *absent* and *stated zero* into one value, and every downstream layer had to guess which one it was.

**Fix.** Resolve it at the model and let every layer carry three states — absent (`nil`, never on the wire, no row anywhere), stated zero (`"0"` + coin context → `$0.00` on both devices), stated amount:

1. `EVMQuote.Transaction.swapFee` is `String?` (memberwise default `nil`, `decodeIfPresent` without `?? "0"`). `evmSwapFeeBigInt` returns `0` for a stated zero and `nil` for absent, so the payload builder attaches coin context to a stated zero. Providers stop writing a default `"0"`: LiFi-EVM states `"0"` only when the app requested a zero integrator fee (otherwise `nil` when the fixed-fee entry is missing); 1inch / SwapKit-generic fall to `nil` through the default.
2. Wire: the writer sends whatever the quote states — a stated `"0"` travels as field 7 with `swap_fee_chain/token_id/decimals`; `nil` writes nothing, so a relayed legacy payload stays byte-identical. Decode reads an empty `swap_fee` as absent, and the KyberSwap proto (which has no fee field) decodes to absent instead of a fabricated `"0"`. The co-signer resolver renders a stated zero on both aggregator payloads; the "generic path still hides a zero" tripwire is inverted.
3. LiFi-Solana states its integrator fee in `toCoin` (`toCoin.raw(dstAmount × integratorFee)`, `"0"` at zero bps) and Jupiter states `platformFee.amount` in the output mint (`"0"` when uncharged, nothing when the fee is collected on the input mint, where the initiator already suppresses the row). Jupiter joins the readers and its builder case attaches the same coin context the EVM case does. `swapFeeCoin` resolves a Solana-source LiFi fee to `toCoin` directly — a native destination has no contract to match on, and the contract-only match would have fallen back to the Solana gas coin (an 18-decimal ETH amount read as 9-decimal SOL on both devices).

`tx.data` is untouched on every path; `swap_fee` and its context are display-only fields.

**Initiator side-effects worth knowing.** `routeFeeString` (Select-route sheet) now shows the provider fee for LiFi-Solana and Jupiter candidates where it was empty, since those quotes now state it. `affiliateFeeFiat` output is unchanged for every route (LiFi-Solana's output-fraction fallback is kept for quotes that arrive without the field and agrees with the stated value by construction).

Closes #5366

## Which feature is affected?
- [ ] Sending
- [x] Swap — display-only change to the keysign payload's `swap_fee` group; no on-device swap was executed for this PR (see "Needs human review")

## Parity

- Android: n/a for this issue — its mapper tests cover the coin context, not zero semantics; not verified further.
- Windows + extension: linked — Windows' `getSwapFeeFromPayload.ts` hides any zero row on join by its own choice, so the parity this PR delivers is iOS↔iOS; the wider "which fee is `swap_fee`" question is vultisig-ios#5357.
- SDK: n/a — `getKeysignSwapFeeFields.ts` already documents an empty `swapFee` as "the payload states no fee", which is the semantics this PR adopts.

## Checklist

## Additional context

Sibling fixes for the same class of bug on other payloads: native #5352, SwapKit #5360. Plan and per-provider table: `wiki/pages/projects/vultisig/ios-bugs/evm-swap-stated-zero-fee-5366-plan.md`.

**Verification.** Three codex-reviewed commits (no findings on any step; whole-branch pass: 1 MAJOR rejected with reason — see "Codex ledger" below). Full iOS gate on a dedicated en_US simulator: `** TEST SUCCEEDED **`, `Executed 7489 tests, with 11 tests skipped and 0 failures` on `iPhone 17 Pro Max, iOS 26.5` (the runtime CI pins). A first run on an iOS 27.0 sim showed only the known `CreateVaultSnapshotTests`/`LimitPriceChartSnapshotTests` pixel-precision artifact (6 failures, all snapshot, none touching this change); it was green on 26.5..

**Codex ledger** (`gpt-5.6-sol`, read-only, per step + whole branch): step 1 — no findings; step 2 — no findings; step 3 — 1 MINOR (a fee-bearing Jupiter response with an absent/malformed `platformFee.amount` states `"0"`) rejected: `swapFee` and the initiator's `platformFee` derive from the same value, so both devices render the same `$0.00`, and nil would recreate the split this PR removes; whole branch — 1 MAJOR (uncharged native-SOL Jupiter output states nil, "hiding a genuine $0.00 row") rejected: the initiator hides that row too (`showAffiliateFeeRow` is false for any `feeOnInput` quote), so nil is the symmetric outcome; whether an uncharged native-SOL swap should show `$0.00` on the initiator is a separate product call.

- **Agent**: Claude Code
- **Issue**: #5366
- **Confidence**: high for commits 1–2 (Kyber repro, wire, co-signer); medium for commit 3 (LiFi-Solana / Jupiter now state a fee that was previously never on the wire)
- **Needs human review for**: keysign payload fields (`swap_fee` + `swap_fee_chain/token_id/decimals` on `OneInchTransaction`) — display-only, `tx.data` unchanged, but it touches what a co-signer sees before signing; and an on-device run of an Ultimate-tier KyberSwap quote showing `$0.00` on both devices and a 1inch quote showing no row on the co-signer, which was not performed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap-fee handling across Jupiter, LiFi, 1inch, and KyberSwap routes.
  * Fees now preserve accurate token units and metadata, including integrator fees.
  * Explicit zero fees are displayed correctly, while missing fees remain undisplayed.
  * Improved fee-token selection for Solana swaps and Jupiter EVM transactions.
* **Tests**
  * Expanded coverage for fee-free, zero-fee, input-fee, output-fee, and integrator-fee scenarios.


johnnyluo (approved):
LGTM

PR #5423, opened 2026-09-16
fix(macOS): restore opaque native window title bar

### LINKED ISSUE #5422: [Fix] Restore opaque native window title bar on macOS 27
## Problem
On macOS 27, Vultisig's native window title bar appears translucent and picks up the content gradient. The previous macOS 26 layout workaround does not explicitly provide an opaque title-bar background.

## Expected behavior
The native title bar should use the app's solid primary background color consistently, including while scrolling and navigating to chain details. Preserve the window title and standard window controls.

This is separate from the empty native tab strip fixed in #5418.

gastonm5:
## Summary
On macOS 27, the native window title bar picks up the content gradient instead of staying opaque. Set its background to `Theme.colors.bgPrimary` and keep it visible at the shared macOS content root, covering both window and document scenes.

Closes #5422

## Validation
- Full `make test` on iOS 26.5: 7,492 tests, 11 skipped, 0 failures.
- Signed macOS 27 build passed.
- Verified in the rebuilt macOS app on Home, while scrolling, and on chain details.
- SwiftLint: unchanged 27 existing warnings, 0 serious violations.

Only two production modifiers are added; no tests or scripts are introduced.

## Summary by CodeRabbit

* **Style**
  * Updated the macOS interface to use the primary background color.
  * Ensured the window toolbar background remains visible.
  * Standardized the macOS application appearance to dark mode for a consistent visual experience.


johnnyluo (approved):
LGTM

PR #5425, opened 2026-09-16
fix(swap): block stale quotes after failed refreshes

### LINKED ISSUE #5354: [Fix] A failed quote refresh leaves the stale swap quote signable [Swap]

gastonm5:
A failed swap refresh could leave the previous quote available to Continue and `makeTransaction()`, even after settings changed. Quote handoff now requires a successful quote-and-fee refresh for the current pair, amount and advanced settings. Transient failures retain the displayed preview but block progression; structural failures clear it. Dismissing the error cannot restore readiness.

Regression coverage includes failed periodic/settings refreshes, structural and empty results, fee failures, in-flight handoff, recovery, and a cancelled request failing after its replacement succeeds.

Closes #5354

Manual provider selection is retained as provider identity while settings invalidate the quote payload, then reattached only to a newly returned quote. A structural failure keeps automatic retries available for a previously quoted request.

Validation: six original regression cases failed against unchanged production code in simulator-hosted tests. Final full iOS 26.5 suite: 7,501 tests, 11 skipped, zero failures, including nine new regressions and existing settings/provider-selection coverage. SwiftLint remains at the baseline 27 warnings with zero serious violations. Read-only Claude step and whole-branch reviews completed; all findings triaged.

Verify refresh/signing exclusion is handled separately in #5424. Tests did not submit a real transaction; physical-device swap smoke testing remains before merge.

Universal macOS build also passed at the tested commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap quote validation and refresh behavior to prevent stale or invalid quotes from being used.
  * Preserved the selected provider while updated quote details are loading, and cleared selections that are no longer available.
  * Continue is no longer blocked solely while quote or fee data is loading.
  * Improved recovery from quote and fee errors, including retry handling and cancellation behavior.
  * Advanced settings changes now correctly refresh and revalidate swap details.

* **Tests**
  * Expanded coverage for quote refreshes, settings changes, provider selection, fee errors, and transaction safeguards.


johnnyluo (approved):
LGTM

